PR #464, opened 2024-05-17
genpolicy-msft: bump version to 3.2.0.azl1.genpolicy0

burgerdev:
This release is what's currently available on AKS.

### DIFF
--- a/packages/by-name/genpolicy-msft/package.nix
+++ b/packages/by-name/genpolicy-msft/package.nix
@@ -4,7 +4,6 @@
 { lib
 , fetchFromGitHub
 , fetchurl
-, fetchpatch
 , applyPatches
 , rustPlatform
 , openssl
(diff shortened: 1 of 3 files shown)


packages/by-name/genpolicy-msft/package.nix
```diff
@@ -60,11 +46,15 @@ rustPlatform.buildRustPackage rec {
...
+  preBuild = ''
+    make src/version.rs
+  '';
+
   passthru = rec {
     settings = fetchurl {
       name = "${pname}-${version}-settings";
-      url = "https://github.com/microsoft/kata-containers/releases/download/${version}/genpolicy-settings.json";
-      hash = "sha256-Q19H7Oj8c7SlPyib96fSRZhx/nJ96HXb8dfb9Y/Rsw8=";
+      url = "https://raw.githubusercontent.com/microsoft/kata-containers/${version}/src/tools/genpolicy/genpolicy-settings.json";
```
katexochen, line 56: Should we add a comment to switch this back to the release artifact once they publish the next genpolicy relelease?

katexochen (approved):

packages/by-name/genpolicy-msft/package.nix
burgerdev: Given that the tag is out for 3 weeks by now, and that `3.2.0.azl1.genpolicy0` and `3.2.0.azl1` track the same commit, I don't expect them to publish these files as release artifacts.

katexochen: Maybe they forgot this time?

burgerdev: Comment added

PR #477, opened 2024-05-21
coordinator: add manifest generation metric

davidweisse:
This adds a metric for the current manifest generation set at the Coordinator.

### DIFF
--- a/coordinator/userapi.go
+++ b/coordinator/userapi.go
@@ -25,6 +25,7 @@ import (
 	"github.com/edgelesssys/contrast/internal/userapi"
 	grpcprometheus "github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


coordinator/userapi.go
```diff
@@ -54,6 +60,12 @@ func newUserAPIServer(mSGetter manifestSetGetter, caGetter certChainGetter, reg
 		),
 	)
 
+	manifestGeneration := promauto.With(reg).NewCounter(prometheus.CounterOpts{
```
burgerdev, line 63: Using a counter might be incorrect here, considering coordinator restarts or distributed coordinators. Should this be a gauge instead?

coordinator/userapi.go
```diff
@@ -122,6 +137,8 @@ func (s *userAPIServer) SetManifest(ctx context.Context, req *userapi.SetManifes
 		return nil, status.Errorf(codes.Internal, "setting manifest: %v", err)
 	}
 
+	s.metrics.manifestGeneration.Inc()
```
burgerdev, line 140:
```suggestion
	s.metrics.manifestGeneration.Set(len(s.manifSetGetter.GetManifests()))
```
Does not make a difference right now, but will when we support recovery.

burgerdev (approved):

burgerdev (approved):

PR #478, opened 2024-05-21
service-mesh: open envoy admin endpoint

davidweisse:
This enables the admin interface of the Envoy Proxy in the Service Mesh. The interface is running by default on port `9901` but is only accessible if the ingress settings for the Proxy are manually set to receive data from the specified port.


docs/docs/architecture/observability.md
```diff
@@ -30,3 +30,24 @@ For the mesh API, the metric names are prefixed with `meshapi_grpc_server_`. The
...
+The [Service Mesh](../components/service-mesh.md) can be configured to expose
+metrics via its [Envoy admin
+interface](https://www.envoyproxy.io/docs/envoy/latest/operations/admin). Be
+aware that the admin interface can expose private information and allows
+destructive operations to be performed. To enable the admin interface for the
+Service Mesh, set the environment variable `EDG_ADMIN_PORT` in the configuration
+of the Service Mesh sidecar. If this variable is set, the admin interface will
+be started on this port.
+
+By default, the Service Mesh is configured to start the admin interface on port
```
burgerdev, line 45: Is that true? The implementation seems to only configure the admin port if it's explicitly set.

davidweisse: Envoy automatically starts an administration interface if an address is specified. If not specified, the server will not start, see [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#envoy-v3-api-msg-config-bootstrap-v3-admin).

e2e/servicemesh/servicemesh_test.go
```diff
@@ -114,6 +114,27 @@ func TestIngressEgress(t *testing.T) {
...
+
+		backendPods, err := c.PodsFromDeployment(ctx, ct.Namespace, "emoji")
+		require.NoError(err)
+		require.Len(backendPods, 1, "pod not found: %s/%s", ct.Namespace, "emoji")
+
+		frontendPods, err := c.PodsFromDeployment(ctx, ct.Namespace, "web")
+		require.NoError(err)
+		require.Len(frontendPods, 1, "pod not found: %s/%s", ct.Namespace, "web")
+
+		argv := []string{"curl", "-sS", frontendPods[0].Status.PodIP + ":9901"}
```
burgerdev, line 134:
```suggestion
		argv := []string{"curl", "-fsS", net.JoinHostPort(frontendPods[0].Status.PodIP, "9901") + "/stats/prometheus"}
```

service-mesh/main.go
```diff
@@ -35,7 +36,10 @@ func run() (retErr error) {
 	ingressProxyConfig := os.Getenv(ingressProxyConfigEnvVar)
 	log.Println("Egress Proxy configuration:", ingressProxyConfig)
 
-	pconfig, err := ParseProxyConfig(ingressProxyConfig, egressProxyConfig)
+	adminPort := os.Getenv(adminPortEnvVar)
+	log.Println("Port for Envoy admin interface:", adminPort)
```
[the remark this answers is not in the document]
davidweisse: I'm trying to be consistent with other log outputs. E.g., when no ingress or egress are configured, you would see that in the logs. Same with the admin port: if no admin port has been set, you would see an empty field for the port.

burgerdev (approved):

burgerdev (approved):

PR #488, opened 2024-05-23
ci: create and update a `versions.json` file

miampf:
Create and update a `versions.json` file that contains released versions with their corresponding files and hash. This file will be used to update parts of nix code. A CI run of the release CI can be seen [here](https://github.com/edgelesssys/contrast/actions/runs/9208085083).


versions.json
katexochen, line 1: putting this file top level is a bit too prominent, maybe put it in packages/by-name/contrast-release/

.github/workflows/release.yml
```diff
@@ -251,6 +261,14 @@ jobs:
...
         run: |
           nix build -L .#cli-release
+      - name: Update versions.json in main
+        uses: ./.github/actions/update_versions_json
+        with:
+          version: "${{ needs.process-inputs.outputs.MAJOR_MINOR_PATCH }}"
+      - name: Commit new versions.json
+        run: |
+          git add versions.json
+          git commit -m "release: update versions.json"
```
katexochen, line 271: I think we should use an artifact for the updated versions.json to get this into the other job.

.github/workflows/release.yml
```diff
@@ -151,20 +151,30 @@ jobs:
...
+          echo "commit-msg=$commit_msg" | tee -a "$GITHUB_OUTPUT"
       - name: Create PR
         uses: peter-evans/create-pull-request@6d6857d36972b65feb161a90e484f2984215f83e # v6.0.5
         with:
           title: Post ${{ needs.process-inputs.outputs.WITHOUT_V }} release updates to main
           body: |
             Updating main as part of the ${{ needs.process-inputs.outputs.WITHOUT_V }} release.
             Only merge after the release is published.
-          commit-message: ${{ steps.bump.outputs.commit-msg }}
+          commit-message: ${{ steps.bump.result == 'skipped' && steps.versions_update.outputs.commit-msg || steps.bump.outputs.commit-msg }}
```
katexochen, line 177: I think we can update the versions.json first and then do the bump. should be ok to have commit-message empty in case we didn't bump, as there shouldn't be anything to commit.

.github/actions/update_versions_json/update_versions.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
```
katexochen, line 1:
```suggestion
#!/usr/bin/env bash
```
Can we have the full script as a nix script instead of an action? (/packages/scripts.nix)

katexochen (commented):
Still need to review the script, but here are some initial thoughts.

.github/actions/update_versions_json/update_versions.sh
miampf: Should not be a problem, I'll rewrite it accordingly.

miampf:
New working workflow with changes is [here](https://github.com/edgelesssys/contrast/actions/runs/9222323187).

packages/scripts.nix
```diff
@@ -242,4 +242,84 @@ with pkgs;
       az vm list-skus --size "$sku" | jq -r '.[] | .locations.[]'
     '';
   };
+
+  update-versions-json = writeShellApplication {
+    name = "update-versions-json";
+    runtimeInputs = [
+      jq
+    ];
+    text = ''
```
katexochen, line 251: Given the size of the script, it might be better to use `builtins.readFile`

packages/update-versions-json.sh
```diff
@@ -0,0 +1,73 @@
...
+
+  # get the file path
+  file=${fields["$field"]}
+
+  echo "[*] Creating hash for $file"
+  hash=$(create_hash "$file")
+  echo "      $hash"
+
+  echo "[*] Updating ./packages/versions.json for $field"
+  update "$field" "$hash"
```
katexochen, line 64: all functions are only ever called once, inlining could improve readability.

packages/update-versions-json.sh
```diff
@@ -0,0 +1,73 @@
...
+# `check_for_version` checks if the given entry already contains a version.
+function check_for_version {
+  out=$(jq --arg NAME "$1" --arg VERSION "$VERSION" '.[$NAME] | map(select(.version == $VERSION))' ./packages/versions.json || exit 1)
+  if [[ ! "$out" = "[]" ]]; then
+    echo "[x] Version $VERSION exists for entry $1"
+    exit 1
+  fi
+}
+
+echo "::group::Updating versions"
```
katexochen, line 25: not that much output, no need to group

packages/update-versions-json.sh
```diff
@@ -0,0 +1,73 @@
...
+# SPDX-License-Identifier: AGPL-3.0-only
+
+# `create_hash` creates a hash for a given file.
+function create_hash {
+  nix hash file --sri --type sha256 "$1" || exit 1
+}
+
+# `update` appends a new version entry to a given section.
+function update {
+  out=$(jq --arg NAME "$1" --arg HASH "$2" --arg VERSION "$VERSION" '.[$NAME] |= . + [{"version": $VERSION,hash: $HASH}]' ./packages/versions.json || exit 1)
```
katexochen, line 12:
```suggestion
  out=$(
      jq --arg NAME "$1" \
          --arg HASH "$2" \
          --arg VERSION "$VERSION" \
          '.[$NAME] |= . + [{"version": $VERSION,hash: $HASH}]' \
          ./packages/versions.json
)
```

Format something like that for better readability

katexochen (changes requested):

katexochen:
- added simplifications requested by shfmt
- renamed file to contrast-releases.json
- pint messages to stderr
- prefix error messages with "Error" for better searchability

katexochen:
e2e test run succeeded, changes look good: https://github.com/edgelesssys/contrast/pull/530

katexochen (approved):

PR #490, opened 2024-05-23
resourcegen: settle on one set for the Coordinator

burgerdev:
We currently have a `Coordinator` (only deployment) and a `CoordinatorRelease` (with service and LB annotation), both in use. Since 

1. there is no release-specific information in the `CoordinatorRelease` set,
2. the coordinator definition is functionally identical and
3. the coordinator is always used with its service

settle on one `CoordinatorBundle` that's used everywhere.


malt3 (approved):
Thanks! Looks cleaner.

PR #491, opened 2024-05-23
coordinator: make metrics endpoint configurable

davidweisse:
This makes the Coordinator metrics endpoint configurable via an environment variable. If `CONTRAST_METRICS_PORT` is set, the metrics server will start on that port. If the variable is not set, the metrics server will not start.

For CI, the endpoint is activated through the `PatchCoordinatorMetrics` function.

### DIFF
--- a/coordinator/main.go
+++ b/coordinator/main.go
@@ -19,6 +19,10 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	metricsPortEnvVar = "CONTRAST_METRICS_PORT"
+)
+
 func main() {
 	if err := run(); err != nil {
 		os.Exit(1)
@@ -43,6 +47,8 @@ func run() (retErr error) {
 		return fmt.Errorf("setting up mount: %w", err)
 	}
 
+	metricsPort := os.Getenv(metricsPortEnvVar)
+
 	caInstance, err := ca.New()
 	if err != nil {
 		return fmt.Errorf("creating CA: %w", err)
@@ -57,7 +63,13 @@ func run() (retErr error) {
 	eg := errgroup.Group{}
 
 	eg.Go(func() error {
-		logger.Info("Starting prometheus /metrics endpoint")
+		if metricsPort == "" {
+			return nil
+		}
+		if metricsPort == userapi.Port || metricsPort == meshapi.Port {
+			return fmt.Errorf("invalid port for metrics endpoint: %s", metricsPort)
+		}
+		logger.Info("Starting prometheus /metrics endpoint on port " + metricsPort)
 		mux := http.NewServeMux()
 		mux.Handle("/metrics", promhttp.InstrumentMetricHandler(
 			promRegistry, promhttp.HandlerFor(
@@ -65,7 +77,7 @@ func run() (retErr error) {
 				promhttp.HandlerOpts{Registry: promRegistry},
 			),
 		))
-		if err := http.ListenAndServe(":9102", mux); err != nil {
+		if err := http.ListenAndServe(":"+metricsPort, mux); err != nil {
 			return fmt.Errorf("serving Prometheus endpoint: %w", err)
 		}
 		return nil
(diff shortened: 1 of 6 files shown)


internal/kuberesource/sets.go
```diff
@@ -638,3 +638,21 @@ func PatchServiceMeshAdminInterface(resources []any, port int32) []any {
 	}
 	return resources
 }
+
+// PatchCoordinatorMetrics enables Coordinator metrics on the specified port.
+func PatchCoordinatorMetrics(resources []any, port int32) []any {
+	for _, resource := range resources {
+		switch r := resource.(type) {
+		case *applyappsv1.DeploymentApplyConfiguration:
+			if *r.Name == "coordinator" {
```
burgerdev, line 647: We are using an annotation in other functions to identify coordinators.

internal/kuberesource/sets.go
```diff
@@ -638,3 +638,21 @@ func PatchServiceMeshAdminInterface(resources []any, port int32) []any {
 	}
 	return resources
 }
+
+// PatchCoordinatorMetrics enables Coordinator metrics on the specified port.
```
burgerdev, line 642: Should this also add the `prometheus.io/scrape` annotation?

internal/kuberesource/sets.go
burgerdev, line 1: I started moving some of the mutating functions to `mutators.go` last week. Maybe rebase and move yours, too.

internal/kuberesource/sets.go
```diff
@@ -638,3 +638,21 @@ func PatchServiceMeshAdminInterface(resources []any, port int32) []any {
...
+func PatchCoordinatorMetrics(resources []any, port int32) []any {
+	for _, resource := range resources {
+		switch r := resource.(type) {
+		case *applyappsv1.DeploymentApplyConfiguration:
+			if *r.Name == "coordinator" {
+				r.Spec.Template.Spec.Containers[0].WithEnv(NewEnvVar("EDG_METRICS_PORT", fmt.Sprint(port)))
+				r.Spec.Template.Spec.Containers[0].WithPorts(
+					ContainerPort().
+						WithName("prometheus").
+						WithContainerPort(9102),
```
burgerdev, line 652:
```suggestion
						WithContainerPort(port),
```

internal/kuberesource/sets.go
davidweisse: As we only use this for E2E-tests, we don't need this annotation because we can discover the coordinator pods by name and port.

coordinator/main.go
```diff
@@ -18,6 +18,10 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	metricsPortEnvVar = "EDG_METRICS_PORT"
```
katexochen, line 22:
```suggestion
	metricsPortEnvVar = "CONTRAST_METRICS_PORT"
```

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -96,6 +96,7 @@ func (ct *ContrastTest) Init(t *testing.T, resources []any) {
 	resources = kuberesource.PatchImages(resources, ct.ImageReplacements)
 	resources = kuberesource.PatchNamespaces(resources, ct.Namespace)
 	resources = kuberesource.PatchServiceMeshAdminInterface(resources, 9901)
+	resources = kuberesource.PatchCoordinatorMetrics(resources, 9102)
```
katexochen, line 99: Not sure about patching this within `contrasttest`. We might want to add a test case where we disable the metrics endpoint to check it is actually not reachable. Same for the envoy admin iface or other config things. Any thoughts on this @burgerdev ?

Anyway, nothing we must address in this PR.

katexochen (approved):

e2e/internal/contrasttest/contrasttest.go
burgerdev: I agree: we should eventually test these scenarios, but until then we can leave it here.

burgerdev (approved):

PR #493, opened 2024-05-24
resourcegen: don't return errors when creating sets

burgerdev:
Our sets are known at compile time and thus creating them should never cause an error. This PR changes the implementation of a couple of mutators to be error-less, by switching them to `MapPodSpec`, and adds tests to ensure that this mapping still works. As a byproduct, `AddInitializer` and `AddServiceMesh` can now be used with any pod-creating resource.


malt3 (approved):
LGTM. Is there a reason this is still a draft?

burgerdev:
Thanks for the review. It's a draft because the diffbase is not merged yet.

PR #496, opened 2024-05-24
genpolicy-msft: add support for volumeDevices

burgerdev:
This is essentially kata-containers/kata-containers#9706, but adopted to Microsoft's fork.

Since the patch touches `rules.rego`, we can't take the one from the URL anymore, so I changed the rules to small derivations that extract the patched `rules.rego` and `genpolicy-settings.json` files, respectively.

### DIFF
--- a/packages/by-name/genpolicy-msft/package.nix
+++ b/packages/by-name/genpolicy-msft/package.nix
@@ -3,8 +3,9 @@
 
 { lib
 , fetchFromGitHub
-, fetchurl
+, fetchpatch
 , applyPatches
+, stdenvNoCC
 , rustPlatform
 , openssl
 , pkg-config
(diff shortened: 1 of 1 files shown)


packages/by-name/genpolicy-msft/package.nix
```diff
@@ -25,6 +26,17 @@ rustPlatform.buildRustPackage rec {
     hash = "sha256-W36RJFf0MVRIBV4ahpv6pqdAwgRYrlqmu4Y/8qiILS8=";
   };
 
+  patches = [
+    # TODO(burgerdev): drop this patch after upstream applied it
```
katexochen, line 30: Can you link the upsteam PR?

packages/by-name/genpolicy-msft/package.nix
```diff
@@ -3,8 +3,9 @@
 
 { lib
 , fetchFromGitHub
-, fetchurl
+, fetchpatch
 , applyPatches
+, pkgs
```
katexochen, line 8: we never pass `pkgs` to a package expression. The idea of `callPackage` is to only pass the args needed.

packages/by-name/genpolicy-msft/package.nix
```diff
@@ -49,14 +61,16 @@ rustPlatform.buildRustPackage rec {
   '';
 
   passthru = rec {
-    settings = fetchurl {
+    settings = pkgs.stdenv.mkDerivation {
```
katexochen, line 64:
```suggestion
    settings = stdenvNoCC.mkDerivation {
```

katexochen (changes requested):

packages/by-name/genpolicy-msft/package.nix
burgerdev: Thanks, changed.

packages/by-name/genpolicy-msft/package.nix
burgerdev: Done

katexochen (approved):

PR #497, opened 2024-05-24
userapi: add seed to protocol; seedengine: implement key derivation

katexochen:
First step for Coordinator recovery. 

- The UserAPI protocol is extended with a field for the seed shares document, which will in the future contain the seed (recovery secret) encrypted once with each workload owner key defined in the manifest
- The seedengine takes an initial seed and salt and deterministically derives the required ECDSA keys and symmetric secrets.

While the seedengine is dead code for now, the userapi protocol change will be live, but no seed will be sent for now, and now file will be printed on the client side.


internal/seedengine/seedengine.go
```diff
@@ -0,0 +1,116 @@
...
+	se.rootCAKey, err = se.generateECDSAPrivateKey(rootCASeed)
+	if err != nil {
+		return nil, fmt.Errorf("generating ECDSA key: %w", err)
+	}
+
+	return se, nil
+}
+
+// DerivePodSecret derives a secret for a pod from the policy hash and the secret seed.
+func (s *SeedEngine) DerivePodSecret(policyHash []byte) ([]byte, error) {
```
malt3, line 78: Not sure how relevant this is in practice, but should we check the length of the input here?

malt3 (approved):
Otherwise LGTM

3u13r (approved):
I would have named the seedengine something like secretengine (or secretfactory in Java speak), since it is there to generate secrets. But I'm also fine with the current naming.

internal/seedengine/seedengine.go
```diff
@@ -0,0 +1,116 @@
...
+// DeriveMeshCAKey derives a secret for a mesh CA from the transaction and the secret seed.
+func (s *SeedEngine) DeriveMeshCAKey(transaction []byte) (*ecdsa.PrivateKey, error) {
+	transactionSecret, err := s.hkdfDerive(s.historySeed, fmt.Sprintf("TRANSACTION SECRET %x", transaction))
+	if err != nil {
+		return nil, err
+	}
+	meshCASeed, err := s.hkdfDerive(transactionSecret, "MESH CA SECRET")
+	if err != nil {
+		return nil, err
+	}
```
burgerdev, line 104: What's the idea behind the second derivation here?

internal/seedengine/seedengine_test.go
burgerdev, line 1: The test vectors used here are effectively a regression detection smoke test that catches us/Go changing something that affects key determinism, right? We should maybe leave a note somewhere that warns of changing the reference values.

What I'm missing is tests for invariants of the construction, such as

* Creating (pod secrets|mesh CA keys) for the same (pod hash|transaction) from two engine instances results in the same secret.
* Derived pod keys and CA keys for different hashes are different.

internal/seedengine/seedengine.go
katexochen: I've switched the input type from `[]byte` to `[32]byte` and check explicitly for an empty array as well as for the hash of an empty byte slice.

internal/userapi/userapi.proto
```diff
@@ -19,6 +19,18 @@ message SetManifestResponse {
...
+}
+
+message SeedShareDocument {
+  repeated SeedShare SeedShares = 1;
+  bytes salt = 2;
+}
+
+message SeedShare {
+  string PublicKey = 1;
+  bytes EncrypteSeed = 2;
```
burgerdev, line 33:
```suggestion
  bytes EncryptedSeed = 2;
```

internal/userapi/userapi.proto
```diff
@@ -19,6 +19,18 @@ message SetManifestResponse {
   bytes RootCA = 1;
   // PEM-encoded certificate
   bytes MeshCA = 2;
+  // Secred seed (share), encrypted with each of the recovery holders' public keys.
```
burgerdev, line 22:
```suggestion
  // Secret seed (share), encrypted with each of the recovery holders' public keys.
```

internal/seedengine/seedengine.go
```diff
@@ -0,0 +1,129 @@
...
+	se.rootCAKey, err = se.generateECDSAPrivateKey(rootCASeed)
+	if err != nil {
+		return nil, fmt.Errorf("generating ECDSA key: %w", err)
+	}
+
+	return se, nil
+}
+
+// DerivePodSecret derives a secret for a pod from the policy hash and the secret seed.
+func (s *SeedEngine) DerivePodSecret(policyHash [32]byte) ([]byte, error) {
```
burgerdev, line 79: I appreciate the use of fixed-size arrays. Should the size be a constant?

internal/seedengine/seedengine.go
```diff
@@ -0,0 +1,129 @@
...
+
+	podStateSeed []byte
+	historySeed  []byte
+
+	rootCAKey             *ecdsa.PrivateKey
+	transactionSigningKey *ecdsa.PrivateKey
+}
+
+// New creates a new SeedEngine from a secret seed and a salt.
+func New(secretSeed []byte, salt []byte) (*SeedEngine, error) {
```
burgerdev, line 36: Do we want to require a fixed length for these inputs, too?

internal/seedengine/seedengine.go
katexochen: Nothing in particular, mostly holding open the possibility to derive other secrets from `transactionSecret`.

3u13r (approved):

internal/seedengine/seedengine_test.go
katexochen: > * Creating (pod secrets|mesh CA keys) for the same (pod hash|transaction) from two engine instances results in the same secret.
>  * Derived pod keys and CA keys for different hashes are different.

Aren't those cases basically covered by the existing tests? First point: if you re-run the test, you will have a different engine instance. The test checks if the same hash will give the same secret as before. Second point: as we have multiple test cases with different hash inputs, we also ensure that they output different secrets.

internal/seedengine/seedengine.go
katexochen: hm, should be `s.hashFun().Size()` but that's not constant. :disappointed:

burgerdev: The internal hash function is not necessarily identical to the pod policy hashes, is it?

internal/seedengine/seedengine.go
katexochen: salt was already checked, I've added secretSeed >= hashLen.

internal/seedengine/seedengine.go
katexochen: You mean we should enforce the len of a hash fun used in another package? That doesn't sound good to me from a design point of view/

internal/seedengine/seedengine_test.go
burgerdev: It's all implicitly there, I agree. Thanks for adding the note.

PR #510, opened 2024-05-28
cli: inject initializer with `contrast generate`

davidweisse:
The `generate` command will now automatically inject the Initializer init container to all resources with the `contrast-cc` runtime. This can be disabled by either specifying the `--skip-initializer` flag when calling `generate`, or with the `contrast.edgeless.systems/skip-initializer` annotation set to `"true"` for a specific resource.

Release run: https://github.com/edgelesssys/contrast/actions/runs/9350903637


internal/kuberesource/mutators.go
```diff
@@ -5,42 +5,78 @@ package kuberesource
...
+
+		// Remove already existing volume mounts on the worker container with unique volume mount name.
+		spec.Containers[0].VolumeMounts = slices.DeleteFunc(spec.Containers[0].VolumeMounts, func(v applycorev1.VolumeMountApplyConfiguration) bool {
+			return *v.Name == *initializer.VolumeMounts[0].Name
+		})
+
 		// Add the volume mount written by the initializer to the worker container.
 		spec.Containers[0].WithVolumeMounts(VolumeMount().
 			WithName(*initializer.VolumeMounts[0].Name).
 			WithMountPath(*initializer.VolumeMounts[0].MountPath))
```
burgerdev, line 76: This needs to be done for all containers, right? E.g., when adding the service mesh but also dealing with some part of TLS manually...

internal/kuberesource/mutators.go
```diff
@@ -195,27 +231,33 @@ func PatchServiceMeshAdminInterface(resources []any, port int32) []any {
 }
 
 // MapPodSpec applies a function to a PodSpec in a Kubernetes resource.
-func MapPodSpec(resource any, f func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration) any {
+func MapPodSpec(
```
burgerdev, line 234: Have you considered adding a `MapPodSpecWithMeta` and implementing the existing `MapPodSpec` with it?

burgerdev (commented):
Thanks! Do you want to add the image replacements file to the released artifact in a future PR?

davidweisse:
> Thanks! Do you want to add the image replacements file to the released artifact in a future PR?

I'll add the changes to the release artifacts here. I'll probably open a separate PR on this branch.

internal/kuberesource/mutators.go
```diff
@@ -215,28 +250,41 @@ func PatchCoordinatorMetrics(resources []any, port int32) []any {
 	return resources
 }
 
-// MapPodSpec applies a function to a PodSpec in a Kubernetes resource.
-func MapPodSpec(resource any, f func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration) any {
+// MapPodSpecWithMeta applies a function to a PodSpec in a Kubernetes resource.
```
burgerdev, line 253: Nit: the meta part is not documented.

internal/kuberesource/mutators_test.go
burgerdev, line 1: Could you please add some tests for the edge cases:

* existing initializer is replaced
* existing volume is used

burgerdev (approved):

cli/cmd/generate.go
```diff
@@ -248,16 +261,67 @@ func patchTargets(paths []string, logger *slog.Logger) error {
...
+			}
+			kubeObjs[i] = resource
+		}
+
+		logger.Debug("Updating resources in yaml file", "path", path)
+		modifiedData, err := kuberesource.EncodeResources(kubeObjs...)
+		if err != nil {
+			return err
+		}
+		if err := os.WriteFile(path, modifiedData, os.ModePerm); err != nil {
```
katexochen, line 281: I think it would be better to only write those files once. Right now, we are writing here with images that don't make sense, and later we read those files again and patch the images. If the command fails in between, the user ends up with partially patched files that contain images that aren't recommended to use.

As discussed out of band, a solution might be to move the `injectInitializer` func into patchTargets.

This would also solve the current naming problem, as it is a bit unclear what `patchTargets` does as we also patch targets in `injectInitializer`.

cli/cmd/generate.go
```diff
@@ -248,16 +261,67 @@ func patchTargets(paths []string, logger *slog.Logger) error {
...
+			return fmt.Errorf("failed to write %s: %w", path, err)
```
katexochen, line 282:
```suggestion
			return fmt.Errorf("writing %s: %w", path, err)
```

cli/cmd/generate.go
```diff
@@ -248,16 +261,67 @@ func patchTargets(paths []string, logger *slog.Logger) error {
...
+	return nil
+}
+
+func patchTargets(paths []string, imageReplacementsFile string, logger *slog.Logger) error {
+	var replacements map[string]string
+	var err error
+	if imageReplacementsFile != "" {
+		f, err := os.Open(imageReplacementsFile)
+		if err != nil {
+			return fmt.Errorf("could not open image replacements file %s: %w", imageReplacementsFile, err)
```
katexochen, line 294:
```suggestion
			return fmt.Errorf("opening image replacements file %s: %w", imageReplacementsFile, err)
```
In general, we don't wrap erros with `failed to ...` or `could not ...`, as errors are usually wrapped multiple times and those phrases will stack up:

```
Error: failed to foo: failed to bar: could not open image replacements file "": could not read file: file does not exist
```
Only the inner error (like error creates with `errors.New`) should use failed.

docs/docs/components/service-mesh.md
```diff
@@ -56,19 +56,11 @@ spec:
...
-              value: coordinator
-          volumeMounts:
-            - name: tls-certs
-              mountPath: /tls-config
         - name: sidecar
           image: "ghcr.io/edgelesssys/contrast/service-mesh-proxy@sha256:..."
           restartPolicy: Always
           volumeMounts:
-            - name: tls-certs
+            - name: contrast-tls-certs-volume
```
katexochen, line 63:
```suggestion
            - name: contrast-tls-certs
```
I don't think the volume name should contain "volume".

docs/docs/deployment.md
```diff
@@ -185,6 +173,36 @@ contrast generate resources/
 Please be aware that runtime policies currently have some blind spots. For example, they can't guarantee the starting order of containers. See the [current limitations](known-limitations.md#runtime-policies) for more details.
 :::
 
+If you don't want the Contrast Initializer to automatically be added to your
+workloads, there are two ways you can skip the Initializer injection step,
+depending on how you want to customize your deployment.
+
+<Tabs groupId="initializer-injection">
+<TabItem value="flag" label="Command-Line Flag">
```
katexochen, line 181:
```suggestion
<TabItem value="flag" label="Command-line flag">
```

docs/docs/deployment.md
```diff
@@ -185,6 +173,36 @@ contrast generate resources/
...
+You can disable the Initializer injection completely by specifying the
+`--skip-initializer` flag in the `generate` command.
+
+```sh
+contrast generate --skip-initializer resources/
+```
+
+</TabItem>
+
+<TabItem value="annotation" label="Per-Workload Annotation">
```
katexochen, line 192:
```suggestion
<TabItem value="annotation" label="Per-workload annotation">
```

internal/kuberesource/mutators.go
```diff
@@ -5,45 +5,80 @@ package kuberesource
...
+		}
+
+		// Remove already existing init containers with unique initializer name.
+		spec.InitContainers = slices.DeleteFunc(spec.InitContainers, func(c applycorev1.ContainerApplyConfiguration) bool {
+			return *c.Name == *initializer.Name
+		})
+		// Add the initializer as first init container.
+		spec.InitContainers = append([]applycorev1.ContainerApplyConfiguration{*initializer}, spec.InitContainers...)
 		if len(initializer.VolumeMounts) < 1 {
 			return spec
```
katexochen, line 46: Maybe we should return an error in this unexpected case, and add a comment that this is rather a sanity check?

internal/kuberesource/mutators.go
```diff
@@ -5,45 +5,80 @@ package kuberesource
...
 func AddInitializer(
 	resource any,
 	initializer *applycorev1.ContainerApplyConfiguration,
-) any {
-	return MapPodSpec(resource, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
-		// Add the initializer as an init container.
-		spec.WithInitContainers(
-			initializer,
-		)
+) (res any, err error) {
```
katexochen, line 30:
```suggestion
) (res any, retErr error) {
```

katexochen (changes requested):

katexochen (approved):

PR #512, opened 2024-05-29
release: embed image replacements

davidweisse:
The image references for all generated components will be embedded in the CLI on release to be used by `contrast generate`.


.github/workflows/release.yml
```diff
@@ -244,10 +244,13 @@ jobs:
...
             base64 -d | sha256sum | cut -d " " -f1 > cli/cmd/assets/coordinator-policy-hash
 
+          cp image-replacements.txt cli/cmd/assets/image-replacements.txt
+
           git config --global user.name "edgelessci"
           git config --global user.email "edgelessci@users.noreply.github.com"
           git add cli/cmd/assets/coordinator-policy-hash
-          git diff --staged --quiet || git commit -m "release: update coordinator policy hash"
+          git add cli/cmd/assets/image-replacements.txt
+          git diff --staged --quiet || git commit -m "release: update coordinator policy hash and embed image replacements"
```
burgerdev, line 253:
```suggestion
          git diff --staged --quiet || git commit -m "release: update cli assets"
```
... in anticipation of other future assets.

cli/cmd/generate.go
```diff
@@ -263,6 +264,11 @@ func injectInitializer(paths []string, imageReplacementsFile string, logger *slo
 		if err != nil {
 			return fmt.Errorf("could not parse image definition file %s: %w", imageReplacementsFile, err)
 		}
+	} else {
+		replacements, err = kuberesource.ImageReplacementsFromFile(bytes.NewReader(releaseImageReplacements))
```
burgerdev, line 268: This is now always going to fail for dev builds (because of the embedded bogus file). Is that intentional? If not, you could prefix the placeholder text with `#` to just yield no replacements.

packages/by-name/cli-release/package.nix
```diff
@@ -12,6 +12,7 @@
     install -D ${lib.getExe genpolicy} cli/cmd/assets/genpolicy
     install -D ${genpolicy.settings}/genpolicy-settings.json cli/cmd/assets/genpolicy-settings.json
     install -D ${genpolicy.rules}/genpolicy-rules.rego cli/cmd/assets/genpolicy-rules.rego
+    echo $"${builtins.readFile ../../../cli/cmd/assets/image-replacements.txt}" > cli/cmd/assets/image-replacements.txt
```
burgerdev, line 15: Are the nix changes necessary? The file should be right there to begin with...

davidweisse: The file is only there locally and not inside the nix build, as I understand it. That's why I have to `touch` the file for dev builds, resulting in an empty image replacement file. Dev builds will not fail, referring to your other comment, because the file is then completely empty.

burgerdev (approved):

PR #514, opened 2024-05-31
e2e: test verify with root CA after cert rotation

burgerdev:
This adds 

* a new test to verify that the root CA is still accepted after certificate rotation. 
* another run of the certificate rotation test that restarts the other deployment

I want these tests to prepare for RFC004 changes to meshauth/ca.

To make these changes easier, I introduced

* a helper function to exec in a random pod of a deployment
* a helper function to construct `s_client` arguments

While I was looking into the test, I noticed that there are a couple of leftovers in the OpenSSL deployment originating from when we used it for manual testing. I don't think we need the OpenSSL pods to communicate without user interaction - Emojivoto should be sufficient to test this kind of thing.

### DIFF
--- a/internal/kuberesource/sets.go
+++ b/internal/kuberesource/sets.go
@@ -55,10 +55,10 @@ func OpenSSL() []any {
 						Container().
 							WithName("openssl-backend").
 							WithImage("ghcr.io/edgelesssys/contrast/openssl:latest").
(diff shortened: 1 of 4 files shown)


internal/kuberesource/sets.go
```diff
@@ -79,30 +79,6 @@ func OpenSSL() []any {
 
 	backendService := ServiceForDeployment(backend)
 
-	client := Deployment("openssl-client", ns).
```
burgerdev, line 82: This removal needs to be reflected in `just wait-for-workload`.

3u13r (approved):
LGTM (if the just target has been also adjusted)

PR #515, opened 2024-06-01
docs: better error message when forgetting to set variables

blenessy:
When doing cluster setup from start you will set the variables. However, if you revisit the guide later and forget to set the variables in question you might get weird (missleading) az error messages like: The api-version '2024-03-02-preview' is invalid.

This will probably save you guys some support issues


katexochen:
Thanks, I've also added these changes to the docs of the current version.

katexochen (approved):

PR #520, opened 2024-06-03
coordinator: don't share CA instance with gRPC servers

burgerdev:
At a high-level, this PR

* moves and renames the `meshAuthority` to `authority.Authority`
* takes the permanent `ca.CA` instance away from the gRPC servers

The idea is that the methods called by the gRPC servers on the CA instance are always coupled to the current active manifest, thus getting the manifest from `meshAuthority` and the certificates from `CA` is inherently racy (think of concurrent calls to `SetManifest` and `NewMeshCert`, for example). This is a first step towards tighter coupling of the two, by handing out CA objects when required and removing the permanent members.

In the meshAPI server, I'm moving all certificate operations to the `ValidateCallback`, and all certificate operations happen on the same CA instance. This does not make a practical difference now, but allows to swap out the permanent CA object with one per manifest generation.

In the userAPI server, I'm adding the CA as a return to the manifest retrieval operation, also in preparation to swap it out in the future.

The API of the `Authority` should be considered transitory while we're working on the persistence. I could imagine it returning a struct with manifest and CA combined.

The race conditions are still present (see associated TODOs), but it should be easier to mitigate them going forward.


coordinator/internal/authority/authority.go
```diff
@@ -23,24 +23,35 @@ import (
...
+	return &Authority{
+		ca:        caInstance,
+		bundles:   make(map[string]Bundle),
 		manifests: new(appendable.Appendable[*manifest.Manifest]),
 		logger:    log.WithGroup("mesh-authority"),
 	}
 }
 
-func (m *meshAuthority) SNPValidateOpts(report *sevsnp.Report) (*validate.Options, error) {
+// SNPValidateOpts verifies that the report HOSTDATA field is registered in the active manifest.
```
katexochen, line 53: That's wrong, it returns any value for validation _beside HOSTDATA_, which is validated through the ValidateCallback.

coordinator/meshapi.go
```diff
@@ -20,27 +24,26 @@ import (
...
 )
 
 type meshAPIServer struct {
-	grpc          *grpc.Server
-	certGet       certGetter
-	caChainGetter certChainGetter
-	ticker        clock.Ticker
-	logger        *slog.Logger
+	grpc   *grpc.Server
+	getter certBundleGetter
```
katexochen, line 35: I'd say `getter` is a bit unspecific...

coordinator/userapi.go
```diff
@@ -252,15 +252,9 @@ func manifestSliceToBytesSlice(s []*manifest.Manifest) ([][]byte, error) {
...
-type certChainGetter interface {
-	GetRootCACert() []byte
-	GetMeshCACert() []byte
-	GetIntermCACert() []byte
-}
-
 type manifestSetGetter interface {
 	SetManifest(*manifest.Manifest) error
-	GetManifests() []*manifest.Manifest
+	GetManifests() ([]*manifest.Manifest, *ca.CA)
```
katexochen, line 257: Rename it to reflect the change? `GetManifestAndActiveCA` ?

katexochen (approved):

coordinator/internal/authority/authority.go
burgerdev: Well, it's not _wrong_: https://github.com/edgelesssys/contrast/blob/de5e0f93f2ba7f0f00ae1f2c79779dbf48a8b55a/coordinator/mesh.go#L94-L98.

But I agree that it does not capture all the function does and adjusted the docs.

PR #522, opened 2024-06-04
coordinator: implement history with fs store backend

katexochen:
This is the history storage module and part of the implementation of RFC 004.

### DIFF
--- a/coordinator/history/fsstore.go
+++ b/coordinator/history/fsstore.go
@@ -1,0 +1,56 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package history
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io/fs"
+	"path/filepath"
+	"sync"
+
+	"github.com/spf13/afero"
+)
+
+type fsStore struct {
+	fs  *afero.Afero
+	mux sync.RWMutex
+}
+
+func newPVStore(fs *afero.Afero) *fsStore {
+	return &fsStore{fs: fs}
+}
+
+func (s *fsStore) Get(key string) ([]byte, error) {
+	s.mux.RLock()
+	defer s.mux.RUnlock()
+	return s.fs.ReadFile(key)
+}
+
+func (s *fsStore) Set(key string, value []byte) error {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	if err := s.fs.MkdirAll(filepath.Base(key), 0o755); err != nil {
+		return fmt.Errorf("creating directory for %q: %w", key, err)
+	}
+	return s.fs.WriteFile(key, value, 0o644)
+}
+
+func (s *fsStore) CompareAndSwap(key string, oldVal, newVal []byte) error {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	current, err := s.fs.ReadFile(key)
+	// Treat non-existing file as empty to allow initial set.
+	if err != nil && !(errors.Is(err, fs.ErrNotExist) && len(oldVal) == 0) {
+		return err
+	}
+	if !bytes.Equal(current, oldVal) {
+		return fmt.Errorf("object %q has changed since last read", key)
+	}
+	if err := s.fs.MkdirAll(filepath.Base(key), 0o755); err != nil {
+		return fmt.Errorf("creating directory for %q: %w", key, err)
+	}
+	return s.fs.WriteFile(key, newVal, 0o644)
+}
--- a/coordinator/history/history.go
+++ b/coordinator/history/history.go
@@ -1,0 +1,231 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package history
+
+import (
+	"bytes"
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"hash"
+
+	"github.com/spf13/afero"
+)
+
+const (
+	hashSize = 32 // byte, History.hashFun().Size()
+	histPath = "/mnt/state/history"
+)
+
+// History is the history of the Coordinator.
+type History struct {
+	store      store
+	hashFun    func() hash.Hash
+	signingKey *ecdsa.PrivateKey
+}
+
+// New creates a new History with the given signing key.
+func New() (*History, error) {
+	osFS := afero.NewOsFs()
+	if err := osFS.MkdirAll(histPath, 0o755); err != nil {
+		return nil, fmt.Errorf("creating history directory: %w", err)
+	}
+	h := &History{
+		store:   newPVStore(&afero.Afero{Fs: afero.NewBasePathFs(osFS, histPath)}),
+		hashFun: sha256.New,
+	}
+	if hashSize != h.hashFun().Size() {
+		return nil, errors.New("mismatch between hashSize and hash function size")
+	}
+	return h, nil
+}
+
+// ConfigureSigningKey sets the signing key for validation and signing of the protected history parts.
+func (h *History) ConfigureSigningKey(signingKey *ecdsa.PrivateKey) {
+	h.signingKey = signingKey
+}
+
+// GetManifest returns the manifest for the given hash.
+func (h *History) GetManifest(hash [hashSize]byte) ([]byte, error) {
+	return h.getContentaddressed("manifests/%s", hash)
+}
+
+// SetManifest sets the manifest and returns its hash.
+func (h *History) SetManifest(manifest []byte) ([hashSize]byte, error) {
+	return h.setContentaddressed("manifests/%s", manifest)
+}
+
+// GetPolicy returns the policy for the given hash.
+func (h *History) GetPolicy(hash [hashSize]byte) ([]byte, error) {
+	return h.getContentaddressed("policies/%s", hash)
+}
+
+// SetPolicy sets the policy and returns its hash.
+func (h *History) SetPolicy(policy []byte) ([hashSize]byte, error) {
+	return h.setContentaddressed("policies/%s", policy)
+}
+
+// GetTransition returns the transition for the given hash.
+func (h *History) GetTransition(hash [hashSize]byte) (*Transition, error) {
+	transitionBytes, err := h.getContentaddressed("transitions/%s", hash)
+	if err != nil {
+		return nil, err
+	}
+	var transition Transition
+	if err := transition.unmarshalBinary(transitionBytes); err != nil {
+		return nil, fmt.Errorf("unmarshaling transition: %w", err)
+	}
+	return &transition, nil
+}
+
+// SetTransition sets the transition and returns its hash.
+func (h *History) SetTransition(transition *Transition) ([hashSize]byte, error) {
+	return h.setContentaddressed("transitions/%s", transition.marshalBinary())
+}
+
+// GetLatest returns the verified transition for the given hash.
+func (h *History) GetLatest() (*LatestTransition, error) {
+	if h.signingKey == nil {
+		return nil, errors.New("signing key not configured")
+	}
+	transitionBytes, err := h.store.Get("transitions/latest")
+	if err != nil {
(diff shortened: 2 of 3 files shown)


coordinator/history/history.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (l *LatestTransition) verify(key *ecdsa.PublicKey) error {
+	if !ecdsa.VerifyASN1(key, l.TransitionHash[:], l.signature) {
+		return errors.New("latest transition signature is invalid")
+	}
+	return nil
+}
+
+// HashMissmatchError is returned when a hash does not match the expected value.
+type HashMissmatchError struct {
```
burgerdev, line 201: typo: `HashMismatchError`

coordinator/history/fsstore.go
```diff
@@ -0,0 +1,39 @@
...
+func (s *fsStore) Get(key string) ([]byte, error) {
+	return s.fs.ReadFile(key)
+}
+
+func (s *fsStore) Set(key string, value []byte) error {
+	return s.fs.WriteFile(key, value, 0o644)
+}
+
+func (s *fsStore) CompareAndSwap(key string, oldVal, newVal []byte) error {
+	// Not atomic at all!
```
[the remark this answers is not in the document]
katexochen: Must be added to all of the funcs, done.

coordinator/history/history.go
```diff
@@ -0,0 +1,214 @@
...
+	if err := newT.sign(h.signingKey); err != nil {
+		return fmt.Errorf("signing latest transition: %w", err)
+	}
+	if err := h.store.CompareAndSwap("transitions/latest", oldT.marshalBinary(), newT.marshalBinary()); err != nil {
+		return fmt.Errorf("setting latest transition: %w", err)
+	}
+	return nil
+}
+
+func (h *History) getCA(hash [hashSize]byte, pathFmt string) ([]byte, error) {
```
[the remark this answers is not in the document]
katexochen: xD that's because you are working on the other CA thingy...

coordinator/history/fsstore.go
```diff
@@ -0,0 +1,39 @@
...
+	return &fsStore{fs: fs}
+}
+
+func (s *fsStore) Get(key string) ([]byte, error) {
+	return s.fs.ReadFile(key)
+}
+
+func (s *fsStore) Set(key string, value []byte) error {
+	return s.fs.WriteFile(key, value, 0o644)
+}
```
burgerdev, line 27: Do we need to `MkdirAll` at some point?

katexochen: Splitted this between constructor and set. Maybe I should also write some unit tests for the fsstore...

coordinator/history/fsstore.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := s.fs.MkdirAll(filepath.Base(key), 0o755); err != nil {
+		return fmt.Errorf("creating directory for %q: %w", key, err)
+	}
+	return s.fs.WriteFile(key, value, 0o644)
+}
+
+func (s *fsStore) CompareAndSwap(key string, oldVal, newVal []byte) error {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	current, err := s.Get(key)
```
3u13r, line 42: The `s.Get()` call also wants to acquire the same mutex. The easiest solution would be to inline the `s.Get()` and `s.Set()` call below (minus their mutex handling).

coordinator/history/history.go
```diff
@@ -0,0 +1,227 @@
...
+	osFS := afero.NewOsFs()
+	if err := osFS.MkdirAll(histPath, 0o755); err != nil {
+		panic(fmt.Errorf("creating history directory: %w", err))
+	}
+	h := &History{
+		store:   newPVStore(&afero.Afero{Fs: afero.NewBasePathFs(osFS, histPath)}),
+		hashFun: sha256.New,
+	}
+	if hashSize != h.hashFun().Size() {
+		panic("mismatch between hashSize and hash function size")
```
3u13r, line 42: Out of curiosity, what is the tradeoff between panic and erroring here? Is having a New function that can return an error discouraged practice?

coordinator/history/history.go
```diff
@@ -0,0 +1,227 @@
...
+	if err := newT.sign(h.signingKey); err != nil {
+		return fmt.Errorf("signing latest transition: %w", err)
+	}
+	if err := h.store.CompareAndSwap("transitions/latest", oldT.marshalBinary(), newT.marshalBinary()); err != nil {
+		return fmt.Errorf("setting latest transition: %w", err)
+	}
+	return nil
+}
+
+func (h *History) getContentaddressed(hash [hashSize]byte, pathFmt string) ([]byte, error) {
```
3u13r, line 123: I think it is common practice for functions that take a format specifier, to have it in the first place.
Or did you tried to make the difference between the set and get version clearer?

coordinator/history/history.go
```diff
@@ -0,0 +1,227 @@
...
+	hashStr := hex.EncodeToString(hash[:])
+	data, err := h.store.Get(fmt.Sprintf(pathFmt, hashStr))
+	if err != nil {
+		return nil, err
+	}
+	dataHash := h.hash(data)
+	if !bytes.Equal(hash[:], dataHash[:]) {
+		return nil, HashMismatchError{Expected: hash[:], Actual: dataHash[:]}
```
3u13r, line 131: This only occurs if the CAS has been corrupted, right? Should we spell this out somewhere, either in the error or at least in a comment above this if?

coordinator/history/history_test.go
```diff
@@ -0,0 +1,524 @@
...
+			transition: Transition{
+				ManifestHash:           strToHash(require.New(t), "2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824"),
+				PreviousTransitionHash: strToHash(require.New(t), "486ea46224d1bb4fb680f34f7c9ad96a8f24ec88be73ea8e5a6c65260e9cb8a7"),
+			},
+			wantHash: "7305db9b2abccd706c256db3d97e5ff48d677cfe4d3a5904afb7da0e3950e1e2",
+			wantFSContent: map[string]string{
+				"transitions/7305db9b2abccd706c256db3d97e5ff48d677cfe4d3a5904afb7da0e3950e1e2": "2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824486ea46224d1bb4fb680f34f7c9ad96a8f24ec88be73ea8e5a6c65260e9cb8a7",
+			},
+		},
+		"ojbect exists": {
```
3u13r, line 242:
```suggestion
		"object exists": {
```

coordinator/history/history.go
3u13r, line 1: I wonder if this is the package where we should implement the state transition in the sense that this package exposes some getter and only one `AdvanceState` function. This function would also receive all policies + the manifest needed for the state transition. Otherwise we would e.g. need to first put the policies into the CAS, before transitioning the state, to ensure that if the coordinator crashes during a transition, the latest manifest has all policies in the CAS.

3u13r (changes requested):

coordinator/history/history.go
burgerdev: There's no meaningful way to handle this error - the package is just broken if this assertion does not hold. With that in mind, I appreciate that panicking gives us an error-return-free `New` function.

coordinator/history/history.go
burgerdev: @katexochen and I discussed that yesterday, and concluded that the right place to do this is in the [`Authority`](https://github.com/edgelesssys/contrast/pull/520/files#diff-ada64dadec60f0da40372b12c7453087f563cece0b121f6158efa005dd024af1). The history struct manages the content-addressing and layout on top of a simple KV engine.

coordinator/history/history.go
katexochen: It is discouraged to panic, that's right. Actually, as there is now a second function that might fail, which isn't a sanity check, we should return the error (sorry @burgerdev :disappointed: , I like constructors without error as well. )

coordinator/history/history.go
```diff
@@ -0,0 +1,227 @@
...
+func (h *History) SetTransition(transition Transition) ([hashSize]byte, error) {
+	return h.setContentaddressed("transitions/%s", transition.marshalBinary())
+}
+
+// GetLatest returns the verified transition for the given hash.
+func (h *History) GetLatest() (LatestTransition, error) {
+	if h.signingKey == nil {
+		return LatestTransition{}, errors.New("signing key not configured")
+	}
+	transitionBytes, err := h.store.Get("transitions/latest")
```
burgerdev, line 95: I believe we forgot to specify that `GetLatest` needs to respect the mutual exclusion properties of `SetLatest`, so that we don't end up reading a half-written file :/

coordinator/history/history.go
```diff
@@ -0,0 +1,227 @@
...
+	hashStr := hex.EncodeToString(hash[:])
+	if err := h.store.Set(fmt.Sprintf(pathFmt, hashStr), data); err != nil {
+		return [hashSize]byte{}, err
+	}
+	return hash, nil
+}
+
+func (h *History) hash(in []byte) [hashSize]byte {
+	hf := h.hashFun()
+	_, _ = hf.Write(in)
```
burgerdev, line 147:
```suggestion
	_, _ = hf.Write(in) // Hash.Write never returns an error.
```

coordinator/history/history.go
katexochen: That should be given by the store's mutex, right?

burgerdev: We could just require the store interface to behave well on concurrent Set and Get, and implement Set by writing to a random file and then atomically `syscall.Rename` it.

burgerdev: > That should be given by the store's mutex, right?

Sorry, missed that.

coordinator/history/history.go
```diff
@@ -0,0 +1,228 @@
...
+
+// SetLatest signs and sets the latest transition if the current latest is equal to oldT.
+func (h *History) SetLatest(oldT, newT LatestTransition) error {
+	if h.signingKey == nil {
+		return errors.New("signing key not configured")
+	}
+	if err := newT.sign(h.signingKey); err != nil {
+		return fmt.Errorf("signing latest transition: %w", err)
+	}
+	if err := h.store.CompareAndSwap("transitions/latest", oldT.marshalBinary(), newT.marshalBinary()); err != nil {
```
burgerdev, line 117: How do we set the first `latest`? If I pass an empty `oldT`, it marshals to 64 zero bytes, but the `CompareAndSwap` compares it with an empty slice.

katexochen: Any ideas?

burgerdev: Accept `oldT` as pointer and treat `nil` as "must be empty"?

coordinator/history/history.go
```diff
@@ -0,0 +1,231 @@
...
+	return &transition, nil
+}
+
+// SetTransition sets the transition and returns its hash.
+func (h *History) SetTransition(transition *Transition) ([hashSize]byte, error) {
+	return h.setContentaddressed("transitions/%s", transition.marshalBinary())
+}
+
+// GetLatest returns the verified transition for the given hash.
+func (h *History) GetLatest() (*LatestTransition, error) {
```
burgerdev, line 91: While implementing I realized that I need to be able to tell the difference between generic errors and "no latest set".

burgerdev: The issue is that I'd like to update my local state from persistency before handling a SetManifest, when I don't yet have something to CompareAndSet. But I can work around that by ignoring errors when my local state suggests emptiness - I will be stopped at the very latest when my update conflicts. So this would be a good extension, but not necessary right now.

burgerdev (approved):

coordinator/history/history.go
katexochen: use `errors.Is(err, os.ErrNotExist)` ?

burgerdev: I can do that for now, but it's somewhat counter to the idea of having pluggable KV stores :)

katexochen: True.

3u13r (approved):
LGTM

PR #529, opened 2024-06-05
cli: inject service mesh

davidweisse:
The service mesh sidecar will be injected to every workload with the `contrast-cc` runtime class. This behavior can be disabled with a command-line flag or an annotation, similar to the initializer injection. Configuration can be done via object annotations:

- `contrast.edgeless.systems/sm-ingress`: Envoy proxy ingress config
- `contrast.edgeless.systems/sm-egress`: Envoy proxy egress config
- `contrast.edgeless.systems/sm-admin-interface-port`: Port for the Envoy admin interface

The annotations will be copied to the corresponding environment variables during `generate`.


rfc/005-generate-injection.md
```diff
@@ -50,7 +50,17 @@ To make `contrast generate` idempotent and handle potential edge cases, we will
 
 ### Service Mesh Injection
 
-The service mesh injection will follow a similar approach to the initializer injection outlined, possibly with additional configuration for labels for service discovery/ports. Details of the service mesh injection will be expanded in a future PR update to this RFC.
+The service mesh injection will follow a similar approach to the initializer injection outlined.
+During the generation process, all containers with the `contrast-cc` runtime class will have a service mesh added as a sidecar init container.
```
burgerdev, line 54: I'm unsure whether service mesh should be enabled by default. Unlike the initializer, without which the deployment is somewhat pointless, I can live without a service mesh quite fine. How about making this opt-in until we have sufficient feedback from users that this should be default? We could also decide whether to inject based on presence of the annotations. cc @katexochen

rfc/005-generate-injection.md
```diff
@@ -50,7 +50,17 @@ To make `contrast generate` idempotent and handle potential edge cases, we will
...
+This behavior can be disabled with the command-line flag `--skip-servicemesh` or the per-workload annotation `contrast.edgeless.systems/skip-servicemesh` set to `true`.
+
+The configuration for the Envoy proxy is handled via Kubernetes object annotations.
+The annotations `contrast.edgeless.systems/sm-ingress`, `contrast.edgeless.systems/sm-egress` and `contrast.edgeless.systems/sm-admin-interface-port` will be written into the environment variables `EDG_INGRESS_PROXY_CONFIG`, `EDG_EGRESS_PROXY_CONFIG` and `EDG_ADMIN_PORT` of the injected service mesh container.
```
burgerdev, line 58: Please spell out `servicemesh` in these annotations, this will make the purpose obvious even for someone unfamiliar with the topic.

internal/kuberesource/sets.go
```diff
@@ -154,10 +156,7 @@ func GetDEnts() ([]any, error) {
 // Emojivoto returns resources for deploying Emojivoto application.
 func Emojivoto(smMode serviceMeshMode) []any {
 	ns := ""
-	var emojiSvcImage, emojiVotingSvcImage, emojiWebImage, emojiWebVoteBotImage, emojiSvcHost, votingSvcHost string
-	smProxyEmoji := ServiceMeshProxy()
-	smProxyWeb := ServiceMeshProxy()
-	smProxyVoting := ServiceMeshProxy()
+	var emojiSvcImage, emojiVotingSvcImage, emojiWebImage, emojiWebVoteBotImage, emojiSvcHost, votingSvcHost, smWebIngress, smWebEgress string
 	switch smMode {
 	case ServiceMeshDisabled:
```
burgerdev, line 161: Don't we need a `skip-servicemesh` annotation here?

rfc/005-generate-injection.md
davidweisse: We would also need some form of configuration to inject a service mesh with default configuration, i.e. no ingress or egress. Maybe just include an annotation but leave it empty?

burgerdev: A service mesh without ingress or egress means ... no service mesh? I'd consider "the default service mesh" to be no ingress exceptions and no egress setup, which is what you get when the sidecar is there but the env vars are empty.

davidweisse: Yes, I meant no special ingress or egress configuration, so the default behavior of the service mesh. If we want to inject the service mesh based on the presence of annotations, we would need a way to tell the CLI to generate a service mesh, even if you don't specify ingress/egress in the annotations.

burgerdev: There's a difference between the annotation being missing, or present but empty. I think taking the existence of one of the annotations as clue for injection would just work. 

There's another discussion about how to disable ingress completely while having egress, but that's unrelated to your proposal.

internal/kuberesource/mutators.go
```diff
@@ -95,17 +85,84 @@ const (
...
-// This function is not idempotent.
+// This function is idempotent.
 func AddServiceMesh(
 	resource any,
 	serviceMeshProxy *applycorev1.ContainerApplyConfiguration,
-) any {
-	return MapPodSpec(resource, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+) (res any, retErr error) {
+	res = MapPodSpecWithMeta(resource, func(meta *applymetav1.ObjectMetaApplyConfiguration, spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+		if spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
```
burgerdev, line 98: Not for you to fix here, but a general observation: we should make `contrast-cc` a constant and replace all the string literals.

internal/kuberesource/mutators.go
```diff
@@ -95,17 +85,84 @@ const (
...
+		}
+
+		if portAnnotation != "" {
+			port, err := strconv.Atoi(portAnnotation)
+			if retErr != nil {
+				retErr = fmt.Errorf("parsing service mesh admin interface port: %w", err)
+				return nil
+			}
+
+			ingressConfig += fmt.Sprintf("##admin#%s#true", portAnnotation)
```
burgerdev, line 128: This behavior should be documented. And if we're not 100% sure that this is always the correct config, we should not do it at all.

internal/kuberesource/mutators.go
```diff
@@ -95,17 +85,84 @@ const (
...
+		})
+
+		retErr = ensureVolumeExists(spec, *serviceMeshProxy.VolumeMounts[0].Name)
+		if retErr != nil {
+			return nil
+		}
+
+		if portAnnotation != "" {
+			port, err := strconv.Atoi(portAnnotation)
+			if retErr != nil {
```
burgerdev, line 123: Did you mean to write
```suggestion
			if err != nil {
```

internal/kuberesource/mutators.go
```diff
@@ -95,17 +85,84 @@ const (
...
+				retErr = fmt.Errorf("parsing service mesh admin interface port: %w", err)
+				return nil
+			}
+
+			ingressConfig += fmt.Sprintf("##admin#%s#true", portAnnotation)
+			serviceMeshProxy.
+				WithEnv(NewEnvVar("EDG_ADMIN_PORT", portAnnotation)).
+				WithPorts(
+					ContainerPort().
+						WithName("admin-interface").
```
burgerdev, line 133: Maybe more specific? The container might have its own admin interface.

```suggestion
						WithName("contrast-admin-interface").
```

internal/kuberesource/mutators.go
```diff
@@ -201,38 +258,18 @@ func PatchNamespaces(resources []any, namespace string) []any {
...
-						r.Spec.Template.Spec.InitContainers[i].WithEnv(
-							NewEnvVar("EDG_INGRESS_PROXY_CONFIG", fmt.Sprintf("admin#%d#true", port)),
-						)
-					}
-				}
+		out = append(out, MapPodSpecWithMeta(resource, func(meta *applymetav1.ObjectMetaApplyConfiguration, spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+			_, ingressOk := meta.Annotations[smIngressConfigAnnotationKey]
+			_, egressOk := meta.Annotations[smEgressConfigAnnotationKey]
+			if ingressOk || egressOk {
+				meta.WithAnnotations(map[string]string{smAdminInterfaceAnnotationKey: fmt.Sprint(port)})
```
burgerdev, line 267: I'm not too happy that this relies on the side effect that the input is mutable. This is also true for most other functions in this package, which claim to return an object but actually do updates in-place, too. But here it's a bit too much for my taste, given that `meta` is not even an output.

internal/kuberesource/sets.go
```diff
@@ -470,15 +445,12 @@ func Emojivoto(smMode serviceMeshMode) []any {
...
-	}
-	if smProxyVoting != nil {
-		AddServiceMesh(voting, smProxyVoting)
-	}
+	emoji.WithAnnotations(map[string]string{smIngressConfigAnnotationKey: ""})
+	voting.WithAnnotations(map[string]string{smIngressConfigAnnotationKey: ""})
+	web.WithAnnotations(map[string]string{
+		smIngressConfigAnnotationKey: smWebIngress,
+		smEgressConfigAnnotationKey:  smWebEgress,
+	})
```
burgerdev, line 453: This should be skipped in case `ServiceMeshDisabled`, shouldn't it?

internal/kuberesource/mutators.go
davidweisse: We could update the `MapPodSpecWithMeta` function to also update the ObjectMeta and not only taking it as an input.

internal/kuberesource/sets.go
davidweisse: In this case, the resources are returned above without the change.

internal/kuberesource/mutators.go
davidweisse: I think we should leave this out. How to access the interface is already documented on the Observability page, including setting up the required ingress.

burgerdev (approved):

PR #535, opened 2024-06-06
coordinator: usability improvements to the history package

burgerdev:
* Export hash size because it is part of the API (via hashRef arguments).
* Export `store` and allow it as an argument for creating `History` (useful for `authority` tests).
* Same for the fs-based implementation.
* Add tests for AferoStore and fix a bug in the implementation (when run on `OsFs`).


katexochen (approved):
Nice!

PR #542, opened 2024-06-07
feat: also print launch digest with version information

miampf:
Add the printing of the launch digest to the `--version`/`-v` flag.


packages/by-name/contrast/package.nix
```diff
@@ -73,11 +73,28 @@ buildGoModule rec {
...
+
   CGO_ENABLED = 0;
   ldflags = [
     "-s"
     "-w"
     "-X main.version=v${version}"
+    "-X main.runtimeHandler=${runtimeHandler}"
+    "-X main.launchDigest=${launchDigest}"
+    "-X main.genpolicyVersion=${genpolicy.version}"
+    "-X main.containerVersions=${imageVersionsString}"
```
davidweisse, line 97: The image replacements are already embedded in the CLI in the `cmd` package. We can just make it public and use that.

cli/main.go
```diff
@@ -26,13 +27,32 @@ func execute() error {
...
 
 func newRootCmd() *cobra.Command {
+	// build the versions string
+	var versionsBuilder strings.Builder
+	versionsBuilder.WriteString(fmt.Sprintf("%s:\n\n", version))
+	versionsBuilder.WriteString(fmt.Sprintf("\truntime handler: %s\n", runtimeHandler))
+	versionsBuilder.WriteString(fmt.Sprintf("\tlaunch digest: %s\n", launchDigest))
+	versionsBuilder.WriteString(fmt.Sprintf("\tgenpolicy version: %s\n", genpolicyVersion))
+	for _, container := range strings.Split(containerVersions, ",") {
+		containerSplit := strings.Split(container, "@")
```
davidweisse, line 46: The released images are also tagged with a version, so the images are of the form `.../<container-name>:v1.2.3@sha256:abc...`, so in that case I would split at the first `:`.

cli/main.go
```diff
@@ -26,13 +27,32 @@ func execute() error {
...
+		containerName := containerSplit[0]
+		containerVersion := containerSplit[1]
+		versionsBuilder.WriteString(fmt.Sprintf("\t%s: %s\n", containerName, containerVersion))
```
davidweisse, line 49: I think we can print the container without splitting it, because we are effectively only adding a space. Maybe add a line saying `image versions:` above and just print all the images?

davidweisse (commented):
As we are now printing the runtime handler in the `version` command, there is no need for the additional `runtime` command. I think we can remove it in a separate PR.

cli/main.go
miampf: Since images are now simply printed, I ignored that.

cli/main.go
```diff
@@ -26,13 +27,33 @@ func execute() error {
...
+	versionsBuilder.WriteString(fmt.Sprintf("\truntime handler: %s\n", runtimeHandler))
+	versionsBuilder.WriteString(fmt.Sprintf("\tlaunch digest: %s\n", launchDigest))
+	versionsBuilder.WriteString(fmt.Sprintf("\tgenpolicy version: %s\n", genpolicyVersion))
+	versionsBuilder.WriteString("\timage versions:\n")
+	imageReplacements := strings.Trim(string(cmd.ReleaseImageReplacements), "\n")
+	for _, image := range strings.Split(imageReplacements, "\n") {
+		if !strings.HasPrefix(image, "#") {
+			image = strings.Split(image, "=")[1]
+		}
+		versionsBuilder.WriteString(fmt.Sprintf("\t\t%s\n", image))
```
davidweisse, line 50:
```suggestion
		if !strings.HasPrefix(image, "#") {
			image = strings.Split(image, "=")[1]
			versionsBuilder.WriteString(fmt.Sprintf("\t\t%s\n", image))
		}
```

That way, we don't get the placeholder text from the file in dev builds.

davidweisse (approved):

cli/main.go
```diff
@@ -26,13 +27,33 @@ func execute() error {
...
+	versionsBuilder.WriteString(fmt.Sprintf("%s:\n\n", version))
+	versionsBuilder.WriteString(fmt.Sprintf("\truntime handler: %s\n", runtimeHandler))
+	versionsBuilder.WriteString(fmt.Sprintf("\tlaunch digest: %s\n", launchDigest))
+	versionsBuilder.WriteString(fmt.Sprintf("\tgenpolicy version: %s\n", genpolicyVersion))
+	versionsBuilder.WriteString("\timage versions:\n")
+	imageReplacements := strings.Trim(string(cmd.ReleaseImageReplacements), "\n")
+	for _, image := range strings.Split(imageReplacements, "\n") {
+		if !strings.HasPrefix(image, "#") {
+			image = strings.Split(image, "=")[1]
+			versionsBuilder.WriteString(fmt.Sprintf("\t\t%s\n", image))
```
katexochen, line 49:
```suggestion
			versionsBuilder.WriteString(fmt.Sprintf("  %s\n", image))
```

cli/main.go
```diff
@@ -26,13 +27,33 @@ func execute() error {
...
+	version          = "0.0.0-dev"
+	runtimeHandler   = "contrast-cc"
+	launchDigest     = "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
+	genpolicyVersion = "0.0.0-dev"
+)
 
 func newRootCmd() *cobra.Command {
+	// build the versions string
+	var versionsBuilder strings.Builder
+	versionsBuilder.WriteString(fmt.Sprintf("%s:\n\n", version))
```
katexochen, line 40:
```suggestion
	versionsBuilder.WriteString(fmt.Sprintf("%s\n\n", version))
```

katexochen (changes requested):
Ideally, we would use something like a tab writer to produce some nice columns.

katexochen (approved):

PR #547, opened 2024-06-10
cli: error checking for struct fields

davidweisse:
During `contrast generate`, the parsed YAML file is not guaranteed to be valid. This checks for all struct fields that are necessary for the generation process if they are nil or not, to avoid runtime panics on an invalid YAML file.

### DIFF
--- a/internal/kuberesource/mutators.go
+++ b/internal/kuberesource/mutators.go
@@ -42,7 +42,7 @@ func AddInitializer(
 
 		// Remove already existing init containers with unique initializer name.
 		spec.InitContainers = slices.DeleteFunc(spec.InitContainers, func(c applycorev1.ContainerApplyConfiguration) bool {
-			return *c.Name == *initializer.Name
+			return c.Name != nil && *c.Name == *initializer.Name
 		})
 		// Add the initializer as first init container.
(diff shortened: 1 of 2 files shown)


internal/kuberesource/mutators.go
```diff
@@ -298,33 +324,73 @@ func MapPodSpecWithMeta(
 		meta *applymetav1.ObjectMetaApplyConfiguration,
 		spec *applycorev1.PodSpecApplyConfiguration,
 	) (*applymetav1.ObjectMetaApplyConfiguration, *applycorev1.PodSpecApplyConfiguration),
-) any {
+) (any, error) {
```
burgerdev, line 327: Did you consider skipping the invalid cases instead of returning an error?

davidweisse: It turns out, genpolicy also checks these fields, so we could just ignore invalid cases. The program would not panic and genpolicy will show a reasonable error message with `missing field 'spec'` and a path like `spec.template.spec`.

burgerdev (approved):
LGTM, thanks for rewriting!

katexochen:
Should we lable this as bug fix? Or put it otherwise in the changelog, I think it is a nice improvement we don't panic on unexpected yaml anymore. :)

PR #551, opened 2024-06-11
coordinator: use history in authority

burgerdev:
This change swaps the in-memory persistence for a filesystem-backed KV store wrapped by `History`.  

Through the use of an internal atomic state pointer, the `Authority` struct is now thread-safe and responds with the correct state (esp. CA certs) even under concurrent calls to `SetManifest`. It also anticipates out-of-band modifications of persistence and syncs accordingly, which should make it ready for a distributed setup.

### DIFF
--- a/coordinator/internal/authority/authority.go
+++ b/coordinator/internal/authority/authority.go
@@ -189,7 +189,7 @@ func (m *Authority) GetCertBundle(peerPublicKeyHashStr string) (Bundle, error) {
 
 // GetManifestsAndLatestCA retrieves the manifest history and the currently active CA instance.
 //
-// If no manifest is configured, it returns an empty slice and a nil CA.
+// If no manifest is configured, it returns an error.
 func (m *Authority) GetManifestsAndLatestCA() ([]*manifest.Manifest, *ca.CA, error) {
 	if m.se.Load() == nil {
 		return nil, nil, ErrNoManifest
--- a/coordinator/internal/authority/authority.go
+++ b/coordinator/internal/authority/authority.go
@@ -9,13 +9,17 @@ import (
 	"crypto/x509"
 	"encoding/asn1"
 	"encoding/hex"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
+	"os"
+	"slices"
 	"sync"
 	"sync/atomic"
 
+	"github.com/edgelesssys/contrast/coordinator/history"
 	"github.com/edgelesssys/contrast/coordinator/internal/seedengine"
-	"github.com/edgelesssys/contrast/internal/appendable"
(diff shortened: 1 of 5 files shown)


coordinator/internal/authority/authority.go
```diff
@@ -181,52 +189,145 @@ func (m *Authority) GetCertBundle(peerPublicKeyHashStr string) (Bundle, error) {
...
-	return m.manifests.All(), m.ca.Load()
+//
+// If no manifest is configured, it returns an empty slice and a nil CA.
+func (m *Authority) GetManifestsAndLatestCA() ([]*manifest.Manifest, *ca.CA, error) {
+	if err := m.sync(); err != nil {
+		return nil, nil, fmt.Errorf("could not sync internal state: %w", err)
+	}
+	state := m.state.Load()
+	if state == nil {
+		return nil, nil, nil
```
katexochen, line 200: returning nil, nil (,nil) is usually discouraged.

coordinator/internal/authority/authority_test.go
katexochen, line 1: Test don't cover many code paths, but better than no tests. :)

katexochen (changes requested):

coordinator/internal/authority/authority.go
burgerdev: Would you rather fail the call, or return an empty slice and nil CA?

katexochen: We are returning an error on the caller side anyway if we get 3x nil, right? So we might want to raise it here.

burgerdev: Added a test for it that revealed a bug :)

coordinator/internal/authority/authority_test.go
burgerdev: I tried to cover some more, but leaving the rest open for a future PR.

katexochen (approved):

katexochen:
Let's wait with the merge until after v0.7.0.

coordinator/internal/authority/authority.go
```diff
@@ -174,59 +181,160 @@ func (m *Authority) GetCertBundle(peerPublicKeyHashStr string) (Bundle, error) {
...
 
 	return bundle, nil
 }
 
 // GetManifestsAndLatestCA retrieves the manifest history and the currently active CA instance.
-func (m *Authority) GetManifestsAndLatestCA() ([]*manifest.Manifest, *ca.CA) {
-	// TODO(burgerdev): The CA should be tied to the manifest.
-	return m.manifests.All(), m.ca.Load()
+//
+// If no manifest is configured, it returns an empty slice and a nil CA.
```
burgerdev, line 192: This is not true anymore.

PR #553, opened 2024-06-11
packages: update genpolicy-kata to 0.3.5

msanft:
This updates the Kata upstream version of the genpolicy tool to the tag 3.5.0. Previously, it relied on Microsoft source for whatever reason. This changes it back to Kata source. It also moves the package into a new directory structure, which will allow for better separation of Microsoft and Kata upstream tools. Additionally, it moves the remaining counterpart, the genpolicy tool from Microsoft, into the new directory structure, too.


packages/by-name/cli-release/package.nix
```diff
@@ -3,8 +3,8 @@
 
 { lib
 , contrast
-, genpolicy-msft
-, genpolicy ? genpolicy-msft
```
katexochen, line 7: @msanft could you please add a commit that exposed the genpolicy version used in the cli as `passthru.genpolicy`, so we can refer to the `settings`/`rules` in `cli-release` via `contrast.genplicy`? That will prevent us from mismatching.

packages/scripts.nix
```diff
@@ -154,7 +154,7 @@
     name = "write-coordinator-policy";
     runtimeInputs = with pkgs; [
       yq-go
-      genpolicy-msft
+      microsoft.src/tools/genpolicy/rules.rego
```
katexochen, line 157:
```suggestion
```

packages/by-name/cli-release/package.nix
msanft: done.

PR #557, opened 2024-06-11
packages: add upstream kata-agent

msanft:
This adds the kata-agent from the Kata upstream sources to our package set. (`kata.kata-agent`) Additionally, it moves the Microsoft kata-agent to `microsoft.kata-agent` (in the new directory structure).


packages/by-name/kata-image/package.nix
```diff
@@ -5,11 +5,12 @@
 , stdenv
 , stdenvNoCC
 , distro ? "cbl-mariner"
+, microsoft
 , bubblewrap
 , fakeroot
 , fetchFromGitHub
 , fetchurl
-, kata-agent
+, kata-agent ? microsoft.kata-agent
```
katexochen, line 13: This doesn't really make sense, as it builds msft's kata-image and you can't just inject the upstream agent. Same for the kernel-uvm.

katexochen (approved):
Otherwise lgtm, thanks. :seal:

packages/by-name/kata-image/package.nix
msanft: What? This imports the microsoft agent? That's what it did before, right?

katexochen: It makes the kata-agent configurable, and let's it default to the microsoft.kata-agent. Instead, we can just always use microsoft.kata-agent (or just kata-agent, when it's in the same subset?).

PR #559, opened 2024-06-11
release: patch latest references in versioned docs

davidweisse:
This patches the download links and container images in the released versioned docs to be tagged with the release version and hash instead of `latest`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -150,11 +150,27 @@ jobs:
           git commit -am "docs: release ${{ needs.process-inputs.outputs.MAJOR_MINOR }}"
           # Clean up auxiliary files, so next steps run on a clean tree
           git clean -fx :/
+      - name: Download image replacements file (from release branch)
+        uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e #v4.1.7
+        with:
+          name: image-replacements.txt
+      - name: Update release urls in docs with tags
+        working-directory: contrast-main
+        env:
+          MAJOR_MINOR: ${{ needs.process-inputs.outputs.MAJOR_MINOR }}
+          VERSION: ${{ inputs.version }}
+        run: nix run .#scripts.update-release-urls
+      - name: Commit updated docs
+        working-directory: contrast-main
+        run: |
+          git add docs/versioned_docs/version-${{ needs.process-inputs.outputs.MAJOR_MINOR }}
+          git commit -m "docs: update release download urls"
+          git clean -fx :/
       - name: Download updated contrast releases file (from release branch)
         uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e #v4.1.7
         with:
           name: contrast-releases.json
-          path: ./contrast-main/packages/contrast-releases.json
+          path: ./contrast-main/packages
       - name: Commit updated contrast-releases.json
         working-directory: contrast-main
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -150,14 +151,29 @@ jobs:
...
+      - name: Patch latest references in docs
+        shell: bash
+        run: |
+          while IFS= read -r line; do
+              target=$(echo "$line" | cut -d"=" -f2)
+              source=$(echo "$line" | sed "s/:.*//" | sed "s/\./\\\./g")
+              find ./contrast-main/docs/versioned_docs/version-$MAJOR_MINOR -type f -exec sed -i "s#$source\(:[a-zA-Z0-9_.-]\{1,\}\(@sha256:[a-fA-F0-9]\{64\}\)\?\|@sha256:[a-fA-F0-9]\{64\}\)#$target#g" {} \;
+          done <"./image-replacements.txt"
+
+          find ./contrast-main/docs/versioned_docs/version-$MAJOR_MINOR -type f -exec sed -i "s#github.com/edgelesssys/contrast/releases/latest/download/#github.com/edgelesssys/contrast/releases/download/${{ inputs.version }}/#g" {} \;
```
katexochen, line 167: Please put this into a nix script in packages/scripts.nix

.github/workflows/release.yml
```diff
@@ -150,14 +151,29 @@ jobs:
...
+      - name: Commit updated docs
+        working-directory: contrast-main
+        run: |
+          git add docs/versioned_docs/version-$MAJOR_MINOR
+          git commit -m "docs: patch latest references"
```
katexochen, line 172:
```suggestion
          git commit -m "docs: update release download urls"
```
Might be a bit clearer?

katexochen, line 172: Can we `git clean -fx :/` after this step?

.github/workflows/release.yml
```diff
@@ -150,14 +151,29 @@ jobs:
           git commit -am "docs: release ${{ needs.process-inputs.outputs.MAJOR_MINOR }}"
           # Clean up auxiliary files, so next steps run on a clean tree
           git clean -fx :/
-      - name: Download updated contrast releases file (from release branch)
+      - name: Download artifacts
         uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e #v4.1.7
         with:
-          name: contrast-releases.json
-          path: ./contrast-main/packages/contrast-releases.json
+          merge-multiple: true
```
katexochen, line 157: Will this always download all artifacts? We should document the problem we faced here and link to the issue.

.github/workflows/release.yml
```diff
@@ -274,8 +295,6 @@ jobs:
         with:
           name: contrast-releases.json
           path: ./packages/contrast-releases.json
-      - name: Clean up git tree
-        run: git clean -fx :/
```
katexochen, line 278: Can we instead modify this statement to only reset checked-in files and keep the build results?

.github/workflows/release.yml
```diff
@@ -150,14 +151,29 @@ jobs:
...
-          name: contrast-releases.json
-          path: ./contrast-main/packages/contrast-releases.json
+          merge-multiple: true
+      - name: Patch latest references in docs
+        shell: bash
+        run: |
+          while IFS= read -r line; do
+              target=$(echo "$line" | cut -d"=" -f2)
+              source=$(echo "$line" | sed "s/:.*//" | sed "s/\./\\\./g")
+              find ./contrast-main/docs/versioned_docs/version-$MAJOR_MINOR -type f -exec sed -i "s#$source\(:[a-zA-Z0-9_.-]\{1,\}\(@sha256:[a-fA-F0-9]\{64\}\)\?\|@sha256:[a-fA-F0-9]\{64\}\)#$target#g" {} \;
```
katexochen, line 164: Please break the long lines. Maybe you could even prepare parts of the sed expression in variables before calling find.

katexochen (changes requested):

katexochen (approved):

PR #562, opened 2024-06-12
release: test in the default namespace

burgerdev:
This brings us closer to the documented instructions (we had a broken release before because of this discrepancy).

Drive-by: use the appropriate way to determine whether an unstructured object is namespaced or not. This allows applying/deleting resources with no (or empty) namespace using kubeclient.

### DIFF
--- a/e2e/release/release_test.go
+++ b/e2e/release/release_test.go
@@ -7,10 +7,8 @@ package release
 
 import (
 	"context"
-	"crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
-	"encoding/hex"
 	"flag"
 	"io"
 	"net"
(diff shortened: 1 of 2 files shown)


e2e/release/release_test.go
```diff
@@ -47,42 +47,44 @@ func TestRelease(t *testing.T) {
 	ctx := context.Background()
 	k := kubeclient.NewForTest(t)
 
-	if *namespace == "" {
-		*namespace = randomNamespace(t)
```
davidweisse, line 51: If we test in the default namespace, we don't need the `randomNamespace` function anymore.

davidweisse (approved):

PR #565, opened 2024-06-12
logging: enable subsystem warn level by default

davidweisse:
This enables the all log subsystems at `warn` level or above. These logs will be get printed regardless of the `CONTRAST_LOG_SUBSYSTEMS` list. The `CONTRAST_LOG_LEVEL` variable is also considered, so when the global log level is set to `error`, only error messages will get printed from the subsystems

### DIFF
--- a/internal/attestation/snp/validator.go
+++ b/internal/attestation/snp/validator.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"log/slog"
 
-	"github.com/edgelesssys/contrast/internal/logger"
 	"github.com/edgelesssys/contrast/internal/oid"
 	"github.com/google/go-sev-guest/abi"
 	"github.com/google/go-sev-guest/proto/sevsnp"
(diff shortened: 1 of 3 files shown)


burgerdev:
@katexochen I wonder if we should enable _all_ log subsystems at _WARN_ level by default, and then ensure some rigour so that actually relevant messages (and only those) are reported at that level.

internal/attestation/snp/validator.go
```diff
@@ -69,7 +68,7 @@ func NewValidatorWithCallbacks(optsGen validateOptsGenerator, kdsGetter trust.HT
 		validateOptsGen: optsGen,
 		callbackers:     callbacks,
 		kdsGetter:       kdsGetter,
-		logger:          slog.New(logger.NewHandler(log.Handler(), "snp-validator")),
```
katexochen, line 72: why are we removing the log subsystem?

katexochen:
> @katexochen I wonder if we should enable _all_ log subsystems at _WARN_ level by default, and then ensure some rigour so that actually relevant messages (and only those) are reported at that level.

Fine for me! :)

internal/attestation/snp/validator.go
davidweisse: The log subsystem is already added on the caller side, so this would create a nested subsystem with the same name.

katexochen:
@davidweisse could you update PR title and description?

PR #566, opened 2024-06-13
resourcegen: cleanup emojivoto env variables

davidweisse:
Only add the necessary environment variables in the case where the Service Mesh is disabled.

### DIFF
--- a/internal/kuberesource/sets.go
+++ b/internal/kuberesource/sets.go
@@ -220,9 +220,6 @@ func Emojivoto(smMode serviceMeshMode) []any {
 							).
 							WithEnv(EnvVar().WithName("GRPC_PORT").WithValue("8080")).
 							WithEnv(EnvVar().WithName("PROM_PORT").WithValue("8801")).
-							WithEnv(EnvVar().WithName("EDG_CERT_PATH").WithValue("/tls-config/certChain.pem")).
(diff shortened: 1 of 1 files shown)


internal/kuberesource/sets.go
```diff
@@ -442,6 +432,19 @@ func Emojivoto(smMode serviceMeshMode) []any {
 	}
 
 	if smMode == ServiceMeshDisabled {
```
burgerdev, line 434: We are using @3u13r  images for ServiceMeshIngress, too. If they are modified, they might need the environment variables, but if they are not we should replace them with vanilla ones.

katexochen: I'm actually not sure why we aren't using the upstream image. Maybe because the endpoint is https, but the bot talks http by default?

3u13r: For the vote-bot (part of the "web" image): yes. It's a http client but it must be a https client.

katexochen: So we don't need the ENV configuration there, right?

3u13r: No, we don't need it. For more details see: https://github.com/3u13r/emojivoto/commit/8ba877681c297721cde63eb7be95c98c4c1186ee

katexochen (approved):

PR #568, opened 2024-06-13
packages: move remaining packages into new directory structure

msanft:
This moves the remaining Microsoft-specific packages into the new directory structure. Unfortunately, `callPackage` still requires us to address them from the top-level attribute, even though they're in the same attribute set. I also made an opinionated change by purposefully not moving runtime-class-files, even though it's currently (obviously) only used on AKS, as it will definitely, and in somewhat of the same form, be needed for the bare-metal setup as well.


katexochen (changes requested):
igvm-tooling, igvmmeasure, igvm-singing-key don't need to be moved, as they are not specific to the microsoft for of kata.

Moving kata-image/package-index.json will require adaption in the update rpm workflow that writes to that file.

.github/workflows/rpm_updates.yml
```diff
@@ -14,13 +14,16 @@ jobs:
         with:
           ref: ${{ !github.event.pull_request.head.repo.fork && github.head_ref || '' }}
           token: ${{ !github.event.pull_request.head.repo.fork && secrets.NUNKI_CI_COMMIT_PUSH_PR || github.token }}
+
```
katexochen, line 17:
```suggestion
```

.github/workflows/rpm_updates.yml
```diff
@@ -14,13 +14,16 @@ jobs:
...
       - uses: ./.github/actions/setup_nix
         with:
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
```
katexochen, line 22:
```suggestion
```

.github/workflows/rpm_updates.yml
```diff
@@ -14,13 +14,16 @@ jobs:
...
       - name: Update rpm packages
         run: |
-           nix run .#rpm-pin-vendor > packages/by-name/kata-image/package-index.json
+           nix run .#rpm-pin-vendor > packages/by-name/microsoft/kata-image/package-index.json
+
```
katexochen, line 26:
```suggestion
```

katexochen (approved):
lgtm otherwise.

PR #570, opened 2024-06-13
recovery: add server and recover authority

katexochen:


burgerdev:
I rebased `burgerdev/history`, so if you rebase you get the `hist.HasLatest`.

burgerdev (approved):

PR #573, opened 2024-06-13
resourcegen: add startup probe to service mesh

davidweisse:
This adds a startup probe to the Service Mesh sidecar container. It probes on the port the Envoy proxy listens on for incoming traffic. Envoy only starts after the iptables rules have been updated, which is when the regular containers should start.

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -301,5 +301,12 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
 			WithPrivileged(true).
 			AddCapabilities("NET_ADMIN").
 			SecurityContextApplyConfiguration,
+		).
+		WithStartupProbe(Probe().
+			WithInitialDelaySeconds(1).
+			WithPeriodSeconds(5).
+			WithFailureThreshold(5).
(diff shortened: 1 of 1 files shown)


internal/kuberesource/parts.go
```diff
@@ -301,5 +301,12 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
 			WithPrivileged(true).
 			AddCapabilities("NET_ADMIN").
 			SecurityContextApplyConfiguration,
+		).
+		WithStartupProbe(Probe().
+			WithInitialDelaySeconds(1).
+			WithPeriodSeconds(5).
```
burgerdev, line 307: I think this could be smaller, say 1s. I measured a couple of container startups and it varied between 0.5s to 0.9s to get to `Invalid path: /tls-config/mesh-ca.pem`, so I'd guess the entire thing should be done in 2s most. But the threshold should be higher accordingly.

davidweisse: If we are probing for the Envoy endpoint, we need to wait until the Envoy proxy has started. Looking at the logs, that takes about 4s to 4.3s for me.

burgerdev (approved):

PR #583, opened 2024-06-14
coordinator: uniform gRPC metric prefix 

burgerdev:
During a refactoring, we discovered that the gRPC metrics include the service in both the metric name and a label:

```
contrast_userapi_grpc_server_handling_seconds_bucket{grpc_method="GetManifests",grpc_service="userapi.UserAPI",grpc_type="unary",le="0.01"} 1
```

Since it's both more consistent to report all gRPC methods under the same metric, and easier to maintain the gRPC metrics when there is only one instance, we decided to move all of them under a common prefix, `contrast_`. 

<details>
<summary>Future list of metrics</summary>
<code>
contrast_attestation_failures 8
contrast_coordinator_manifest_generation 1
contrast_grpc_server_handled_total
contrast_grpc_server_handling_seconds_bucket
contrast_grpc_server_handling_seconds_count
contrast_grpc_server_handling_seconds_sum
contrast_grpc_server_msg_received_total
contrast_grpc_server_msg_sent_total
contrast_grpc_server_started_total
</code>
</details>

Below is the original PR description with the motivation for refactoring userapi etc.

---

This PR moves the gRPC handlers for `userapi` and `recoveryapi` into the `authority` module. In the main module remain gRPC server setup (credentials, metrics, listener).

I'm mostly driven by the increasing complexity of changing code in both `authority` and `userapi`. We had to change the `Authority` API a couple of times, and adjusting the `userapi` test accordingly is cumbersome. Given that the `Authority` API is closely coupled to the gRPC API, it is probably for the best to merge the two and remove the current `Authority` API. 

To keep things simple, I'm refactoring in several steps. This is the first step, which does the bare minimum to keep the boat afloat after the move. There are still some rough edges (I left a couple of TODOs), but it already allows to resolve [this](https://github.com/edgelesssys/contrast/blob/131b5d82bcaac23010ae1ffd1e0f0be99614c789/coordinator/userapi.go#L148) and [that](https://github.com/edgelesssys/contrast/blob/131b5d82bcaac23010ae1ffd1e0f0be99614c789/coordinator/internal/authority/authority.go#L355) TODO, which are blocking RFC004 - the former here and the latter in a PoC commit [here].

Next, I need to consolidate the former `Authority` API with the `userapi`, which will mostly involve moving code around, and adapt the tests. Fortunately, these changes can now be interleaved with the recovery effort.

### Review Recommendation

I split the PR into 

* a pure move of `userapi*.go` and `recoveryapi.go` 
* subsequent changes to adapt to their new habitat

It is probably easier to restrict the `Changes from` in the review tab to the second commit, so that you only see the actual changes. Otherwise, git does not consider the files related and the diffs are the entire files.

**TODO**:

- [x] rename PR for changelog (user facing is the change in metric names)
- [x] rename (and keep) first commit 
- [x] squash fixups


coordinator/internal/authority/userapi_test.go
```diff
@@ -95,9 +89,11 @@ func TestManifestSet(t *testing.T) {
 					[]byte("c"),
 				},
 			},
-			mSGetter: &stubManifestSetGetter{},
-			wantErr:  true,
+			wantErr: true,
 		},
+		// TODO(burgerdev): add test for dysfunctional history backend
+		// TODO(burgerdev): add test for workload owner keys in req	(present|absent|wrong)
```
burgerdev, line 95: This is actually done below.

coordinator/internal/authority/userapi_test.go
```diff
@@ -95,9 +89,11 @@ func TestManifestSet(t *testing.T) {
...
+		// TODO(burgerdev): add test for last manifest (no owner keys)
```
burgerdev, line 96: This TODO refers to the situation where a manifest with no workload owner keys is configured.

coordinator/main.go
```diff
@@ -122,3 +116,34 @@ func run() (retErr error) {
...
 	return eg.Wait()
 }
+
+func newGRPCServer(reg *prometheus.Registry, log *slog.Logger) *grpc.Server {
+	issuer := snp.NewIssuer(logger.NewNamed(log, "snp-issuer"))
+	credentials := atlscredentials.New(issuer, nil)
+
+	grpcUserAPIMetrics := grpcprometheus.NewServerMetrics(
+		grpcprometheus.WithServerCounterOptions(
+			grpcprometheus.WithSubsystem("contrast_userapi"),
```
katexochen, line 126: metrics name are now wrong, as server is used by both user and mesh api.

katexochen (approved):
Maybe we should keep the first commit to retain history.

coordinator/main.go
katexochen: Also remember that this needs to be updated in the docs when changed.

burgerdev: Good catch! I think I also made a mistake below resulting in no metrics being registered at all. Let me check that and see if I can work around that.

burgerdev: Addressed in 6e1c6b87081f251ea04b7f62d2241a62a8b87b45

PR #586, opened 2024-06-17
docs: update graphics

m1ghtym0:
* Add concept design to readme and intro
* Update components with black instead of purple on green
* Backport both to 0.7


katexochen:
There is also https://github.com/edgelesssys/contrast/blob/main/docs/docs/_media/personas.svg with the purple on green.

katexochen (approved):
Seems to comply with the style guidelines now. I'm not a fan of the black boxes, looks quite depressive and also a bit hard to read. Maybe we can just use white in the future?

PR #587, opened 2024-06-17
genpolicy: allow contrast env vars for coordinator

davidweisse:
The default Coordinator policy now allows environment variables of the form `CONTRAST_*`, to allow enabling and configuring debug logging. For the dev setup, these changes are also applied to the default policy settings. This allows configuring logging for the Coordinator as well as the Initializer or Service Mesh, but also allows changing the Proxy config for Envoy, so for the release the patches are only applied to the Coordinator policy.


packages/by-name/microsoft/genpolicy/genpolicy_msft_settings_coordinator.patch
```diff
@@ -0,0 +1,21 @@
...
+index 7d35862a..4eacc7cd 100644
+--- a/genpolicy-settings.json
++++ b/genpolicy-settings.json
+@@ -307,7 +307,8 @@
+                 "^AZURE_CLIENT_ID=[A-Fa-f0-9-]*$",
+                 "^AZURE_TENANT_ID=[A-Fa-f0-9-]*$",
+                 "^AZURE_FEDERATED_TOKEN_FILE=/var/run/secrets/azure/tokens/azure-identity-token$",
+-                "^AZURE_AUTHORITY_HOST=https://login\\.microsoftonline\\.com/$"
++                "^AZURE_AUTHORITY_HOST=https://login\\.microsoftonline\\.com/$",
++                "^CONTRAST_[A-Z_]*=.*$"
```
burgerdev, line 11:
```suggestion
+                "^CONTRAST_[A-Z0-9_]*=.*$"
```
... just as a precaution, if we ever wanted digits in our env vars.

burgerdev (approved):

katexochen (approved):

PR #590, opened 2024-06-17
manifest: add utilities for encrypted SeedShares

burgerdev:
This centralizes the algorithms used for managing seed shares and owner keys.

### DIFF
--- a/internal/manifest/crypto.go
+++ b/internal/manifest/crypto.go
@@ -1,0 +1,59 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package manifest
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/x509"
+	"fmt"
+
+	"github.com/edgelesssys/contrast/internal/userapi"
+)
+
+// MarshalSeedShareOwnerKey converts a public key into the format for userapi.SetManifestRequest.
+func MarshalSeedShareOwnerKey(pubKey *rsa.PublicKey) HexString {
+	return NewHexString(x509.MarshalPKCS1PublicKey(pubKey))
+}
+
+// ParseSeedShareOwnerKey reads a public key embedded in a userapi.SetManifestRequest.
+func ParseSeedShareOwnerKey(pubKeyHex HexString) (*rsa.PublicKey, error) {
+	pubKeyBytes, err := pubKeyHex.Bytes()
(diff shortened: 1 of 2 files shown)


katexochen:
I don't like how this change puts business logic into a pure crypto library. Imo the package shouldn't be aware of seed shares or import the api packages. Maybe we can move this somewhere else?

burgerdev:
My first pick would have been the `userapi` package, but you can make a similar argument there (it's a pure transport layer package). Manifest sounds good.

burgerdev:
Fixed, ptal.

internal/manifest/crypto.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/edgelesssys/contrast/internal/userapi"
+)
+
+// MarshalSeedShareOwnerKey converts a public key into the format for userapi.SetManifestRequest.
+func MarshalSeedShareOwnerKey(pubKey *rsa.PublicKey) HexString {
+	return NewHexString(x509.MarshalPKCS1PublicKey(pubKey))
+}
+
+// ParseSeedShareOwnerKey reads a public key embedded in a userapi.SetManifestRequest.
+func ParseSeedShareOwnerKey(pubKeyHex HexString) (*rsa.PublicKey, error) {
```
katexochen, line 22: UnmarshalSeed... ?

internal/manifest/crypto_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"testing"
+
+	"github.com/edgelesssys/contrast/internal/userapi"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEncryptDecryptSingleKey(t *testing.T) {
```
katexochen, line 16: Remember that we use table driven tests by default (but no need to change this now).

katexochen (approved):

internal/manifest/crypto.go
burgerdev: It's called `Marshal...` and `Parse...` in the `x509` package, though. Which inconsistency would you rather have?

internal/manifest/crypto_test.go
burgerdev: I tried to use the `testCase` pattern, but eventually realized that we want the full cartesian product. The tests for wrong key and wrong label are somewhat special and don't seem to fit naturally into a test table.

3u13r (approved):

PR #591, opened 2024-06-17
packages: package kata-kernel

msanft:
This packages a kernel that should resemble what Kata uses upstream for their confidential computing deployments. It does not reflect it 1-to-1 (for example, we do not supply initramfs sources), but very closely. Tested to work in a Kata Kubernetes deployment.


packages/by-name/kata/kata-kernel/package.nix
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 Edgeless Systems GmbH
+# SPDX-License-Identifier: AGPL-3.0-only
+
+{ fetchurl
+, linuxManualConfig
+}:
+linuxManualConfig rec {
```
katexochen, line 7:
```suggestion
}:

linuxManualConfig rec {
```

packages/by-name/kata/kata-kernel/package.nix
```diff
@@ -0,0 +1,22 @@
...
+  version = "6.7";
+
+  src = fetchurl {
```
katexochen, line 10:
```suggestion

  # See https://github.com/kata-containers/kata-containers/blob/5f11c0f144037d8d8f546c89a0392dcd84fa99e2/versions.yaml#L198-L201
  src = fetchurl {
```

packages/by-name/kata/kata-kernel/package.nix
```diff
@@ -0,0 +1,22 @@
...
+    url = "https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-${version}.tar.xz";
+    sha256 = "sha256-7zEUSiV20IDYwxaY6D7J9mv5fGd/oqrw1bu58zRbEGk=";
+  };
+
+  # TODO(msanft): possibly package the config generation in Nix.
+  # Built from Kata upstream via:
+  # kata-containers/tools/packaging/kernel/build-kernel.sh -a x86_64 -m -t kvm -x -f setup
+  # Note: initramfs sources are removed, as we do not supply them in the build.
```
katexochen, line 18:
```suggestion
  # Built from Kata upstream via:
  # kata-containers/tools/packaging/kernel/build-kernel.sh -a x86_64 -m -t kvm -x -f setup
  # Note: initramfs sources are removed, as we do not supply them in the build.
  # TODO(msanft): possibly package the config generation in Nix.
```

katexochen (changes requested):

katexochen:
I think we should call the package `kata-kernel-uvm`, similar to the microsoft one.

katexochen (approved):

PR #592, opened 2024-06-17
e2e: cleanup when test is cancelled

davidweisse:
Adds an extra cleanup step to both the OpenSSL and Service Mesh test to clean up the resources in the CI cluster if the test was cancelled.

### DIFF
--- a/.github/workflows/e2e_openssl.yml
+++ b/.github/workflows/e2e_openssl.yml
@@ -6,7 +6,8 @@ on:
         skip-undeploy:
           description: "Skip undeploy"
           required: false
-          default: "false"
+          type: boolean
+          default: false
     pull_request:
       paths-ignore:
         - docs/**
@@ -60,4 +61,11 @@ jobs:
           just coordinator initializer openssl port-forwarder node-installer
       - name: E2E Test
         run: |
-          nix shell .#contrast.e2e --command openssl.test -test.v workspace/just.containerlookup
+          nix shell .#contrast.e2e --command openssl.test -test.v \
+            --image-replacements workspace/just.containerlookup \
+            --namespace-file workspace/e2e.namespace \
+            --skip-undeploy="${{ inputs.skip-undeploy && 'true' || 'false' }}"
+      - name: Cleanup
+        if: cancelled() && !inputs.skip-undeploy
+        run: |
+          kubectl delete ns $(cat workspace/e2e.namespace) --timeout 5m
--- a/.github/workflows/e2e_servicemesh.yml
+++ b/.github/workflows/e2e_servicemesh.yml
@@ -6,7 +6,8 @@ on:
         skip-undeploy:
           description: "Skip undeploy"
           required: false
-          default: "false"
+          type: boolean
+          default: false
     pull_request:
       paths-ignore:
         - docs/**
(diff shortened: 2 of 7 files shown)


.github/workflows/e2e_openssl.yml
```diff
@@ -61,3 +63,7 @@ jobs:
       - name: E2E Test
         run: |
           nix shell .#contrast.e2e --command openssl.test -test.v workspace/just.containerlookup
+      - name: Cleanup
+        if: cancelled() && inputs.skip-undeploy != 'true'
```
burgerdev, line 67: Can we please make `skip-undeploy` a boolean? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputsinput_idtype

.github/workflows/e2e_servicemesh.yml
```diff
@@ -61,7 +63,7 @@ jobs:
       - name: E2E Test
         run: |
           nix shell .#contrast.e2e --command servicemesh.test -test.v workspace/just.containerlookup
-      - name: Undeploy
-        if: always() && inputs.skip-undeploy != 'true'
+      - name: Cleanup
+        if: cancelled() && inputs.skip-undeploy != 'true'
         run: |
-          just undeploy
+          kubectl delete ns $(cat e2e.namespace) --timeout 5m
```
burgerdev, line 69: I think contrasttest puts that file into a tempdir.

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -81,19 +81,22 @@ func (ct *ContrastTest) Init(t *testing.T, resources []any) {
...
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	err = ct.Kubeclient.Apply(ctx, namespace...)
 	cancel()
 	require.NoError(err)
 
 	t.Cleanup(func() {
-		// Deleting the namespace may take some time due to pod cleanup, but we don't want to wait until the test times out.
-		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
-		defer cancel()
+		if os.Getenv("SKIP_UNDEPLOY") != "true" {
```
burgerdev, line 92: Did you consider making this a boolean flag? That would allow for more flexibility in evaluation (e.g. `--skip-undeploy`, `--skip-undeploy=0`, `--skip-undeploy=TRUE`).

.github/workflows/e2e_openssl.yml
```diff
@@ -60,4 +61,11 @@ jobs:
           just coordinator initializer openssl port-forwarder node-installer
       - name: E2E Test
         run: |
-          nix shell .#contrast.e2e --command openssl.test -test.v workspace/just.containerlookup
+          nix shell .#contrast.e2e --command openssl.test -test.v \
+            --image-replacements workspace/just.containerlookup \
+            --namespace-file workspace/e2e.namespace \
+            --skip-undeploy=${{ inputs.skip-undeploy }}
+      - name: Cleanup
+        if: cancelled() && inputs.skip-undeploy != true
```
burgerdev, line 69:
```suggestion
        if: cancelled() && inputs.skip-undeploy
```

Now that we have a bool, this should be enough, right?

burgerdev (approved):

.github/workflows/e2e_openssl.yml
burgerdev: Sorry, forgot the negation, but you know what I mean :)

.github/workflows/e2e_openssl.yml
```diff
@@ -60,4 +61,11 @@ jobs:
           just coordinator initializer openssl port-forwarder node-installer
       - name: E2E Test
         run: |
-          nix shell .#contrast.e2e --command openssl.test -test.v workspace/just.containerlookup
+          nix shell .#contrast.e2e --command openssl.test -test.v \
```
burgerdev, line 64:
```suggestion
          nix shell .#contrast.e2e --command openssl.test -test.v -args \
```

that should fix the errors

burgerdev (approved):

.github/workflows/e2e_openssl.yml
davidweisse: The `-args` is only a flag when using `go test`, I think. The problem is that the input boolean gets parsed as an empty string and not as `false`.

burgerdev: Oh, I see.  How about `--skip-undeploy="${{ inputs.skip-undeploy && "true" || "false" }}"`?

PR #593, opened 2024-06-17
genpolicy: add flag for cache path

3u13r:
Upstream PR: https://github.com/kata-containers/kata-containers/pull/9864
For now, this should have no change to the user. 
The question is, how we now want to use this option. Also have a similar flag in our CLI or just have a general home/.config path.

TODO: 

- [x] Wait for first upstream feedback
- [x] Use `git format-patch` to create initial patch file

### DIFF
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -25,6 +25,7 @@ const (
 	settingsFilename     = "settings.json"
 	seedSharesFilename   = "seed-shares.json"
 	rulesFilename        = "rules.rego"
+	layersCacheFilename  = "layers-cache.json"
 	verifyDir            = "verify"
 	cacheDirEnv          = "CONTRAST_CACHE_DIR"
 )
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -66,6 +66,7 @@ subcommands.`,
 
 	cmd.Flags().StringP("policy", "p", rulesFilename, "path to policy (.rego) file")
 	cmd.Flags().StringP("settings", "s", settingsFilename, "path to settings (.json) file")
+	cmd.Flags().StringP("genpolicy-cache-path", "c", layersCacheFilename, "path to cache for the cache (.json) file containing the image layers")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
 	cmd.Flags().String("reference-values", "", "set the default reference values used for attestation (one of: aks)")
 	cmd.Flags().StringArrayP("workload-owner-key", "w", []string{workloadOwnerPEM},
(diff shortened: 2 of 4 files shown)


cli/cmd/generate.go
```diff
@@ -391,7 +391,7 @@ func addWorkloadOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) e
 func generatePolicyForFile(ctx context.Context, genpolicyPath, regoPath, policyPath, yamlPath string, logger *slog.Logger) ([32]byte, error) {
 	args := []string{
 		"--raw-out",
-		"--use-cached-files",
+		"--layers-cache-file-path=./layers-cache.json",
```
katexochen, line 394: this should be made configurable through a flag, defaulting to the file being at a default location within the configured workspaceDir.

3u13r:
Done. Do we already have a use for this e.g. somewhere in the CI? We currently don't list explicit CLI features in the docs, so there's also nothing to add here, I assume.

cli/cmd/common.go
```diff
@@ -25,6 +25,7 @@ const (
 	settingsFilename     = "settings.json"
 	seedSharesFilename   = "seed-shares.json"
 	rulesFilename        = "rules.rego"
+	cachePath            = "./layers-cache.json"
```
katexochen, line 28:
```suggestion
	layersCacheFilename            = "layers-cache.json"
```

cli/cmd/generate.go
```diff
@@ -66,6 +66,7 @@ subcommands.`,
 
 	cmd.Flags().StringP("policy", "p", rulesFilename, "path to policy (.rego) file")
 	cmd.Flags().StringP("settings", "s", settingsFilename, "path to settings (.json) file")
+	cmd.Flags().StringP("cache-path", "c", cachePath, "path to cache for the cache (.json) file containing the image layers")
```
katexochen, line 69: I think we should give the flag a prefix like "layers-" or "genpolicy-". Contrast has itself a cache directory (currently controlled via env vars) which might be confusing.

katexochen (approved):
> Do we already have a use for this e.g. somewhere in the CI?

As you correctly merge cache-path with workspace-dir, this change will already have effect on the dev workflow where the workspace is set. We will now generate on every `just generate` invocation, where we would previously often use the cache. A shift towards more correctness at the cost of performance, let's see how it goes.

cli/cmd/generate.go
3u13r: Yeah, I also stumbled over the "other" cache dir.  Since it may be that we need to pass more flags to genpolicy eventually, I'll go with `genpolicy-`.

PR #597, opened 2024-06-18
cli: fix autocomplete

### LINKED ISSUE #596: cli: tab completion not working
There is an issue with the tab completion

The completion is not initialized correctly. Basically, the generated completion script is missing the command it's intended for: "contrast". See the header of the generated script:

$fpath[1]/_contrast
```sh
compdef _
```
vs. 
```sh
compdef _constellation constellation
```

You can see the same behavior on the help command that shows incomplete "completion" commands. It misses the "contrast" prefix to the "[contrast] completion ...":

```sh
$ contrast completion bash -h
...
To load completions in your current shell session:

        source <( completion zsh)

To load completions for every new session, execute once:

#### Linux:

         completion zsh > "${fpath[1]}/_
```

Same for the bash commands.

This is due to a missing "Use" field in the Cobra initialization: https://github.com/edgelesssys/contrast/blob/main/cli/main.go#L32

m1ghtym0:
Fixes #596

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -30,6 +30,7 @@ var version = "0.0.0-dev"
 
 func newRootCmd() *cobra.Command {
 	root := &cobra.Command{
+		Use:              "contrast",
 		Short:            "contrast",
 		PersistentPreRun: preRunRoot,
 		Version:          version,


katexochen (approved):
can't test this locally right now, but change lgtm.

burgerdev (approved):
Tested - completion works now and other output is unaffected.

PR #599, opened 2024-06-18
Change CN: Constellation -> Contrast

blenessy:
The self-signed x.509 cert of the Coordinator has currently the CN: Constellation. Seems like a copy & paste error so I changed it to Contrast.


katexochen (approved):
Good catch, thanks!

PR #600, opened 2024-06-18
packages: package Kata upstream UVM image

msanft:
This packages a CBL-Mariner-based UVM image, resembling what's used upstream by Kata. For now, it uses almost the same build instructions as the Microsoft image, but eventually, we'll might want to adapt it as per the specific requirements of AKS and bare-metal appliances.

It also enables the CI to update the RPM packages inside the root filesystem for both the AKS and bare-metal UVM images.


katexochen (approved):
Build isn't reproducible right now, but ok to work on this in the future.

PR #601, opened 2024-06-18
just: add e2e target

davidweisse:
Instead of transforming the current `just default` target into an E2E-test, which disrupts the dev workflow by doing all steps like `generate`, `apply`, `set` and `verify` at once in the E2E test, this adds a new `just e2e` target which can run both the `openssl` and `servicemesh` tests without deleting the namespace afterward.


burgerdev (approved):
Very nice, thanks!

PR #603, opened 2024-06-18
treewide: prepare separation of node installers

msanft:
To support multiple deployment "flavours" (i.e. AKS and bare-metal QEMU on TDX), we need to provide separate node installer deployments, as both deployment scenarios require different components to be installed on the hosts. This prepares this separation, by moving the Microsoft-specific runtime-class-files into the `microsoft` package set, and including the "flavour" of the deployment in the node installer image name.


.github/workflows/e2e_getdents.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           just get-credentials
       - name: Build and prepare deployments
         run: |
-          just node-installer
+          just microsoft-node-installer
```
katexochen, line 51: please change to node-installer-microsoft, which is more intuitive for tab completion.

internal/kuberesource/parts.go
```diff
@@ -42,12 +42,12 @@ func NodeInstaller(namespace string) *NodeInstallerConfig {
 			).
 			WithTemplate(PodTemplateSpec().
 				WithLabels(map[string]string{"app.kubernetes.io/name": name}).
-				WithAnnotations(map[string]string{"contrast.edgeless.systems/pod-role": "contrast-node-installer"}).
+				WithAnnotations(map[string]string{"contrast.edgeless.systems/pod-role": "contrast-node-installer-microsoft"}).
```
katexochen, line 45: Why do we need to change this?

katexochen (changes requested):
Commit title should not end with a dot.

msanft:
Closing in favor of #611

PR #604, opened 2024-06-18
coordinator: enable returning seed to user

burgerdev:
This PR consolidates `Authority.setManifest` (authority) and `Authority.SetManifest` (userapi) after the refactoring in #583. This is more or less a move of the function body from one file to another, except that the generated seed could now be returned to the user (for that to work, however, bae8e7f needs to change the `Manifest` definition first). 

I tried to ensure that all test cases removed from `authority_test.go` are reflected in `userapi_test.go`, and it turned out we already covered most of it there.

### DIFF
--- a/coordinator/internal/authority/authority.go
+++ b/coordinator/internal/authority/authority.go
@@ -13,7 +13,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"os"
 	"slices"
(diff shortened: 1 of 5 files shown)


katexochen (approved):
Really hard to evaluate to overall picture with all the open todos. We should do a final review after we have everything merged.

PR #606, opened 2024-06-19
kubeclient: combine duplicate code for WaitFor* functions

davidweisse:
The functions `WaitForDeployment`, `WaitForDaemonSet` and `WaitForStatefulSet` are combined into one `WaitFor` function, which takes the resource kind as an input and performs all the logic that the previous functions had.

Additionally, the context error is now checked when receiving from the watcher channel, which is closed when the context expires, so you don't get error messages like this: https://github.com/edgelesssys/contrast/actions/runs/9522518982/job/26252214876#step:12:118

### DIFF
--- a/e2e/internal/kubeclient/deploy.go
+++ b/e2e/internal/kubeclient/deploy.go
@@ -18,7 +18,82 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
 )
+
+// ResourceWaiter is implemented by resources that can be waited for with WaitFor.
+type ResourceWaiter interface {
+	kind() string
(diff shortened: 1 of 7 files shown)


e2e/internal/kubeclient/deploy.go
```diff
@@ -28,181 +37,127 @@ func (c *Kubeclient) WaitForPod(ctx context.Context, namespace, name string) err
...
-					return nil
-				}
-			case watch.Deleted:
-				return fmt.Errorf("pod %s/%s was deleted while waiting for it", namespace, name)
-			default:
-				c.log.Warn("ignoring unexpected watch event", "type", evt.Type, "object", evt.Object)
+		evt, ok := <-watcher.ResultChan()
+		if !ok {
+			if ctx.Err() == nil {
+				return fmt.Errorf("watcher for Pod %s/%s unexpectantly closed", namespace, name)
```
burgerdev, line 43:
```suggestion
				return fmt.Errorf("watcher for Pod %s/%s closed unexpectedly", namespace, name)
```

burgerdev (approved):
Great, thanks!

PR #608, opened 2024-06-20
cli: use manifest reference values for attestation

davidweisse:
This changes the Coordinator validate options to use the reference values from the local manifest in the `set` and `verify` command. The verify command takes an additional flag `--manifest` for this to provide the path to the manifest.

The default workspace directory for the `verify` command is no longer `./verify`, because the manifest lies in the current directory. Instead, the output files for the `verify` command are always written to the `verify` directory in the current workspace directory.

### DIFF
--- a/cli/cmd/verify.go
+++ b/cli/cmd/verify.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"net"
 	"os"
@@ -16,9 +17,6 @@ import (
 	"github.com/edgelesssys/contrast/internal/grpc/dialer"
 	"github.com/edgelesssys/contrast/internal/manifest"
 	"github.com/edgelesssys/contrast/internal/userapi"
-	"github.com/google/go-sev-guest/abi"
-	"github.com/google/go-sev-guest/kds"
-	"github.com/google/go-sev-guest/validate"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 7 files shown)


cli/cmd/verify.go
```diff
@@ -131,18 +148,25 @@ func parseVerifyFlags(cmd *cobra.Command) (*verifyFlags, error) {
...
 		coordinator:  coordinator,
 		workspaceDir: workspaceDir,
 		policy:       policy,
 	}, nil
 }
 
-func newCoordinatorValidateOptsGen(hostData []byte) *snp.StaticValidateOptsGenerator {
-	defaultManifest := manifest.Default()
-	trustedMeasurement, err := defaultManifest.ReferenceValues.TrustedMeasurement.Bytes()
+func newCoordinatorValidateOptsGen(mnfst manifest.Manifest, hostData []byte) (*snp.StaticValidateOptsGenerator, error) {
```
burgerdev, line 166: This is now very similar to `authority.SNPValidateOpts`, we should imho share that code.

davidweisse: I moved the shared code to the `manifest` package, to generate a `validate.Options` struct from the manifest.

msanft:
Could you make it so that `runtimeHandler` is not passed to the CLI in the build process anymore, but deduced from the trusted measurement too?

katexochen:
> Could you make it so that `runtimeHandler` is not passed to the CLI in the build process anymore, but deduced from the trusted measurement too?

Should be done in a separate PR.

burgerdev (approved):
We should follow up with

1. A documentation change mentioning the manifest flag.
2. `verify` should fail if the given manifest is not latest.
3. Tests for SNPValidateOpts.

PR #611, opened 2024-06-20
Initial Support for bare-metal QEMU-TDX runtime deployment

msanft:
This adds initial support for deploying the Contrast runtime on bare-metal TDX machines with QEMU.

Notice that this does not yet contain full support for running Contrast on such platforms. Things that are left out as of now contain (non-comprehensive list):
- TDX attestation
- Policy generation and verification through the attestation
- Snapshotter support (Nydus / Tardev)
- Correct generation of the deployments with the right runtime handler name, which should be fixed with the follow-up for [reading the runtime handler from the Manifest](https://github.com/edgelesssys/contrast/pull/608).

Due to the size of this PR, I encourage reviewers to see the commit messages for a more detailed description of what each commit does, and thus to review on a per-commit basis.


katexochen:
PR is pretty big, can we split this into smaller chunks?

msanft:
> PR is pretty big, can we split this into smaller chunks?

As mentioned in the PR description, I split this into digestible commits already. What benefit do you see in splitting this up into multiple PRs instead of reviewing each commit?

node-installer/internal/constants/configuration-qemu-tdx.toml
```diff
@@ -0,0 +1,61 @@
...
+# upstream source: https://github.com/kata-containers/kata-containers/blob/0f2a4d202e90b39b50074725b2cfe9c3088a4e20/src/runtime/config/configuration-qemu-tdx.toml.in
+[hypervisor.qemu]
+path = "/usr/bin/qemu-system-x86_64"
+kernel = "/opt/kata/share/kata-containers/vmlinuz-confidential.container"
+image = "/opt/kata/share/kata-containers/kata-containers-confidential.img"
+machine_type = "q35"
+tdx_quote_generation_service_socket_port = 4050
+rootfs_type="erofs"
+confidential_guest = true
+enable_annotations = ["enable_iommu", "virtio_fs_extra_args", "kernel_params", "default_vcpus", "default_memory", "default_vcpus"]
```
katexochen, line 11: Why does this list `default_vcpus` twice?

node-installer/internal/constants/configuration-qemu-tdx.toml
```diff
@@ -0,0 +1,61 @@
...
+valid_hypervisor_paths = ["/usr/bin/qemu-system-x86_64"]
+kernel_params = ""
+firmware = "/usr/share/ovmf/OVMF.fd"
```
katexochen, line 14: Should we bring our own firmware? This will be part of the measurement, right?

katexochen (changes requested):
Please also address the comments from https://github.com/edgelesssys/contrast/pull/603

Reviewed up to and including 93b01654b6ec5d44d65e7d531bb154d7b67c8c08

node-installer/internal/constants/configuration-qemu-tdx.toml
msanft: Yeah, we probably should.

internal/flavours/flavours.go
```diff
@@ -0,0 +1,44 @@
...
+// Flavour is a type that represents a deployment flavour of Contrast.
+type Flavour int
+
+const (
+	// Unknown is the default value for the Flavour type.
+	Unknown Flavour = iota
+	// AKSCLHSNP represents a deployment with Cloud-Hypervisor on SEV-SNP AKS.
+	AKSCLHSNP
+	// BareMetalQEMUTDX represents a deployment with QEMU on bare-metal TDX.
+	BareMetalQEMUTDX
```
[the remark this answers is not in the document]
msanft: I'd advocate against this to not name-clash with Equinix Metal

node-installer/internal/constants/configuration-qemu-tdx.toml
msanft: Not sure, it's also the case in the [upstream config](https://github.com/kata-containers/kata-containers/blob/81d23a18651d31704beba67c0bc7d58781931f1f/src/runtime/Makefile#L756), but it shouldn't hurt to remove it.

msanft:
> Please also address the comments from https://github.com/edgelesssys/contrast/pull/603

I think we should keep the pod-role annotation unique, as it is - besides the container image URL - the only way to see which installer is used.

katexochen:
> I think we should keep the pod-role annotation unique, as it is - besides the container image URL - the only way to see which installer is used.

We can add another annotation with the string value of what is now called flavour instead.

PR #612, opened 2024-06-20
generate: add flag for aks reference values

davidweisse:
This allows for fields in the manifest to have the value `"null"`, in which case the attestation will fail when trying to parse the validation options from the manifest. This is helpful when generating the manifest for the first time with `contrast generate` where the SNP reference values are now left empty (`"null"`) by default. The default reference values used for AKS can then optionally be generated with the `--reference-values aks` flag.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -63,6 +63,7 @@ subcommands.`,
 	cmd.Flags().StringP("policy", "p", rulesFilename, "path to policy (.rego) file")
 	cmd.Flags().StringP("settings", "s", settingsFilename, "path to settings (.json) file")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
+	cmd.Flags().String("reference-values", "", "set the default reference values used for attestation (one of: aks)")
 	cmd.Flags().StringArrayP("workload-owner-key", "w", []string{workloadOwnerPEM}, "path to workload owner key (.pem) file")
 	cmd.Flags().BoolP("disable-updates", "d", false, "prevent further updates of the manifest")
 	cmd.Flags().String("image-replacements", "", "path to image replacements file")
@@ -115,6 +116,9 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	}
 
 	defaultManifest := manifest.Default()
+	if flags.referenceValues == "aks" {
+		defaultManifest = manifest.DefaultAKS()
+	}
 	defaultManifestData, err := json.MarshalIndent(&defaultManifest, "", "  ")
 	if err != nil {
 		return fmt.Errorf("marshaling default manifest: %w", err)
@@ -445,6 +449,7 @@ type generateFlags struct {
 	policyPath            string
 	settingsPath          string
 	manifestPath          string
+	referenceValues       string
 	workloadOwnerKeys     []string
 	disableUpdates        bool
 	workspaceDir          string
@@ -464,6 +469,13 @@ func parseGenerateFlags(cmd *cobra.Command) (*generateFlags, error) {
 	manifestPath, err := cmd.Flags().GetString("manifest")
 	if err != nil {
 		return nil, err
+	}
+	referenceValues, err := cmd.Flags().GetString("reference-values")
+	if err != nil {
+		return nil, err
+	}
+	if !slices.Contains([]string{"", "aks"}, referenceValues) {
+		return nil, fmt.Errorf("unknown reference values")
 	}
 	workloadOwnerKeys, err := cmd.Flags().GetStringArray("workload-owner-key")
 	if err != nil {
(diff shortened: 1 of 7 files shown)


cli/cmd/generate.go
```diff
@@ -515,6 +522,15 @@ func parseGenerateFlags(cmd *cobra.Command) (*generateFlags, error) {
 	}, nil
 }
 
+// getDefaultManifest returns the default manifest without reference values, or
+// it includes the default reference values for AKS.
+func getDefaultManifest(aksReferenceValues bool) manifest.Manifest {
```
burgerdev, line 527: This will need to change as soon as we introduce another set of default reference values. I'd just inline this into `runGenerate`.

cli/cmd/generate.go
```diff
@@ -63,6 +63,7 @@ subcommands.`,
 	cmd.Flags().StringP("policy", "p", rulesFilename, "path to policy (.rego) file")
 	cmd.Flags().StringP("settings", "s", settingsFilename, "path to settings (.json) file")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
+	cmd.Flags().Bool("aks-reference-values", false, "use the default reference values for AKS")
```
burgerdev, line 66: How about `--reference-values=(aks|default|...)`?

coordinator/internal/authority/authority_test.go
```diff
@@ -183,13 +183,13 @@ func newManifest(t *testing.T) (*manifest.Manifest, []byte, [][]byte) {
...
 	policyHash := sha256.Sum256(policy)
 	policyHashHex := manifest.NewHexString(policyHash[:])
-	mnfst := &manifest.Manifest{
-		Policies:                map[manifest.HexString][]string{policyHashHex: {"test"}},
-		WorkloadOwnerKeyDigests: []manifest.HexString{keyDigest},
-	}
+
+	mnfst := manifest.DefaultAKS()
+	mnfst.Policies = map[manifest.HexString][]string{policyHashHex: {"test"}}
+	mnfst.WorkloadOwnerKeyDigests = []manifest.HexString{keyDigest}
```
burgerdev, line 189: If we need to change this (I don't think we need to), could we use `manifest.Default` instead?

internal/manifest/manifest.go
```diff
@@ -38,37 +38,51 @@ type SNPReferenceValues struct {
...
 // SNPTCB represents a set of SNP TCB values.
 type SNPTCB struct {
-	BootloaderVersion SVN
-	TEEVersion        SVN
-	SNPVersion        SVN
-	MicrocodeVersion  SVN
+	BootloaderVersion NullSVN
+	TEEVersion        NullSVN
+	SNPVersion        NullSVN
+	MicrocodeVersion  NullSVN
```
burgerdev, line 44: How is this different from using `*SVN`?

davidweisse: I guess I was overthinking it. I thought we can avoid nil checks on the `UInt8` function this way, but I suppose we can just use a pointer since otherwise you would need to check if the value is valid before calling `UInt8`, as we are doing it now.

coordinator/internal/authority/authority_test.go
davidweisse: The default manifest does not include the reference values, so when testing `authority.SNPValidateOpts`, it would fail.

burgerdev: But... it's working now without reference values, isn't it? (Admitted, there's also no test.)

davidweisse: Before, the reference values would default to 0, but now they will default to `null`, in which case `manifest.SNPValidateOpts` returns an error when using `manifest.Default()`.

burgerdev: Understood, thanks.

cli/cmd/generate.go
```diff
@@ -63,6 +63,7 @@ subcommands.`,
 	cmd.Flags().StringP("policy", "p", rulesFilename, "path to policy (.rego) file")
 	cmd.Flags().StringP("settings", "s", settingsFilename, "path to settings (.json) file")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
+	cmd.Flags().String("reference-values", "default", "set the default reference values used for attestation (one of: default|aks)")
```
katexochen, line 66:
```suggestion
	cmd.Flags().String("reference-values", "", "set the default reference values used for attestation (one of: aks)")
```
If default isn't a case we want to explicitly select, I'd opt to use default when flag isn't set instead of making it a string opt.

cli/cmd/generate.go
```diff
@@ -63,6 +63,7 @@ subcommands.`,
 	cmd.Flags().StringP("policy", "p", rulesFilename, "path to policy (.rego) file")
 	cmd.Flags().StringP("settings", "s", settingsFilename, "path to settings (.json) file")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
+	cmd.Flags().String("reference-values", "", "set the default reference values used for attestation (one of: aks)")
```
burgerdev, line 66: `(one of: aks)` reads a bit strange. We could print a detailed list of embedded defaults with `--version`.

burgerdev (approved):

cli/cmd/generate.go
katexochen: @davidweisse please create a follow up ticket to explicitly print the relevant values in a separate section as AKS defaults.

PR #615, opened 2024-06-21
verify: verify active manifest at Coordinator

davidweisse:
The `verify` command already takes the manifest file as an input. On `verify`, the CLI will now check if the local manifest matches the active manifest on the Coordinator.

### DIFF
--- a/cli/cmd/verify.go
+++ b/cli/cmd/verify.go
@@ -4,6 +4,7 @@
 package cmd
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
@@ -99,7 +100,7 @@ func runVerify(cmd *cobra.Command, _ []string) error {
 	}
 	log.Debug("Got response")
 
-	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Successfully verified coordinator")
+	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Successfully verified Coordinator CVM based on reference values from manifest")
 
 	filelist := map[string][]byte{
 		coordRootPEMFilename: resp.RootCA,
@@ -118,6 +119,13 @@ func runVerify(cmd *cobra.Command, _ []string) error {
 	}
 
 	fmt.Fprintf(cmd.OutOrStdout(), "✔️ Wrote Coordinator configuration and keys to %s\n", filepath.Join(flags.workspaceDir, verifyDir))
+
+	currentManifest := resp.Manifests[len(resp.Manifests)-1]
+	if !bytes.Equal(currentManifest, manifestBytes) {
+		return fmt.Errorf("manifest active at Coordinator does not match expected manifest")
+	}
+
+	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Manifest active at Coordinator matches expected manifest")
 	fmt.Fprintln(cmd.OutOrStdout(), "  Please verify the manifest history and policies")
 
 	return nil
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -168,7 +168,7 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to marshal manifest: %w", err)
 	}
-	if err := os.WriteFile(flags.manifestPath, manifestData, 0o644); err != nil {
+	if err := os.WriteFile(flags.manifestPath, append(manifestData, '\n'), 0o644); err != nil {
 		return fmt.Errorf("failed to write manifest: %w", err)
 	}
 


cli/cmd/verify.go
```diff
@@ -96,7 +97,13 @@ func runVerify(cmd *cobra.Command, _ []string) error {
 	}
 	log.Debug("Got response")
 
-	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Successfully verified coordinator")
+	currentManifest := resp.Manifests[len(resp.Manifests)-1]
+	// Local manifest will have a newline if the user has edited the file.
```
katexochen, line 101: Why isn't that the case for the manifest we set?

msanft: This could (*Presumably*) be an editor-specific problem. I know that VSCode does append a newline to files by default upon editing. For the default manifest, if not edited, this should not be the case.  
I think we might want to strip all whitespace here for a canonical comparison.

burgerdev: I think we should be verifying the manifest byte-for-byte here, and rather fix the JSON encoding in `generate`.

davidweisse: The manifest is encoded in the GetManifest response, so it does not matter how the encoding is in the `generate` step. If we want to compare the manifest byte-for-byte, we need to add or remove a newline somewhere in the verify step or the GetManifest encoding, or convert to a canonical JSON encoding when comparing the manifests (e.g. `json.Compact`)

burgerdev: Ah, I see. This should not be the case anymore since #613 - can you try with that base?

davidweisse: Still an issue, since we are comparing the marshaled manifest to the bytes read from the file. I also noticed that the `SeedShareOwnerPubKeys` field is null after `generate` but then missing in the obtained history from `verify`...

burgerdev: > Still an issue, since we are comparing the marshaled manifest to the bytes read from the file.

There should be no marshalling anymore - the manifest should make the full round-trip as `[]byte`, and if it doesn't that's a bug I need to fix.

> I also noticed that the SeedShareOwnerPubKeys field is null after generate but then missing in the obtained history from verify...

This code path is not complete yet, I'm afraid.

davidweisse: No marshalling except for the initial marshalling in `generate`, so yeah, we can fix the encoding here.

> This code path is not complete yet, I'm afraid.

Forgot to rebuild, now it works.

But is it a good idea to compare byte-for-byte? The verification then depends on, e.g., the user's JSON indentation.

katexochen: canonicalization might be a good idea, feel free to create a ticket. I don't think the policies have a canonical form.

burgerdev: The `SetManifest` method [accepts manifest and policies as `[]byte`](https://github.com/edgelesssys/contrast/blob/afdc62e96a220d43f807fcfb28013904057abf9d/internal/userapi/userapi.proto#L13-L14) and the `GetManifests` method [returns them as `[]byte`](https://github.com/edgelesssys/contrast/blob/afdc62e96a220d43f807fcfb28013904057abf9d/internal/userapi/userapi.proto#L39-L40). We don't need to care about the user's JSON indentation if we ensure that the manifest can round trip without being recoded. 

> I don't think the policies have a canonical form.

They don't, but the policy hash requires them being byte-for-byte identical.

katexochen: > We don't need to care about the user's JSON indentation if we ensure that the manifest can round trip without being recoded

Changes could occur if the manifest is handed out-of-band from the workload owner to the relying party and then used on verify.

> They don't, but the policy hash requires them being byte-for-byte identical.

Canonicalization would be applied (if it would exist) before hashing, so that wouldn't be a problem.

burgerdev (approved):

cli/cmd/verify.go
```diff
@@ -96,7 +97,12 @@ func runVerify(cmd *cobra.Command, _ []string) error {
 	}
 	log.Debug("Got response")
 
-	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Successfully verified coordinator")
+	currentManifest := resp.Manifests[len(resp.Manifests)-1]
+	if bytes.Equal(currentManifest, manifestBytes) {
+		fmt.Fprintln(cmd.OutOrStdout(), "✔️ Successfully verified coordinator")
+	} else {
+		fmt.Fprintln(cmd.OutOrStdout(), "❌Manifest at Coordinator does not match local manifest")
```
katexochen, line 104: we should finally fail in this case

katexochen (changes requested):

katexochen:
I'm still confused by this change. If we have a manifest we trust, then we also already audited the policies referenced in this manifest. Why should we still write all the files on disk then and ask the user to audit it again?

burgerdev:
The `set` command does not unpack the policies, so it is somewhat convenient to get them through `verify`. But I agree that we should not ask the user to audit them.

katexochen:
> The `set` command does not unpack the policies, so it is somewhat convenient to get them through `verify`. But I agree that we should not ask the user to audit them.

Remember those two calls are (potentially) called from different entities! The data owner still has to review both the manifest and the policies. If the result of the verify should be "you definitely can trust this Coordinator, and no further steps are required" then we must assume the policies were also communicated out of band and the data owner already verified them (so there is no reason to output them). I think we should keep the output of files and state even more clear to the data owner what to do.

cli/cmd/verify.go
```diff
@@ -115,6 +116,13 @@ func runVerify(cmd *cobra.Command, _ []string) error {
 	}
 
 	fmt.Fprintf(cmd.OutOrStdout(), "✔️ Wrote Coordinator configuration and keys to %s\n", filepath.Join(flags.workspaceDir, verifyDir))
+
+	currentManifest := resp.Manifests[len(resp.Manifests)-1]
+	if !bytes.Equal(currentManifest, manifestBytes) {
+		return fmt.Errorf("active manifest at coordinator does not match expected manifest")
```
katexochen, line 122:
```suggestion
		return fmt.Errorf("manifest active at Coordinator does not match expected manifest")
```

katexochen (approved):
otherwise lgtm

msanft (approved):

PR #616, opened 2024-06-21
manifest: check for empty trusted measurement

davidweisse:
Previously, if the TrustedMeasurement field in the manifest was left empty, no error would be produced and the attestation would simply skip the measurement check. The correct way to check for the measurement would be to check if it is empty and not `nil`. In that case, the value is overridden with 48 zero bytes and the attestation should fail.

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -132,7 +132,7 @@ func (m *Manifest) SNPValidateOpts() (*validate.Options, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to convert TrustedMeasurement from manifest to byte slices: %w", err)
 	}
-	if trustedMeasurement == nil {
+	if len(trustedMeasurement) == 0 {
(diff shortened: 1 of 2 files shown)


burgerdev (commented):
If the trusted measurement is empty but non-nil, does this mean that the user explicitly configured it? Shouldn't we respect that?

katexochen:
> If the trusted measurement is empty but non-nil, does this mean that the user explicitly configured it? Shouldn't we respect that?

Why? To allow the user to disable fundamental security checks?

burgerdev:
I'd argue that a user who actively resets the measurement in the generated manifest should have the right to do so, but I see your point that this is a footgun with no concrete use case.

davidweisse:
If the user should not be able to disable the trusted measurement check, then #630 already covers this and we can close this PR.

burgerdev (approved):
Let's do this - if a use case should come up, allowing it should be easy enough.

PR #618, opened 2024-06-21
cli: derive runtime class name from launch digest

davidweisse:
Instead of injecting the runtime handler into the CLI, we can directly derive it from the trusted measurement.

### DIFF
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"context"
 	_ "embed"
+	"fmt"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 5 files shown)


cli/cmd/common.go
```diff
@@ -34,6 +35,8 @@ var (
 	defaultGenpolicySettings []byte
 	//go:embed assets/genpolicy-rules.rego
 	defaultRules []byte
+	// This value is injected at build time.
+	launchDigest = "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
msanft, line 39: This should be taken from the manifest, shouldn't it?

davidweisse: In the `generate` command, we don't have a manifest yet, so we need to embed this somewhere. We could use the embedded value from the manifest module and make this public, so we have a centralized place where this value comes from.

msanft: I think this might be a good idea. What do you think @katexochen?

msanft (approved):

PR #620, opened 2024-06-24
e2e: add genpolicy regression tests

burgerdev:


.github/workflows/e2e_genpolicy.yml
```diff
@@ -0,0 +1,65 @@
+name: genpolicy regression tests
+
+on:
+  schedule:
+    - cron: "16 6 * * 6" # 6:16 on Saturdays
```
katexochen, line 5: Should we combine this with the getdents regression test? For example, put the tests in one workflow, using one job per test?

internal/kuberesource/sets.go
```diff
@@ -151,6 +152,37 @@ func GetDEnts() ([]any, error) {
...
+				WithMatchLabels(map[string]string{"app.kubernetes.io/name": badLayer}),
+			).
+			WithTemplate(PodTemplateSpec().
+				WithLabels(map[string]string{"app.kubernetes.io/name": badLayer}).
+				WithSpec(PodSpec().
+					WithRuntimeClassName(runtimeHandler).
+					WithContainers(
+						Container().
+							WithName(badLayer).
+							WithImage("ghcr.io/edgelesssys/contrast/genpolicy-bad-layer:1").
```
katexochen, line 173: Where is the definition of this image?

burgerdev: This is a hand-crafted image, adding a known-bad layer to `busybox:3.16.1`. That one is pretty hard to reproduce.

On the other hand, I think this issue should be reproducible with a simple non-ASCII link name. In that case, I could write a Dockerfile for it. But where should we put that? Do we have a definition for the getdents image?

katexochen: > But where should we put that? Do we have a definition for the getdents image?

I'm fine with putting it into e2e/<testname>/Containerfile or into tools/e2e-containers. I don't think we have the source of the getdents image, but I'd be really happy to fix it.

burgerdev: I'm using an `httpd` image from dockerhub for now, don't think we need the source for that.

.github/workflows/e2e_genpolicy.yml
burgerdev: Done.

internal/kuberesource/sets.go
```diff
@@ -151,6 +152,36 @@ func GetDEnts() ([]any, error) {
 	return []any{tester}, nil
 }
 
+// GenpolicyRegressionTests returns deployments for regression testing genpolicy.
+func GenpolicyRegressionTests() map[string]*applyappsv1.DeploymentApplyConfiguration {
+	out := make(map[string]*applyappsv1.DeploymentApplyConfiguration)
+
+	badLayer := "bad-layer"
+	out[badLayer] = Deployment(badLayer, "").
```
katexochen, line 160: Can we link to the releated bug?

katexochen (approved):

internal/kuberesource/sets.go
burgerdev: Done

PR #629, opened 2024-06-24
ci: add reproducibility test for runtime components

msanft:
This tests whether the Contrast runtime components are reproducible between two distinct ubuntu-{20.04,22.04} systems.

Example Runs: https://github.com/edgelesssys/contrast/actions/runs/9644569372, https://github.com/edgelesssys/contrast/actions/runs/9646306170

### DIFF
--- a/.github/workflows/runtime-reproducibility.yml
+++ b/.github/workflows/runtime-reproducibility.yml
@@ -1,0 +1,76 @@
+name: verify runtime stack reproducibility
+
+on:
+    workflow_dispatch:
+    schedule:
+      - cron: '0 19 * * 0' # 7pm UTC on Sundays
+
+jobs:
+  os-matrix:
+    strategy:
+      matrix:
+        os: [ubuntu-22.04, ubuntu-20.04]
+        # As we do not use the Cachix cache containing the artifacts built by developers in this workflow,
+        # building the node-installer-image ensures that the whole transitive closure of the packages we build (i.e. everything
+        # except what's in cache.nixos.org). Therefore, building this ensures that the kernel, image, IGVM, runtime and the image itself
+        # is reproducible across individual builds (as the --rebuild flag is used, causing Nix to rebuild the node-installer-image derivation)
+        # and across independent builds on Ubuntu 20.04 and 22.04 (which also test the reproducibility of the transitive closure of our packages, as no shared
+        # cache is present between the two machines)
+        build-target: [contrast-node-installer-image]
+    # Usually we would define the matrix outputs here, but as GitHub Actions don't seem to allow per-combination outputs,
+    # we'll write the outputs without defining them here. See https://github.com/orgs/community/discussions/17245#discussioncomment-3814009.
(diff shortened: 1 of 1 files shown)


.github/workflows/runtime-reproducibility.yml
```diff
@@ -0,0 +1,73 @@
...
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - uses: nicknovitski/nix-develop@a2060d116a50b36dfab02280af558e73ab52427d # v1.1.0
+      - name: Build
+        id: build
+        run: |
+            nix build .#${{ matrix.build-target }} --option substituters https://cache.nixos.org --builders ""
+            reference_checksum="$(cat result/index.json | jq  -r '.manifests[0].digest')"
+            echo "reference-checksum-${{ matrix.os }}-${{ matrix.build-target}}=$reference_checksum" >> "$GITHUB_OUTPUT"
+
+            nix build .#${{ matrix.build-target }} --rebuild --option substituters https://cache.nixos.org --builders "" -o rebuild
```
katexochen, line 35: Using nix build --rebuild, we will only rebuild the referenced derivation, not its dependencies. Why are we doing this? I mean sure, makes sense to check that building the OCI is reproducible, but I think we'd expect most issues further down. Not saying this is bad, but we should document why we are doing this and what guarantees it provides us.

.github/workflows/runtime-reproducibility.yml
```diff
@@ -0,0 +1,73 @@
...
+                  if not checksum in seen:
+                      seen[checksum] = []
+                  seen[checksum].append(matrix)
+
+          if len(seen) > 1:
+            print("At least one checksum mismatched:")
+            print(json.dumps(seen, indent=2))
+            exit(1)
+
+          print("All checksums were equal")
```
katexochen, line 73: why do we need python here? check out https://github.com/edgelesssys/constellation/blob/main/.github/workflows/reproducible-builds.yml#L191 for a concise way of doing this with bash.

.github/workflows/runtime-reproducibility.yml
msanft: Didn't think about this

.github/workflows/runtime-reproducibility.yml
msanft: We don't have files, and if we want files, we'd need to write them from the output here first, which would require us to parse it, etc., or use artifact sharing between the jobs - both incurring quite a lot of overhead if done in bash

I think this is more readable, too. But that is my personal preference

.github/workflows/runtime-reproducibility.yml
katexochen: >  but we should document why we are doing this and what guarantees it provides us.

is still unaddressed.

katexochen (approved):
I think documentation of what the --rebuild step does (and doesn't) is still missing, otherwise lgtm.

PR #630, opened 2024-06-25
manifest: validate manifest fields

davidweisse:
This adds a function `Validate()` to the manifest to check if every field in the manifest struct contains a reasonable value or if any required fields are missing.

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -126,20 +126,60 @@ func (p Policy) Hash() HexString {
 	return NewHexString(hashBytes[:])
 }
 
+// Validate checks the validity of all fields in the reference values.
+func (r ReferenceValues) Validate() error {
+	if r.SNP.MinimumTCB.BootloaderVersion == nil {
+		return fmt.Errorf("field BootloaderVersion in manifest cannot be empty")
+	} else if r.SNP.MinimumTCB.TEEVersion == nil {
+		return fmt.Errorf("field TEEVersion in manifest cannot be empty")
+	} else if r.SNP.MinimumTCB.SNPVersion == nil {
+		return fmt.Errorf("field SNPVersion in manifest cannot be empty")
+	} else if r.SNP.MinimumTCB.MicrocodeVersion == nil {
+		return fmt.Errorf("field MicrocodeVersion in manifest cannot be empty")
+	}
+
+	if len(r.TrustedMeasurement) != abi.MeasurementSize*2 {
+		return fmt.Errorf("trusted measurement has invalid length: %d (expected %d)", len(r.TrustedMeasurement), abi.MeasurementSize*2)
+	}
+
+	return nil
+}
+
+// Validate checks the validity of all fields in the manifest.
+func (m *Manifest) Validate() error {
+	for policyHash := range m.Policies {
+		if _, err := policyHash.Bytes(); err != nil {
+			return fmt.Errorf("decoding policy hash %s: %w", policyHash, err)
+		} else if len(policyHash) != sha256.Size*2 {
+			return fmt.Errorf("policy hash %s has invalid length: %d (expected %d)", policyHash, len(policyHash), sha256.Size*2)
+		}
+	}
+
+	if err := m.ReferenceValues.Validate(); err != nil {
+		return fmt.Errorf("validating reference values: %w", err)
+	}
+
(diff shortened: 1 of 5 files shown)


internal/manifest/manifest.go
```diff
@@ -125,21 +125,53 @@ func (p Policy) Hash() HexString {
...
 }
 
+// Validate checks the validity of all fields in the manifest.
+func (m *Manifest) Validate() error {
+	for policyHash := range m.Policies {
+		if len(policyHash) != sha256.Size*2 {
+			return fmt.Errorf("policy hash %s has invalid length: %d (expected %d)", policyHash, len(policyHash), sha256.Size*2)
+		} else if _, err := hex.DecodeString(policyHash.String()); err != nil {
+			return fmt.Errorf("decoding policy hash %s: %w", policyHash, err)
+		}
```
burgerdev, line 135: I'd have a weak preference for decoding first and then checking length vs. `sha256.Size`.

internal/manifest/manifest.go
```diff
@@ -125,21 +125,53 @@ func (p Policy) Hash() HexString {
 	return NewHexString(hashBytes[:])
 }
 
+// Validate checks the validity of all fields in the manifest.
+func (m *Manifest) Validate() error {
+	for policyHash := range m.Policies {
+		if len(policyHash) != sha256.Size*2 {
+			return fmt.Errorf("policy hash %s has invalid length: %d (expected %d)", policyHash, len(policyHash), sha256.Size*2)
+		} else if _, err := hex.DecodeString(policyHash.String()); err != nil {
```
burgerdev, line 133:
```suggestion
		} else if _, err := policyHash.Bytes(); err != nil {
```

internal/manifest/manifest.go
```diff
@@ -125,21 +125,53 @@ func (p Policy) Hash() HexString {
...
+		return fmt.Errorf("field TEEVersion in manifest cannot be empty")
+	} else if tcb.SNPVersion == nil {
+		return fmt.Errorf("field SNPVersion in manifest cannot be empty")
+	} else if tcb.MicrocodeVersion == nil {
+		return fmt.Errorf("field MicrocodeVersion in manifest cannot be empty")
+	}
+
+	if len(m.ReferenceValues.TrustedMeasurement) != 96 {
+		return fmt.Errorf("trusted measurement has invalid length: %d (expected 96)", len(m.ReferenceValues.TrustedMeasurement))
+	}
```
burgerdev, line 151: How about making this `ReferenceValues.Validate`?

internal/manifest/manifest.go
```diff
@@ -125,21 +125,53 @@ func (p Policy) Hash() HexString {
...
+
+	for _, keyDigest := range m.WorkloadOwnerKeyDigests {
+		if len(keyDigest) != sha256.Size*2 {
+			return fmt.Errorf("workload owner key digest %s has invalid length: %d (expected %d)", keyDigest, len(keyDigest), sha256.Size*2)
+		} else if _, err := hex.DecodeString(keyDigest.String()); err != nil {
+			return fmt.Errorf("decoding key digest %s: %w", keyDigest, err)
+		}
+	}
```
burgerdev, line 159: Same comments as above

internal/manifest/manifest.go
```diff
@@ -126,21 +126,61 @@ func (p Policy) Hash() HexString {
...
+		return fmt.Errorf("field BootloaderVersion in manifest cannot be empty")
+	} else if r.SNP.MinimumTCB.TEEVersion == nil {
+		return fmt.Errorf("field TEEVersion in manifest cannot be empty")
+	} else if r.SNP.MinimumTCB.SNPVersion == nil {
+		return fmt.Errorf("field SNPVersion in manifest cannot be empty")
+	} else if r.SNP.MinimumTCB.MicrocodeVersion == nil {
+		return fmt.Errorf("field MicrocodeVersion in manifest cannot be empty")
+	}
+
+	if len(r.TrustedMeasurement) != 96 {
```
burgerdev, line 141: Nit: we should replace all raw 48s with [`abi.MeasurementSize`](https://pkg.go.dev/github.com/google/go-sev-guest/abi#pkg-constants).

burgerdev (approved):

burgerdev (approved):

PR #631, opened 2024-06-25
scripts: add tools to debug policy failures

katexochen:


packages/scripts.nix
```diff
@@ -225,4 +225,31 @@
...
+    name = "extract-policies";
+    runtimeInputs = with pkgs; [ yq-go ];
+    text = ''
+      set -euo pipefail
+      while read -r line; do
+          name=$(echo "$line" | cut -d' ' -f1)
+          namespace=$(echo "$line" | cut -d' ' -f2)
+          echo "Extracting policy for $namespace.$name" >&2
+          echo "$line" | cut -d' ' -f3 | base64 -d > "$namespace.$name.rego"
+      done < <(yq '.metadata.name + " " + .metadata.namespace // "default" + " " + .spec.template.metadata.annotations["io.katacontainers.config.agent.policy"]')
```
burgerdev, line 252: Nit: this won't work for pods. Not too bad, but maybe we can fake it with concatenation of 

```
.spec.template.metadata.annotations["io.katacontainers.config.agent.policy"]
.metadata.annotations["io.katacontainers.config.agent.policy"]
```

burgerdev (approved):

packages/scripts.nix
burgerdev: TIL - nice!

burgerdev (approved):

PR #633, opened 2024-06-26
generate: translate genpolicy logs, show warnings

katexochen:
Previously, we would collect the output with `RUST_LOG=info`, but only print it in case the invocation failed, thus swallowing any warnings. This PR introduces a log translator, that parses the output of genpolicy line by line and filters the output with regard to the currently configured log level. As a consequence, warnings are now shown by default, and when we configure debug logging in the Contrast CLI, the info logging of genpolicy is shown. However, in case genpolicy fails and the CLI debug level is info, we will now get slightly less output.

Example output (CLI loglevel info):

```
✔️ Patched targets
time=2024-06-26T08:03:50.284+02:00 level=WARN msg="Skipping symlink with long link name (NetLock_Arany_=Class_Gold=_Főtanúsítvány.pem, 48 bytes, NetLock_Arany_=Class_Gold=_Ftanstvny.pem, 40 bytes): etc/ssl/certs/988a38cb.0"
time=2024-06-26T08:03:50.284+02:00 level=WARN msg="Skipping symlink with long link name (/usr/share/ca-certificates/mozilla/NetLock_Arany_=Class_Gold=_Főtanúsítvány.crt, 83 bytes, /usr/share/ca-certificates/mozilla/NetLock_Arany_=Class_Gold=_Ftanstvny.crt, 75 bytes): etc/ssl/certs/NetLock_Arany_=Class_Gold=_Főtanúsítvány.pem"
time=2024-06-26T08:04:28.788+02:00 level=WARN msg="Failed to parse user as u32, using uid = 0 - error invalid digit found in string" position=genpolicy::registry
✔️ Generated workload policy annotations
✔️ Updated manifest workspace/manifest.json
```

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -4,6 +4,7 @@
 package cmd
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/ecdsa"
@@ -16,10 +17,12 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


msanft (approved):

cli/cmd/generate.go
```diff
@@ -388,6 +391,60 @@ func addWorkloadOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) e
...
+	}
+	l.startTranslate()
+	return l
+}
+
+func (l logTranslator) Write(p []byte) (n int, err error) {
+	return l.w.Write(p)
+}
+
+var genpolicyLogPrefixReg = regexp.MustCompile(`^\[[^\]\s]+\s+(\w+)\s+([^\]\s]+)\] (.*)`)
```
burgerdev, line 417: Could you please add a test for this regexp? That would also serve to document what it's supposed to capture.

cli/cmd/generate.go
```diff
@@ -388,6 +391,60 @@ func addWorkloadOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) e
...
+				l.logger.Warn(line)
+			} else {
+				switch match[1] {
+				case "ERROR":
+					l.logger.Error(match[3], "position", match[2])
+				case "WARN":
+					l.logger.Warn(match[3], "position", match[2])
+				case "INFO": // prints quite a lot, only show on debug
+					l.logger.Debug(match[3], "position", match[2])
+				}
```
burgerdev, line 437: How is this different from setting `RUST_LOG` based on the configured Contrast log level and then logging all of stderr?

burgerdev: sgtm

burgerdev (approved):
Thanks!

PR #634, opened 2024-06-26
cli: add recover command

katexochen:
This introduces a new command `recover` that let's you recover the Coordinator state after a restart.

TODO: link doc page about recovery.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -10,6 +10,7 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
+	"crypto/rsa"
 	"crypto/sha256"
 	"crypto/x509"
 	"encoding/hex"
@@ -67,7 +68,10 @@ subcommands.`,
 	cmd.Flags().StringP("settings", "s", settingsFilename, "path to settings (.json) file")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
 	cmd.Flags().String("reference-values", "", "set the default reference values used for attestation (one of: aks)")
-	cmd.Flags().StringArrayP("workload-owner-key", "w", []string{workloadOwnerPEM}, "path to workload owner key (.pem) file")
+	cmd.Flags().StringArrayP("workload-owner-key", "w", []string{workloadOwnerPEM},
+		"path to workload owner key (.pem) file (can be passed more than once)")
+	cmd.Flags().StringArray("seedshare-owner-key", []string{seedshareOwnerPEM},
+		"path to seedshare owner key (.pem) file (can be passed more than once)")
 	cmd.Flags().BoolP("disable-updates", "d", false, "prevent further updates of the manifest")
 	cmd.Flags().String("image-replacements", "", "path to image replacements file")
 	cmd.Flags().Bool("skip-initializer", false, "skip injection of Contrast Initializer")
@@ -117,6 +121,9 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	if err := generateWorkloadOwnerKey(flags); err != nil {
 		return fmt.Errorf("generating workload owner key: %w", err)
 	}
+	if err := generateSeedshareOwnerKey(flags); err != nil {
+		return fmt.Errorf("generating seedshare owner key: %w", err)
+	}
 
 	defaultManifest := manifest.Default()
 	if flags.referenceValues == "aks" {
@@ -149,6 +156,13 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 		}
 	}
 	slices.Sort(manifest.WorkloadOwnerKeyDigests)
+
+	for _, keyPath := range flags.seedshareOwnerKeys {
+		if err := addSeedshareOwnerKeyToManifest(manifest, keyPath); err != nil {
+			return fmt.Errorf("adding seedshare owner key to manifest: %w", err)
+		}
+	}
+	slices.Sort(manifest.SeedshareOwnerPubKeys)
 
 	manifestData, err = json.MarshalIndent(manifest, "", "  ")
 	if err != nil {
@@ -394,6 +408,31 @@ func addWorkloadOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) e
 	return nil
 }
 
+func addSeedshareOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) error {
+	keyData, err := os.ReadFile(keyPath)
+	if err != nil {
+		return fmt.Errorf("reading seedshare owner key: %w", err)
+	}
+	block, _ := pem.Decode(keyData)
+	if block == nil {
+		return errors.New("failed to decode PEM block")
+	}
+	var publicKey manifest.HexString
+	switch block.Type {
+	case "RSA PRIVATE KEY":
+		privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+		if err != nil {
+			return fmt.Errorf("parsing RSA private key: %w", err)
+		}
+		publicKey = manifest.MarshalSeedShareOwnerKey(&privateKey.PublicKey)
+	default:
+		return fmt.Errorf("unsupported PEM block type: %s", block.Type)
+	}
+	manifst.SeedshareOwnerPubKeys = append(manifst.SeedshareOwnerPubKeys, publicKey)
+
+	return nil
+}
+
 type logTranslator struct {
 	r         *io.PipeReader
 	w         *io.PipeWriter
@@ -492,13 +531,27 @@ func generateWorkloadOwnerKey(flags *generateFlags) error {
 	}
 	keyPath := flags.workloadOwnerKeys[0]
 
-	if err := createFileWithDefault(keyPath, newKeyPair); err != nil {
+	if err := createFileWithDefault(keyPath, newECDSAKeyPair); err != nil {
 		return fmt.Errorf("creating default workload owner key file: %w", err)
 	}
 	return nil
 }
 
-func newKeyPair() ([]byte, error) {
+func generateSeedshareOwnerKey(flags *generateFlags) error {
+	if len(flags.seedshareOwnerKeys) != 1 {
+		// No need to generate keys
+		// the user has provided a set of (presumably already generated) public keys
+		return nil
+	}
+	keyPath := flags.seedshareOwnerKeys[0]
+
+	if err := createFileWithDefault(keyPath, newRSAKeyPair); err != nil {
+		return fmt.Errorf("creating default seedshare owner key file: %w", err)
+	}
+	return nil
+}
+
+func newECDSAKeyPair() ([]byte, error) {
 	privateKey, err := ecdsa.GenerateKey(elliptic.P384(), rand.Reader)
 	if err != nil {
 		return nil, fmt.Errorf("generating private key: %w", err)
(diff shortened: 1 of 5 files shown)


cli/cmd/generate.go
```diff
@@ -64,6 +65,7 @@ subcommands.`,
 	cmd.Flags().StringP("settings", "s", settingsFilename, "path to settings (.json) file")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
 	cmd.Flags().StringArrayP("workload-owner-key", "w", []string{workloadOwnerPEM}, "path to workload owner key (.pem) file")
+	cmd.Flags().StringArray("seedshare-owner-key", []string{seedshareOwnerPEM}, "path to seedshare owner key (.pem) file")
```
msanft, line 68:
```suggestion
	cmd.Flags().StringArrayP("workload-owner-key", "w", []string{workloadOwnerPEM}, "paths to workload owner key (.pem) files")
	cmd.Flags().StringArray("seedshare-owner-key", []string{seedshareOwnerPEM}, "paths to seedshare owner key (.pem) files")
```

cli/cmd/generate.go
```diff
@@ -384,6 +396,32 @@ func addWorkloadOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) e
 	return nil
 }
 
+func addSeedshareOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) error {
```
msanft, line 399:
```suggestion
func addSeedshareOwnerKeyToManifest(manifest *manifest.Manifest, keyPath string) error {
```
Or is the naming on purpose to not collide with the package name? This should not be a compiler error though, as it is scope-aware

cli/cmd/set.go
```diff
@@ -237,7 +235,7 @@ func policyMapToBytesList(m []deployment) [][]byte {
 	return policies
 }
 
-func loadWorkloadOwnerKey(path string, manifst manifest.Manifest, log *slog.Logger) (*ecdsa.PrivateKey, error) {
+func loadWorkloadOwnerKey(path string, manifst *manifest.Manifest, log *slog.Logger) (*ecdsa.PrivateKey, error) {
```
msanft, line 238: Same here

msanft (approved):
I don't have the expertise to review the actual recovery logic itself, so I only took a look at the Go code

katexochen:
> I don't have the expertise to review the actual recovery logic itself, so I only took a look at the Go code

@msanft please take some time to familiarize yourself with the recovery logic.

cli/cmd/generate.go
burgerdev: The package is also needed further down, so we can't shadow it.

cli/cmd/generate.go
burgerdev: I would rather add `(can be passed more than once)` or something like that.

cli/cmd/generate.go
```diff
@@ -384,6 +396,32 @@ func addWorkloadOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) e
...
+	case "RSA PRIVATE KEY":
+		privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+		if err != nil {
+			return fmt.Errorf("parsing RSA private key: %w", err)
+		}
+		publicKey = x509.MarshalPKCS1PublicKey(&privateKey.PublicKey)
+	default:
+		return fmt.Errorf("unsupported PEM block type: %s", block.Type)
+	}
+	manifst.SeedshareOwnerPubKeys = append(manifst.SeedshareOwnerPubKeys, manifest.NewHexString(publicKey))
```
burgerdev, line 421: It would be good if we could use `manifest.MarshalSeedShareOwnerKey` so that we have a single place where algorithms are chosen.

cli/cmd/generate.go
```diff
@@ -384,6 +396,32 @@ func addWorkloadOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) e
...
+		return fmt.Errorf("reading seedshare owner key: %w", err)
+	}
+	block, _ := pem.Decode(keyData)
+	if block == nil {
+		return errors.New("failed to decode PEM block")
+	}
+	var publicKey []byte
+	switch block.Type {
+	case "PUBLIC KEY":
+		publicKey = block.Bytes
```
burgerdev, line 411: I'm not sure whether this is the right public key encoding. https://pkg.go.dev/crypto/x509#ParsePKCS1PublicKey says the PEM block should be `RSA PUBLIC KEY`, and https://pkg.go.dev/crypto/x509#ParsePKIXPublicKey says `PUBLIC KEY` but would be the wrong DER.

cli/cmd/generate.go
burgerdev, line 399: Since the function does not need the manifest as input, it might be better to return a `HexString, error` here and inline the `append`.

cli/cmd/recover.go
```diff
@@ -0,0 +1,218 @@
...
+			return nil, nil, fmt.Errorf("decoding seed share key: %w", err)
+		}
+		shareKey, err := x509.ParsePKCS1PublicKey(shareKeyBytes)
+		if err != nil {
+			return nil, nil, fmt.Errorf("parsing seed share key: %w", err)
+		}
+		if !pub.Equal(shareKey) {
+			continue
+		}
+		seed, err := rsa.DecryptOAEP(sha256.New(), nil, key, share.EncryptedSeed, []byte("seedshare"))
```
burgerdev, line 156: Please use `manifest.DecryptSeedShare`.

cli/cmd/recover.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	for _, share := range seedShareDoc.SeedShares {
+		shareKeyBytes, err := hex.DecodeString(share.PublicKey)
+		if err != nil {
+			return nil, nil, fmt.Errorf("decoding seed share key: %w", err)
+		}
+		shareKey, err := x509.ParsePKCS1PublicKey(shareKeyBytes)
+		if err != nil {
+			return nil, nil, fmt.Errorf("parsing seed share key: %w", err)
+		}
```
burgerdev, line 152: You could convert `pub` to `HexString` outside the loop and compare the hex.

cli/cmd/set.go
```diff
@@ -256,20 +254,25 @@ func loadWorkloadOwnerKey(path string, manifst manifest.Manifest, log *slog.Logg
...
+		ownerKeyHex := manifest.NewHexString(ownerKeyHash[:])
+		if len(manifst.WorkloadOwnerKeyDigests) == 0 {
+			log.Warn("No workload owner keys in manifest. Further manifest updates will be rejected by the coordinator")
+			return workloadOwnerKey, nil
+		}
+		log.Debug("Workload owner keys in manifest", "keys", manifst.WorkloadOwnerKeyDigests)
+		if !slices.Contains(manifst.WorkloadOwnerKeyDigests, ownerKeyHex) {
+			log.Warn("Workload owner key not found in manifest. This may lock you out from further updates")
+		}
 	}
```
burgerdev, line 274: Should this be part of the manifest validation? cc @davidweisse

cli/cmd/recover.go
```diff
@@ -0,0 +1,218 @@
...
+		RunE: withTelemetry(runRecover),
+	}
+
+	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
+	cmd.Flags().StringP("coordinator", "c", "", "endpoint the coordinator can be reached at")
+	must(cobra.MarkFlagRequired(cmd.Flags(), "coordinator"))
+	cmd.Flags().String("coordinator-policy-hash", DefaultCoordinatorPolicyHash, "override the expected policy hash of the coordinator")
+	cmd.Flags().String("workload-owner-key", workloadOwnerPEM, "path to workload owner key (.pem) file")
+	cmd.Flags().String("seedshare-owner-key", seedshareOwnerPEM, "private key to decrypt the seed share")
+	cmd.Flags().String("seed", seedSharesFilename, "file with the encrypted seed shares")
```
burgerdev, line 48: Some of these are leftovers, I assume.

cli/cmd/generate.go
katexochen: As my code used the pkix thingy before rebase, I'm pretty sure it would work. But that code is removed anyway now.

cli/cmd/generate.go
katexochen: We should follow up and also move the workload owner key stuff into the manifest package.

cli/cmd/generate.go
burgerdev: That's what I meant - it'd be using PKIX, but expected is PKCS1.

cli/cmd/generate.go
katexochen: wanted to keep it close to `addWorkloadOwnerKeyToManifest`. I've added a check if the key is already present, so now the manifest is used. ;)

burgerdev: We might want to allow duplicate seedshare owner keys when we eventually split the seed.

cli/cmd/recover.go
katexochen: Why? I think we need all those flags for recovery.

cli/cmd/set.go
davidweisse: If this is possibly wanted behavior, then I don't think so. Also, the manifest validation only checks if all fields contain valid data and shouldn't require any extra keys for comparison. We should instead check if the seed share owner pub keys are valid public keys.

cli/cmd/recover.go
burgerdev: Sorry, I forgot that we need the manifest for attestation config.

cli/cmd/recover.go
```diff
@@ -0,0 +1,210 @@
...
+	workloadOwnerKeyPath, err := cmd.Flags().GetString("workload-owner-key")
+	if err != nil {
+		return recoverFlags{}, err
+	}
+	manifestPath, err := cmd.Flags().GetString("manifest")
+	if err != nil {
+		return recoverFlags{}, err
+	}
+
+	return recoverFlags{
```
davidweisse, line 183: Can you prepend the workspace directory here if it is set, like in the other commands?

davidweisse:
I think we don't need this anymore: [recoveryapi.Port](https://github.com/edgelesssys/contrast/blob/b5707f20847c19bf1dc19c9a7a3d78680d5e6a0c/internal/recoveryapi/recoveryapi.go#L9).

cli/cmd/recover.go
```diff
@@ -0,0 +1,233 @@
...
+
+func decryptedSeedFromShares(seedSharesPath, seedShareOwnerKeyPath string) ([]byte, []byte, error) {
+	key, err := loadSeedshareOwnerKey(seedShareOwnerKeyPath)
+	if err != nil {
+		return nil, nil, err
+	}
+	pub, ok := key.Public().(*rsa.PublicKey)
+	if !ok {
+		return nil, nil, fmt.Errorf("could not get public key from seedshare owner key")
+	}
```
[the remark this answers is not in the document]
burgerdev: (or just inline below)

cli/cmd/set.go
```diff
@@ -256,20 +254,25 @@ func loadWorkloadOwnerKey(path string, manifst manifest.Manifest, log *slog.Logg
...
+		}
+		ownerKeyHash := sha256.Sum256(pubKey)
+		ownerKeyHex := manifest.NewHexString(ownerKeyHash[:])
+		if len(manifst.WorkloadOwnerKeyDigests) == 0 {
+			log.Warn("No workload owner keys in manifest. Further manifest updates will be rejected by the Coordinator")
+			return workloadOwnerKey, nil
+		}
+		log.Debug("Workload owner keys in manifest", "keys", manifst.WorkloadOwnerKeyDigests)
+		if !slices.Contains(manifst.WorkloadOwnerKeyDigests, ownerKeyHex) {
+			log.Warn("Workload owner key not found in manifest. This may lock you out from further updates")
```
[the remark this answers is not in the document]
katexochen: When else should we warn about it? During generate? What if a user edits the manifest manually?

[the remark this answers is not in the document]
katexochen: Hm, this is somewhat an edge case. Maybe Malte's intention was that if you own both keys (e.g. key rotation), you might not actually lock yourself out.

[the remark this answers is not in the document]
burgerdev: Sounds about right.

PR #639, opened 2024-06-26
recoveryapi: cleanup and remove

katexochen:
We ended up adding the recovery call to the userapi. There isn't really a need to have this in a separate proto then either, so this PR does some cleanup.


burgerdev (approved):
lgtm, thanks

PR #640, opened 2024-06-26
release: publish emojivoto as yml instead of zip

davidweisse:
This changes the released Emojivoto deployment from a zip file to a single YAML file.

### DIFF
--- a/docs/docs/examples/emojivoto.md
+++ b/docs/docs/examples/emojivoto.md
@@ -35,17 +35,11 @@ where their votes are processed without leaking to the platform provider or work
 
 ### Downloading the deployment
 
-The emojivoto deployment files are part of a zip file in the Contrast release. You can download the
+The emojivoto deployment files are part of Contrast release. You can download the
 latest deployment by running:
 
 ```sh
-curl -fLO https://github.com/edgelesssys/contrast/releases/latest/download/emojivoto-demo.zip
-```
-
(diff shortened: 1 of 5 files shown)


burgerdev (approved):
The change lgtm, but I wonder whether we should just generate all the released yaml with the CLI in the first place. cc @katexochen

docs/docs/examples/emojivoto.md
```diff
@@ -35,17 +35,11 @@ where their votes are processed without leaking to the platform provider or work
...
 
 ```sh
-curl -fLO https://github.com/edgelesssys/contrast/releases/latest/download/emojivoto-demo.zip
-```
-
-After that, unzip the `emojivoto-demo.zip` file to extract the `deployment/` directory.
-
-```sh
-unzip emojivoto-demo.zip
+curl -fLO https://github.com/edgelesssys/contrast/releases/latest/download/emojivoto-demo.yml
```
katexochen, line 42: I think we should still download this into the deployment dir. The way it is right now, the demodir + e2e won't fit the docs.

katexochen:
> The change lgtm, but I wonder whether we should just generate all the released yaml with the CLI in the first place. cc @katexochen

We have tickets for that, but iirc we decided to postpone implementation.

katexochen (approved):

PR #641, opened 2024-06-26
packages: bare-metal node-installer image

msanft:
This adds a container image for a bare-metal node-installer (called `node-installer-kata`, as it's build is derived from the Kata upstream sources more directly than the Microsoft one)

One thing to notice is that we currently source QEMU through a binary package from the Ubuntu sources, as it's hard to built a QEMU that can be used on non-Nix systems from Nix.


internal/kuberesource/parts.go
```diff
@@ -47,7 +47,7 @@ func NodeInstaller(namespace string) *NodeInstallerConfig {
 					WithHostPID(true).
 					WithInitContainers(Container().
 						WithName("installer").
-						WithImage("ghcr.io/edgelesssys/contrast/node-installer:latest").
+						WithImage("ghcr.io/edgelesssys/contrast/node-installer-microsoft:latest").
```
katexochen, line 50: One thing that is still unclear to me is how we gonna know which node-installer-image supports which platform variant.

packages/by-name/contrast/package.nix
```diff
@@ -5,9 +5,9 @@
 , buildGoModule
 , buildGoTest
 , microsoft
+, runtime ? microsoft
```
katexochen, line 8: I don't think we want to make this configurable, but rather embed multiple reference value sets in case in case we support multiple. So passing microsoft should be enough.

packages/by-name/contrast/package.nix
```diff
@@ -27,9 +27,9 @@ let
     subPackages = [ "e2e/genpolicy" "e2e/getdents" "e2e/openssl" "e2e/servicemesh" "e2e/release" ];
   };
 
-  launchDigest = builtins.readFile "${runtime-class-files}/launch-digest.hex";
+  launchDigest = builtins.readFile "${runtime.runtime-class-files}/launch-digest.hex";
```
burgerdev, line 30: Does this mean we need to build dedicated for an environment? The idea was to allow multiple reference value sets (and multiple defaults, afaict) in a single binary.

packages/by-name/kata/runtime-class-files/package.nix
```diff
@@ -0,0 +1,43 @@
...
+, kata
+, fetchurl
+, OVMF
+}:
+
+let
+  image = kata.kata-image;
+  kernel = "${kata.kata-kernel-uvm}/bzImage";
+
+  # TODO(msanft): building a non-NixOS QEMU is hard, investigate later and pin it for now.
```
burgerdev, line 14: That's fine, but please document where the binary is from.

packages/by-name/kata/runtime-class-files/package.nix
```diff
@@ -0,0 +1,43 @@
...
+  qemu-bin = fetchurl {
+    url = "https://cdn.confidential.cloud/contrast/node-components/1718800762/qemu-system-x86_64";
+    sha256 = "sha256-7MS/tK6q4D8y/FH6VcfARQLhIuvtNP6TsGfy+0o9kSc=";
+  };
```
katexochen, line 18: In the guest, we solve this by copying the closure into the image. In theory, we could do the same on the host, right?

packages/by-name/kata/runtime-class-files/package.nix
```diff
@@ -0,0 +1,43 @@
...
+  inherit (kata.kata-image) version;
+
+  dontUnpack = true;
+
+  # TODO(msanft): perform the actual launch digest calculation.
+  buildPhase = ''
+    mkdir -p $out
+    sha256sum ${image} ${kernel} ${qemu-bin} ${containerd-shim-contrast-cc-v2} | sha256sum | cut -d " " -f 1 > $out/launch-digest.hex
+    echo -n "contrast-cc-" > $out/runtime-handler
+    cat $out/launch-digest.hex | head -c 32 >> $out/runtime-handler
```
burgerdev, line 37: Nit: I find constructs like this easier to read

```sh
printf "contrast-cc-%s" "$(cat $out/launch-digest.hex | head -c 32)" > $out/runtime-handler
```

packages/by-name/kata/runtime-class-files/package.nix
katexochen, line 1: Please adapt to changes made to the msft version.

packages/scripts.nix
```diff
@@ -159,12 +159,13 @@
...
     text = ''
       imageRef=$1
+      platform=$2
 
       tmpdir=$(mktemp -d)
       trap 'rm -rf $tmpdir' EXIT
 
       echo "ghcr.io/edgelesssys/contrast/coordinator:latest=$imageRef" > "$tmpdir/image-replacements.txt"
-      resourcegen --image-replacements "$tmpdir/image-replacements.txt" --add-load-balancers coordinator > "$tmpdir/coordinator_base.yml"
+      resourcegen --image-replacements "$tmpdir/image-replacements.txt" --add-load-balancers --platform $platform coordinator > "$tmpdir/coordinator_base.yml"
```
katexochen, line 168: The coordinator is still platform independent, right? I think this snippet shouldn't take platform input.

.github/workflows/release.yml
```diff
@@ -272,9 +272,10 @@ jobs:
       - name: Create portable coordinator resource definitions
         run: |
           mkdir -p workspace deployment
-          nix run .#scripts.write-coordinator-yaml -- "${coordinatorImgTagged}" > workspace/coordinator.yml
-          nix shell .#contrast --command resourcegen --image-replacements ./image-replacements.txt --namespace kube-system runtime > workspace/runtime.yml
-          nix shell .#contrast --command resourcegen --image-replacements ./image-replacements.txt --add-load-balancers emojivoto-sm-ingress > deployment/emojivoto-demo.yml
+          # TODO(msanft): Release for all platforms once stable
+          nix run .#scripts.write-coordinator-yaml -- "${coordinatorImgTagged}" AKS-CLH-SNP > workspace/coordinator.yml
+          nix shell .#contrast --command resourcegen --image-replacements ./image-replacements.txt --namespace kube-system --platform AKS-CLH-SNP runtime > workspace/runtime.yml
+          nix shell .#contrast --command resourcegen --image-replacements ./image-replacements.txt --add-load-balancers --platform AKS-CLH-SNP emojivoto-sm-ingress > deployment/emojivoto-demo.yml
```
burgerdev, line 278: Is the `--platform` flag already implemented in `resourcegen`?

justfile
```diff
@@ -1,5 +1,5 @@
 # Undeploy, rebuild, deploy.
-default target=default_deploy_target cli=default_cli: soft-clean coordinator initializer openssl port-forwarder service-mesh-proxy node-installer runtime (apply "runtime") (deploy target cli) set verify (wait-for-workload target)
+default target=default_deploy_target platform=default_platform cli=default_cli: soft-clean coordinator initializer openssl port-forwarder service-mesh-proxy (node-installer platform) (runtime platform) (apply "runtime") (deploy target cli) set verify (wait-for-workload target)
```
burgerdev, line 2: Does the CLI built here match the platform already?

msanft: The CLI should be platform-agnostic. Or what do you mean exactly?

.github/workflows/release.yml
msanft: This will be implemented in a follow-up, where I implement the actual logic in the node-installer and the generation for its deployment files. So theoretically, we could leave it out of this PR for now.

justfile
```diff
@@ -21,13 +21,29 @@ service-mesh-proxy: (push "service-mesh-proxy")
...
 
+# Build the node-installer, containerize and push it.
+node-installer platform=default_platform:
+    #!/usr/bin/env bash
+    case {{ platform }} in
+        "AKS-CLH-SNP")
+            just push "node-installer-microsoft"
+            exit 0
+        ;;
+        "BareMetal-QEMU-TDX")
```
katexochen, line 37: This is not a platform as defined by the platforms package.

justfile
```diff
@@ -21,13 +21,29 @@ service-mesh-proxy: (push "service-mesh-proxy")
...
+            just push "node-installer-kata"
+            exit 0
```
katexochen, line 39:
```suggestion
        ;;
        "BareMetal-QEMU-TDX")
            just push "node-installer-kata"
```

katexochen (changes requested):

packages/by-name/contrast/package.nix
msanft: The node-installer is platform-agnostic, but the container image we ship for it (including the runtime files) would get unreasonably large if we'd include all platforms.

justfile
burgerdev: If the CLI is platform agnostic my comment is void. Continuing in the other thread.

packages/by-name/kata/runtime-class-files/package.nix
msanft: Agreed in person to do this in a follow-up PR

packages/by-name/kata/runtime-class-files/package.nix
msanft: Not really feasible here, as we need the let binding to make the attributes available in the mkDerivation as well as the passthrough scope

internal/kuberesource/parts.go
msanft: Microsoft -> AKS
Kata -> Bare-Metal

Or what's the question?

katexochen: Kata -> K3sQEMUTDX + RKE2QEMUTDX, as there is no "bare-metal" platform.

msanft: There will only be a Single bare-Metal Node-installer though

katexochen (approved):

PR #644, opened 2024-06-27
e2e: add recovery test

burgerdev:


davidweisse (approved):

katexochen:
I think we should also restart one of the workload pods and repeat the openssl test logic, so we are assured that a pod joining after the recovery is able to talk to a pod from before the recovery.

PR #650, opened 2024-06-27
generate: check for existing public key in manifest

davidweisse:
If you call `generate` twice with the same seed share owner public key, the public key would just have been appended to the list instead of checking if the public key is already present in the manifest. Now, calling `generate` twice will result in the same manifest.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -65,8 +65,8 @@ subcommands.`,
 	cmd.Flags().String("reference-values", "", "set the default reference values used for attestation (one of: aks)")
 	cmd.Flags().StringArrayP("add-workload-owner-key", "w", []string{workloadOwnerPEM},
 		"add a workload owner key from a PEM file to the manifest (pass more than once to add multiple keys)")
-	cmd.Flags().StringArray("seedshare-owner-key", []string{seedshareOwnerPEM},
-		"path to seedshare owner key (.pem) file (can be passed more than once)")
+	cmd.Flags().StringArray("add-seedshare-owner-key", []string{seedshareOwnerPEM},
+		"add a seedshare owner key from a PEM file to the manifest (pass more than once to add multiple keys)")
 	cmd.Flags().BoolP("disable-updates", "d", false, "prevent further updates of the manifest")
 	cmd.Flags().String("image-replacements", "", "path to image replacements file")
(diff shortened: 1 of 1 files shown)


katexochen:
> I'm not sure about this. If I generate twice like this:
> 
> ```shell
> contrast generate --seedshare-owner-key=alice.pem deploy.yml
> contrast generate --seedshare-owner-key=bob.pem deploy.yml
> ```
> 
> Would I really expect both alice and bob to be in the manifest? I think it might be less surprising if we cleared out the keys and then added what's given as flag.

We should  match the behavior of workload owner keys (or update it, too).

davidweisse:
> I'm not sure about this. If I generate twice like this:
> 
> ```shell
> contrast generate --seedshare-owner-key=alice.pem deploy.yml
> contrast generate --seedshare-owner-key=bob.pem deploy.yml
> ```
> 
> Would I really expect both alice and bob to be in the manifest? I think it might be less surprising if we cleared out the keys and then added what's given as flag.

This is the current behavior for workload owner keys, as far as I can see.

Freax13:
> I'm not sure about this. If I generate twice like this:
> 
> ```shell
> contrast generate --seedshare-owner-key=alice.pem deploy.yml
> contrast generate --seedshare-owner-key=bob.pem deploy.yml
> ```
> 
> Would I really expect both alice and bob to be in the manifest? I think it might be less surprising if we cleared out the keys and then added what's given as flag.

We could rename the flag to something like `--add-seedshare-owner-key` to make it more explicit to the user that we only add keys and don't clear them.

katexochen:
> We could rename the flag to something like `--add-seedshare-owner-key` to make it more explicit to the user that we only add keys and don't clear them.

I say that's a good solution, wdyt @burgerdev ?

burgerdev:
Calling the flags `--add-*-key` should manage expectations, I agree, so if there's no concrete use-case for idempotency of generate runs I won't object. When we get to actually implement seed sharing, I wonder what the UX would look like for many keys, though - maybe users would rather modify the manifest directly?

burgerdev (approved):

Freax13:
I also applied the change suggested in https://github.com/edgelesssys/contrast/pull/670#discussion_r1665199505 to the `add-seedshare-owner-key` flag here.

PR #651, opened 2024-06-28
node-installer: support bare-metal platforms

msanft:
This adds support for the K3s- / RKE2-based bare-metal TDX deployment platforms in the node-installer. Depending on which platform is provided to the node-installer command line, different units are restarted, configs are written, etc. This does not yet add support of writing the appropriate commandline in the node-installer deployment, as this will be done in a follow-up PR. Thus, for now, to keep main healthy and deploy-able, we default to AKS-CLH-SNP in the node-installer if no platform is specified.

### DIFF
--- a/node-installer/node-installer.go
+++ b/node-installer/node-installer.go
@@ -6,6 +6,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"flag"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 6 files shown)


node-installer/node-installer.go
```diff
@@ -173,40 +291,48 @@ func restartHostContainerd(containerdConfigPath string) error {
...
 	// Note that "--timestamp=unix" is not supported in the installed version of systemd (v250) at the time of writing.
-	containerdStartTime, err := exec.Command(
+	serviceStartTime, err := exec.Command(
 		"nsenter", "--target", "1", "--mount", "--",
-		"systemctl", "show", "--timestamp=utc", "--property=ActiveEnterTimestamp", "containerd",
+		"systemctl", "show", "--timestamp=utc", "--property=ActiveEnterTimestamp", service,
 	).CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("getting containerd start time: %w %q", err, containerdStartTime)
+		return fmt.Errorf("getting service start time: %w %q", err, serviceStartTime)
```
katexochen, line 299: include service name

katexochen (changes requested):

PR #652, opened 2024-06-28
resourcegen: deploy platform-specific node-installer 

msanft:
This is the follow-up mentioned in https://github.com/edgelesssys/contrast/commit/b324c2f29113a4e449bb2d358e46d4e1232f3ecd which adds the proper templating support for the runtime deployment to the resourcegen tool. The tool now takes a platform flag, which is used to define which node-installer should be included in the runtime deployment.

As the runtime deployment now provides the node-installer with the correct platform flag, the fallback to AKS-CLH-SNP is not required anymore and can be removed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,11 @@ go 1.22.0
 
 toolchain go1.22.4
 
+replace github.com/edgelesssys/contrast/node-installer => ./node-installer
+
 require (
 	filippo.io/keygen v0.0.0-20230306160926-5201437acf8e
+	github.com/edgelesssys/contrast/node-installer v0.0.0-00010101000000-000000000000
 	github.com/google/go-github/v62 v62.0.0
 	github.com/google/go-sev-guest v0.11.1
 	github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus v1.0.1
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -4,8 +4,10 @@
 package kuberesource
 
 import (
+	"fmt"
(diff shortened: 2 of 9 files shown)


go.mod
```diff
@@ -4,8 +4,11 @@ go 1.22.0
 
 toolchain go1.22.4
 
+replace github.com/edgelesssys/contrast/node-installer => ./node-installer
```
katexochen, line 7: As previously discussed in-person, I don't think using a replace here makes any sense. If you need that code as part of the main module, just make it part of the main module.

msanft: See the rationale behind this in https://github.com/edgelesssys/contrast/commit/249d627b443ba28cd81a66a4ff66243b4a108006

I think this is the best we can currently get to not have to rebuild the node-installer each time contrast changes.

internal/kuberesource/parts.go
```diff
@@ -31,9 +33,21 @@ type NodeInstallerConfig struct {
...
 
 // NodeInstaller constructs a node installer daemon set.
-func NodeInstaller(namespace string) *NodeInstallerConfig {
+func NodeInstaller(namespace string, platform platforms.Platform) (*NodeInstallerConfig, error) {
 	name := "contrast-node-installer"
 
+	var podRole, imageURL string
+	switch platform {
+	case platforms.AKSCloudHypervisorSNP:
+		podRole = "contrast-node-installer-microsoft"
```
katexochen, line 42: I still think we should split this, keep the role as is and instead introduce an additional platform annotation.

katexochen (approved):

PR #654, opened 2024-06-30
dev-docs: fix links in life-of-a-confidential-container.md

blenessy:


burgerdev (approved):
Thanks for fixing these, much appreciated!

PR #655, opened 2024-07-01
ca: include SubjectKeyId and AuthorityKeyId in certificates

burgerdev:
We accidentally used the certificate templates instead of the final certificates for signing intermediate and leaf, which caused the `*KeyId` fields to be missing. These fields are mandatory in TLSv3, and our certs did not pass strict verification accordingly.

This PR also adds additional tests for the recovery scenario, enforces `x509_strict` on the OpenSSL tests and removes unnecessary CA and template fields.

### DIFF
--- a/e2e/openssl/openssl_test.go
+++ b/e2e/openssl/openssl_test.go
@@ -155,7 +155,7 @@ func TestOpenSSL(t *testing.T) {
 			stdout, stderr, err := c.ExecDeployment(ctx, ct.Namespace, opensslFrontend, []string{"/bin/bash", "-c", opensslConnectCmd("openssl-backend:443", "mesh-ca.pem")})
 			t.Log("openssl with wrong certificates:", stdout)
 			require.Error(err)
-			require.Contains(stderr, "certificate signature failure")
+			require.Contains(stderr, "self-signed certificate in certificate chain")
 
 			// Connect from backend to fronted, because the frontend does not require client certs.
 			// This should succeed because the root cert did not change.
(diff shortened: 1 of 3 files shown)


e2e/openssl/openssl_test.go
```diff
@@ -120,6 +120,22 @@ func TestOpenSSL(t *testing.T) {
...
+		require := require.New(t)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+		defer cancel()
+
+		c := kubeclient.NewForTest(t)
+
+		require.NoError(c.WaitFor(ctx, kubeclient.Deployment{}, ct.Namespace, opensslBackend))
+
+		cmd := `openssl verify -x509_strict -CAfile /tls-config/coordinator-root-ca.pem -untrusted /tls-config/certChain.pem /tls-config/certChain.pem`
```
katexochen, line 133: Could you add a comment similar to the one on the previous test case explaining what exactly we are checking here (and why it isn't covered by the previous commend)? I always find these openssl commands hard to understand.

katexochen (approved):

e2e/openssl/openssl_test.go
burgerdev: Thanks for the suggestion - after reading through the manpages again, I realized that I can add `x509_strict` to the `s_client` command, which should then be equivalent to the checks I introduced here. Removing this test case again and adding the option to the `opensslConnectCmd`.

katexochen (approved):

PR #656, opened 2024-07-01
flake: add more packages to dev shell

Freax13:
These packages are used in some Just targets. To make onboarding easier they should be part of the devshell.


katexochen (approved):
If we want to add more to our dev shell, we might want a second shell for CI runs or pick the required dependencies per-job so we don't have to download stuff we don't need.

PR #660, opened 2024-07-02
cli: move workload & seedshare owner key handling into manifest

Freax13:
This PR removes code related to workload and seedshare key handling from the CLI subcommands and consolidates it into the manifest package. This PR also adds some unit tests for the new functions.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -8,10 +8,8 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
-	"crypto/x509"
 	"encoding/hex"
 	"encoding/json"
-	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
@@ -395,23 +393,11 @@ func addSeedshareOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) 
 	if err != nil {
 		return fmt.Errorf("reading seedshare owner key: %w", err)
 	}
-	block, _ := pem.Decode(keyData)
-	if block == nil {
-		return errors.New("failed to decode PEM block")
-	}
-	var publicKey manifest.HexString
-	switch block.Type {
-	case "RSA PRIVATE KEY":
-		privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
-		if err != nil {
-			return fmt.Errorf("parsing RSA private key: %w", err)
-		}
-		publicKey = manifest.MarshalSeedShareOwnerKey(&privateKey.PublicKey)
-	default:
-		return fmt.Errorf("unsupported PEM block type: %s", block.Type)
+	publicKey, err := manifest.ExtractSeedshareOwnerPublicKey(keyData)
+	if err != nil {
+		return fmt.Errorf("extracting seed share public key: %w", err)
 	}
 	manifst.SeedshareOwnerPubKeys = append(manifst.SeedshareOwnerPubKeys, publicKey)
-
 	return nil
 }
 
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -374,25 +374,9 @@ func addWorkloadOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string) e
 	if err != nil {
 		return fmt.Errorf("reading workload owner key: %w", err)
 	}
-	block, _ := pem.Decode(keyData)
-	if block == nil {
-		return errors.New("failed to decode PEM block")
-	}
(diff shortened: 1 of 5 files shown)


cli/cmd/generate.go
```diff
@@ -415,21 +393,11 @@ func addSeedshareOwnerKeyToManifest(manifst *manifest.Manifest, keyPath string)
...
-	switch block.Type {
-	case "RSA PRIVATE KEY":
-		privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
-		if err != nil {
-			return fmt.Errorf("parsing RSA private key: %w", err)
-		}
-		publicKey = manifest.MarshalSeedShareOwnerKey(&privateKey.PublicKey)
-	default:
-		return fmt.Errorf("unsupported PEM block type: %s", block.Type)
+	privateKey, err := manifest.ParseSeedshareOwnerPrivateKey(keyData)
```
burgerdev, line 396: Could this be more similar to the workload owner key function, like `ExtractSeedshareOwnerPublicKey`?

internal/manifest/crypto_test.go
```diff
@@ -96,3 +96,29 @@ func newKey(t *testing.T, bits int) *rsa.PrivateKey {
...
+}
+
+func TestWorkloadOwnerKeyParseMarshal(t *testing.T) {
+	keyData, err := NewWorkloadOwnerKey()
+	require.NoError(t, err)
+
+	privateKey, err := ParseWorkloadOwnerPrivateKey(keyData)
+	require.NoError(t, err)
+
+	HashWorkloadOwnerKey(&privateKey.PublicKey)
```
burgerdev, line 120: Does this need testing, e.g. `size > 0`?

internal/manifest/crypto_test.go
```diff
@@ -96,3 +96,29 @@ func newKey(t *testing.T, bits int) *rsa.PrivateKey {
...
+
+	_, err = ExtractWorkloadOwnerPublicKey(keyData)
```
burgerdev, line 122: Could we add a test for the `PUBLIC KEY` case?

burgerdev (commented):
Nice, thanks!

internal/manifest/crypto_test.go
Freax13: Sure.

internal/manifest/crypto_test.go
Freax13: My main intention with putting this in the test was to make sure the function won't panic, but sure, we can also test for the length.

internal/manifest/crypto_test.go
```diff
@@ -96,3 +98,43 @@ func newKey(t *testing.T, bits int) *rsa.PrivateKey {
...
+	keyData, err := NewSeedShareOwnerPrivateKey()
+	require.NoError(t, err)
+
+	key, err := ParseSeedshareOwnerPrivateKey(keyData)
+	require.NoError(t, err)
+
+	pubHexStr := MarshalSeedShareOwnerKey(&key.PublicKey)
+
+	_, err = ParseSeedShareOwnerKey(pubHexStr)
+	require.NoError(t, err)
```
burgerdev, line 112: This looks like a leftover (the function is not defined?).

internal/manifest/crypto_test.go
```diff
@@ -96,3 +98,43 @@ func newKey(t *testing.T, bits int) *rsa.PrivateKey {
...
+
+	publicKeyHexStr, err := ExtractSeedshareOwnerPublicKey(keyData)
+	require.NoError(t, err)
+
+	publicKeyBytes, err := publicKeyHexStr.Bytes()
+	require.NoError(t, err)
+	publicKeyPem := pem.EncodeToMemory(&pem.Block{Type: "PUBLIC KEY", Bytes: publicKeyBytes})
+	_, err = ExtractSeedshareOwnerPublicKey(publicKeyPem)
```
burgerdev, line 120: Nit: we could also test that the extracted public key is the same in both cases.

burgerdev (approved):

internal/manifest/crypto_test.go
burgerdev: Nevermind, missed that it already existed.

Freax13: The function is defined [here](https://github.com/edgelesssys/contrast/blob/81bbc6bf5acd92a1230abfffd55e25a617ae08bf/internal/manifest/crypto.go#L75).

PR #661, opened 2024-07-02
docs: use --reference-values aks flag on generate

Freax13:
Without the --reference-values flag, the generate command will produce an invalid manifest.

### DIFF
--- a/docs/docs/deployment.md
+++ b/docs/docs/deployment.md
@@ -159,7 +159,7 @@ After that, it will generate the execution policies and add them as annotations 
 A `manifest.json` with the reference values of your deployment will be created.
 
 ```sh
-contrast generate resources/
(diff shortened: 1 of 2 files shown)


docs/docs/troubleshooting.md
```diff
@@ -89,7 +89,7 @@ To ensure there isn't a mismatch between Kubernetes resource YAML and the annota
 policies, rerun
 
 ```sh
-contrast generate
```
katexochen, line 92: If we re-run, we shouldn't need to re-initialize the reference values, right?

katexochen (commented):
Please prefix commit title and PR title with `docs: ...`, for example

`docs: use --reference-values aks flag on generate`

katexochen (approved):

PR #665, opened 2024-07-03
e2e: add readiness probes to emojivoto pods

burgerdev:
We sometimes see EOF or other connectivity problems in e2e tests, which suggest server unavailability. It turns out that we don't have a lot of readiness probes in the emojivoto demo, so the `WaitFor` calls in the e2e test are not quite sufficient to ensure availability. My suspicion is that this started getting visible after #573, which had the (side) effect of serializing container startup.

* Adds HTTP readiness probe to the Emojivoto frontend.
* Adds TCP readiness probes to the Emojivoto backends. 
  - Ideally, these would be gRPC probes, but the demo does not register a health server.  
  - Next preferred would be HTTP probes on the metrics endpoint, but the service mesh enforces mTLS and I can't easily change that because we rely on that in tests.
* Removes the egress-only service mesh variant, which is not actively used and somewhat broken since we introduced sidecar injection. Not necessary for the purpose of the PR, but I was touching these files anyway.

### DIFF
--- a/internal/kuberesource/mutators.go
+++ b/internal/kuberesource/mutators.go
@@ -79,8 +79,6 @@ type serviceMeshMode string
 const (
 	// ServiceMeshIngressEgress sets the service mesh mode to ingress and egress.
 	ServiceMeshIngressEgress serviceMeshMode = "service-mesh-ingress-egress"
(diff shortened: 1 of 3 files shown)


Freax13 (approved):
Looks good to me!

PR #668, opened 2024-07-03
just: refactor e2e target, use undeploy target for e2e

katexochen:
- Add support for undeploying e2e runs with `just undeploy`
- Simplify e2e target. This allows running all existing e2e tests, at the cost of slightly worse error message if the test does not exist. Run undeploy as a dependency, remove special undeploy handling.

### DIFF
--- a/justfile
+++ b/justfile
@@ -44,26 +44,13 @@ node-installer platform=default_platform:
         ;;
     esac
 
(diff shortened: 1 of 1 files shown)


Freax13 (approved):
LGTM!

PR #670, opened 2024-07-03
generate: rename --workload-owner-key to --add-workload-owner-key

Freax13:
This makes it more obvious to the user that this flag only ever adds keys and never removes existing keys.

This change was made following the discussion in https://github.com/edgelesssys/contrast/pull/650#pullrequestreview-2145590619.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -63,8 +63,8 @@ subcommands.`,
 	cmd.Flags().StringP("genpolicy-cache-path", "c", layersCacheFilename, "path to cache for the cache (.json) file containing the image layers")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
 	cmd.Flags().String("reference-values", "", "set the default reference values used for attestation (one of: aks)")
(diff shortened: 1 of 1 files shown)


cli/cmd/generate.go
```diff
@@ -63,7 +63,7 @@ subcommands.`,
 	cmd.Flags().StringP("genpolicy-cache-path", "c", layersCacheFilename, "path to cache for the cache (.json) file containing the image layers")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
 	cmd.Flags().String("reference-values", "", "set the default reference values used for attestation (one of: aks)")
-	cmd.Flags().StringArrayP("workload-owner-key", "w", []string{workloadOwnerPEM},
+	cmd.Flags().StringArrayP("add-workload-owner-key", "w", []string{workloadOwnerPEM},
 		"path to workload owner key (.pem) file (can be passed more than once)")
```
burgerdev, line 67: The docstring should describe the behavior. Maybe

> add a workload owner key from a PEM file to the manifest (pass more than once to add multiple keys)

burgerdev (approved):

PR #673, opened 2024-07-03
packages/contrast{,-releases}: install shell completions

miampf:


katexochen:
completion for releases isn't working yet as it requires https://github.com/edgelesssys/contrast/commit/d61a1168a96aa8fe60041d6485b6a1f60ecad3e6 which wasn't released yet.

katexochen (approved):

PR #685, opened 2024-07-05
packages: package bare-metal kata-runtime tool

msanft:
This packages a statically compiled version of the `kata-runtime` command line tool that's especially useful for debugging into Kata CVMs. On AKS, this is not necessary as of now, as the nodes contain the upstream version of the tool, and including the statically compiled version woul only bloat the node-installer image.

### DIFF
--- a/packages/by-name/kata/contrast-node-installer-image/package.nix
+++ b/packages/by-name/kata/contrast-node-installer-image/package.nix
@@ -58,6 +58,10 @@ let
               url = "file:///opt/edgeless/bin/containerd-shim-contrast-cc-v2";
               path = "/opt/edgeless/${runtime-handler}/bin/containerd-shim-contrast-cc-v2";
             }
+            {
+              url = "file:///opt/edgeless/bin/kata-runtime";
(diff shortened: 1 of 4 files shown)


packages/by-name/kata/contrast-node-installer-image/package.nix
```diff
@@ -115,6 +119,15 @@ let
     ];
   };
 
+  kata-runtime = ociLayerTar {
```
katexochen, line 122: I think we can place this in the same layer as containerd-shim (maybe use use kata-runtime as name), as those files come from the same build and will likely be updated togehter.

packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -44,9 +42,14 @@ buildGoModule rec {
     git
   ];
 
-  CGO_ENABLED = 0;
+  CGO_ENABLED = 1;
```
katexochen, line 45:
```suggestion
```

Is the default.

katexochen (changes requested):

katexochen:
@msanft needs rebase

katexochen (approved):

PR #690, opened 2024-07-08
issuer: don't open sev-guest device

Freax13:
We only used that device to query the Product, but we don't actually need the device to do that. Not opening the device devices saves us from having to mount it in the first place.


katexochen (approved):
Nice!

PR #694, opened 2024-07-09
microsoft.genpolicy: use git format-patch for patches

katexochen:


packages/by-name/microsoft/genpolicy/package.nix
```diff
@@ -31,12 +31,16 @@ rustPlatform.buildRustPackage rec {
     patches = [
       # TODO(burgerdev): drop after Microsoft ported https://github.com/kata-containers/kata-containers/pull/9706
       (fetchpatch {
-        name = "genpolicy_device_support.patch";
+        name = "0001-genpolicy-add-rules-and-types-for-volumeDevices.patch";
```
burgerdev, line 34: I'll take a note to vendor this patch, too.

burgerdev (approved):
Thanks!

packages/by-name/microsoft/genpolicy/package.nix
katexochen: If we want that I can easily do it now. I thought it wound't be necessary to vendor it.

katexochen: Ah wait, the patch we are fetching isn't part of the upstream repo, got it.

PR #697, opened 2024-07-10
node-installer: configure and run tardev-snapshotter

katexochen:
With this change, we improve the stability of Contrast by deploying our own tardev-snapshotter on the host as part of the node-installer deployment. Previously, the Contrast runtime would use the tardev-snapshotter installed on the host, which is managed by Azure. However, the snapshotter is highly coupled with the guest image, and breaking changes to the snaphotter led to workload pods not being able to start in Contrast. Therefore, we are shipping a pinned snapshotter as part of the runtime. This will also allow running multiple Contrast runtime next to each other that require different versions of the snapshotter.

### DIFF
--- a/node-installer/node-installer.go
+++ b/node-installer/node-installer.go
@@ -199,19 +199,23 @@ func patchContainerdConfig(runtimeName, basePath, configPath string, platform pl
 		existing = constants.ContainerdBaseConfig()
 	}
 
+	snapshotterName := "no-snapshotter"
 	// Add tardev snapshotter, only required for AKS
 	if platform == platforms.AKSCloudHypervisorSNP {
 		if existing.ProxyPlugins == nil {
 			existing.ProxyPlugins = make(map[string]config.ProxyPlugin)
 		}
-		if _, ok := existing.ProxyPlugins["tardev"]; !ok {
-			existing.ProxyPlugins["tardev"] = constants.TardevSnapshotterConfigFragment()
+		snapshotterName = fmt.Sprintf("tardev-%s", runtimeName)
+		socketName := fmt.Sprintf("/run/containerd/tardev-snapshotter-%s.sock", runtimeName)
+		existing.ProxyPlugins[snapshotterName] = config.ProxyPlugin{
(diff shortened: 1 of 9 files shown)


Freax13 (commented):
One of the commit messages has a typo:

```diff
-  microsoft.genpolicy: don't check laryer-src-prefix 
+  microsoft.genpolicy: don't check layer-src-prefix 
```

node-installer/node-installer.go
```diff
@@ -199,19 +199,23 @@ func patchContainerdConfig(runtimeName, basePath, configPath string, platform pl
 		existing = constants.ContainerdBaseConfig()
 	}
 
+	snapshotterName := "tardev"
```
Freax13, line 202: In the function below you use `"no-snapshotter"` when no snapshotter is needed? Should we also use this here?

Freax13 (approved):

node-installer/node-installer.go
```diff
@@ -199,19 +199,23 @@ func patchContainerdConfig(runtimeName, basePath, configPath string, platform pl
 		existing = constants.ContainerdBaseConfig()
 	}
 
+	snapshotterName := "no-snapshotter"
```
burgerdev, line 202: Why is this not `tardev-$hash`?

katexochen: It is set to that a few lines below in case we are on a platform that uses a snapshotter.

burgerdev: Ah, I see - thanks.

burgerdev (approved):

node-installer/node-installer.go
burgerdev: Interestingly, we're already switching on the platform outside these functions: https://github.com/edgelesssys/contrast/blob/59826da4deb184be42231a0b0bbf52634080a8db/node-installer/node-installer.go#L128-L141.

katexochen: That's indeed confusing (but necessary with the current structure, as some platforms won't call patchContainerdConfig.

PR #702, opened 2024-07-10
nix: build static QEMU

Freax13:
This is based on nixpkgs PRs #30070, #309929, and #314998.


katexochen (approved):
Can we add the upstream source of a patch also in a comment where we reference it, including a note if/why it was modified?

Otherwise lgtm, great work!

PR #710, opened 2024-07-11
ci: fix missing "current version" in commit message from `bump_version` action

daniel-weisse:
The action was missing an initialization of the `current` variable.
This leads to the current version being omitted from the commit message.
See https://github.com/edgelesssys/contrast/commit/f7c10e613c8b495b744cb3e377008bd57531bf41 for example.

Load current version from `version.txt` before writing the new version to fix this.


katexochen (approved):
Oh wow, this has been broken for quite some time. :sweat_smile: Good catch and thanks for the fix!

PR #719, opened 2024-07-12
docs: move "Components" to subpage and other minor fixes

thomasten:
A category page should not provide essential information because it can easily be overlooked.
(For category pages, both the generated index or a summary of the subpages are acceptable.)


burgerdev (approved):
lgtm, thanks for backporting the fixes.

docs/versioned_docs/version-0.6/components/overview.md
katexochen, line 1: IMO such improvements shouldn't be backported to release versions of the docs. Users might link to parts of this page, moving it will break links and expectations.

katexochen (commented):
I'm not happy with this change, I liked the content-full category pages and we were actually moving into the direction of having this for more categories. The index pages are pretty useless and especially annoying when using the "next" button on the bottom of a page.
I think the actual issue is that the arrow button of a category will uncollaps it without showing the category page. If you click on the category title, the category page is shown so you can't miss it in that case.

burgerdev:
We had this discussion about Constellation docs yesterday. The UX of tree navigation is just not great - it turns out that some people don't expect useful content in non-leaf tabs. It might be a better idea overall to make the groups _not clickable_ and not render the automatic pages.

docs/versioned_docs/version-0.6/components/overview.md
thomasten: Agreed. Reverted this change. (Though I think people will rarely link to versions because you can't get a permalink to the current version :(

katexochen:
> We had this discussion about Constellation docs yesterday. The UX of tree navigation is just not great - it turns out that some people don't expect useful content in non-leaf tabs. It might be a better idea overall to make the groups _not clickable_ and not render the automatic pages.

I was thinking the same, but didn't find an easy way to disable the category page completely.

thomasten:
If the pages can't be disabled and we want to avoid the index pages, I'd prefer pages like
https://docs.edgeless.systems/constellation/getting-started/examples
https://docs.edgeless.systems/constellation/overview/performance
that summarize the category.

katexochen:
What about the landing page? That's also a category page currently.

If we want to go back to overview-only category pages, we should also use the feature to autogenerate them again instead of placing empty index.md files everywhere.

thomasten:
> make the groups _not clickable_ and not render the automatic pages

:heavy_check_mark:

katexochen (approved):
Thanks!

PR #721, opened 2024-07-12
golangci-lint: add more linters

katexochen:

PR #722, opened 2024-07-12
meshapi: follow best practice for metric names

katexochen:
Counters should have a _total suffix, so we are renaming
`contrast_meshapi_attestation_failures` -> `contrast_meshapi_attestation_failures_total`.

This will also be enforced by https://github.com/edgelesssys/contrast/pull/721 in the future.

### DIFF
--- a/coordinator/meshapi.go
+++ b/coordinator/meshapi.go
@@ -49,7 +49,7 @@ func newMeshAPIServer(meshAuth *authority.Authority, bundleGetter certBundleGett
 
 	attestationFailuresCounter := promauto.With(reg).NewCounter(prometheus.CounterOpts{
 		Subsystem: "contrast_meshapi",
-		Name:      "attestation_failures",
+		Name:      "attestation_failures_total",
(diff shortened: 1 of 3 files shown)


coordinator/meshapi.go
```diff
@@ -49,7 +49,7 @@ func newMeshAPIServer(meshAuth *authority.Authority, bundleGetter certBundleGett
 
 	attestationFailuresCounter := promauto.With(reg).NewCounter(prometheus.CounterOpts{
 		Subsystem: "contrast_meshapi",
-		Name:      "attestation_failures",
+		Name:      "attestation_failures_total",
```
Freax13, line 52: Have you considered creating a constant containing this value? We could reuse the constant in the tests.

Freax13 (approved):

coordinator/meshapi.go
katexochen: Might be a good idea if we want to use this in more places in the future, or must reference more label names elsewhere. But I think in general building up tests on metrics as insight isn't that good of a design.

PR #726, opened 2024-07-15
e2e: fix reproducibility test, don't fail fast

katexochen:
Run: https://github.com/edgelesssys/contrast/actions/runs/9935664109


msanft:
I'm not exactly sure about not failing fast. If one of the matrix instantiations fails, the comparison (and thus the entire test) is going to fail too, isn't it?

katexochen:
> I'm not exactly sure about not failing fast. If one of the matrix instantiations fails, the comparison (and thus the entire test) is going to fail too, isn't it?

Yes, but we can still check whether the image builds/rebuilds on a single node.

msanft (approved):

PR #732, opened 2024-07-17
coordinator: use random key for intermediate CA

burgerdev:
When we introduced Coordinator recovery, we decided to derive all key material deterministically from the seed. This implies that whoever owns the seed can derive these keys offline, too. However, since the seed needs to be held by the workload owner for recovery, the data owner can't trust in the mesh CA cert without trusting the workload owner.

This PR reverts to a random mesh/intermediate key, which is inaccessible to the workload owner. The trade-off is that the mesh key changes after recovery, and all workloads need to be restarted in order to pick up this change.


e2e/openssl/openssl_test.go
```diff
@@ -201,15 +201,24 @@ func TestOpenSSL(t *testing.T) {
...
+			if err != nil {
+				t.Logf("openssl with %q after recovery:\n%s", rootCAFile, stdout)
+			}
+			assert.NoError(t, err, "stderr: %q", stderr)
+		})
+
+		require.NoError(c.Restart(ctx, kubeclient.Deployment{}, ct.Namespace, opensslBackend))
+		require.NoError(c.WaitFor(ctx, kubeclient.Deployment{}, ct.Namespace, opensslBackend))
+
+		t.Run("mesh CA after coordinator recovery is accepted when workloads are restarted", func(t *testing.T) {
```
katexochen, line 215: Should we also check mesh CA is not accepted before the restart?

katexochen (approved):
lgtm, thanks!

e2e/openssl/openssl_test.go
burgerdev: Makes sense, added a test.

edgelessci:
Successfully created backport PR for `release/v0.8`:
- #733

PR #738, opened 2024-07-17
attestation: add TDX issuer

msanft:
This adds an attestation issuer for TDX and makes the node-installer work with both SNP and TDX. See commit messages for more detailed descriptions.

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -26,8 +26,8 @@ environment variables `CONTRAST_LOG_LEVEL`, `CONTRAST_LOG_FORMAT` and
   format (defaults to `text`).
 - `CONTRAST_LOG_SUBSYSTEMS` is a comma-seperated list of subsystems that should
   be enabled for logging, which are disabled by default. Subsystems include:
-  `snp-issuer`, `kds-getter`, and `snp-validator`. To enable all subsystems, use
-  `*` as the value for this environment variable.
+   `kds-getter`, `issuer` and `validator`.
+  To enable all subsystems, use `*` as the value for this environment variable.
   Warnings and error messages from subsystems get printed regardless of whether
   the subsystem is listed in the `CONTRAST_LOG_SUBSYSTEMS` environment variable.
 
--- a/node-installer/node-installer.go
+++ b/node-installer/node-installer.go
@@ -63,26 +63,18 @@ func run(ctx context.Context, fetcher assetFetcher, platform platforms.Platform,
 	if err := config.Validate(); err != nil {
 		return fmt.Errorf("validating config: %w", err)
 	}
+	fmt.Printf("Using config: %+v\n", config)
 
 	runtimeBase := filepath.Join("/opt", "edgeless", config.RuntimeHandlerName)
-	binDir := filepath.Join(hostMount, runtimeBase, "bin")
 
-	// Create directory structure
-	if err := os.MkdirAll(binDir, os.ModePerm); err != nil {
-		return fmt.Errorf("creating runtime bin directory: %w", err)
-	}
-	if err := os.MkdirAll(filepath.Join(hostMount, runtimeBase, "share/qemu"), os.ModePerm); err != nil {
-		return fmt.Errorf("creating runtime share directory: %w", err)
-	}
-	if err := os.MkdirAll(filepath.Join(hostMount, runtimeBase, "etc"), os.ModePerm); err != nil {
-		return fmt.Errorf("creating runtime etc directory: %w", err)
-	}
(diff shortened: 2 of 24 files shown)


packages/by-name/kata/runtime-class-files/package.nix
```diff
@@ -32,18 +54,19 @@ stdenvNoCC.mkDerivation {
   # TODO(msanft): perform the actual launch digest calculation.
   buildPhase = ''
     mkdir -p $out
-    sha256sum ${image} ${kernel} ${qemu-bin} ${qemu-share}/kvmvapic.bin ${qemu-share}/linuxboot_dma.bin ${qemu-share}/efi-virtio.rom ${containerd-shim-contrast-cc-v2} ${ovmf} | sha256sum | cut -d " " -f 1 > $out/launch-digest.hex
+    sha256sum ${image} ${kernel} ${qemu-snp.bin} ${qemu-tdx.bin} ${containerd-shim-contrast-cc-v2} ${ovmf-snp} ${ovmf-tdx} | sha256sum | cut -d " " -f 1 > $out/launch-digest.hex
```
Freax13, line 57: This removes the files in `share/` from the (admittedly fake) launch digest.

Freax13 (approved):

packages/by-name/kata/kata-runtime/0003-runtime-agent-verify-the-agent-policy-hash.patch
katexochen, line 1: Please don't edit patches by hand. Apply the patches, do your modification and recreate the patch set instead.

packages/by-name/kata/runtime-class-files/package.nix
msanft: I'm aware of this, but it can only ever actually get relevant if we have a change in the QEMU share directories, but not the QEMU binary itself, which I don't expect to happen before we get an actual launch digest. I think it'd be excessive to measure all of the components of both here.

packages/by-name/kata/kata-runtime/0003-runtime-agent-verify-the-agent-policy-hash.patch
msanft: I just noticed that this patch doesn't apply on the 3.6.0 tree at all

docs/docs/troubleshooting.md
```diff
@@ -26,8 +26,8 @@ environment variables `CONTRAST_LOG_LEVEL`, `CONTRAST_LOG_FORMAT` and
   format (defaults to `text`).
 - `CONTRAST_LOG_SUBSYSTEMS` is a comma-seperated list of subsystems that should
   be enabled for logging, which are disabled by default. Subsystems include:
-  `snp-issuer`, `kds-getter`, and `snp-validator`. To enable all subsystems, use
-  `*` as the value for this environment variable.
+  `snp-issuer`, `tdx-issuer`, `kds-getter`, `snp-validator` and `tdx-validator`.
+  To enable all subsystems, use `*` as the value for this environment variable.
```
katexochen, line 30: I think this is getting confusing. Maybe we can set the attestation type as attribute (`WithAttrs()`) instead of creating new named loggers for each attestation variant?

docs/docs/troubleshooting.md
katexochen, line 1: Can be squashed into prev commit.

docs/docs/troubleshooting.md
msanft: I agree. Should we do this in a separate PR?

katexochen: let's do it in this one, as we must touch the docs anyway.

katexochen:
```
reading bin directory "/host/opt/edgeless/contrast-cc-d8ebd654bdea606472dbb3f5e34215f9/tdx/bin": open /host/opt/edgeless/contrast-cc-d8ebd654bdea606472dbb3f5e34215f9/tdx/bin: no such file or directory
```

node-installer/node-installer.go
```diff
@@ -93,35 +85,47 @@ func run(ctx context.Context, fetcher assetFetcher, platform platforms.Platform,
...
-	}
 
-	for _, item := range items {
-		if !item.Type().IsRegular() {
-			continue
+	// Fix-up the permissions of executables
+	binDirs := []string{
+		filepath.Join(hostMount, runtimeBase, "bin"),
+		filepath.Join(hostMount, runtimeBase, "tdx", "bin"),
+		filepath.Join(hostMount, runtimeBase, "snp", "bin"),
```
katexochen, line 93: This must be constructed based on what platform we are running on.

katexochen (approved):

PR #742, opened 2024-07-18
dev-docs: document patching conventions

katexochen:


dev-docs/patches.md
```diff
@@ -0,0 +1,95 @@
...
+  (fetchpatch {
+    url = "https://github.com/...";
+    stripLen = 2; # Sometimes the patch strip level must be configured
+    hash = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";
+  })
+];
+```
+
+If there aren't any changes needed to an upstream patch, it's good practice to fetch the patch from remote. If the changes to upstream patches are needed or patches aren't available upstream, the
+*full patch set* should be vendored, so it can easily be worked with.
```
burgerdev, line 52: Can you define "full patch set" here?

dev-docs/patches.md
```diff
@@ -0,0 +1,95 @@
...
+patches = [
+  (fetchpatch {
+    url = "https://github.com/...";
+    stripLen = 2; # Sometimes the patch strip level must be configured
+    hash = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";
+  })
+];
+```
+
+If there aren't any changes needed to an upstream patch, it's good practice to fetch the patch from remote. If the changes to upstream patches are needed or patches aren't available upstream, the
```
burgerdev, line 51: I'd argue that a patch needs to be in some reliable branch (main, release/xxx) to qualify for remote inclusion, to minimize the chance of losing it due to garbage collection etc.

dev-docs/patches.md
```diff
@@ -0,0 +1,95 @@
...
+
+When done, recreate the patch set:
+
+```sh
+git format-patch -o $pkgDir $rev
+```
+
+# Patch documentation conventions
+
+Patches need thorough documentation. Each reference of a patch must have a comment.
```
burgerdev, line 88: > Each reference of a patch must have a comment.

You mean a comment in the `.patch` file, after the `---`?

burgerdev (approved):

dev-docs/patches.md
katexochen: While I see the point of garbage collection on the GitHub side, I don't think this is too big of a problem. From my experience they don't run aggressive gc on repos and commit stay around for quite some time. The reason I'd rather not restrict this further is that it is quite common to reference a commit from an open PR that we've just opened, which is ok imo.

Actually did some research, seems like at least in some scenarios dangling commits are kept around for quite some time: https://stackoverflow.com/questions/4367977/how-to-remove-a-dangling-commit-from-github

dev-docs/patches.md
katexochen: tried to clarify this

dev-docs/patches.md
katexochen: I've added an example and some reasoning about it.

dev-docs/patches.md
burgerdev: Clear now, thanks!

dev-docs/patches.md
burgerdev: I like how it's phrased now, and thanks for doing the research.

burgerdev (approved):

PR #749, opened 2024-07-19
atls: try to attest with all validators

msanft:
Our aTLS code needs to be adjusted to allow for multiple attestation variants. While we did allow one to have multiple validators before, we tried to validate the first matching one and instantly errored out if validation for that validator failed. Now, we can have multiple matching validators and try all until one successfully validated the document, or until the list of applicable validators is exhausted, where we can return an error.

### DIFF
--- a/internal/atls/atls.go
+++ b/internal/atls/atls.go
@@ -21,6 +21,7 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/edgelesssys/contrast/internal/attestation"
 	"github.com/edgelesssys/contrast/internal/crypto"
 )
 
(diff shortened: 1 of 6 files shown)


Freax13 (approved):
LGTM!

internal/atls/atls.go
```diff
@@ -205,19 +211,56 @@ func processCertificate(rawCerts [][]byte, _ [][]*x509.Certificate) (*x509.Certi
...
+	if !foundExtension {
+		return ErrNoValidAttestationExtensions
+	}
+
+	if !foundMatchingValidator {
+		return ErrNoMatchingValidators
+	}
+
+	// If we're here, an error must've happened during validation.
+	return
```
burgerdev, line 263:
```suggestion
	return retErr
```
https://go.dev/wiki/CodeReviewComments#named-result-parameters

burgerdev (approved):

PR #750, opened 2024-07-19
just: add get-credentials-tdxbm

katexochen:


msanft (approved):

katexochen:
cc @burgerdev

justfile
```diff
@@ -224,6 +224,18 @@ get-credentials-ci:
         --name "contrast-ci" \
         --admin
 
+get-credentials-tdxbm:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    tmpConfig=$(mktemp)
+    gcloud secrets versions access \
+        projects/796962942582/secrets/m50-ganondorf-kubeconf/versions/2 \
+        --out-file="$tmpConfig"
```
burgerdev, line 233: Do we really need this in the `justfile`? It sounds like it should be part of Github Actions or something.

justfile
```diff
@@ -224,6 +224,18 @@ get-credentials-ci:
...
+    mergedConfig=$(mktemp)
+    export KUBECONFIG=$tmpConfig:~/.kube/config
+    kubectl config view --flatten > $mergedConfig
+    mv $mergedConfig ~/.kube/config
```
burgerdev, line 237: This should respect an existing `KUBECONFIG` var.

justfile
katexochen: It enables devs to quickly setup the needed creds, without setting up ssh keys and searching for kube config in other users home directory. Only requirement is to be in the tailnet (should add a command with this requirement).

burgerdev: On the other hand, this is only useful to a particular group of devs - those who have access to this gcp secret. I simply wonder how much org-specific things we should put into OSS, but feel free to go ahead with it.

justfile
katexochen: I can use 
```sh
export KUBECONFIG=$tmpConfig:${KUBECONFIG:-~/.kube/config}
```
for respecting it as input, but how can I respect it when writing the config? Or is it okay to always update .kube/config, even if KUBECONFIG != .kube/config?

justfile
```diff
@@ -224,6 +224,18 @@ get-credentials-ci:
...
+    #!/usr/bin/env bash
+    set -euo pipefail
+    tmpConfig=$(mktemp)
+    gcloud secrets versions access \
+        projects/796962942582/secrets/m50-ganondorf-kubeconf/versions/2 \
+        --out-file="$tmpConfig"
+    mergedConfig=$(mktemp)
+    export KUBECONFIG=$tmpConfig:${KUBECONFIG:-~/.kube/config}
+    kubectl config view --flatten > $mergedConfig
+    mv $mergedConfig ~/.kube/config
```
burgerdev, line 237:
```suggestion
    env KUBECONFIG=$tmpConfig:${KUBECONFIG:-~/.kube/config} kubectl config view --flatten > $mergedConfig
    mv $mergedConfig "${KUBECONFIG}"
```

justfile
burgerdev: Made a suggestion.

justfile
katexochen: What if KUBECONFIG=./foo:./bar?

justfile
```diff
@@ -224,6 +224,18 @@ get-credentials-ci:
...
+
```
katexochen, line 238: Reminder to add a step to set active context.

justfile
burgerdev: `${KUBECONFIG%%:*}`?

justfile
```diff
@@ -224,6 +224,20 @@ get-credentials-ci:
...
+    tmpConfig=$(mktemp)
+    gcloud secrets versions access \
+        projects/796962942582/secrets/m50-ganondorf-kubeconf/versions/2 \
+        --out-file="$tmpConfig"
+    mergedConfig=$(mktemp)
+    KUBECONFIG_BAK=${KUBECONFIG:-~/.kube/config}
+    KUBECONFIG=$tmpConfig:${KUBECONFIG_BAK} kubectl config view --flatten > $mergedConfig
+    export newContext=$(yq -r '.contexts.[0].name' $tmpConfig)
+    yq -i '.current-context = env(newContext)' $mergedConfig
+    mv $mergedConfig ${KUBECONFIG_BAK%%:*:}
```
burgerdev, line 239: Did you put the final `:` deliberately? My suggestion was meant to take everything up to exclduing the first `:` char.

burgerdev (approved):

justfile
katexochen: no idea how that slipped in

burgerdev (approved):

PR #752, opened 2024-07-19
cli: embed multiple reference values

msanft:
This adds support for embedding a more versatile format of reference values (i.e. a structured type) into the Contrast binaries. This will allow us to embed all reference values at build-time from a single source (the Nix build file) rather than having SVNs in Go code and inserting trusted measurements via the go build commandline. It will now embed a JSON file containing the reference values, which is unmarshaled at first default manifest generation.


internal/manifest/constants.go
```diff
@@ -3,32 +3,31 @@
...
-		},
+	"github.com/edgelesssys/contrast/node-installer/platforms"
+)
+
+// Default returns a default manifest with reference values for the given platform.
+func Default(platform platforms.Platform) Manifest {
+	if embeddedReferenceValues == nil {
+		// If we're here, this is the first time this function is called, and the global state is not
+		// yet initialized. So let's unmarshal the embedded reference values.
+		if err := json.Unmarshal(EmbeddedReferenceValuesJSON, &embeddedReferenceValues); err != nil {
```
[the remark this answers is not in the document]
msanft: As `embeddedReferenceValues` have a fixed value at compile time, this is idempotent. But we might want to add a comment saying that it is

cli/cmd/generate.go
```diff
@@ -101,23 +101,41 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
+	var mnf *manifest.Manifest
+	if existingManifest, err := os.ReadFile(flags.manifestPath); err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			mnf = manifest.Default(flags.referenceValuesPlatform)
+		} else {
+			if err := json.Unmarshal(existingManifest, &mnf); err != nil {
+				return fmt.Errorf("unmarshal existing manifest: %w", err)
+			}
+		}
+	}
```
Freax13, line 115: The `else` branch is at the wrong level.

internal/manifest/manifest.go
```diff
@@ -128,6 +60,23 @@ func (p Policy) Hash() HexString {
...
+		}
+	}
+
+	if r.BareMetalTDX != nil {
+		if err := r.BareMetalTDX.Validate(); err != nil {
+			return fmt.Errorf("validating bare metal TDX reference values: %w", err)
+		}
+	}
+
+	return fmt.Errorf("no reference values provided")
```
Freax13, line 75: This fails unconditionally regardless of whether one or more set of reference values was set.

internal/manifest/manifest.go
```diff
@@ -128,6 +60,23 @@ func (p Policy) Hash() HexString {
...
+		}
+	} else if r.BareMetalTDX != nil {
+		if err := r.BareMetalTDX.Validate(); err != nil {
+			return fmt.Errorf("validating bare metal TDX reference values: %w", err)
+		}
+	} else {
+		return fmt.Errorf("reference values in manifest cannot be empty")
+	}
+
+	return nil
```
Freax13, line 75: This only validates `AKS` or `BareMetalTDX`, but never both.

Freax13 (approved):
LGTM!

cli/cmd/verify.go
```diff
@@ -175,7 +175,7 @@ func parseVerifyFlags(cmd *cobra.Command) (*verifyFlags, error) {
 }
 
 func newCoordinatorValidateOptsGen(mnfst manifest.Manifest, hostData []byte) (*snp.StaticValidateOptsGenerator, error) {
-	validateOpts, err := mnfst.SNPValidateOpts()
+	validateOpts, err := mnfst.AKSValidateOpts()
```
burgerdev, line 178: Should a manifest just have a `ValidateOpts` method that returns opts for all configured reference values?

cli/main.go
```diff
@@ -45,8 +45,7 @@ func buildVersionString() string {
 	}
 	fmt.Fprint(versionsWriter, "\n")
 	fmt.Fprintf(versionsWriter, "reference values for %s platform:\n", platforms.AKSCloudHypervisorSNP.String())
-	fmt.Fprintf(versionsWriter, "\truntime handler:\tcontrast-cc-%s\n", manifest.TrustedMeasurement[:32])
-	fmt.Fprintf(versionsWriter, "\tlaunch digest:\t%s\n", manifest.TrustedMeasurement)
+	fmt.Fprintf(versionsWriter, "\tembedded reference values:\t%s\n", manifest.EmbeddedReferenceValuesJSON)
```
burgerdev, line 48: This is imho a good use case for [`json.MarshalIndent`](https://pkg.go.dev/encoding/json#MarshalIndent).

internal/kuberesource/resourcegen/main.go
```diff
@@ -42,18 +42,30 @@ func main() {
 		var err error
 		switch set {
 		case "coordinator":
-			subResources = kuberesource.CoordinatorBundle()
+			subResources, err = kuberesource.CoordinatorBundle(platform)
+			if err != nil {
+				log.Fatalf("Error generating coordinator: %v", err)
+			}
```
burgerdev, line 48: There's an error handler outside the switch statement, would be enough to rely on that.

internal/kuberesource/parts.go
```diff
@@ -217,7 +228,12 @@ type CoordinatorConfig struct {
 }
 
 // Coordinator constructs a new CoordinatorConfig.
-func Coordinator(namespace string) *CoordinatorConfig {
+func Coordinator(namespace string, platform platforms.Platform) (*CoordinatorConfig, error) {
+	runtimeHandler, err := manifest.DefaultPlatformHandler(platform)
+	if err != nil {
+		return nil, fmt.Errorf("getting default runtime handler: %w", err)
+	}
```
burgerdev, line 235: The coordinator resource does not really depend on the platform, so I'd prefer not to have the `platform` argument and to patch the `runtimeClassName` outside of this constructor. Then you would not have to pass down the argument through all functions.

Same goes for all the other resources (openssl etc.).

internal/manifest/constants.go
```diff
@@ -3,32 +3,37 @@
...
-			SNPVersion:        toPtr(SVN(8)),
-			MicrocodeVersion:  toPtr(SVN(115)),
-		},
+	mnfst := Manifest{}
+	switch platform {
+	case platforms.AKSCloudHypervisorSNP:
+		mnfst.ReferenceValues.AKS = embeddedReferenceValues.AKS
+	case platforms.RKE2QEMUTDX, platforms.K3sQEMUTDX:
+		mnfst.ReferenceValues.BareMetalTDX = embeddedReferenceValues.BareMetalTDX
 	}
```
burgerdev, line 32: The switch statement is not exhaustive - do we want to hand out an empty manifest for an unsupported platform?

internal/manifest/constants.go
```diff
@@ -3,32 +3,37 @@
...
-	return mnfst
+	return &mnfst
 }
 
-func toPtr[T any](t T) *T {
-	return &t
+// DefaultPlatformHandler is a short-hand for getting the default runtime handler for a platform.
+func DefaultPlatformHandler(platform platforms.Platform) (string, error) {
+	return Default(platform).RuntimeHandler(platform)
```
burgerdev, line 38: This looks odd - why does the `RuntimeHandler` func need a platform argument?

internal/manifest/manifest.go
```diff
@@ -128,6 +60,26 @@ func (p Policy) Hash() HexString {
...
+		}
+	}
+	if r.BareMetalTDX != nil {
+		if err := r.BareMetalTDX.Validate(); err != nil {
+			return fmt.Errorf("validating bare metal TDX reference values: %w", err)
+		}
+	}
+
+	if r.BareMetalTDX == nil && r.AKS == nil {
+		return fmt.Errorf("reference values in manifest cannot be empty")
```
burgerdev, line 75: Maybe hint to an unsupported platform choice in the error message?

internal/manifest/constants.go
```diff
@@ -3,32 +3,37 @@
...
-		},
+	"github.com/edgelesssys/contrast/node-installer/platforms"
+)
+
+// Default returns a default manifest with reference values for the given platform.
+func Default(platform platforms.Platform) *Manifest {
+	if embeddedReferenceValues == nil {
+		// If we're here, this is the first time this function is called, and the global state is not
+		// yet initialized. So let's unmarshal the embedded reference values. This does not need to be
+		// locked, as embeddedReferenceValues has a fixed value at compile time, making this idempotent.
```
burgerdev, line 19: Could this be done in an `init()` function?

msanft: I did this before, but @katexochen doesn't like init functions :D

internal/manifest/constants.go
msanft: Because the runtime-handler is platform-specific (as its name is derived from the launch measurement)

internal/kuberesource/parts.go
msanft: What'd be the benefit of this though? In reality, the resource won't work without the runtime handler, so it does depend on it in some sense?

cli/cmd/verify.go
msanft: This would need to be generic then, which might get.. ugly?
I'd advocate for doing this once we _need_ the TDX validate opts at all.

internal/kuberesource/parts.go
burgerdev: The resources work without a runtime handler, with the kata-cc runtime handler, with older runtime handlers, etc. Imho, the runtime handler is a property  of the environment and should thus be added where the environment is set up, i.e. in test runners etc. And this would have the upside that you don't need to add and maintain these platform arguments everywhere.

cli/cmd/verify.go
burgerdev: Fine with me.

internal/manifest/constants.go
burgerdev: You mean the manifest could contain multiple sets of reference values, so it can't decide for which one to return the RuntimeHandler?

msanft: I think it was the plan to allow this, yes. (We'll also have a problem once we have multiple allowed trusted measurements for one platform though. No clue how to handle that)

internal/kuberesource/parts.go
msanft: Sounds reasonable. Where would you implement the patching?

internal/manifest/constants.go
burgerdev: I mean, as far as I understand the data model assigning a pointer _should_ be safe, but still counts as a race condition. init would have the benefit of implicit synchronization.

internal/kuberesource/parts.go
burgerdev: In the resourcegen binary and the test runners, probably?

cli/main.go
```diff
@@ -45,8 +46,9 @@ func buildVersionString() string {
 	}
 	fmt.Fprint(versionsWriter, "\n")
 	fmt.Fprintf(versionsWriter, "reference values for %s platform:\n", platforms.AKSCloudHypervisorSNP.String())
-	fmt.Fprintf(versionsWriter, "\truntime handler:\tcontrast-cc-%s\n", manifest.TrustedMeasurement[:32])
-	fmt.Fprintf(versionsWriter, "\tlaunch digest:\t%s\n", manifest.TrustedMeasurement)
+	if refValues, err := json.MarshalIndent(manifest.EmbeddedReferenceValuesJSON, "", " "); err == nil {
```
burgerdev, line 49: Sorry I did not add enough context. What I meant to say is to do something like this:

```go
json.MarshalIndent(manifest.embeddedReferenceValues, "\t\t\t\t", "  ")
``` 

And this should marshal the object, not the byte array.

msanft: Oh. This makes sense now..

cli/main.go
```diff
@@ -43,11 +43,10 @@ func buildVersionString() string {
 			fmt.Fprintf(versionsWriter, "\t%s\n", image)
 		}
 	}
-	fmt.Fprint(versionsWriter, "\n")
-	fmt.Fprintf(versionsWriter, "reference values for %s platform:\n", platforms.AKSCloudHypervisorSNP.String())
-	fmt.Fprintf(versionsWriter, "\truntime handler:\tcontrast-cc-%s\n", manifest.TrustedMeasurement[:32])
-	fmt.Fprintf(versionsWriter, "\tlaunch digest:\t%s\n", manifest.TrustedMeasurement)
-	fmt.Fprintf(versionsWriter, "\tgenpolicy version:\t%s\n", constants.GenpolicyVersion)
+	if refValues, err := json.MarshalIndent(manifest.EmbeddedReferenceValues(), "\t", "  "); err == nil {
```
burgerdev, line 46:
```suggestion
	if refValues, err := json.MarshalIndent(manifest.EmbeddedReferenceValues(), "\t\t\t\t", "  "); err == nil {
```
... to align nicely: https://go.dev/play/p/6GY8oxwZTsd.

burgerdev (approved):

cli/main.go
msanft: Huh?

I tested it so that it aligns:
```
[14:32] moritzs:☸ msanft-contrast in contrast @ msanft/tdx-trusted-measurement-embedding [$!] took 1m15s 
❯ nix run .#contrast.cli -- -v
warning: Git tree '/home/moritzs/dev/edgeless/contrast' is dirty
contrast version 0.9.0-pre

container image versions:
embedded reference values:    {
                                "AKS": {
                                  "SNP": {
                                    "MinimumTCB": {
                                      "BootloaderVersion": 3,
                                      "TEEVersion": 0,
                                      "SNPVersion": 8,
                                      "MicrocodeVersion": 115
                                    }
                                  },
                                  "TrustedMeasurement": "e56b58bdc033c24c28bac08e6afe094e248824cfaaf885aad7c2c832a5c3672645bc83b0e1bfec309b00fd6a090637cb"
                                },
                                "BareMetalTDX": {
                                  "TrustedMeasurement": "b534c4fb9a7d4283d1030cfd71f00997f3b19f613171526b3e241abf92d3712d"
                                }
                              }
genpolicy version:            3.2.0.azl1.genpolicy0
```

internal/manifest/manifest.go
```diff
@@ -194,17 +154,29 @@ func (m *Manifest) SNPValidateOpts() (*validate.Options, error) {
...
 	}, nil
 }
+
+// RuntimeHandler returns the runtime handler for the given platform.
+func (m *Manifest) RuntimeHandler(platform platforms.Platform) (string, error) {
+	switch platform {
+	case platforms.AKSCloudHypervisorSNP:
+		return fmt.Sprintf("contrast-cc-%s", m.ReferenceValues.AKS.TrustedMeasurement[:32]), nil
+	case platforms.K3sQEMUTDX, platforms.RKE2QEMUTDX:
+		return fmt.Sprintf("contrast-cc-%s", m.ReferenceValues.BareMetalTDX.TrustedMeasurement[:32]), nil
```
Freax13, line 178: If `BareMetalTDX` is nil, this line here causes a panic. I can trigger this with the `generate` subcommand.

Freax13: Nvm

internal/manifest/manifest.go
```diff
@@ -175,13 +135,13 @@ func (m *Manifest) Validate() error {
 	return nil
 }
 
-// SNPValidateOpts returns validate options populated with the manifest's
-// SNP reference values and trusted measurement.
-func (m *Manifest) SNPValidateOpts() (*validate.Options, error) {
+// AKSValidateOpts returns validate options populated with the manifest's
+// AKS reference values and trusted measurement.
+func (m *Manifest) AKSValidateOpts() (*validate.Options, error) {
```
Freax13, line 140: If `AKS` is `nil` and `BareMetalTDX` is not nil and valid, this function will still access `AKS` and cause a null-pointer panic.

cli/main.go
burgerdev: Sorry, I did not consider the underying tabwriter.

cli/main.go
```diff
@@ -43,11 +43,10 @@ func buildVersionString() string {
 			fmt.Fprintf(versionsWriter, "\t%s\n", image)
 		}
 	}
-	fmt.Fprint(versionsWriter, "\n")
-	fmt.Fprintf(versionsWriter, "reference values for %s platform:\n", platforms.AKSCloudHypervisorSNP.String())
-	fmt.Fprintf(versionsWriter, "\truntime handler:\tcontrast-cc-%s\n", manifest.TrustedMeasurement[:32])
```
katexochen, line 48: Runtime handler isn't printed anymore.

cli/main.go
```diff
@@ -43,11 +43,10 @@ func buildVersionString() string {
...
-	fmt.Fprintf(versionsWriter, "\tlaunch digest:\t%s\n", manifest.TrustedMeasurement)
-	fmt.Fprintf(versionsWriter, "\tgenpolicy version:\t%s\n", constants.GenpolicyVersion)
+	if refValues, err := json.MarshalIndent(manifest.EmbeddedReferenceValues(), "\t", "  "); err == nil {
+		fmt.Fprintf(versionsWriter, "embedded reference values:\t%s\n", refValues)
```
katexochen, line 47: This reduced readability of the output a lot. If we want to keep this  as json output, we should move it to the end and start the json doc in a separate line so you can cut it and process it with jq if needed.

Freax13:
CI's failing.

internal/kuberesource/resourcegen/main.go
```diff
@@ -42,18 +49,17 @@ func main() {
...
-			if err != nil {
-				log.Fatalf("Error generating runtime: %v", err)
-			}
 		case "openssl":
-			subResources = kuberesource.OpenSSL()
+			subResources = kuberesource.PatchRuntimeHandlers(kuberesource.OpenSSL(), runtimeHandler)
 		case "emojivoto":
-			subResources = kuberesource.Emojivoto(kuberesource.ServiceMeshDisabled)
+			subResources, err = kuberesource.Emojivoto(kuberesource.ServiceMeshDisabled)
+			subResources = kuberesource.PatchRuntimeHandlers(subResources, runtimeHandler)
```
katexochen, line 59: What if the first call fails and subResources is nil/empty?

internal/kuberesource/sets.go
```diff
@@ -130,7 +132,7 @@ func OpenSSL() []any {
 }
 
 // GetDEnts returns a set of resources for testing getdents entry limits.
-func GetDEnts() ([]any, error) {
+func GetDEnts() []any {
```
katexochen, line 135: Why are we removing the error here but adding it for Emojivoto?

internal/manifest/referencevalues.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var (
+	// EmbeddedReferenceValuesJSON contains the embedded reference values in JSON format.
+	// At startup, they are unmarshaled into a globally-shared ReferenceValues struct.
+	//
+	//go:embed assets/reference-values.json
+	EmbeddedReferenceValuesJSON []byte
+
+	embeddedReferenceValues *ReferenceValues
```
katexochen, line 21: Why do we need this global state?

internal/manifest/constants.go
```diff
@@ -3,32 +3,61 @@
...
+// EmbeddedReferenceValuesIfUninitialized returns the reference values embedded in the binary, initializing
+// the global state if it is not yet initialized.
+func setReferenceValuesIfUninitialized() ReferenceValues {
+	if embeddedReferenceValues == nil {
+		// If we're here, this is the first time this function is called, and the global state is not
+		// yet initialized. So let's unmarshal the embedded reference values. This does not need to be
+		// locked, as embeddedReferenceValues has a fixed value at compile time, making this idempotent.
+		if err := json.Unmarshal(EmbeddedReferenceValuesJSON, &embeddedReferenceValues); err != nil {
+			fmt.Printf("Failed to unmarshal embedded reference values: %s\n", err)
+			os.Exit(1)
```
katexochen, line 59: os.Exit is only allowed in top-level functions (main package).

internal/manifest/manifest.go
```diff
@@ -25,72 +23,6 @@ type Manifest struct {
...
-// String returns the string representation of the HexString.
-func (h HexString) String() string {
-	return string(h)
-}
-
-// Bytes returns the byte slice representation of the HexString.
-func (h HexString) Bytes() ([]byte, error) {
-	return hex.DecodeString(string(h))
-}
-
```
katexochen, line 93: why does this need to be moved? Moving code around makes it super hard to review and causes loss of history at nearly no benefit (as its destination is still in the same package).

internal/manifest/referencevalues.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// ReferenceValues contains the workload-independent reference values for each platform.
+type ReferenceValues struct {
+	// AKS holds the reference values for AKS.
+	AKS *AKSReferenceValues
+	// BareMetalTDX holds the reference values for TDX on bare metal.
+	BareMetalTDX *BareMetalTDXReferenceValues
```
katexochen, line 29: both should have omitempty for json marshalling.

katexochen (changes requested):
Ticket said 

> ReferenceValues section should be a list of the same type

Which isn't reflected in this PR. Will that be done in a follow-up?

internal/manifest/manifest.go
msanft: To not have a super large file, which makes editing hard

internal/manifest/referencevalues.go
msanft: To not unmarshal it over and over?

cli/main.go
msanft: Because there's no singular valid runtime handler anymore

msanft:
> Which isn't reflected in this PR. Will that be done in a follow-up?

@katexochen With the current design, I don't think this is necessary anymore? Also, we discussed in-person that a structured format makes more sense, and you agreed

internal/manifest/constants.go
msanft: This would just add additional code complexity as we'd have to bubble up and handle the error everywhere. What about using an init, as @burgerdev suggested? I just don't see the advantage with this

internal/manifest/referencevalues.go
katexochen: How often would we need to unmarshal it? What is the cost?

msanft: Every time someone needs the reference values. Which happens every time the runtime handler is required, at least. So at generation time (with the current code - for each generation of a resource), thus also in each e2e test

I just don't see any value handling a static value as if it were dynamic

katexochen:
> > Which isn't reflected in this PR. Will that be done in a follow-up?
> 
> @katexochen With the current design, I don't think this is necessary anymore? Also, we discussed in-person that a structured format makes more sense, and you agreed

We discussed the _embedding_ of the reference values in a structured format (json). 

In the manifest, we still target a list of reference value configurations, as discussed in previous meetings and documented in the ticket. This will allow us to add multiple platforms of different configuration (heterogeneous clusters with different types of SNP CPUs, for example), which is something we must support but cannot be enabled by the manifest structure given in this PR.

msanft:
@katexochen I think there was a misunderstanding of the tickets on my side. Consider this PR to handle [4459](https://dev.azure.com/Edgeless/Edgeless/_backlogs/backlog/Edgeless%20Team/Features?workitem=4459), while a follow-up PR should handle [4299](https://dev.azure.com/Edgeless/Edgeless/_backlogs/backlog/Edgeless%20Team/Features?workitem=4299)

cli/main.go
katexochen: That was already anticipated, that's why there was a platform-specific section in the output.

msanft: So do you want to roll back to how it was before, but showing everything for SNP and TDX?

internal/kuberesource/resourcegen/main.go
msanft: `range` over `nil` / empty should do nothing. So `subResources` will just stay `nil` / empty.

e2e/openssl/openssl_test.go
```diff
@@ -43,9 +44,17 @@ var (
...
+	platform := platforms.AKSCloudHypervisorSNP
+
+	runtimeHandler, err := manifest.DefaultPlatformHandler(platform)
+	require.NoError(t, err)
 
+	resources := kuberesource.OpenSSL()
 	coordinator := kuberesource.CoordinatorBundle()
+
+	resources = kuberesource.PatchRuntimeHandlers(resources, runtimeHandler)
+
```
Freax13, line 57: The line has to be moved below `resources = append(resources, coordinator...)`. AFAICT this is why CI fails.

packages/by-name/contrast/package.nix
```diff
@@ -41,10 +38,29 @@ let
...
+        };
+        trustedMeasurement = lib.removeSuffix "\n" (
+          builtins.readFile "${microsoft.runtime-class-files}/launch-digest.hex"
+        );
+      };
+      bareMetalTDX = {
+        trustedMeasurement = lib.removeSuffix "\n" (
+          builtins.readFile "${kata.runtime-class-files}/launch-digest.hex"
+        );
+      };
```
Freax13, line 62: This uses the launch measurement calculated for SNP. Should we maybe use a different fake value until we do the launch calculations for TDX? My fear is that using the launch measurement for SNP might lead to confusion.

internal/manifest/manifest.go
katexochen: I don't see how 60 lines make a file difficult to edit. If you need to do a refactoring, do it as initial work before implementing your feature, in a separate commit. These are exactly the changes that unnecessary blow up the size of your PRs and make the whole thing unpleasant to review.

internal/manifest/constants.go
katexochen: No, no init. We even run a linter to prevent you from using it.

If you cannot live without global state, just use panic here instead of os.Exit and cover it with a unit test.

cli/main.go
katexochen: I'm just saying you shouldn't throw away the work other developers put in there without a thought. Should users construct the runtime name themselves in the future and count the offset of the prefix we are using? Is the length of the offset documented anywhere?

msanft: I'm fine with changing it back, but then I'd also suggest we don't print the JSON at all. Are you fine with "reverting" that @burgerdev?

internal/manifest/constants.go
burgerdev: We could also just keep it simple and parse the JSON every time `EmbeddedReferenceValues` is called - which is not going to be that often, in one CLI run? And panic sounds fine since we're dealing with constant, predictable data and should fail loudly in case it is not.

cli/main.go
burgerdev: Sounds like we need to discuss the question of runtime handlers vs. reference values first - we want this to be a 1:1 mapping, right? So, should it go into `ReferenceValues` fields?

cli/main.go
burgerdev: I'd rather have a `--json` for this use case.

katexochen: That would mean a refactoring of the command structure, as we are currently using the builtin --version flag, not a command,  right?

internal/manifest/constants.go
msanft: Ok, will do

cli/main.go
msanft: Implicitly, I'd say it's already there due to the presence of `TrustedMeasurement` in each of the `ReferenceValues`. But I think this is something so adjacent to the general discussion of how we want to handle these that we should discuss it in the sync today.

packages/by-name/contrast/package.nix
msanft: Currently this combines the hashes of SNP and TDX software in the launch digest, which, imo, should be acceptable for now, since we don't actually deploy SNP and TDX side-by-side just yet. Making them individual digests obviously needs to happen eventually, but we'll need to tweak quite some code in the node-installer to accept multiple runtime-handlers and install based on configuration here, which, I think, is out of scope for this PR.

Freax13: Ah, fair enough, I forgot that https://github.com/edgelesssys/contrast/pull/730/commits/5527f8d47054777533a51ff53c420920fce5d4fc hasn't been merged yet, my bad.

Freax13:
What's the status on this PR?

msanft:
> What's the status on this PR?

I'd like to merge it, but requires a review from @katexochen

cli/main.go
msanft: Created [a ticket](https://dev.azure.com/Edgeless/Edgeless/_backlogs/backlog/Edgeless%20Team/Features?workitem=4464) for this

katexochen (approved):

PR #757, opened 2024-07-23
just: add get-credentials-snpbm

Freax13:
Like #750, but for discovery.


burgerdev (approved):

katexochen:
Should we use a just function for this?

burgerdev (approved):
Nice

PR #758, opened 2024-07-23
node-installer: add kernel_params for setting up dm-verity

Freax13:
Until now, we didn't check the integrity of the boot image on bare-metal platforms. Starting with this PR, we create a dm-verity device and use that as root instead of `/dev/vda1`.


node-installer/internal/config/config.go
```diff
@@ -17,6 +17,8 @@ type Config struct {
 	Files []File `json:"files"`
 	// RuntimeHandlerName is the name of the runtime handler (containerd runtime) to create.
 	RuntimeHandlerName string `json:"runtimeHandlerName"`
+	// QemuDmVerityArgs is an additional set of kernel parameters to pass kernels started with QEMU.
+	QemuDmVerityArgs string `json:"QemuDmVerityArgs"`
```
msanft, line 21:
```suggestion
	// // QemuExtraKernelParams is an additional set of kernel parameters to pass kernels started with QEMU.
	QemuExtraKernelParams string `json:"QemuExtraKernelParams"`
```

node-installer/internal/constants/constants.go
```diff
@@ -42,7 +42,7 @@ var (
 const CRIFQDN = "io.containerd.grpc.v1.cri"
 
 // KataRuntimeConfig returns the Kata runtime configuration.
-func KataRuntimeConfig(baseDir string, platform platforms.Platform, debug bool) (*config.KataRuntimeConfig, error) {
+func KataRuntimeConfig(baseDir string, platform platforms.Platform, qemuDmVerityArgs string, debug bool) (*config.KataRuntimeConfig, error) {
```
msanft, line 45: s.a.

node-installer/internal/constants/constants.go
```diff
@@ -64,13 +64,15 @@ func KataRuntimeConfig(baseDir string, platform platforms.Platform, debug bool)
 		config.Hypervisor["qemu"]["image"] = filepath.Join(baseDir, "share", "kata-containers.img")
 		config.Hypervisor["qemu"]["kernel"] = filepath.Join(baseDir, "share", "kata-kernel")
 		config.Hypervisor["qemu"]["valid_hypervisor_paths"] = []string{filepath.Join(baseDir, "tdx", "bin", "qemu-system-x86_64")}
+		kernelParams := qemuDmVerityArgs
```
msanft, line 67:
```suggestion
		kernelParams, ok := config.Hypervisor["qemu"]["kernel_params"].(string)
		if !ok {
		...
		}
		kernelParams += qemuDmVerityArgs
```

node-installer/internal/constants/constants.go
```diff
@@ -85,13 +87,15 @@ func KataRuntimeConfig(baseDir string, platform platforms.Platform, debug bool)
...
 		if debug {
 			config.Hypervisor["qemu"]["enable_debug"] = true
-			config.Hypervisor["qemu"]["kernel_params"] = " agent.log=debug initcall_debug"
 			config.Agent["kata"]["enable_debug"] = true
 			config.Agent["kata"]["debug_console_enabled"] = true
 			config.Runtime["enable_debug"] = true
+			kernelParams += " agent.log=debug initcall_debug"
 		}
+		config.Hypervisor["qemu"]["kernel_params"] = kernelParams
```
msanft, line 98: s.a.

packages/by-name/kata/kata-kernel-uvm/package.nix
```diff
@@ -22,7 +22,8 @@ let
     postPatch = ''
       config=$(find . -regex '.*/config-[0-9.-]+-confidential')
       substituteInPlace $config \
-        --replace-fail 'CONFIG_INITRAMFS_SOURCE="initramfs.cpio.gz"' 'CONFIG_INITRAMFS_SOURCE=""'
+        --replace-fail 'CONFIG_INITRAMFS_SOURCE="initramfs.cpio.gz"' 'CONFIG_INITRAMFS_SOURCE=""' \
+        --replace-fail '# CONFIG_DM_INIT is not set' 'CONFIG_DM_INIT=y'
```
msanft, line 26: We should probably add a little comment here explaining why (both of) the patches are necessary.
For the first one, we should mention that we're not using an initrd.

node-installer/internal/config/config.go
msanft: Not a fan of specifying the whole command line. We should rather provide extra settings, I guess? Or make the settings a map.

msanft (changes requested):

node-installer/internal/constants/constants.go
Freax13: We already know that the default value is an empty string, so what's the point in appending? Also, if the default value *wasn't* an empty string, the kernel command line wouldn't be the same we're using to calculate the launch measurement, and so this would cause problems when verifying the attestation reports.

node-installer/internal/config/config.go
Freax13: We need the command line the **exactly** match the one we're using to calculate the launch measurement, so my idea was to construct it ones and then use that in all required places. That way they can't go out of sync.

node-installer/internal/constants/constants.go
msanft: Valid, but then let's add a comment that reasons about this.

node-installer/internal/config/config.go
msanft: Fine. But let's call it QemuKernelPrams then. It's got nothing to do with dm-verity per se

node-installer/internal/config/config.go
```diff
@@ -17,6 +17,8 @@ type Config struct {
 	Files []File `json:"files"`
 	// RuntimeHandlerName is the name of the runtime handler (containerd runtime) to create.
 	RuntimeHandlerName string `json:"runtimeHandlerName"`
+	// QemuExtraKernelParams is an additional set of kernel parameters to pass kernels started with QEMU.
+	QemuExtraKernelParams string `json:"QemuExtraKernelParams"`
```
msanft, line 21:
```suggestion
	QemuKernelParams string `json:"QemuKernelParams"`
```
Now it's not an "extra" anymore ;D

Freax13: You could argue it is though: The kata shim already passes a bunch of kernel parameters and the parameters here are appended at the end.

msanft (approved):

PR #760, opened 2024-07-24
vale: enable Microsoft.FirstPerson and Microsoft.Terms

thomasten:
Suggesting to enable two more Vale rules. What do you think?


katexochen (approved):
Nice, thanks! :)

PR #761, opened 2024-07-26
rfc: genpolicy on diverse platforms

burgerdev:


rfc/006-genpolicy-platforms.md
```diff
@@ -0,0 +1,111 @@
...
+
+Due to the nature of the settings file, Contrast can't rely on a single vendor to provide this.
+On the other hand, we don't want to force our users to make modifications for our standard target platforms.
+Thus, we develop settings for our known target platforms, validate them through CI and bundle them with the CLI.
+
+The upstream rules should eventually be compatible with all non-tardev platforms.
+This is a shared interest throughout the wider CoCo community, but we can accelerate it through contributions.
+
+For the tardev snapshotter platforms (that is, AKS for now), we start with the Microsoft rule set.
+Once the necessary patches grow uncomfortable (for example reference value divergence), we port the missing parts to a patch for the upstream rules.
```
m1ghtym0, line 97: Just for my understanding:
In other words, this means bringing all the msft changes to upstream except the tardev support.
Leaving tardev as the only difference/reason for the existence of the msft fork?

burgerdev: Microsoft has other patches that might not make it upstream, e.g. for their CSI drivers. My idea here is that we would port these so that they _would_ apply to upstream Kata, but apply them in the Contrast build. However,  this may not become necessary if upstream is "compatible enough".

m1ghtym0: Upstream would not accept these patches since they're specific for Azure's CSI-Plugins in this case?
What happens if we support other platforms such as GCP or AWS.
Would we have patches for things such as CSI for all of them?

katexochen (approved):
Thanks for writing this down!

rfc/006-genpolicy-platforms.md
```diff
@@ -0,0 +1,111 @@
...
+## Problem Statement
+
+Generated policies are a key building block of Confidential Containers and Contrast.
+However, the only implementation that guarantees basic confidentiality levels is the fork maintained by Microsoft.
+The upstream Kata policy is lacking security fixes that haven't been backported, and yet lacks a mechanism for ensuring image integrity.
+On the other hand, some features that we contributed and would like to use are only available upstream with no ETA of adoption by Microsoft.
+
+We're using the Microsoft fork successfully on AKS-CoCo, but it's tailored to that platform and unlikely to ever be portable.
+Portability will become important for our bare-metal efforts, though.
+Bare-metal cluster will diverge from the standardized AKS environment, requiring flexibility in policy generation.
```
msanft, line 14:
```suggestion
Bare-metal clusters will diverge from the standardized AKS environment, requiring flexibility in policy generation.
```

rfc/006-genpolicy-platforms.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Parts of genpolicy
+
+The workload policy mechanism can be divided into four parts:
+
+* The tool that generates reference values from workload YAML and composes the workload policy (`genpolicy`).
```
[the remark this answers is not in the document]
msanft: We should then move this point below (as of now) point 3, so that `rules.rego` is introduced first

rfc/006-genpolicy-platforms.md
```diff
@@ -0,0 +1,111 @@
...
+* The settings file for the tool (`genpolicy-settings.json`).
+* The policy rules themselves (`rules.rego`).
+* The policy engine that's queried in the agent.
+
+These components are somewhat independent, and the following subsections describe how they fit together.
+
+### The tool
+
+The idea of the `genpolicy` tool is to take a set of Kubernetes workload descriptors and derive the Kata agent API requests that a legitimate runtime would issue to spawn these workloads.
```
msanft, line 29:
```suggestion
The idea of the `genpolicy` tool is to take a set of Kubernetes workload descriptors and derive the Kata agent API requests that a legitimate runtime would issue to spawn, stop, and communicate with these workloads.
```
nit: it doesn't only spawn the workloads, does it?

rfc/006-genpolicy-platforms.md
```diff
@@ -0,0 +1,111 @@
...
+It appends these anticipated requests as data to the rules file to produce a complete policy.
```
msanft, line 30:
```suggestion
It appends rules that allow exactly and only these anticipated requests to the rules file to produce a complete policy.
```

rfc/006-genpolicy-platforms.md
```diff
@@ -0,0 +1,111 @@
...
+
+Guessing the transformations from Kubernetes to CRI is easy because it's somewhat stable, and the path from Kata runtime to the agent is under our control.
```
msanft, line 32:
```suggestion
Guessing the transformations from Kubernetes to CRI is easy because it's somewhat stable, and the path from Kata runtime to the agent is under our control.
```
I don't really get this. What does it mean?

rfc/006-genpolicy-platforms.md
```diff
@@ -0,0 +1,111 @@
...
+The rules consist of predicates that evaluate to a Boolean decision for all agent service methods.
+Broadly generalizing, each rule compares the input request with reference values.
+If the input request diverges from expectations, the output is `false` and the request is rejected.
+
+Assuming a relatively stable format of the reference values, the rules can be written independently of the other parts.
+For example, if `genpolicy` computes both `dm-verity` hashes for tardev layers and includes the corresponding pinned images, rules can be written to accommodate both the tardev-snapshotter and any guest-pull mechanism.
+
+### The agent
+
+The functionality in the agent is very simple: an in-memory policy engine is equipped with the incoming policy, and subsequent agent requests are subject to policy validation.
```
msanft, line 53:
```suggestion
The functionality in the agent is very simple: an in-memory policy engine is equipped with the incoming policy, and subsequent agent requests are subject to policy validation. The agent starts from a default policy that's available in the UVM image.
```

rfc/006-genpolicy-platforms.md
```diff
@@ -0,0 +1,111 @@
...
+This is a shared interest throughout the wider CoCo community, but we can accelerate it through contributions.
+
+For the tardev snapshotter platforms (that is, AKS for now), we start with the Microsoft rule set.
+Once the necessary patches grow uncomfortable (for example reference value divergence), we port the missing parts to a patch for the upstream rules.
+
+## Alternatives considered
+
+### Bundle both Microsoft's and Kata's tool
+
+While this approach seems easiest on the surface, it's going to explode the size of the CLI binary, which is already significant.
```
msanft, line 103: I see the second point, but I think there could be ways around a large CLI (e.g. by sourcing static builds of these tools at runtime, only for the required platform)

Just a note though, not advocating against your proposal

msanft (commented):
Thanks a lot for writing this!

rfc/006-genpolicy-platforms.md
burgerdev: True, let me just write _manage_.

rfc/006-genpolicy-platforms.md
burgerdev: This is not true - there are request objects (or, parts of them) literally appended to the rules. Let me try if I can make that clearer.

rfc/006-genpolicy-platforms.md
burgerdev: Let me try to explain this better.

rfc/006-genpolicy-platforms.md
msanft: brilliant!

rfc/006-genpolicy-platforms.md
burgerdev: I'm not sure what Kata's strategy is wrt. insecure volume mounts in confidential containers, but I don't expect it to be a priority and it may thus take considerable time until adoption.

Neither do I think we should make it _our_ priority, because there's just not a good use case that can't be addressed with block storage, which is much easier to check with policies.

rfc/006-genpolicy-platforms.md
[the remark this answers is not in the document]
burgerdev: Restructured

m1ghtym0 (approved):
Thanks!

PR #764, opened 2024-07-26
cli: factor out genpolicy

burgerdev:
This PR moves the invocation of the genpolicy tool into its own package, to reduce the amount of code in `cmd` and allow for better testing. 

This new package also knows which settings are appropriate for which platform. We would like to keep backward compatibility with respect to the cli flags for rules and settings. This is why the config is separate from the runner: the CLI gets the default config, (maybe) writes it to the files, and then starts the runner with the files.


cli/genpolicy/genpolicy.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return nil
+}
+
+// Teardown cleans up temporary files and should be called after the last Run.
+func (r *Runner) Teardown() error {
+	var err error
+	for _, install := range []embedbin.Installed{r.genpolicy, r.settings, r.rules} {
+		err = errors.Join(err, install.Uninstall())
+	}
```
Freax13, line 122: This will cause a null-pointer exception if `Runner.Run` has never been called and `genpolicy` etc haven't been initialized.

cli/genpolicy/genpolicy.go
```diff
@@ -0,0 +1,124 @@
...
+		r.genpolicy, err = e.Install("", genpolicyBin)
+		outerErr = errors.Join(outerErr, err)
+		r.settings, err = e.Install("", r.Settings)
+		outerErr = errors.Join(outerErr, err)
+		r.rules, err = e.Install("", r.Rules)
+		outerErr = errors.Join(outerErr, err)
+	})
+	if outerErr != nil {
+		return fmt.Errorf("creating temp files: %w", outerErr)
+	}
```
Freax13, line 88: AFAICT this error will only be caught on the first `Run` call, on subsequent calls no error is generated and so the code down below will fail with a null-pointer exception. 

As of right now, AFAICT there's no scenario where this could happen, but still, it'd be nice to return a proper error message instead of the null-pointer one.

cli/genpolicy/genpolicy.go
burgerdev: I'm checking for nil now.

cli/genpolicy/genpolicy.go
burgerdev: I rewrote this part, don't need the `sync.Once` anymore.

cli/genpolicy/assets/allow-all.rego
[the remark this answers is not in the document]
burgerdev: Done

burgerdev:
Hey folks! I rewrote the PR - it's now fully backwards compatible.

packages/by-name/cli-release/package.nix
```diff
@@ -11,9 +11,10 @@
 (contrast.overrideAttrs (
   _finalAttrs: previousAttrs: {
     prePatch = ''
-      install -D ${lib.getExe genpolicy} cli/cmd/assets/genpolicy
-      install -D ${contrast.settings}/genpolicy-settings.json cli/cmd/assets/genpolicy-settings.json
-      install -D ${contrast.rules}/genpolicy-rules.rego cli/cmd/assets/genpolicy-rules.rego
+      install -D ${lib.getExe genpolicy} cli/genpolicy/assets/genpolicy
+      install -D ${contrast.settings}/genpolicy-settings.json cli/genpolicy/assets/genpolicy-settings.json
+      install -D ${contrast.rules}/genpolicy-rules.rego cli/genpolicy/assets/genpolicy-rules.rego
+      install -D ${genpolicy.src}/src/kata-opa/allow-all.rego cli/genpolicy/assets/allow-all.rego
```
katexochen, line 17: Do we really need to embed the allow-all policy into the release cli? Are we planning to do a release with this policy?

Freax13 (approved):

packages/by-name/cli-release/package.nix
burgerdev: We should not release in this state, yes. If I just remove the file here, the release build will fail - is that acceptable? Alternatively, we could fail `NewConfig` for unsupported platforms (by checking for an empty embedded ruleset, for example).

katexochen: Ah, I thought the build wouldn't fail due to the placeholder, but we don't include that in src.

katexochen (approved):

PR #767, opened 2024-07-29
pause-bundle: init; kata.kata-image: optionally include pause-bundle 

katexochen:


packages/by-name/kata/kata-image/package.nix
```diff
@@ -38,25 +40,30 @@ let
...
-
-      tar --sort=name --mtime="@$SOURCE_DATE_EPOCH" -cvf /build/rootfs-extra-tree.tar -C /build/rootfs .
-
-      mv /build/rootfs-extra-tree.tar $out
-
-      runHook postBuild
-    '';
+    buildPhase =
+      ''
+          runHook preBuild
```
burgerdev, line 45:
```suggestion
        runHook preBuild
```

packages/by-name/kata/pause-bundle/package.nix
burgerdev, line 1: Did you consider taking `nixpkgs#kubernetes.pause` as rootfs?

burgerdev (approved):

packages/by-name/kata/pause-bundle/package.nix
katexochen: Good point, I think we could do that if we aren't concerned about matching the exact version kata is using.

burgerdev: I would probably still use umoci for that (new -> copy closure of pause -> repack -> config -> unpack).

burgerdev (approved):
Nice!

PR #768, opened 2024-07-29
attestation: add TDX validator

msanft:
This adds a _preliminary_ TDX validator. It is not yet used throughout the code, nor does it implement certificate extensions like the SNP one does. But to not have the PR size explode, and since this is an independent component, I opted to put it into a separate PR with this.

Sent a review request to @daniel-weisse for the `go-tdx-guest` code.


internal/attestation/tdx/validator.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/google/go-tdx-guest/verify"
+	"github.com/google/go-tdx-guest/verify/trust"
+	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/protobuf/proto"
+)
+
+// source: https://certificates.trustedservices.intel.com/Intel_SGX_Provisioning_Certification_RootCA.pem
+//
+//go:embed Intel_SGX_Provisioning_Certification_RootCA.pem
+var tdxRootCert []byte
```
burgerdev, line 29: Nit: The cert has SGX in the name, but the var has TDX. Even if Intel's declared purpose for the cert is to validate TDX (which would make sense), I'd appreciate an explanation of what this is and why we need it here.

internal/attestation/tdx/validator.go
```diff
@@ -0,0 +1,165 @@
...
+// StaticValidateOptsGenerator returns validate.Options generator that returns
+// static validation options.
+type StaticValidateOptsGenerator struct {
+	Opts *validate.Options
+}
+
+// SNPValidateOpts return the TDX validation options.
+func (v *StaticValidateOptsGenerator) SNPValidateOpts(_ *tdx.QuoteV4) (*validate.Options, error) {
+	return v.Opts, nil
+}
```
burgerdev, line 62: Why does it say SNP here?

burgerdev (commented):
I understand that there are validations to be added in subsequent PRs. Could you please leave a TODO for them to document that something is missing and what this is?

internal/attestation/tdx/validator.go
```diff
@@ -0,0 +1,165 @@
...
+	// Validate the report data.
+
+	reportDataExpected := reportdata.Construct(peerPublicKey, nonce)
+	validateOpts, err := v.validateOptsGen.TDXValidateOpts(quote)
+	if err != nil {
+		return fmt.Errorf("generating validation options: %w", err)
+	}
+	validateOpts.TdQuoteBodyOptions.ReportData = reportDataExpected[:]
+
+	// Build the validation options.
```
Freax13, line 138: Nit: These two comments look like they have been swapped.

msanft:
> I understand that there are validations to be added in subsequent PRs. Could you please leave a TODO for them to document that something is missing and what this is?

There should not be (at least not intentionally) any validations missing. The only thing that's missing _related_ to validation is the options generator which takes a set of reference values and transforms them into something that can be understood by the TDX library. As this is required to use the validator implemented in this PR by design (i.e. since no other type implements the required interface, as of now), I don't think leaving a TODO for that makes much sense.

The certificate extension mechanism I talk about in the PR description refers to the embedding of SNP attestation data into the certificate issued by the mesh authority, which, to my understanding, is not integral to an MVP for TDX attestation. Since it'll require quite a lot of code, and I want to keep my PRs easy to review, I simply opted not to do that in this PR.

Freax13 (approved):

internal/attestation/tdx/validator.go
```diff
@@ -0,0 +1,170 @@
...
+			if v.metrics.attestationFailures != nil {
+				v.metrics.attestationFailures.Inc()
+			}
+		}
+	}()
+
+	// Parse the attestation document.
+
+	quote := &tdx.QuoteV4{}
+	if err := proto.Unmarshal(attDocRaw, quote); err != nil {
```
daniel-weisse, line 106: Why does the Issuer return a marshalled proto quote, instead of just the raw quote?

msanft: I'm not sure actually. I think the raw report should even be shorter, as it's non-self-describing too

Freax13:
What's the status of this PR?

msanft:
> What's the status of this PR?

Theoretically we *could* rebase and merge this already, without actually using the validator.

However, since #783 looks ok to be reviewed and merged too (which I expected to not be the case this fast), i'd advocate to merge that one first.

Freax13:
CI is not happy yet.

Freax13:
Rebased onto main.

katexochen (approved):

PR #769, opened 2024-07-30
kata-agent: always use 256-bit hashes and never skip policy validation

Freax13:
This PR changes the kata patches to always use sha-256 - even on TDX. Previously we used sha-384 on TDX, but this lead to problems because we'd need to store two different hashes in the manifest. The init-data proposal in upstream kata will also allow hashes with sizes that don't match the host-data/mrconfigid fields.

This PR also fixes some of the surrounding TODOs to ensure that the policy hash is always enforced.

### DIFF
--- a/packages/by-name/kata/kata-runtime/0003-runtime-agent-verify-the-agent-policy-hash.patch
+++ b/packages/by-name/kata/kata-runtime/0003-runtime-agent-verify-the-agent-policy-hash.patch
@@ -1,4 +1,4 @@
-From caf72c1574aed9c501894323bca868aec400e288 Mon Sep 17 00:00:00 2001
+From f4d02f06c8a36b1e33e39a7fd1e94a5c48b2ccf4 Mon Sep 17 00:00:00 2001
 From: Tom Dohrmann <erbse.13@gmx.de>
 Date: Mon, 8 Jul 2024 07:51:20 +0000
 Subject: [PATCH] runtime: agent: verify the agent policy hash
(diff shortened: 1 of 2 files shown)


msanft (commented):
Patch LGTM, but still need to test this

Freax13:
> Patch LGTM, but still need to test this

I tested this by changing the shim to append a `1` to hash instead of appending only `0`. This caused an error with the expected message. Testing this is not trivial because bare-metal is already not working and is blocked on a bunch of issues.

msanft:
> Testing this is not trivial because bare-metal is already not working on master and is blocked on a bunch of issues.

For reference: bare-metal TDX _works_ as in _being deployable and containers starting_ on main. Policies don't, however

Freax13:
> For reference: bare-metal TDX _works_ as in _being deployable and containers starting_ on main. 

Only the coordinator pod can start though. The other pods are blocked because the coordinator can't verify their attestation reports.

packages/by-name/kata/kata-runtime/0003-runtime-agent-verify-the-agent-policy-hash.patch
```diff
@@ -309,8 +309,7 @@ index d709515ff..98b7552e1 100644
 +        info!(sl!(), "policy: SNP expected digest ({:?})", expected_digest);
 +        verify_sha_256(policy, expected_digest.as_slice())
 +    } else {
-+        warn!(sl!(), "policy: integrity has not been verified!");
-+        Ok(())
++        bail!("integrity has not been verified!");
```
burgerdev, line 326: Nit: this message could hint to the cause (no host data found).

packages/by-name/kata/kata-runtime/0003-runtime-agent-verify-the-agent-policy-hash.patch
```diff
@@ -296,70 +296,48 @@ index d709515ff..fe1461e5c 100644
...
++        // The MrConfigId used with TDX is longer than the host-data field used
++        // with SNP, but we don't want to use different hashes for different
++        // platforms. Instead we truncate the MrConfigId to 256-bit and always
++        // use sha-256.
++        let (expected_digest, trailing_data) =
++            expected_digest.split_at(<Sha256 as Digest>::output_size());
++
++        ensure!(
++            trailing_data.iter().all(|&d| d == 0),
++            "hash isn't followed by zeros: MrConfigId={expected_digest:?}"
```
burgerdev, line 317:
```suggestion
+            "hash isn't padded with zeros: MrConfigId={expected_digest:?}"
```

burgerdev (approved):

PR #770, opened 2024-07-30
verify: strip white-space from manifest before comparing

Freax13:
Previously we failed verification when the white-space between the two manifest files didn't match.

### DIFF
--- a/cli/cmd/verify.go
+++ b/cli/cmd/verify.go
@@ -123,8 +123,19 @@ func runVerify(cmd *cobra.Command, _ []string) error {
 
 	fmt.Fprintf(cmd.OutOrStdout(), "✔️ Wrote Coordinator configuration and keys to %s\n", filepath.Join(flags.workspaceDir, verifyDir))
 
-	currentManifest := resp.Manifests[len(resp.Manifests)-1]
-	if !bytes.Equal(currentManifest, manifestBytes) {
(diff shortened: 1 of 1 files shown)


burgerdev (changes requested):
If we wanted to introduce canonicalization, I would rather use e.g. RFC 8785, but we'd need to use that everywhere the manifest is interpreted.

However, I don't understand the use case:

1. The workload owner has the manifest sitting in their workspace. 
2. The data owner should have received the expected manifest from the workload owner.

In both cases, I don't see why the manifest would have changed in whitespace.

We had a discussion about a similar proposal a few weeks back: https://github.com/edgelesssys/contrast/pull/615#discussion_r1650443181.

msanft:
> In both cases, I don't see why the manifest would have changed in whitespace.

Without knowing any other context, I can think of a scenario where one editor appends a newline at the end, or uses spaces for tabs, and the other doesn't.

I generally think comparisons of structured data should happen in a canonical format, but wold also favor RFC-canonicalization over stripping.

Freax13:
Closed for the following reasons:
- There's no demand for this/there's no real use case.
- Comparing the manifests byte-for-byte is very simple and therefor secure. Canonicalizations can potentially be a source for bugs, so we should avoid them unless there's a good reason.

PR #771, opened 2024-07-30
genpolicy: hide logs by default

Freax13:
Logs may contain sensitive data and so they should be hidden by default. Upstream kata hides logs by default, but the Microsoft fork made them visible. This patch reverts back to hiding the logs. This won't affect the coordinator though because it uses a different set of settings. Note that this settings file is only included in release builds (.#cli-release).

Logs can be re-enabled by changing the `ReadStreamRequest` in settings.json to `true`.

### DIFF
--- a/packages/by-name/microsoft/genpolicy/genpolicy_msft_settings_prod.patch
+++ b/packages/by-name/microsoft/genpolicy/genpolicy_msft_settings_prod.patch
@@ -1,0 +1,13 @@
+diff --git a/genpolicy-settings.json b/genpolicy-settings.json
+index 7d35862af..4e468d78e 100644
+--- a/genpolicy-settings.json
++++ b/genpolicy-settings.json
+@@ -318,7 +318,7 @@
+             "regex": []
+         },
+         "CloseStdinRequest": false,
+-        "ReadStreamRequest": true,
++        "ReadStreamRequest": false,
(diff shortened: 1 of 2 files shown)


burgerdev (approved):
I think this is a good idea, but we should take a roadmap item to allow overriding this, maybe even on a per-workload basis. This also raises questions on whether I can hide the default config from users in #764.

Freax13:
> I think this is a good idea, but we should take a roadmap item to allow overriding this, maybe even on a per-workload basis. This also raises questions on whether I can hide the default config from users in #764.

Without #764, the user can just change the settings, though not on a per-workload basis. It might make sense to expose the whole settings file to the user because it also allows them to change more useful settings such as `ExecProcessRequest` which might be useful for debugging.

burgerdev:
> > I think this is a good idea, but we should take a roadmap item to allow overriding this, maybe even on a per-workload basis. This also raises questions on whether I can hide the default config from users in #764.
> 
> Without #764, the user can just change the settings, though not on a per-workload basis. It might make sense to keep exposing the whole settings file to the user because it also allows them to change more useful settings such as `ExecProcessRequest` which might be useful for debugging.

Acknowledged, will consider that in my PR.

burgerdev:
I think we can merge this, but should probably label as `breaking change`.

PR #772, opened 2024-07-30
treewide: change runtime-handler naming scheme 

msanft:
The Contrast runtime handlers are now named in the format `contrast-cc-<platform>-<hash>`, where `hash` is the hash of the relevant runtime components for `platform` and `platform` is the lowercase variant of the deployed platform.


go.mod
```diff
@@ -72,6 +72,7 @@ require (
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f // indirect
+	github.com/pelletier/go-toml/v2 v2.2.2 // indirect
```
Freax13, line 75: Was this dependency added intentionally?

node-installer/internal/constants/constants.go
```diff
@@ -14,6 +14,9 @@ import (
 )
 
 var (
+	// Version value is injected at build time.
+	Version = "0.0.0-dev"
```
Freax13, line 18: AFAICT this will use the same value for every development build. This is bad because the node-installer will skip some of its install actions if it detects that a runtime handler with the same name already exists.

internal/constants/constants.go
Freax13, line 1: The linker flags to inject the version information have to be updated.

node-installer/internal/constants/constants.go
Freax13, line 18: We already inject the version in `internal/constants/constants.go`. Why do we need to inject it here again?

msanft: I know, this will be handled in a follow-up for pseudo-versioning

msanft: Because Contrast imports the node-installer. We could place all versioning in the node-installer

go.mod
msanft: Yes, it's due to the node-installer being imported

node-installer/internal/constants/constants.go
Freax13: Is there a reason the node-installer can't just use the existing values? Either way it seems silly to inject the version in two different places if we could just use one.

Freax13: Personally, I'd rather have this be part of the same PR and not break people working on main.

msanft: I mean, you can still deploy, but you just need to clean up afterwards, or change the version.


I'm fine with implementing it, but want to keep the PR reviewable

msanft: If the node-installer imports from contrast, we have a cycle (which isn't allowed)

I'm fine with only specifying the version in the node-installer. @katexochen do you want to decide?

katexochen:
> The Contrast runtime handlers are now named in the format `contrast-cc-<version>-<platform>`, where `version` and `platform` are lowercase variants of the Contrast version

Why is this placing the version in the runtime handler name and not the hash of the runtime files?

node-installer/runtimehandler/runtimehandler.go
```diff
@@ -0,0 +1,26 @@
...
+	platformName := strings.ToLower(platform.String())
+
+	if strings.EqualFold(platformName, platforms.Unknown.String()) {
+		return "", fmt.Errorf("unsupported platform %s", platform)
+	}
+
+	// Replace dots to ensure a readable directory name used by the node-installer.
+	version := strings.ReplaceAll(constants.Version, ".", "-")
+
+	return fmt.Sprintf("contrast-cc-%s-%s", version, platformName), nil
```
burgerdev, line 25: Just so I understand it correctly: the prefix up to excluding the platformName is unique per build? If the middle `%s` depends on the right `%s`, I'd prefer to swap them.

Freax13:
The documentation needs to be updated to reflect the new name: https://github.com/edgelesssys/contrast/blob/319bd5091d90c7b0b2c939b36427affb3628156b/docs/docs/examples/emojivoto.md?plain=1#L79.

node-installer/internal/config/config.go
```diff
@@ -25,19 +22,6 @@ type Config struct {
...
-	if len(c.RuntimeHandlerName) > 63 {
-		return errors.New("runtimeHandlerName must be 63 characters or fewer")
-	}
-	matched, err := regexp.Match(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`, []byte(c.RuntimeHandlerName))
-	if err != nil {
-		return err
-	}
-	if !matched {
-		return errors.New("runtimeHandlerName must be a lowercase RFC 1123 subdomain")
-	}
```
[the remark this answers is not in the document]
msanft: Release tests makes sense. I'm not even sure how much sense it'd make to put this into the node-installer, since it's deployed with the actual `runtimeClass` Kubernetes resource, which should [fail to deploy](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) in that case?

packages/by-name/contrast/package.nix
```diff
@@ -113,14 +127,15 @@ buildGoModule rec {
     install -D ${genpolicy.rules}/genpolicy-rules.rego cli/genpolicy/assets/genpolicy-rules.rego
     install -D ${genpolicy.src}/src/kata-opa/allow-all.rego cli/genpolicy/assets/allow-all.rego
     install -D ${embeddedReferenceValues} internal/manifest/assets/reference-values.json
+    install -D ${embeddedPlatformHandlerMapping} internal/manifest/assets/platform-handler-mapping.json
```
Freax13, line 130: This also needs to be added to `packages/by-name/cli-release/package.nix`.

Freax13 (approved):

burgerdev (approved):
LGTM, thanks!

nodeinstaller/node-installer.go
```diff
@@ -204,12 +213,17 @@ func containerdRuntimeConfig(basePath, configPath string, platform platforms.Pla
...
 	existingRaw, existing, err := parseExistingContainerdConfig(configPath)
 	if err != nil {
 		existing = constants.ContainerdBaseConfig()
 	}
 
+	runtimeName, err := manifest.RuntimeHandler(platform)
+	if err != nil {
+		return fmt.Errorf("getting runtime name: %w", err)
+	}
+
```
katexochen, line 226: I think it is fine to keep on taking the runtime handler as input here. We shouldn't start to move calls to such global magic all the way down but rather try to call it somewhere top and pass the value down. This will also keep our test stable and independent from how the runtime handler embedding works, the change of test values in this PR is unnecessary.

Freax13:
We still need to change this path here:
https://github.com/edgelesssys/contrast/blob/81b98bed3d633da96975307d8c2a2b8cdc8114bd/packages/by-name/kata/contrast-node-installer-image/package.nix#L29-L31

internal/manifest/referencevalues.go
```diff
@@ -11,11 +11,17 @@ import (
...
+var (
+	// EmbeddedPlatformRuntimeMappingJSON is a 1:1 mapping of platform names to their runtime component hashes.
+	//
+	//go:embed assets/platform-runtime-mapping.json
+	EmbeddedPlatformRuntimeMappingJSON []byte
+
+	// EmbeddedReferenceValuesJSON contains the embedded reference values in JSON format.
+	//
+	//go:embed assets/reference-values.json
+	EmbeddedReferenceValuesJSON []byte
```
katexochen, line 23: I propose we make this a `map[string]ReferenceValues`, mapping the runtime handler name to the reference values of that platform. We can then drop the PlatformRuntimeMapping completely and instead iterate over the keys of this map and look for the known platform strings in the runtime handlers.

In a future PR, we can then transform this into a `map[string][]ReferenceValues`, allowing us to embed multiple values per platform, while transforming the ReferenceValues struct into something platform independent

```go
type ReferenceValues struct {
    SNP
    TDX
}
```

internal/manifest/referencevalues.go
```diff
@@ -25,6 +31,13 @@ type ReferenceValues struct {
 	BareMetalTDX *BareMetalTDXReferenceValues `json:"bareMetalTDX,omitempty"`
 }
 
+// PlatformRuntimeMapping is a 1:1 mapping of platform names to their runtime component hashes.
+type PlatformRuntimeMapping struct {
+	AKS          string `json:"aks,omitempty"`
+	BareMetalTDX string `json:"bareMetalTDX,omitempty"`
+	BareMetalSNP string `json:"bareMetalSNP,omitempty"`
```
katexochen, line 38: I don't like that this is silently introducing sort of a "PlatformGroup" mechanism without a clean notation.

packages/by-name/kata/contrast-node-installer-image/package.nix
```diff
@@ -40,62 +37,61 @@ let
           files = [
             {
               url = "file:///opt/edgeless/share/kata-containers.img";
-              path = "/opt/edgeless/${runtime-handler}/share/kata-containers.img";
+              path = "/share/kata-containers.img";
```
katexochen, line 40: Instead of just using the suffix here, should we use a placeholder to be more explicit instead?

katexochen (changes requested):

packages/by-name/microsoft/genpolicy/0004-genpolicy-regex-check-contrast-specific-layer-src-pr.patch
```diff
@@ -14,13 +14,12 @@ index 25c16bada..4f622a9f7 100644
 +++ b/src/tools/genpolicy/rules.rego
 @@ -887,7 +887,7 @@ allow_storage_options(p_storage, i_storage, layer_ids, root_hashes) {
      i_count == p_count + 3
- 
+
      print("allow_storage_options 2: i_storage.options[0] =", i_storage.options[0])
 -    i_storage.options[0] == "io.katacontainers.fs-opt.layer-src-prefix=/var/lib/containerd/io.containerd.snapshotter.v1.tardev/layers"
-+    regex.match(`io\.katacontainers\.fs-opt\.layer-src-prefix=/var/lib/containerd/io\.containerd\.snapshotter\.v1\.tardev-contrast-cc-[a-f0-9]{32}/layers`, i_storage.options[0])
- 
++    regex.match(`io\.katacontainers\.fs-opt\.layer-src-prefix=/var/lib/containerd/io\.containerd\.snapshotter\.v1\.tardev-contrast-cc-[^\/]*/layers`, i_storage.options[0])
```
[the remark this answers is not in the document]
msanft: What should this regex check? I wondered when looking at it

packages/by-name/kata/contrast-node-installer-image/package.nix
msanft: I don't really see the advantage. We control both the consumer and producer of this value and can align them. If it was a public API, I'd see the point, but I think this will just be unnecessary complexity here.

internal/manifest/referencevalues.go
msanft: I definitely agree to making `ReferenceValues` a platform-independent structure like you described.

Looking at how an implementation of `map[string]ReferenceValues` would look like, I'm not that big of a fan. This would require us to produce the runtime-handler values for each platform in Nix, as well as parsing the platforms out of the runtime handler names in Go. Both of these are doable, but I think the complexity for doing so outweighs the additional binary size costs of embedding the mapping.

packages/by-name/microsoft/genpolicy/0004-genpolicy-regex-check-contrast-specific-layer-src-pr.patch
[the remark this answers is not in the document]
katexochen: iirc this is passed directly as option to mount, so we need to check the value carefully.

internal/manifest/referencevalues.go
msanft: Decided to just go for the single embedding suggested by @katexochen. I think with a clever implementation, we can build this with reasonable complexity

Freax13: I don't particularly like the idea of extracting the platform the runtime handler name, but I also don't have any better ideas, so it's fine with me ig.

Freax13:
`DefaultPlatformHandler` and by extension `Manifest.RuntimeHandler` are no longer needed.

packages/by-name/microsoft/genpolicy/0004-genpolicy-regex-check-contrast-specific-layer-src-pr.patch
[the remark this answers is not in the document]
burgerdev: Actually, I don't think this is used at all by the agent, but rather by the snapshotter. The path would not be meaningful to the agent, anyway.

internal/manifest/referencevalues.go
```diff
@@ -25,6 +27,10 @@ type ReferenceValues struct {
 	BareMetalTDX *BareMetalTDXReferenceValues `json:"bareMetalTDX,omitempty"`
 }
 
+// EmbeddedReferenceValues is a map of runtime handler names to reference values, as
+// are embedded in the binary.
+type EmbeddedReferenceValues map[string]ReferenceValues
```
Freax13, line 32: This sentence is missing a word somewhere.

packages/by-name/kata/contrast-node-installer-image/package.nix
katexochen: The current design makes it impossible to place a file outside of /opt/edgelessys/....

msanft: Since we don't need that right now?

I will just implement it, but I don't see the point in generalizing things too early that don't need to be general.

burgerdev: How about making the paths relative? That would allow adding absolute paths, should we need them.

msanft: > How about making the paths relative? That would allow adding absolute paths, should we need them.

I favor the explicit replacing in that case. I think having conditional behaviour depending on whether the path's relative or not might be confusing

packages/by-name/contrast/package.nix
```diff
@@ -40,28 +40,55 @@ let
...
+          let
+            digest = builtins.substring 0 8 (builtins.readFile "${launchDigestFile}");
+          in
+          "contrast-cc-${platform}-${digest}"
+        );
+
+      aks-clh-snp-handler = runtimeHandler "aks-clh-snp" "${microsoft.runtime-class-files}/launch-digest.hex";
+      k3s-qemu-tdx-handler = runtimeHandler "k3s-qemu-tdx" "${kata.runtime-class-files}/launch-digest-tdx.hex";
+      rke2-qemu-tdx-handler = runtimeHandler "rke2-qemu-tdx" "${kata.runtime-class-files}/launch-digest-tdx.hex";
+      k3s-qemu-snp-handler = runtimeHandler "k3s-qemu-snp" "${kata.runtime-class-files}/launch-digest-snp.hex";
```
katexochen, line 58: this should use runtime-hash instead of launch-digest

packages/by-name/contrast/package.nix
```diff
@@ -40,28 +40,55 @@ let
...
-            snpVersion = 8;
-            microcodeVersion = 115;
+  embeddedReferenceValues =
+    let
+      runtimeHandler =
+        platform:
+        (
+          launchDigestFile:
+          let
+            digest = builtins.substring 0 8 (builtins.readFile "${launchDigestFile}");
```
Freax13, line 50:
```suggestion
            digest = builtins.substring 0 8 (builtins.readFile launchDigestFile);
```

Freax13 (approved):

katexochen (approved):

PR #775, opened 2024-07-31
Share Go module between node-installer and contrast

msanft:
Since the Contrast main package (i.e. cli, etc.) will need to share a significant portion of code to enable runtime-handler calculation for both, we transition to using the top-level module for the node-installer to enable imports of the internal packages of the Contrast main module.

Alternative options considered included code-sharing, which would introduce many duplications accross node-installer and main module, and would have required a duplicated injection of the reference values, and thus was considered infeasible. Another solution tried consisted of moving the platforms package to a top-level exported package in the main module, but since that node-installer is a child of the main module too, a replace tag does not work in this case, as the node-installer in `github.com/edgelesssys/contrast/node-installer` is unable to use a replacement for `github.com/edgelesssys/contrast`.


burgerdev (commented):
I do

internal/platforms/platforms.go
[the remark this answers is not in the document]
msanft: Since the runtime handler calculation depends on the embedded reference values (as they include the hashes of the runtime components).

This would require us to inject the reference values into both the node-installer and the main module, also requiring code duplication for the calculation.

We could also move the reference values (and the embedding) into the node-installer entirely, but that felt out-of-place, since the main usage of these happens in the manifest.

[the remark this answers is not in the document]
burgerdev: But then we still need to duplicate the asset injection, don't we?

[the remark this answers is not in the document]
msanft: As far as I could image now, we'd only need to inject into the node-installer then?

packages/by-name/contrast-node-installer/package.nix
```diff
@@ -16,21 +16,27 @@ buildGoModule {
...
+        (fileset.fileFilter (file: hasSuffix ".toml" file.name) node-installer)
+        (fileset.fileFilter (file: hasSuffix ".toml.tmpl" file.name) node-installer)
+        (fileset.fileFilter (file: hasSuffix ".go" file.name) node-installer)
+        # To prevent unnecessary rebuilds, and since the platforms package is the only
+        # imported package from the Contrast main module by the node-installer, we explicitly
+        # supply it here, rather than including "internal" as a whole.
+        (path.append root "internal/platforms")
       ];
     };
+  sourceRoot = "${src.name}/node-installer";
```
katexochen, line 36: use subPackages instead.

packages/by-name/contrast-node-installer/package.nix
```diff
@@ -16,21 +16,27 @@ buildGoModule {
...
-        (fileset.fileFilter (file: hasSuffix ".toml" file.name) root)
-        (fileset.fileFilter (file: hasSuffix ".toml.tmpl" file.name) root)
-        (fileset.fileFilter (file: hasSuffix ".go" file.name) root)
+        (fileset.fileFilter (file: hasSuffix ".toml" file.name) node-installer)
+        (fileset.fileFilter (file: hasSuffix ".toml.tmpl" file.name) node-installer)
+        (fileset.fileFilter (file: hasSuffix ".go" file.name) node-installer)
+        # To prevent unnecessary rebuilds, and since the platforms package is the only
+        # imported package from the Contrast main module by the node-installer, we explicitly
+        # supply it here, rather than including "internal" as a whole.
+        (path.append root "internal/platforms")
```
katexochen, line 33: I don't think this is a good idea, it will break the next person importing another package in a very cursed way.

katexochen (commented):
Why aren't we just building the node-installer as output of contrast then?

packages/by-name/contrast-node-installer/package.nix
msanft: It's a decision between having a lot of rebuilds and being developer-friendly.

I think having less rebuilds is better here, since the node-installer is less likely to be changed (especially with further additions from the Contrast package) than the other way around. What about adding a comment that explains precisely what to do to add another package instead?

packages/by-name/contrast/package.nix
```diff
@@ -130,6 +131,8 @@ buildGoModule rec {
     runHook postCheck
   '';
 
+  doCheck = false;
+
```
katexochen:
```suggestion

```

katexochen (approved):

PR #777, opened 2024-07-31
rfc: workload secrets

burgerdev:


rfc/007-workload-secrets.md
```diff
@@ -0,0 +1,108 @@
+# RFC 007: Workload secrets
+
+## Objective
+
+Provide Contrast workloads with secret key material, such that:
+
+1. Workloads can use key material to derive additional secrets autonomously.
+2. The key material is stable across workload upgrades.
+3. The key material is stable across manifest generations.
+4. The key material is stable across Contrast upgrades.
```
Freax13, line 10: Should we also use this secret key material to derive the private keys for our TLS CA?

rfc/007-workload-secrets.md
```diff
@@ -0,0 +1,108 @@
...
+### Extension to encrypted storage
+
+A persistent workload key could be used to set up transparently encrypted block storage.
+Due to the required permissions involved, and Kubernetes' volume sharing limitations, this would best be handled by a CVM-level service (a guest component in CoCo terms).
+
+### Manifest updates and workload owner trust
+
+Due to being derived from the secret seed, the workload key is stable across manifest updates.
+This implies that no migration is needed, as long as the `WorkloadSecretID` doesn't change.
+On the other hand, this means that the workload secret can be obtained by the workload owner, and should be treated accordingly.
```
m1ghtym0, line 103: Definitely something we should keep in mind when documenting this.

m1ghtym0 (approved):
Another great RFC, thanks!

rfc/007-workload-secrets.md
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+}
+```
+
+### Implementation
+
+After successful workload verification, the Coordinator derives a key with HKDF, setting the info argument to `workload-key:$WorkloadSecretID`.
+This key is returned as a new field `WorkloadKey` in the `NewMeshCertResponse`.
+The initializer writes the key to a file in the shared tmpfs, from where it can be used by the workload.
```
thomasten, line 90: This part is similar to MarbleRun. In addition, MarbleRun allows to inject the secret at different places and in different formats. With this, you can often avoid adapting your workload. If we would want something similar for Contrast, I guess it should be build upon the primitive you introduce with this RFC, right? For example, one could create an init container that transforms the secret such that the actual workload wouldn't need to be modified. Do you think this works or is there anything we'd need to consider for this RFC to enable this use case?

rfc/007-workload-secrets.md
burgerdev: We already do for the root CA, and the mesh CA is deliberately random.

rfc/007-workload-secrets.md
```diff
@@ -0,0 +1,108 @@
...
+[Hashicorp Vault]: https://www.hashicorp.com/products/vault
+[OpenBao]: https://openbao.org/
+[storage sealing]: https://developer.hashicorp.com/vault/docs/concepts/seal
+
+## Proposal
+
+### Secret source
+
+Workload secrets would need to be either persisted or deterministically derived.
+The only secure persistence option for the Coordinator is encryption with a key derived from the secret seed.
```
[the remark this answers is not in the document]
burgerdev: I like the idea of recovering the coordinator with hardware-sealed keys! Do you think such a mechanism should be used for workload secrets, too, or would it be sufficient to recover the Coordinator seed and derive the workload key?

rfc/007-workload-secrets.md
burgerdev: I guess it's a good sign that the proposals converge, even though I did not peek at the MarbleRun implementation :) 

The scheme proposed here is intentionally designed to support derivation of more keys, and these use cases should be enabled by it. 

The initializer (and other init containers) is somewhat limited in terms of what it can inject. Shared files are definitely possible, but env vars are not. I think the annotation pattern we use for the service mesh would be a good starting point for this.

rfc/007-workload-secrets.md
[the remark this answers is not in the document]
katexochen: I like it, but that should be part of another RFC. ;) Switching the source afterwards shouldn't impact the design much.

katexochen (approved):
lgtm, thanks! 

Implementation of 1) secret derivation and distribution and 2) workload identity in manifest should be possible in parallel.
Will the design of a guest-component for disk encryption be part of another RFC?

thomasten (approved):

rfc/007-workload-secrets.md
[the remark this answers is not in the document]
Freax13: > Do you think such a mechanism should be used for workload secrets, too, or would it be sufficient to recover the Coordinator seed and derive the workload key?

I didn't think about storing workload secrets, that's an interesting idea! I'm not quite sure what the differences are in practice.

3u13r, line 38: I think we should reference the recovery RFC here, since this is where this "secret seed" comes from, right?

3u13r (approved):
LGTM

rfc/007-workload-secrets.md
burgerdev: Done

burgerdev:
> Will the design of a guest-component for disk encryption be part of another RFC?

I believe it should be in a separate RFC, because delivering a secret seed is useful on its own (see use case). By the way, the capabilities of CDH are evolving rapidly and might be worth taking a look at first.

PR #785, opened 2024-08-02
manifest: add WorkloadSecretID field

3u13r:
This field will be used as part of the workload secret feature (RFC 007) to generate secrets.

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -16,10 +16,16 @@ import (
 // Manifest is the Coordinator manifest and contains the reference values of the deployment.
 type Manifest struct {
 	// policyHash/HOSTDATA -> commonName
-	Policies                map[HexString][]string
+	Policies                map[HexString]PolicyEntry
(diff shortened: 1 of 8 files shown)


internal/manifest/manifest.go
```diff
@@ -16,12 +16,18 @@ import (
...
+	Policies                map[HexString]PolicyEntry
 	ReferenceValues         ReferenceValues
 	WorkloadOwnerKeyDigests []HexString
 	SeedshareOwnerPubKeys   []HexString
 }
 
+// PolicyEntry is a policy entry in the manifest. It contains further information the user wants to associate with the policy.
+type PolicyEntry struct {
+	SANs             []string
+	WorkloadSecretID string
```
katexochen, line 28: Should this have omit empty for now?

katexochen (approved):

internal/manifest/manifest.go
3u13r: Sure, why not.

PR #791, opened 2024-08-06
coordinator: pass state as AuthInfo to meshapi

burgerdev:
This fixes a race condition between `SNPValidateOpts`, `ValidateCallback` and `SetManifest`. It should also make it easier to evolve validation logic based on state. 

Now that we have the current state available in the meshapi handler, we can move the certificate issuance there and not deal with that in the validation path.

### DIFF
--- a/coordinator/internal/authority/credentials.go
+++ b/coordinator/internal/authority/credentials.go
@@ -1,0 +1,134 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package authority
+
+import (
+	"context"
+	"encoding/asn1"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net"
+	"time"
+
+	"github.com/edgelesssys/contrast/internal/atls"
+	"github.com/edgelesssys/contrast/internal/attestation/snp"
+	"github.com/edgelesssys/contrast/internal/logger"
+	"github.com/edgelesssys/contrast/internal/memstore"
+	"github.com/google/go-sev-guest/proto/sevsnp"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"google.golang.org/grpc/credentials"
+	"k8s.io/utils/clock"
+)
+
+// Credentials are gRPC transport credentials that dynamically update with the Coordinator state.
+type Credentials struct {
+	issuer   atls.Issuer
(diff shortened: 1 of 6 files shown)


coordinator/internal/authority/credentials.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+		logger:                     a.logger,
+		attestationFailuresCounter: attestationFailuresCounter,
+		kdsGetter:                  kdsGetter,
+	}, ticker.Stop
+}
+
+// ServerHandshake implements an aTLS handshake for the latest state.
+//
+// If successful, the state will be passed to gRPC as AuthInfo.
```
msanft, line 66:
```suggestion
// If successful, the state will be passed to gRPC as [AuthInfo].
```
nit

coordinator/internal/authority/credentials.go
```diff
@@ -0,0 +1,135 @@
...
+// Info provides information about the protocol.
+func (c *Credentials) Info() credentials.ProtocolInfo {
+	return credentials.NewTLS(nil).Info()
+}
+
+// Clone the credentials object.
+func (c *Credentials) Clone() credentials.TransportCredentials {
+	cloned := *c
+	return &cloned
+}
```
msanft, line 107: Looking at the [upstream code](https://github.com/grpc/grpc-go/blob/e6b6318ad90847fa8b5ed987b964f1494179c89f/credentials/alts/alts.go#L270), I'm unsure whether a shallow copy is sufficeint here.

coordinator/internal/authority/credentials.go
burgerdev: Are these brackets actually interpreted by something, or do you suggest them for readability?

msanft: They are interpreted by `gopls`, and, when hovering, [show the context information for that type](https://tip.golang.org/doc/comment#package).

coordinator/internal/authority/credentials.go
burgerdev: I agree, and it seems like a bad idea to clone the KDS getter due to the embedded ticker. I checked the gRPC source and found that credentials are only cloned when client-side load balancing, so I decided to remove the implementation.

coordinator/internal/authority/credentials.go
burgerdev: Thanks, adapted.

msanft (approved):
Thanks! This is a huge improvement.

PR #792, opened 2024-08-07
just: credential getter as nix script

burgerdev:
Packaging the script with nix makes it work even if `gcloud` is not in the path.


katexochen (approved):
nice :)

PR #797, opened 2024-08-08
e2e: improve the regression test

miampf:
Improve the regression test by deploying a multitude of deployments with contrast.

Note: All deployments except `redis.yaml` and `nginx.yaml` use [quay](https://quay.io) as a package registry. Redis and Nginx use ghcr.io ([redis](https://ghcr.io/edgelesssys/redis), [nginx](https://ghcr.io/edgelesssys/nginx-unprivileged)). The reason for redis is that I couldn't find an official deployment on quay that also uses alpine. The [nginx/nginx-unprivileged](https://quay.io/nginx/nginx-unprivileged) image on quay seems to overwrite/delete hashes.

You can run this test locally by executing `just e2e regression`. A working CI run can be viewed [here](https://github.com/edgelesssys/contrast/actions/runs/10405416765/job/28816100112?pr=797)


e2e/regression/regression_test.go
```diff
@@ -0,0 +1,87 @@
...
+			c := kubeclient.NewForTest(t)
+			ct := contrasttest.New(t, imageReplacementsFile, namespaceFile, platform, false)
+			ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+			defer cancel()
+
+			resources := kuberesource.CoordinatorBundle()
+
+			yaml, err := os.ReadFile(yamlDir + file.Name())
+			require.NoError(err)
+			yaml = bytes.ReplaceAll(yaml, []byte("REPLACE_NAMESPACE"), []byte(ct.Namespace))
```
katexochen, line 54:
```suggestion
			yaml = bytes.ReplaceAll(yaml, []byte("@@REPLACE_NAMESPACE@@"), []byte(ct.Namespace))
```
Placeholders often use `@@` to be easier to spot.

e2e/regression/regression_test.go
```diff
@@ -0,0 +1,87 @@
...
+			resources = kuberesource.PatchRuntimeHandlers(resources, runtimeHandler)
+
+			resources = kuberesource.AddPortForwarders(resources)
+
+			ct.Init(t, resources)
+
+			require.True(t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
+			require.True(t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
+			require.True(t.Run("set", ct.Set), "contrast set needs to succeed for subsequent tests")
+			require.True(t.Run("contrast verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
```
katexochen, line 69:
```suggestion
			require.True(t.Run("verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
```

e2e/regression/test-data/mongodb.yaml
```diff
@@ -0,0 +1,35 @@
...
+      containers:
+        - name: mongodb
+          securityContext:
+            runAsUser: 101
+          image: quay.io/mongodb/mongodb-community-server@sha256:8b73733842da21b6bbb6df4d7b2449229bb3135d2ec8c6880314d88205772a11
+          volumeMounts:
+            - mountPath: /data/db
+              name: db
+            - mountPath: /data/configdb
+              name: configdb
```
katexochen, line 25: We should document all the workaround we had to make the the yaml to get things running. Or should we even revert them and let the test fail until these things are fixed? :thinking:

miampf: I would say we should leave the fixes in since they should not break things when the bug is fixed eventually. If we document this, the best place would probably be our troubleshooting section in the documentation, I would handle that in a seperate PR if it's fine for you?

miampf: I would say we should leave the fixes in, since they should not break things when the bug is fixed eventually. If we document this, the best place would probably be our troubleshooting section in the documentation, I would handle that in a separate PR if it's fine for you?

katexochen (approved):

PR #798, opened 2024-08-09
node-installer: run nydus snapshotter on bare metal platforms

katexochen:
Run the [nydus-snapshotter](https://github.com/containerd/nydus-snapshotter) on all bare-metal platforms. This snapshotter is used to implement the Kata image guest-pull workflow, where the image is pulled directly inside the CVM. The snapshotter on the host will add the require information to requests the shim will then send to the agent. The agent will use image-rs to pull the image from a registry.

The snapshotter currently requires setting a config options of the containerd plugin. As the table header of this option is part of the k3s/RKE base template, we need to disable the templating to not add the header twice to the document. Opposed to what is written in the k3s docs, it isn't possible to just delete the template and write the config itself, so we copy the content of the rendered config into the template and modify the template onward.

The piece missing in this PR is the installation of `nydus-overlayfs` on the host. This is a mount helper used to create a snapshot on the host while maintaining the custom mount labels that should be delivered to the kata-agent. It is difficult to version as it isn't possible to call it by path, and it's name is relied upon in the agent. It will be installed as part of a followup PR (present on our TDX dev machine for testing purpose).


nodeinstaller/node-installer.go
```diff
@@ -202,21 +186,38 @@ func containerdRuntimeConfig(basePath, configPath string, platform platforms.Pla
...
 	}
 
-	snapshotterName := "no-snapshotter"
-	// Add tardev snapshotter, only required for AKS
-	if platform == platforms.AKSCloudHypervisorSNP {
-		if existing.ProxyPlugins == nil {
-			existing.ProxyPlugins = make(map[string]config.ProxyPlugin)
-		}
+	// Enable containerd debug logging.
+	existing.Debug.Level = "debug"
```
burgerdev, line 195: Should we keep debug logging off on AKS?

packages/by-name/nydus-snapshotter/package.nix
```diff
@@ -33,6 +35,22 @@ buildGoModule rec {
     "-X github.com/containerd/nydus-snapshotter/version.Version=${version}"
   ];
 
+  passthru = {
+    # Based on https://github.com/confidential-containers/operator/blob/main/install/pre-install-payload/remote-snapshotter/nydus-snapshotter/config-coco-guest-pulling.toml
```
burgerdev, line 39: Consider a permalink.

burgerdev (approved):

nodeinstaller/node-installer.go
katexochen: I've changed this so it is only enabled for debug runtimes on all platforms.

burgerdev: Nice, thanks.

PR #801, opened 2024-08-09
runtime-class-files: remove, calculate runtime hash from all files

katexochen:
Previously, we tried to calc the hash over the "runtime files", but actually all files and the config itself are relevant to the runtime (that's why we are packing them into the image), so they should all be included in the hash (we were specifically missing everything snapshotter related). Sadly, we cannot include the node-installer at this point, as we embed the runtimeHash into it (but we should! and pass the hash via flag afterwards?). But the other files should now automatically be included in case we add something.


msanft (commented):
Seems reasonable to have this, but we should then disallow overwriting the config through a configmap, as this would not result in an altered hash, even though the config is different.

katexochen:
> Seems reasonable to have this, but we should then disallow overwriting the config through a configmap, as this would not result in an altered hash, even though the config is different.

We didn't cover that previously in the hash, so I don't see how it is relevant now that the config can be overwritten.

msanft:
> > Seems reasonable to have this, but we should then disallow overwriting the config through a configmap, as this would not result in an altered hash, even though the config is different.
> 
> We didn't cover that previously in the hash, so I don't see how it is relevant now that the config can be overwritten.

Ok, fair enough. So we'll just consider changes made by the user to be their responsibility.

msanft (approved):
Refactoring seems good to me, change does too. LGTM if tests pass.

As you said, we should probably create a follow-up for building a reproducible node-installer without build-time embeddings of the hash.

packages/by-name/microsoft/kata-igvm/package.nix
```diff
@@ -4,53 +4,68 @@
...
   igvm-tooling,
   igvm-signing-keygen,
+  igvmmeasure,
 }:
 
-stdenv.mkDerivation rec {
-  pname = "kata-igvm";
-  inherit (microsoft.genpolicy) src version;
+let
+  igvm = stdenv.mkDerivation rec {
```
Freax13, line 15: What's the point of moving this into a let binding?

katexochen: Allows access to the attribute in passthru, as let bindings are recursive. But I guess we could also access it through 
`microsoft.igvm` from the input attrs.

packages/by-name/microsoft/kata-igvm/package.nix
```diff
@@ -49,6 +51,15 @@ stdenv.mkDerivation rec {
     runHook postBuild
   '';
 
+  passthru.launch-digest = stdenvNoCC.mkDerivation {
+    name = "runtime-class-files";
```
Freax13, line 55:
```suggestion
    name = "launch-digest";
```

Freax13 (approved):

PR #803, opened 2024-08-09
authority: remove unused Bundle struct

3u13r:
This struct no longer is used since https://github.com/edgelesssys/contrast/pull/791.


burgerdev (approved):
Thanks!

PR #809, opened 2024-08-15
resourcegen: add flag to configure logging;  snp/issuer: log quote provider in use

katexochen:
Add the supported env vars to the deployment so configuring them is easier. This also let's the Coordinator default to log level debug for now.

### DIFF
--- a/e2e/internal/contrasttest/contrasttest.go
+++ b/e2e/internal/contrasttest/contrasttest.go
@@ -128,7 +128,7 @@ func (ct *ContrastTest) Init(t *testing.T, resources []any) {
 	resources = kuberesource.PatchNamespaces(resources, ct.Namespace)
 	resources = kuberesource.PatchServiceMeshAdminInterface(resources, 9901)
 	resources = kuberesource.PatchCoordinatorMetrics(resources, 9102)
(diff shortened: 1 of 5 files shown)


Freax13:
Did you intend to include 3708e91b17971e08a1c5ce0dfaf0b84b311a3e46 in this pr?

internal/kuberesource/parts.go
```diff
@@ -325,6 +325,11 @@ func Coordinator(namespace string) *CoordinatorConfig {
 							).
 							WithResources(ResourceRequirements().
 								WithMemoryLimitAndRequest(100),
+							).
+							WithEnv(
+								NewEnvVar("CONTRAST_LOG_LEVEL", "debug"),
+								NewEnvVar("CONTRAST_LOG_FORMAT", "text"),
+								NewEnvVar("CONTRAST_LOG_SUBSYSTEMS", "*"),
```
Freax13, line 332: Is it safe to do this in production? Is there a way we can enable this only for dev builds?

katexochen: refactored it.

Freax13 (approved):

PR #812, opened 2024-08-16
e2e: specify multiple reference values in openssl test

msanft:
The openssl test now additionally uses multiple reference values to test attestation with multiple validators. Only the second one of the provided validators should successfully validate the attestation document.

### DIFF
--- a/e2e/openssl/openssl_test.go
+++ b/e2e/openssl/openssl_test.go
@@ -265,16 +265,21 @@ func toPtr[T any](t T) *T {
 	return &t
 }
 
+// patchReferenceValues modifies the manifest to contain multiple reference values for testing
+// cases with multiple validators, as well as filling in bare-metal SNP-specific values.
 func patchReferenceValues(t *testing.T, platform platforms.Platform, ct *contrasttest.ContrastTest) {
+	manifestBytes, err := os.ReadFile(ct.WorkDir + "/manifest.json")
+	require.NoError(t, err)
+	var m manifest.Manifest
+	require.NoError(t, json.Unmarshal(manifestBytes, &m))
+
+	// Duplicate the reference values to test multiple validators by having at least 2.
+	m.ReferenceValues.SNP = append(m.ReferenceValues.SNP, m.ReferenceValues.SNP[len(m.ReferenceValues.SNP)-1])
+
+	// Fill in bare-metal-SNP-specific values.
 	if platform == platforms.K3sQEMUSNP {
 		// The generate command doesn't fill in all required fields when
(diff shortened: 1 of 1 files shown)


Freax13:
Don't we already have multiple reference values on bare metal?

msanft:
> Don't we already have multiple reference values on bare metal?

How come? I don't think so.

katexochen (approved):

Freax13:
> > Don't we already have multiple reference values on bare metal?
> 
> How come? I don't think so.

We have multiple reference values for baremetal: Milan and Genoa.

Freax13:
See https://github.com/edgelesssys/contrast/pull/807

msanft:
> We have multiple reference values for baremetal: Milan and Genoa.

Thanks, i missed that. However, adding it for both AKS and bare-metal doesn't hurt, and makes the code a little less dependant on the embedding of multiple reference values being present, which I like

e2e/openssl/openssl_test.go
```diff
@@ -265,26 +265,38 @@ func toPtr[T any](t T) *T {
 	return &t
 }
 
-func patchReferenceValues(t *testing.T, platform platforms.Platform, ct *contrasttest.ContrastTest) {
+// patchManifest modifies the manifest to contain multiple reference values for testing
+// cases with multiple validators, as well as filling in bare-metal SNP-specific values.
+func patchManifest(t *testing.T, platform platforms.Platform, ct *contrasttest.ContrastTest) {
```
Freax13, line 270: Why was this function renamed?

e2e/openssl/openssl_test.go
```diff
@@ -265,26 +265,38 @@ func toPtr[T any](t T) *T {
...
+
+	// Duplicate the reference values to test multiple validators.
+	m.ReferenceValues.SNP = append(m.ReferenceValues.SNP, m.ReferenceValues.SNP[0])
+	// Make the first set of reference values invalid by changing the SVNs.
+	m.ReferenceValues.SNP[0].MinimumTCB = manifest.SNPTCB{
+		BootloaderVersion: toPtr(manifest.SVN(255)),
+		TEEVersion:        toPtr(manifest.SVN(255)),
+		SNPVersion:        toPtr(manifest.SVN(255)),
+		MicrocodeVersion:  toPtr(manifest.SVN(255)),
+	}
```
Freax13, line 284: The code below undoes these changes rendering the reference values valid again (though only on bare metal).

Freax13 (approved):

PR #814, opened 2024-08-16
node-installer: fix debug runtime measurement

katexochen:


packages/by-name/kata/contrast-node-installer-image/package.nix
```diff
@@ -16,7 +16,7 @@
   OVMF-SNP,
   OVMF,
 
-  debugRuntime ? false,
+  debugRuntime ? true,
```
Freax13, line 19: Did you intend to set the default to `true`?

Freax13 (approved):

PR #816, opened 2024-08-19
ci: only update packages in rpm update

msanft:
This fixes unexpected files, such as the btrfs image used in CI to be committed into the PR branch, as when no paths are specified, all are added.

[Example Run](https://github.com/edgelesssys/contrast/actions/runs/10448369515)

### DIFF
--- a/.github/actions/setup_nix/action.yml
+++ b/.github/actions/setup_nix/action.yml
@@ -23,9 +23,9 @@ runs:
       shell: bash
       run: |
         sudo mkdir /nixbld
-        truncate -s 3G btrfs.img
-        sudo mkfs.btrfs -f btrfs.img
-        sudo mount btrfs.img /nixbld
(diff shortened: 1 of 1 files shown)


katexochen:
Shouldn't we create the btrfs file outside of the git repo?

.github/actions/setup_nix/action.yml
```diff
@@ -23,9 +23,9 @@ runs:
       shell: bash
       run: |
         sudo mkdir /nixbld
-        truncate -s 3G btrfs.img
-        sudo mkfs.btrfs -f btrfs.img
-        sudo mount btrfs.img /nixbld
+        truncate -s 3G /btrfs.img
```
katexochen, line 26:
```suggestion
        sudo truncate -s 3G /btrfs.img
```

PR #817, opened 2024-08-19
manifest: add CPU model (aka product name) to reference values

Freax13:
The minimum TCB values are different for different CPU models, so we need to pin the reference values to a single CPU model.

Previously, we used a single set of `verify.Options` for all reference values and used different `validate.Options` for different sets of reference values. We can no longer do that, because it's not possible to validate a CPU model through `validate.Options`, this is only possible through `verify.Options`. Now, we generate a new pair of `verify.Options` and `validate.Options` for each set of reference values. A big part of the changes in this PR is related to moving away from the single global `verify.Options` to unique `verify.Options`s for each set of reference values.


internal/manifest/manifest.go
```diff
@@ -99,6 +102,12 @@ func (r SNPReferenceValues) Validate() error {
 		return fmt.Errorf("field MicrocodeVersion in manifest cannot be empty")
 	}
 
+	switch r.ProductName {
+	case Milan, Genoa:
```
msanft, line 106: I'd favor a comment here saying that this is meant to not do anything in that case. If we leave it like this, it might look like the case has just been forgotten.

msanft (approved):

katexochen (approved):

PR #820, opened 2024-08-20
ci: always cleanup runner, create nix builder volume under /mnt

katexochen:
github hosted runners have a second disk under /mnt that we can use for the builder volume.

### DIFF
--- a/.github/actions/setup_nix/action.yml
+++ b/.github/actions/setup_nix/action.yml
@@ -22,11 +22,23 @@ runs:
     - name: use btrfs for nix builds
       shell: bash
       run: |
-        sudo mkdir /nixbld
-        truncate -s 3G btrfs.img
-        sudo mkfs.btrfs -f btrfs.img
-        sudo mount btrfs.img /nixbld
+        sudo df -h
+
+        echo "Removing unwanted software... "
+        sudo rm -rf /usr/share/dotnet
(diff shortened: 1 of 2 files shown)


.github/actions/setup_nix/action.yml
```diff
@@ -22,11 +22,23 @@ runs:
...
-        sudo mount btrfs.img /nixbld
+        sudo df -h
+
+        echo "Removing unwanted software... "
+        sudo rm -rf /usr/share/dotnet
+        sudo rm -rf /usr/local/lib/android
+        sudo rm -rf /opt/ghc
+        sudo rm -rf /opt/hostedtoolcache/CodeQL
+        sudo docker system prune --all --force
+        sudo df -h
```
msanft, line 33: I think we might want to drop the `df`s here. Calculating the disk usage takes quite some time

msanft (approved):
Nice one! I wasn't aware

.github/actions/setup_nix/action.yml
katexochen: The full action takes 20s, I don't think this needs optimization, especially as we likely need further work on it.

katexochen: Maybe you're confusing `df` with `du`?

Freax13 (approved):

PR #825, opened 2024-08-23
e2e: upload container logs as artifact

miampf:
Upload container logs of the e2e namespace as an artifact in each e2e test. This also adds a script in `packages/scripts.nix` that can be used to run log collection locally (execute `nix run .#scripts.get-logs <PATH_TO_NAMESPACE_FILE>`).


packages/scripts.nix
```diff
@@ -317,4 +317,34 @@
...
+          logfile="workspace/namespace-logs/$pod.log"
+          if ! [[ -f "$logfile" ]]; then
+            {
+              touch "$logfile" # prevents creation of to much processes
+              # wait for all containers of the pod to come online, then collect the logs
+              kubectl wait pod --all --for=condition=Ready --timeout="-1s" -n "$namespace" "$pod" 1>/dev/null 2>/dev/null
+              kubectl logs -f --all-containers=true -n "$namespace" "$pod" > "$logfile"
+            } &
+          fi
+        done <<< "$pods"
```
Freax13, line 345:
```suggestion
        for pod in $pods; do
          logfile="workspace/namespace-logs/$pod.log"
          if ! [[ -f "$logfile" ]]; then
            {
              touch "$logfile" # prevents creation of to much processes
              # wait for all containers of the pod to come online, then collect the logs
              kubectl wait pod --all --for=condition=Ready --timeout="-1s" -n "$namespace" "$pod" 1>/dev/null 2>/dev/null
              kubectl logs -f --all-containers=true -n "$namespace" "$pod" > "$logfile"
            } &
          fi
        done
```

Is there a reason we can't just use a simple for loop?

miampf: No, I believe that this is an artifact from when I still was mapping out the script. I changed it :)

Freax13 (approved):

PR #826, opened 2024-08-23
manifest: use tdx validator

Freax13:
This PR implements the necessary functions to generate validation options for TDX and adds the TDX validators to the list of validators. This is enough to make `set` work.

Some caveats/TODOs:
- We don't cache certificates yet.
- We don't validate the launch measurements (we only validate the TCB).
- The coordinator doesn't add the certificate extensions yet.


msanft (approved):
LGTM

katexochen (approved):

PR #832, opened 2024-08-27
nydus-snapshotter: disable database flock timeout

Freax13:
Removing this timeout prevents the nydus-snapshotter from erroring out if there's already another instance running.


katexochen (approved):
Nice!

PR #836, opened 2024-08-27
certcache: don't cache CRL URLs

Freax13:
Unlike VCEKs, these can change over time and the KDS doesn't rate-limit requests to these. Caching and returning old CRLs leads to problems because the go-sev-guest library assumes that it always gets a fresh CRL.


msanft (approved):
Caching CRLs sounds - in all cases - like something that can ultimately lead to accepting a revoked certificate. So LGTM

PR #838, opened 2024-08-28
ci: don't fail fast

Freax13:
The matrix strategy defaults to failing fast: If any one of the test cases fails all other cases get cancelled as well. The problem with this is that sometimes tests fail not because the code is bad, but because the test is flaky. In an ideal world, we'd make the tests less flaky, but until then we shouldn't abort all other tests in the matrix. The advantage of not failing fast is that when we inevitably have to manually push the re-run button, we don't have to wait for all test cases to complete, but only for the one that failed.

### DIFF
--- a/.github/workflows/e2e_kubernetes.yaml
+++ b/.github/workflows/e2e_kubernetes.yaml
@@ -26,6 +26,7 @@ jobs:
     strategy:
       matrix:
         test_name: [servicemesh, openssl, policy, workloadsecret]
+      fail-fast: false
 
     runs-on: ubuntu-22.04
     permissions:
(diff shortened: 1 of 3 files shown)


Freax13:
> The disadvantage is that all tests will take way longer if the code is bad, right?

Following this change failing tests would take the same amount (or less) time as succeeding tests.

msanft:
> > The disadvantage is that all tests will take way longer if the code is bad, right?
> 
> Following this change failing tests would take the same amount (or less) time as succeeding tests.

Won't we run all payloads, even if all will fail?

katexochen (approved):

PR #855, opened 2024-09-05
just: always use -L flag on nix run

katexochen:


Freax13 (approved):
It'd be nice to have an environment variable for this in Nix.

PR #858, opened 2024-09-05
packages/cloud-api-adaptor: init at 0.9.0

katexochen:
Packaging the CAA binaries, based on https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/Makefile

### DIFF
--- a/packages/by-name/cloud-api-adaptor/package.nix
+++ b/packages/by-name/cloud-api-adaptor/package.nix
@@ -1,0 +1,67 @@
+# Copyright 2024 Edgeless Systems GmbH
+# SPDX-License-Identifier: AGPL-3.0-only
+
+{
+  lib,
+  buildGoModule,
+  fetchFromGitHub,
+  pkg-config,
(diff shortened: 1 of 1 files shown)


packages/by-name/cloud-api-adaptor/package.nix
```diff
@@ -0,0 +1,54 @@
...
+
+  # List of supported cloud providers
+  builtinCloudProviders ? [
+    "aws"
+    "azure"
+    "gcp"
+    "ibmcloud"
+    "vsphere"
+    # "libvirt"
+    # "docker"
```
Freax13, line 14: Why are these disabled?

katexochen: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/Makefile#L31

Freax13 (approved):

PR #861, opened 2024-09-07
e2e: add test for luks encryted state disk

3u13r:
This is a precursor to the end-user documentation on how to use the workloadsecret to mount a luks-encrypted CSI disk.

### DIFF
--- a/e2e/volumestatefulset/volumestatefulset_test.go
+++ b/e2e/volumestatefulset/volumestatefulset_test.go
@@ -1,0 +1,116 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//go:build e2e
+
+package volumestatefulset
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/edgelesssys/contrast/e2e/internal/contrasttest"
+	"github.com/edgelesssys/contrast/e2e/internal/kubeclient"
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/platforms"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	imageReplacementsFile, namespaceFile, platformStr string
+	skipUndeploy                                      bool
+)
+
+// TestWorkloadSecrets tests that secrets are correctly injected into workloads.
+func TestVolumeStatefulSet(t *testing.T) {
+	platform, err := platforms.FromString(platformStr)
+	require.NoError(t, err)
+	ct := contrasttest.New(t, imageReplacementsFile, namespaceFile, platform, skipUndeploy)
+
+	runtimeHandler, err := manifest.RuntimeHandler(platform)
+	require.NoError(t, err)
+
+	resources := kuberesource.VolumeStatefulSet()
+
+	coordinator := kuberesource.CoordinatorBundle()
+
+	resources = append(resources, coordinator...)
+
+	resources = kuberesource.PatchRuntimeHandlers(resources, runtimeHandler)
+
+	resources = kuberesource.AddPortForwarders(resources)
+
+	ct.Init(t, resources)
+
+	require.True(t, t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
+
+	require.True(t, t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
+
+	require.True(t, t.Run("set", ct.Set), "contrast set needs to succeed for subsequent tests")
+
+	require.True(t, t.Run("contrast verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
+
+	require.True(t, t.Run("deployments become available", func(t *testing.T) {
+		require := require.New(t)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+		defer cancel()
+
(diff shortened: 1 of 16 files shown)


internal/kuberesource/sets.go
```diff
@@ -496,3 +497,128 @@ func Emojivoto(smMode serviceMeshMode) []any {
...
+# workload_secret_path is the path to the Contrast workload secret.
+workload_secret_path="/contrast/secrets/workload-secret-seed"
+# tmp_key_path is the path to a temporary key file.
+tmp_key_path="/dev/shm/key"
+# disk_encryption_key_path is the path to the disk encryption key.
+disk_encryption_key_path="/dev/shm/disk-key"
+
+# If the device is already a LUKS device, format it.
+if ! cryptsetup isLuks "${device}"; then
+	# Generate a random key for the first initialization.
```
Freax13, line 519: Have you considered removing the key slot with this key after initialization?

justfile
```diff
@@ -101,10 +104,11 @@ generate cli=default_cli platform=default_platform:
     nix run -L .#{{ cli }} -- generate \
         --workspace-dir ./{{ workspace_dir }} \
         --image-replacements ./{{ workspace_dir }}/just.containerlookup \
-        --reference-values {{ platform }}\
+        --reference-values {{ platform }} \
+        --genpolicy-cache-path ./{{ workspace_dir }}/../layers-cache.json \
         ./{{ workspace_dir }}/deployment/*.yml
 
-    # On baremetal SNP, we don't have default values for MinimumTCB, so we need to set some here.
+    # On baremetal SNP, we do not have default values for MinimumTCB, so we need to set some here.
```
Freax13, line 111: Why was this changed?

packages/by-name/microsoft/genpolicy/package.nix
```diff
@@ -45,6 +45,8 @@ rustPlatform.buildRustPackage rec {
       #
       # This patch is not upstreamable.
       ./0004-genpolicy-regex-check-contrast-specific-layer-src-pr.patch
+      ./0005-genpolicy-propagate-mount_options-for-empty-dirs.patch
+      ./0006-genpolicy-support-HostToContainer-mount-propagation.patch
```
Freax13, line 49: Can we add some comments here for the new patches?

packages/containers.nix
```diff
@@ -78,6 +78,27 @@ let
...
+        bash
+        bashInteractive
+        coreutils
+        ncurses
+        procps
+        cryptsetup
+        e2fsprogs # mkfs.ext4
+        mount
+        util-linux # blkid
+        openssl
```
Freax13, line 94: Are all of these needed?

packages/by-name/microsoft/genpolicy/package.nix
burgerdev: In particular what the upstream status is.

justfile
3u13r: The `'` broke my syntax highlighting, so I changed it temporarily. Since it's unrelated I'll revert this.

justfile
```diff
@@ -101,10 +104,11 @@ generate cli=default_cli platform=default_platform:
     nix run -L .#{{ cli }} -- generate \
         --workspace-dir ./{{ workspace_dir }} \
         --image-replacements ./{{ workspace_dir }}/just.containerlookup \
-        --reference-values {{ platform }}\
+        --reference-values {{ platform }} \
+        --genpolicy-cache-path ./{{ workspace_dir }}/../layers-cache.json \
```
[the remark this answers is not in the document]
3u13r: Ah sorry, I oversaw this. But generally we currenly save the layers-cache in the workspace folder, which we clean up on every `just default` command and I wanted some deployment speed. Since this is the justfile we might discuss doing such a change (in another PR) though.

e2e/volumestatefulset/volumestatefulset_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+
+		pods, err := ct.Kubeclient.PodsFromOwner(ctx, ct.Namespace, "StatefulSet", "volume-tester")
+		require.NoError(err)
+		require.Len(pods, 1)
+
+		require.NoError(ct.Kubeclient.Restart(ctx, kubeclient.StatefulSet{}, ct.Namespace, "volume-tester"))
+		require.NoError(ct.Kubeclient.WaitFor(ctx, kubeclient.StatefulSet{}, ct.Namespace, "volume-tester"))
```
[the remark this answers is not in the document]
3u13r: Good (c/p)atch though. Thanks!

packages/containers.nix
3u13r: Combined the bash/shell related ones to busybox. Other than that, yes. I haven't measured it, but I tried not having to include openssl here, but I didn't find another option of doing hkdf "in a shell". We _might_ convert this to a Go program but we'd have to link againt libcryptsetup which is a bit of a pain in Constellation.

internal/kuberesource/sets.go
3u13r: Ok that was easier than imagined. I thought I needed multiple commands to find out and then delete the key slot. Apparently not.

packages/containers.nix
burgerdev: The bash program does a better job at illustrating the ideas, though.

internal/kuberesource/sets.go
burgerdev: Excuse my ignorance: what's the benefit of doing this, as opposed to initializing with the derived key right away?

burgerdev (approved):

internal/kuberesource/sets.go
Freax13: We derive the key from the UUID of the LUKS partition and we don't know the UUID before we create it.

Freax13 (approved):

PR #867, opened 2024-09-10
docs: mention the need to override the coordinator policy hash

Freax13:


docs/docs/deployment.md
```diff
@@ -303,6 +303,10 @@ This will use the reference values from the manifest file to attest the Coordina
 After this step, the Coordinator will start issuing TLS certificates to the workloads. The init container
 will fetch a certificate for the workload and the workload is started.
 
+:::warning
+On bare metal, the coordinator policy hash must be overwritten when using the `set`, `verify`, and `recover` commands. Look up the corresponding value for your platform in the `coordinator-policy.hash` release file and pass it using `--coordinator-policy-hash`.
+:::
+
```
burgerdev, line 309: Can we move this message to a dedicated section in, say, `docs/docs/components/policies.md` and only link it here? This would also make adding the callout to every subsection where it's relevant less cumbersome.

In the doc section over there, we could then

* explain what the released policy hashes are,
* why there is more than one and
* how to work on an unsupported platform.

@katexochen

Freax13: >   * how to work on an unsupported platform.

What do you mean by "unsupported platform"?

burgerdev: Sorry, the term platform was misleading here. What I meant is a system on which the published policies won't work, e.g. due to unexpected containerd settings. The user might be able to work around this with the genpolicy settings, but this probably warrants a dedicated docs PR.

docs/docs/components/policies.md
```diff
@@ -74,3 +74,15 @@ Contrast verifies its confidential containers following these steps:
...
 After the last step, we know that the policy hasn't been tampered with and, thus, that the workload matches expectations and may receive mesh certificates.
+
+## Platform Differences
+
+Contrast uses different rules and data sections for different platforms.
+This results in different policy hashes for different platforms.
+
+The `generate` command automatically derives the correct set of rules and data sections from the `reference-values` flag.
+
+The `verify`, `set`, and `recover` commands need to know the coordinator's policy hash for in order to securely communicate with it.
```
burgerdev, line 85:
```suggestion
The `verify`, `set`, and `recover` commands need to know the coordinator's expected policy hash to verify its identity.
```

burgerdev (approved):

PR #871, opened 2024-09-11
infra: terraform resources for peerpods on Azure

katexochen:
Initial setup to get peer pods working on AKS, based on https://confidentialcontainers.org/docs/cloud-api-adaptor/azure/

### DIFF
--- a/infra/azure-peerpods/main.tf
+++ b/infra/azure-peerpods/main.tf
@@ -1,0 +1,184 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "4.1.0"
+    }
+    azuread = {
+      source  = "hashicorp/azuread"
+      version = "2.53.1"
+    }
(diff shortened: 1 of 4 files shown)


infra/azure-peerpods/main.tf
```diff
@@ -0,0 +1,184 @@
...
+  }
+}
+
+resource "local_file" "kubeconfig" {
+  depends_on = [azurerm_kubernetes_cluster.cluster]
+  filename   = "./kube.conf"
+  content    = azurerm_kubernetes_cluster.cluster.kube_config_raw
+}
+
+resource "local_file" "env" {
```
msanft, line 133: I assume that this is required by the further setup steps of CoCo? Or why do we have that here?

katexochen: Yes, output is still wip, we need values from the env file to put them elsewhere.

msanft (approved):

PR #872, opened 2024-09-13
tdx: calculate & verify MRTD

Freax13:
This PR adds a new tool, tdx-measure, uses it to pre-calculate the MRTD and verifies the MRTD.

### DIFF
--- a/packages/by-name/tdx-measure/package.nix
+++ b/packages/by-name/tdx-measure/package.nix
@@ -1,0 +1,36 @@
+# Copyright 2024 Edgeless Systems GmbH
+# SPDX-License-Identifier: AGPL-3.0-only
+
+{ buildGoModule }:
+
+buildGoModule rec {
+  pname = "tdx-measure";
+  version = "0.1.0";
+
+  # The source of the main module of this repo. We filter for Go files so that
+  # changes in the other parts of this repo don't trigger a rebuild.
+  src = ../../../tools/tdx-measure;
+
+  proxyVendor = true;
+  vendorHash = "sha256-NmgZdeYVeoG/r2i6AsoZQrikTk5qRK17uaXsY9ks0qg=";
+
+  subPackages = [ "." ];
+
+  CGO_ENABLED = 0;
+  ldflags = [
+    "-s"
+    "-X main.version=v${version}"
+  ];
+
+  preCheck = ''
+    export CGO_ENABLED=1
+  '';
+
+  checkPhase = ''
+    runHook preCheck
+    go test -race ./...
+    runHook postCheck
+  '';
+
+  meta.mainProgram = "tdx-measure";
+}
--- a/tools/tdx-measure/main.go
+++ b/tools/tdx-measure/main.go
@@ -1,0 +1,83 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package main
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	"os"
+
+	"github.com/edgelesssys/contrast/tdx-measure/tdvf"
+	"github.com/spf13/cobra"
+)
+
+var version = "0.0.0-dev"
+
+func main() {
+	if err := execute(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func execute() error {
+	cmd := newRootCmd()
+	return cmd.ExecuteContext(context.Background())
+}
+
+func newRootCmd() *cobra.Command {
+	root := &cobra.Command{
+		Version: version,
+		Use:     "tdx-measure",
+		Short:   "tdx-measure",
+	}
+	root.SetOut(os.Stdout)
+
+	root.InitDefaultVersionFlag()
+	root.AddCommand(
+		newMrTdCmd(),
+	)
+
+	return root
+}
+
+// newMrTdCmd creates the tdx-measure mrtd subcommand.
+func newMrTdCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "mrtd -f OVMF.fd",
+		Short: "calculate the MRTD for a firmware file",
+		Long: `Calculate the MRTD for a firmware file.
+
+		This will parse the firmware according to the TDX Virtual Firmware Design Guide and
+		pre-calculate the MRTD.`,
+		RunE: runMrTd,
+	}
+	cmd.Flags().StringP("firmware", "f", "OVMF.fd", "path to firmware file")
+	if err := cmd.MarkFlagFilename("firmware", "fd"); err != nil {
+		panic(err)
+	}
+	return cmd
+}
+
+func runMrTd(cmd *cobra.Command, _ []string) error {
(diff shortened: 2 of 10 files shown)


packages/by-name/tdx-measure/package.nix
```diff
@@ -0,0 +1,49 @@
...
+
+  proxyVendor = true;
+  vendorHash = "sha256-NmgZdeYVeoG/r2i6AsoZQrikTk5qRK17uaXsY9ks0qg=";
+
+  subPackages = [ "." ];
+
+  CGO_ENABLED = 0;
+  ldflags = [
+    "-s"
+    "-w"
```
katexochen, line 34:
```suggestion
```

tools/tdx-measure/main.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	firmware, err := os.ReadFile(firmwarePath)
+	if err != nil {
+		return fmt.Errorf("can't read firmware file: %w", err)
+	}
+
+	digest, err := tdvf.CalculateMrTd(firmware)
+	if err != nil {
+		panic(err)
```
katexochen, line 76: don't panic, return the error

tools/tdx-measure/main.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := cmd.MarkFlagFilename("firmware", "fd"); err != nil {
+		panic(err)
+	}
+	return cmd
+}
+
+func runGenerate(cmd *cobra.Command, args []string) error {
+	firmwarePath, err := cmd.Flags().GetString("firmware")
+	if err != nil {
+		return err
```
katexochen, line 66: wrap error with some info

tools/tdx-measure/mrtd/mrtd.go
katexochen, line 1: Please add some comments, references to the spec etc.

tools/tdx-measure/tdvf/tdvf.go
katexochen, line 1: Please add some comments, references to the spec etc.

tools/tdx-measure/tdvf/tdvf.go
```diff
@@ -0,0 +1,138 @@
...
+
+	metadata := firmare[len(firmare)-int(offset):]
+	metadataHeader := metadata[:16]
+
+	if string(metadataHeader[:4]) != metadataSignature {
+		return nil, errors.New("unexpected signature")
+	}
+
+	length := binary.LittleEndian.Uint32(metadataHeader[4:][:4])
+	_ = length
```
katexochen, line 89: Comment why this is needed?

tools/tdx-measure/tdvf/tdvf.go
```diff
@@ -0,0 +1,138 @@
...
+		section.MemoryAddress = binary.LittleEndian.Uint64(sectionData[8:][:8])
+		section.MemoryDataSize = binary.LittleEndian.Uint64(sectionData[16:][:8])
+		section.Type = tdvfSectionType(binary.LittleEndian.Uint32(sectionData[24:][:4]))
+		section.Attributes = tdvfSectionAttributes(binary.LittleEndian.Uint32(sectionData[28:][:4]))
+	}
+
+	return sections, nil
+}
+
+func CalculateMrTd(firmware []byte) ([48]byte, error) {
```
katexochen, line 113: nit: If would have expected func CalculateMrTd in mrtd.go xD

katexochen (commented):
Please add module to go.work

tools/tdx-measure/tdvf/tdvf.go
Freax13: This function is implemented as laid out in the TDVF spec (I'll add spec links). The specs containing MRTD specifics are guest agnostic and don't contain any suggestions on how to setup a guest. IMHO this function belongs here.

tools/tdx-measure/tdvf/tdvf.go
```diff
@@ -0,0 +1,136 @@
...
+var (
+	tableFooterGUID       = []byte{0xde, 0x82, 0xb5, 0x96, 0xb2, 0x1f, 0xf7, 0x45, 0xba, 0xea, 0xa3, 0x66, 0xc5, 0x5a, 0x08, 0x2d}
+	tdxMetadataOffsetGUID = []byte{0x35, 0x65, 0x7a, 0xe4, 0x4a, 0x98, 0x98, 0x47, 0x86, 0x5e, 0x46, 0x85, 0xa7, 0xbf, 0x8e, 0xc2}
+	metadataSignature     = "TDVF"
+)
+
+// Find the offset for the TDVF metadata as described in Intel® TDX Virtual
+// Firmware Design Guide, 11.1 TDVF Metadata Location.
+func findTdvfMetadataOffset(firmare []byte) (uint32, error) {
+	footerGUID := firmare[len(firmare)-0x30:][:0x10]
```
burgerdev, line 26: Why are some indices in this file base16 and some base10?

tools/tdx-measure/tdvf/tdvf.go
```diff
@@ -0,0 +1,138 @@
...
+	PageAug
+)
+
+func parseTdvfSections(firmare []byte) ([]tdvfSection, error) {
+	offset, err := findTdxMetadataOffset(firmare)
+	if err != nil {
+		return nil, fmt.Errorf("can't locate TDX firmware metadata offset: %w", err)
+	}
+
+	metadata := firmare[len(firmare)-int(offset):]
```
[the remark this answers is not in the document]
Freax13: The reset vector for x86 is at `0xffff_fff0`, but the firmware usually ends at `0x1_0000_0000`, so the vast majority of code is placed before the reset vector. It makes sense for firmware to that as a fixpoint and have all offsets be relative to the end of the firmware.

tools/tdx-measure/tdvf/tdvf.go
Freax13: I don't think there's a good reason. The Intel spec uses base 10 for the TDVF descriptor/sections and a single base 16 number for describing the metadata offset. IIRC there's another SEV spec that goes into a bit more detail about the metadata offset, but I can't find it right now, though I suspect that maybe that spec uses base 16.

I changed all offsets to be in base 10.

tools/tdx-measure/tdvf/tdvf.go
burgerdev: +1

burgerdev (approved):

tools/tdx-measure/tdvf/tdvf.go
Freax13: It's not needed :D
https://github.com/edgelesssys/contrast/blob/af91458e5c6a7ed2a75b8edd2f5ef468ad740ca0/tools/tdx-measure/tdvf/tdvf.go#L83-L85

I just put that there for completeness sake should we ever need it for some reason.

burgerdev: It would help if it said why it's not needed in the docstring, though. Is it because the information is redundant with the number of entries 8 bytes further down?

PR #873, opened 2024-09-14
Simplify Azure cluster setup

blenessy:
It is not necessary to create secondary node pool if you create the AKS with KataCcIsolation enabled.

### DIFF
--- a/docs/docs/getting-started/cluster-setup.md
+++ b/docs/docs/getting-started/cluster-setup.md
@@ -85,10 +85,7 @@ az group create \
 
 ## Create AKS cluster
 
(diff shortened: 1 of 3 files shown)


Freax13 (changes requested):
Thank you for your contribution! This looks good to me.

Can you update [create-coco-aks.sh](https://github.com/edgelesssys/contrast/blob/main/packages/create-coco-aks.sh) script accordingly?

blenessy:
Glad to help @Freax13 ! Updated the script in question and scripts.nix accordingly.

Freax13 (approved):
LGTM, thanks :shipit:!

CI is failing because our workflows apparently don't work well with contributors outside the edgelesssys org. I ran the default target and the e2e openssl target manually; everything worked as expected.

PR #879, opened 2024-09-20
package QEMU with TDX patches

Freax13:
This replaces the old binary we took from Ubuntu. Packaging this ourselves has two advantages:
1. We can build a static binary that works on all distributions.
2. We can add our own patches.


packages/by-name/dtc-static/package.nix
katexochen, line 1: Shouldn't this be fixed upstream now?

Freax13: Right, I forgot about that. We'll have to update our nix lockfile first. I'll do that now.

packages/by-name/qemu-tdx-static/0001-avoid-duplicate-definitions.patch
```diff
@@ -0,0 +1,336 @@
...
+     } else if (!strcmp(digest, "none")) {
+         iscsi_set_header_digest(iscsi, ISCSI_HEADER_DIGEST_NONE);
+-    } else if (!strcmp(digest, "crc32c-none")) {
++    } else if (!strcmp(digest, "__crc32c-none")) {
+         iscsi_set_header_digest(iscsi, ISCSI_HEADER_DIGEST_CRC32C_NONE);
+-    } else if (!strcmp(digest, "none-crc32c")) {
++    } else if (!strcmp(digest, "none-__crc32c")) {
+         iscsi_set_header_digest(iscsi, ISCSI_HEADER_DIGEST_NONE_CRC32C);
+     } else {
+         error_setg(errp, "Invalid header-digest setting : %s", digest);
```
burgerdev, line 50: This is not related to the C function name clash, is it?

packages/by-name/qemu-tdx-static/0001-avoid-duplicate-definitions.patch
```diff
@@ -0,0 +1,336 @@
...
+diff --git a/roms/u-boot/fs/btrfs/ctree.c b/roms/u-boot/fs/btrfs/ctree.c
+index 5ffced91..2a01990c 100644
+--- a/roms/u-boot/fs/btrfs/ctree.c
++++ b/roms/u-boot/fs/btrfs/ctree.c
+@@ -16,7 +16,7 @@ static const struct btrfs_csum {
+ 	u16 size;
+ 	const char name[14];
+ } btrfs_csums[] = {
+-	[BTRFS_CSUM_TYPE_CRC32]		= {  4, "crc32c" },
++	[BTRFS_CSUM_TYPE_CRC32]		= {  4, "__crc32c" },
```
burgerdev, line 143: This also looks like it should not be replaced.

packages/by-name/qemu-tdx-static/package.nix
```diff
@@ -0,0 +1,60 @@
...
+    ];
+
+    postPatch = ''
+      while read patch; do
+        patch="''${patch%%#*}"
+        if [[ $patch == "" ]]; then
+          continue
+        fi
+        patch -p1 < ${tdxPatches}/patches/$patch
+      done < <(cat ${tdxPatches}/patches/series)
```
burgerdev, line 56: Shouldn't these patches be applied before ours?

burgerdev (commented):
Aside from my two comments about semantic differences, there are also a couple of unrelated/unnecessary replacements in the crc32c patch.

packages/by-name/qemu-tdx-static/0001-avoid-duplicate-definitions.patch
Freax13: It is. There's a linker error due to conflicting definitions.

Freax13: Oh, wait, no, of course not, you're right.

packages/by-name/qemu-tdx-static/package.nix
Freax13: That's probably better yes, otherwise we can't apply patches *after* the TDX patches and we'll definitively want to do that.

burgerdev (approved):

PR #887, opened 2024-09-24
docs: add instructions to raise user.max_inotify_instances limit

Freax13:
If this limit isn't raised, kata will fail with `failed to create containerd task: failed to create shim task: Creating watcher returned error too many open files: unknown` when running many containers.

### DIFF
--- a/docs/docs/getting-started/bare-metal.md
+++ b/docs/docs/getting-started/bare-metal.md
@@ -26,6 +26,8 @@ Follow Canonical's instructions on [setting up Intel TDX on Ubuntu 24.04](https:
 </TabItem>
 </Tabs>
 
+Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inotify_instances=8192` to `/etc/sysctl.d/99-sysctl.conf` and running `sysctl --system`.
+
 ## K3s Setup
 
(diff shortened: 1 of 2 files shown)


docs/docs/getting-started/bare-metal.md
```diff
@@ -26,6 +26,8 @@ Follow Canonical's instructions on [setting up Intel TDX on Ubuntu 24.04](https:
 </TabItem>
 </Tabs>
 
+Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inotify_instances=8192` to `/etc/sysctl.d/99-sysctl.conf` and running `sysctl -p`.
```
burgerdev, line 29: According to my manpage, it should be `sysctl --system` to read from files in `/etc/sysctl.d/`, but I rarely use the tool.

burgerdev (approved):

docs/docs/getting-started/bare-metal.md
Freax13: Yup, you're right, `-p` loads from `/etc/sysctl.conf` by default and this *won't* include `/etc/sysctl.d/99-sysctl.conf`. `--system` is better here.

PR #889, opened 2024-09-25
kata.kata-kernel-uvm: disable CONFIG_MODULE_SIG

Freax13:
This config option enables signing of kernel modules. This is a source of non-reproducability. Disable it.


burgerdev (approved):
Nice catch!

msanft (approved):

PR #893, opened 2024-09-27
tdx-measure: assemble Boot0000 by hand

Freax13:
It's always nicer to see how a value is constructed/structured rather than just hard-coding it. Now that we know how that value is derived, we can also be more confident that it won't change in the future (it's mostly comprised of constants).

### DIFF
--- a/tools/tdx-measure/rtmr/rtmr.go
+++ b/tools/tdx-measure/rtmr/rtmr.go
@@ -63,6 +63,19 @@ func (r *Rtmr) extendSeparator() {
 	r.hashAndExtend([]byte{0, 0, 0, 0})
 }
 
+const (
+	mediaProtocolType         = 4
+	piwgFirmwareFileSubType   = 6
+	piwgFirmwareVolumeSubType = 7
+	endOfPathType             = 0x7f
+)
+
+const (
+	loadOptionActive      = 1 << 0
+	loadOptionHidden      = 1 << 3
+	loadOptionCategoryApp = 1 << 8
+)
+
 var (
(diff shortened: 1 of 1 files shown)


tools/tdx-measure/rtmr/rtmr.go
```diff
@@ -78,8 +91,89 @@ var (
...
+	if err := addNode(mediaProtocolType, piwgFirmwareFileSubType, fileGUID[:]); err != nil {
+		return nil, err
+	}
+	if err := addNode(endOfPathType, 0xff, []byte{}); err != nil {
+		return nil, err
+	}
+	return buffer.Bytes(), nil
+}
+
+// buildEfiLoadOption assembles the EFI_LOAD_OPTION passed by QEMU for boot option Boot0000.
```
burgerdev, line 141: How did you figure this out - is it documented somewhere or does one need to read the source to understand what's happening? Can we leave some references in case this changes with future qemu releases?

burgerdev (approved):

tools/tdx-measure/rtmr/rtmr.go
Freax13: I looked at the UEFI and PI specs and dissected the bytes manually. The constants are found in both QEMU and OVMF, so I don't think QEMU changing them is even possible without breaking older OVMFs.

burgerdev: Thanks - do you think we could link the relevant parts?

Freax13: I added some spec links for the data structures.

PR #895, opened 2024-09-27
ci: send notifications to teams channel

miampf:
Add the `post_to_teams` github action that allows a workflow to post into the `Contrast-CI-Notifications` channel. Furthermore, integrate this action into weekly running actions.


.github/actions/post_to_teams/action.yml
```diff
@@ -0,0 +1,26 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Post to teams
+      shell: bash
+      run: |
+        curl \
+          -H "Content-Type: application/json" \
+          -d '{"workflow_name": "${{github.workflow}}", "job_id": "${{github.job}}", "run_id": "${{github.run_id}}", "ref": "${{github.ref_name}}", "message": "${{inputs.message}}"}' \
```
burgerdev, line 25: Can we quote the message string, or at least document what it may and may not contain?

burgerdev (approved):

katexochen:
We should use a card like with Constellation and provide a direct link/button to the logs.

miampf:
> We should use a card like with Constellation and provide a direct link/button to the logs.

Cards are now posted in the corresponding channel.

.github/actions/post_to_teams/action.yml
miampf: Seems like quoted strings just disappear from the message, should I just document this in the description of the workflow run?

miampf: Documented this behavior in the description of the `message` parameter.

.github/actions/post_to_teams/action.yml
```diff
@@ -0,0 +1,29 @@
...
+    required: true
+  message:
+    description: "The message to be sent. Note that you may not use quotes or escaped quotes in your message."
+    required: true
+
+outputs:
+  commit-msg:
+    description: "Commit message"
+    value: ${{ steps.bump-version.outputs.commit-msg }}
+
```
Freax13, line 16:
```suggestion
```

katexochen:
> > We should use a card like with Constellation and provide a direct link/button to the logs.
> 
> Cards are now posted in the corresponding channel.

Button to run seems to be broken?

miampf:
> > > We should use a card like with Constellation and provide a direct link/button to the logs.
> > 
> > 
> > Cards are now posted in the corresponding channel.
> 
> Button to run seems to be broken?

I don't think the buttons are broken, its just that the messages I sent contained quotes which seemingly broke the parsing of the workflow and caused the URL to be replaced with the last word of the quoted text. If you look at the cardwith run_id 11177841920 (which uses the same adaptive card layout) you'll see that the button works there :)

Freax13 (approved):

katexochen (approved):

PR #896, opened 2024-09-27
kata: 3.8.0 -> 3.9.0

katexochen:


dev-docs/patches.md
```diff
@@ -82,7 +82,7 @@ When updating a package, you might need to rebase the current patch set.
 When done, recreate the patch set:
 
 ```sh
-git format-patch -o $pkgDir $rev
+git format-patc -N --no-signature -o $pkgDir $rev
```
Freax13, line 85:
```suggestion
git format-patch -N --no-signature -o $pkgDir $rev
```

packages/by-name/kata/kata-runtime/0003-runtime-agent-verify-the-agent-policy-hash.patch
```diff
@@ -591,27 +590,28 @@ index 6b2b6b02d..bcb6cd2ad 100644
...
- func prepareObjectWithTdxQgs(object Object) string {
+ func prepareTDXObject(object Object) string {
  	qgsSocket := SocketAddress{"vsock", fmt.Sprint(VsockHostCid), fmt.Sprint(object.QgsPort)}
--	tdxObject := TdxQomObject{string(object.Type), object.ID, qgsSocket, nil}
-+	tdxObject := TdxQomObject{
-+		QomType:               string(object.Type),
-+		Id:                    object.ID,
-+		QuoteGenerationSocket: qgsSocket,
-+	}
++	septVEDisable := false
```
Freax13, line 605: AFAICT we didn't set this before, did we? Is there a reason for setting it to `false`? IIRC `true` is more secure.

katexochen: The field wasn't initialized previously, and the default value for a bool in Go is false.

Freax13: It was (and still is) a pointer though, so the default should have been `nil` or am I missing something?

burgerdev: I guess the question is rather what the qemu default is? 

Regarding the desired value, I understand that Linux wants to disable it: https://lkml.org/lkml/2024/8/28/510.

burgerdev (approved):
lgtm % `SEPT_VE_DISABLE` discussion

Freax13 (approved):

PR #901, opened 2024-09-30
Align policy hash verification between SNP and TDX 

burgerdev:
*BREAKING*: The Prometheus counter `contrast_meshapi_attestation_failures_total` is not incremented anymore if a workload has matching reference values but an untrusted policy hash. These rejections are now visible in the `contrast_grpc_server_handled_total` counter with field `grpc_code=PermissionDenied`. 

---

The policy hash is passed as `HOSTDATA` on SNP and as `MRCONFIGID` on TDX. In the SNP validation, we used to check the `HOSTDATA` too, but we did not introduce this for TDX. As a consequence, the metrics observed on SNP and TDX are different: on SNP, an unknown hash ends up in `attestation_failures_total`, while on TDX it ends up in `grpc_server_handled_total` with an internal error status. Thus, we could not have a simple portable policy test that relied on the metrics.

This PR unifies handling of the policy hash between SNP and TDX by removing the additional validation from SNP. Existence of the hash in the manifest is now only checked in the `NewMeshCert` handler, where failures increment the `grpc_server_handled_total` with a `PermissionDenied` status. Now we can run the policy e2e test on all platforms.

Drive-by fix: the CLI now verifies coordinators on TDX correctly.

### DIFF
--- a/coordinator/meshapi.go
+++ b/coordinator/meshapi.go
@@ -19,8 +19,10 @@ import (
 	grpcprometheus "github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus"
 	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
 )
 
 type meshAPIServer struct {
@@ -108,7 +110,7 @@ func (i *meshAPIServer) NewMeshCert(ctx context.Context, _ *meshapi.NewMeshCertR
 	hostData := manifest.NewHexString(report.HostData())
 	entry, ok := state.Manifest.Policies[hostData]
 	if !ok {
-		return nil, fmt.Errorf("report data %s not found in manifest", hostData)
+		return nil, status.Errorf(codes.PermissionDenied, "policy hash %s not found in manifest", hostData)
 	}
 	dnsNames := entry.SANs
 
--- a/e2e/policy/policy_test.go
+++ b/e2e/policy/policy_test.go
@@ -173,15 +173,20 @@ func getFailures(ctx context.Context, t *testing.T, ct *contrasttest.ContrastTes
 	// parse the logs
 	metrics, err := (&expfmt.TextParser{}).TextToMetricFamilies(strings.NewReader(metricsString))
 	require.NoError(err)
-	failures := -1
-	for k, v := range metrics {
-		if k == "contrast_meshapi_attestation_failures_total" {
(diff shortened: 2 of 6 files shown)


coordinator/meshapi.go
```diff
@@ -108,7 +110,7 @@ func (i *meshAPIServer) NewMeshCert(ctx context.Context, _ *meshapi.NewMeshCertR
 	hostData := manifest.NewHexString(report.HostData())
 	entry, ok := state.Manifest.Policies[hostData]
 	if !ok {
-		return nil, fmt.Errorf("report data %s not found in manifest", hostData)
+		return nil, status.Errorf(codes.PermissionDenied, "policy hash %s not found in manifest", hostData)
```
Freax13, line 113: I'm not too happy about `PermissionDenied`. To me, `PermissionDenied` primarily indicates an authentication problem and not a problem related to unexpected inputs. To quote the official docs:

> It must not be used if the caller cannot be identified (use Unauthenticated instead for those errors).

Personally I'd prefer either `InvalidArgument` or `NotFound`.

e2e/policy/policy_test.go
```diff
@@ -175,13 +175,25 @@ func getFailures(ctx context.Context, t *testing.T, ct *contrasttest.ContrastTes
 	require.NoError(err)
 	failures := -1
 	for k, v := range metrics {
-		if k == "contrast_meshapi_attestation_failures_total" {
-			failures = int(v.GetMetric()[0].GetCounter().GetValue())
+		if k != "contrast_grpc_server_handled_total" {
+			continue
+		}
```
Freax13, line 180: Unrelated to your PR, but why aren't do doing a map lookup here instead of looping?

coordinator/meshapi.go
burgerdev: The part you quoted does not apply, though: we don't have an _authentication_ error here, because we know the caller's policy hash, which is the identity for Contrast.

`InvalidArgument` is not applicable because it "indicates arguments that are problematic regardless of the state of the system". The suggested alternative `FailedPrecondition` would be an option, but imho `PermissionDenied` is more precise: "you don't have permission to get a certificate for this Contrast generation". How about I change the error string to make this more clear?

e2e/policy/policy_test.go
burgerdev: Good point, I'll fix that.

coordinator/meshapi.go
burgerdev: Sorry, I meant to also say in the first paragraph that `PermissionDenied` is an _authorization_ error and `Unauthenticated` is the *authentication* error. Maybe it makes more sense that way?

Freax13: Sure, let's keep using `PermissionDenied`.

Freax13 (approved):

PR #903, opened 2024-10-01
infra: add terraform lock file

davidweisse:


packages/scripts.nix
```diff
@@ -361,4 +361,25 @@
...
+    runtimeInputs = with pkgs; [ terraform ];
+    text = ''
+      set -euo pipefail
+      find infra -type f -name "*.tf" -exec dirname {} \; | sort -u | while read -r dir; do
+        if [[ -f "$dir/.terraform.lock.hcl" ]]; then
+          rm "$dir/.terraform.lock.hcl"
+        fi
+
+        terraform -chdir="$dir" init
+        terraform -chdir="$dir" providers lock -platform=linux_arm64 \
```
katexochen, line 376:
```suggestion
        terraform -chdir="$dir" providers lock \
          -platform=linux_arm64 \
```

packages/scripts.nix
```diff
@@ -361,4 +361,25 @@
       wait
     '';
   };
+
+  update-terraform-lock-files = writeShellApplication {
+    name = "update-terraform-lock-files";
+    runtimeInputs = with pkgs; [ terraform ];
+    text = ''
+      set -euo pipefail
+      find infra -type f -name "*.tf" -exec dirname {} \; | sort -u | while read -r dir; do
```
katexochen, line 370: Compared to the script in constellation we are not filtering out submodules of the same module, but should be ok for now.

katexochen (approved):
@davidweisse can you add this as additional job in workflows/static.yml in a follow up PR, so the lockfile automatically updates and commits the lockfile changes on renovate PR? Checkout the go-source job as example.

PR #907, opened 2024-10-02
e2e: run regression tests on bare metal

burgerdev:
Ensure that known test cases work, regardless of platform, and fix a couple of issues for the tests to [succeeded](https://github.com/edgelesssys/contrast/actions/runs/11141168124):

### MongoDB

* The size limit of the database volume was too small and led to pod eviction.
* The image is 1.2GiB, too large for the default memory limits when using guest pull.

### MySQL

The MySQL images need additional configuration to run, otherwise they end up crash-looping and cause sporadic errors.

### Varnish

The Varnish we're using is a [source-to-image image](https://github.com/sclorg/varnish-container/blob/caed5bc11fa4e1fd38cf4ed4185b42fd6493332c/7/s2i/bin/usage#L14-L22) which does not actually run a server in the default configuration. Since I don't know how we would want to configure Varnish in the first place, I'm removing the image. cc @miampf 

### Image build

The `CONFIDENTIAL_GUEST` environment variable for the rootfs build is set upstream [when COCO guest components are provided](https://github.com/kata-containers/kata-containers/blob/cdaaf708a18da8e5f7e2b9824fa3e43b524893a5/tools/osbuilder/rootfs-builder/rootfs.sh#L468). Since we don't provide them, we need to set it manually so that we get the features  enabled by it. 

For now, the only feature guarded by this variable is an increased `/run` mount size. However, there seems to be a bug in the implementation, because the [systemd target that's edited](https://github.com/kata-containers/kata-containers/blob/cdaaf708a18da8e5f7e2b9824fa3e43b524893a5/tools/osbuilder/rootfs-builder/rootfs.sh#L723) is not yet in the new root, so `systemd-remount-fs` is not enabled and thus modifying `fstab` has no effect. We work around this by adding the targets before the rootfs build and disabling a cleanup step in the build script.


.github/workflows/e2e_regression.yml
```diff
@@ -24,20 +24,41 @@ env:
...
+            runner: TDX
+            self-hosted: true
         case:
           - getdents
           - genpolicy
           - regression
+        exclude:
+          # getdents is a regression test for tardev-snapshotter
+          - platform:
+              self-hosted: true
```
Freax13, line 49: I'm surprised that this works, I always assumed it wouldn't.

packages/by-name/kata/kata-runtime/0014-tools-don-t-clean-build-root-when-generating-rootfs.patch
```diff
@@ -0,0 +1,39 @@
...
+@@ -719,9 +713,12 @@ EOF
+ 			[ -e ${fstab_file} ] && sed -i '/\/run/d' ${fstab_file}
+ 			echo "tmpfs /run tmpfs nodev,nosuid,size=50% 0 0" >> ${fstab_file}
+ 
++            info "Enabling systemd-remount-fs.service"
+ 			kata_systemd_target="${ROOTFS_DIR}/usr/lib/systemd/system/kata-containers.target"
++            ls -l ${kata_systemd_target}
+ 			grep -qE "^Requires=.*systemd-remount-fs.service.*" ${kata_systemd_target} || \
+ 				echo "Requires=systemd-remount-fs.service" >> ${kata_systemd_target}
++            info "Done enabling systemd-remount-fs.service"
```
Freax13, line 36: Were these debug statements checked in intentionally?

Freax13 (approved):

packages/by-name/kata/kata-runtime/0014-tools-don-t-clean-build-root-when-generating-rootfs.patch
burgerdev: Nope, didn't mean to.

.github/workflows/e2e_regression.yml
burgerdev: You mean the partial matching of objects? I did not know either - could be better documented, I guess.

miampf (approved):
LGTM :)

PR #908, opened 2024-10-02
e2e: test verify on coordinator with different policy

davidweisse:
This tests that the `contrast verify` command does not verify a coordinator with a different policy by changing the expected coordinator policy hash before calling `verify`.

### DIFF
--- a/e2e/policy/policy_test.go
+++ b/e2e/policy/policy_test.go
@@ -137,6 +137,24 @@ func TestPolicy(t *testing.T) {
 		// errors should happen
 		require.Greater(newFailures, initialFailures)
 	})
+
+	t.Run("cli does not verify coordinator with unexpected policy hash", func(t *testing.T) {
+		require := require.New(t)
+
+		// change expected coordinator policy hash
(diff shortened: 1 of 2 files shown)


e2e/policy/policy_test.go
```diff
@@ -145,6 +145,21 @@ func TestPolicy(t *testing.T) {
...
 		require.Greater(newFailures, initialFailures)
 	})
+
+	t.Run("cli does not verify coordinator with unexpected policy hash", func(t *testing.T) {
+		require := require.New(t)
+
+		// change expected coordinator policy hash
+		policyHash, err := os.ReadFile(path.Join(ct.WorkDir, "coordinator-policy.sha256"))
+		require.NoError(err)
+		require.NoError(os.WriteFile(path.Join(ct.WorkDir, "coordinator-policy.sha256"), []byte("0000000000000000000000000000000000000000000000000000000000000000"), 0o644))
```
burgerdev, line 155: I would prefer a more plausible hash - all-zeroes could ostensibly have some special meaning. Could we maybe just flip a bit in the expected one?

burgerdev (approved):

burgerdev (approved):

PR #910, opened 2024-10-04
ci: rework e2e tests

Freax13:
This achieves the following goals:
1. Only run the openssl tests for PRs.
2. Only run the servicemesh tests when its files have been touched.
3. Run all other tests every day at night (and send notifications on failure).
4. Keep the manual dispatch workflow.

I don't think it's possible to reasonably achieve all of those with a single workflow file (though I'm open to suggestions), so I've turned `e2e.yaml` into a reusable workflow that's invoked by other workflow files with different parameters.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,20 +1,23 @@
 name: e2e test
 
 on:
-  workflow_dispatch:
+  workflow_call:
     inputs:
       skip-undeploy:
         description: "Skip undeploy"
-        required: false
         type: boolean
-        default: false
-  pull_request:
-    paths-ignore:
-      - dev-docs/**
-      - docs/**
-      - rfc/**
-      - tools/asciinema/**
-      - tools/vale/**
+      test-name:
+        description: "Test Name"
+        type: string
+      platform:
+        description: "Platform"
+        type: string
+      runner:
+        description: "Runner"
+        type: string
+      self-hosted:
+        description: "Self Hosted"
+        type: boolean
 
 env:
   container_registry: ghcr.io/edgelesssys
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e_manual.yaml
```diff
@@ -0,0 +1,74 @@
...
+      test-name:
+        description: "Test Name"
+        required: true
+        type: choice
+        options:
+          - openssl
+          - servicemesh
+          - policy
+          - workloadsecret
+          - volumestatefulset
```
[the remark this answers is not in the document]
Freax13: Seems reasonable to me.

We expect those tests to work on bare metal, don't we? Currently, we only run them on AKS.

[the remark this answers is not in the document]
burgerdev: They should all run on bare metal, but some of them are specific to the tardev snapshotter and thus not tested where we use Nydus.

.github/workflows/e2e.yaml
katexochen, line 1: we use .yml extension for all YAML files.

.github/workflows/e2e_regression.yml
```diff
@@ -39,82 +24,27 @@ jobs:
...
-          kubectl delete ns "$(cat workspace/e2e.namespace)" --timeout 5m
+    name: "${{ matrix.platform.name }}"
+    uses: ./.github/workflows/e2e.yaml
+    with:
+      skip-undeploy: false
+      test-name: ${{ matrix.test-name }}
+      platform: ${{ matrix.platform.name }}
+      runner: ${{ matrix.platform.runner }}
+      self-hosted: ${{ matrix.platform.self-hosted }}
+      send-failure-notifications: true
```
burgerdev, line 46: Should only send notifications if running on schedule, right?

burgerdev (approved):

.github/workflows/e2e_regression.yml
Freax13: Yup, good catch. I removed `send-failure-notifications` in favor of just always checking for `github.event_name == 'schedule'`.

burgerdev (approved):

PR #911, opened 2024-10-07
e2e: apply backoff after failed port-forwarding attempt

burgerdev:
Right now the port-forwarding is retried immediately, but we should give the system some time to rectify if there was some sort of connectivity problem.

### DIFF
--- a/e2e/internal/kubeclient/portforward.go
+++ b/e2e/internal/kubeclient/portforward.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 
 	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/transport/spdy"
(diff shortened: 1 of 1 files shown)


e2e/internal/kubeclient/portforward.go
```diff
@@ -20,6 +21,8 @@ import (
 func (k *Kubeclient) WithForwardedPort(ctx context.Context, namespace, podName, remotePort string, f func(addr string) error) error {
 	var funcErr error
 	for i := range 3 {
+		// Apply backoff after first attempt.
+		time.Sleep(time.Duration(i) * time.Second)
```
Freax13, line 25: The `ctx` may expire before the sleep completes. Do we care about this?

burgerdev: Maybe not at the current backoff levels, but you're right that it should be handled - let me fix that.

katexochen (approved):

Freax13 (approved):

PR #913, opened 2024-10-07
use .yml consistently

Freax13:
Inspired by https://github.com/edgelesssys/contrast/pull/910#discussion_r1789710430


katexochen (commented):
Otherwise lgtm, we should squash those commits.

Freax13:
@katexochen I need an approval before I can merge.

The two failing CI jobs are the regression tests with the know nydus-snapshotter issues.

PR #925, opened 2024-10-10
ci: use card template for teams post

miampf:
Rewrite the `post_to_teams` action to use a JSON template for posting to teams.


Freax13 (approved):
@miampf can you intentionally cause a test failure of one of the e2e so that we can see the generated Teams message?

miampf:
> I want to see the message for an e2e failure specifically.

Made the e2e tests fail, look in the channel for a few messages

PR #927, opened 2024-10-11
qemu-static: add patches for VFIO support on SNP

Freax13:
Upstream QEMU can't launch SNP VMs with VFIO devices. Apply the patches from this patch series: https://patchwork.kernel.org/project/kvm/cover/20240725072118.358923-1-chenyi.qiang@intel.com/. We also need another patch to support large devices (such as GPUs).

Note that I didn't yet test GPU support with the binary built by qemu-static, but I did test the patches on a separate non-nix QEMU build. We might need additional changes to make GPU support work for Contrast.


katexochen:
> Note that I didn't yet test GPU support with the binary built by qemu-static, but I did test the patches on a separate non-nix QEMU build. We might need additional changes to make GPU support work for Contrast.

So this PR is still work in progress?

Freax13:
My idea was to merge this now and test it later, but I'm also fine with delaying this until we can actually test it. I just wanted to make sure we don't lose this work while waiting for other stuff.

msanft (approved):
Can't say too much regarding the individual patches, but I can vouch for the change working.

PR #931, opened 2024-10-15
automate peerpods image upload & use that in Terraform configs

Freax13:
This PR adds a script/just target for automating the image upload of peerpod VM images and adjusts our Terraform configs to use the image id of the uploaded image.

### DIFF
--- a/infra/azure-peerpods/vars.tf
+++ b/infra/azure-peerpods/vars.tf
@@ -9,3 +9,7 @@ variable "image_resource_group_name" {
 variable "subscription_id" {
   type = string
 }
+
+variable "image_id" {
+  type = string
+}
--- a/justfile
+++ b/justfile
@@ -192,6 +192,11 @@ create platform=default_platform:
         "AKS-PEER-SNP")
             just upload-image
 
+            # Populate Terraform variables.
+            echo "name_prefix = \"$azure_resource_group\"" > infra/azure-peerpods/just.auto.tfvars
+            echo "image_resource_group_name = \"$azure_resource_group\"" >> infra/azure-peerpods/just.auto.tfvars
+            echo "subscription_id = \"$azure_subscription_id\"" >> infra/azure-peerpods/just.auto.tfvars
+
             nix run -L .#terraform -- -chdir=infra/azure-peerpods init
             nix run -L .#terraform -- -chdir=infra/azure-peerpods apply
         ;;
--- a/justfile
+++ b/justfile
@@ -184,6 +184,9 @@ create platform=default_platform:
             :
         ;;
         "AKS-PEER-SNP")
+            # Ensure that the resource group exists.
+            az group create  --name "${azure_resource_group}_caa_cluster" --location "$azure_location"
+
             nix run -L .#terraform -- -chdir=infra/azure-peerpods init
             nix run -L .#terraform -- -chdir=infra/azure-peerpods apply
         ;;
@@ -306,6 +309,8 @@ destroy platform=default_platform:
         ;;
         "AKS-PEER-SNP")
             nix run -L .#terraform -- -chdir=infra/azure-peerpods destroy
+
+            az group delete --name "${azure_resource_group}_caa_cluster" --yes
         ;;
         *)
             echo "Unsupported platform: {{ platform }}"
@@ -355,6 +360,8 @@ azure_resource_group=""
 
 # Azure location for the resource group and AKS cluster.
 azure_location="westeurope"
+# Azure subscription id.
+azure_subscription_id="0d202bbb-4fa7-4af8-8125-58c269a05435"
 # Namespace suffix, can be empty. Will be used when patching namespaces.
 namespace_suffix=""
 # Cache directory for the CLI.
--- a/justfile
+++ b/justfile
@@ -172,6 +172,9 @@ undeploy:
         kubectl delete namespace $ns
     fi
 
+upload-image:
+    nix run -L .#scripts.upload-image -- --subscription-id="$azure_subscription_id" --location="$azure_location" --resource-group="${azure_resource_group}_caa_cluster"
+
 # Create a CoCo-enabled AKS cluster.
 create platform=default_platform:
     #!/usr/bin/env bash
@@ -309,6 +312,9 @@ destroy platform=default_platform:
         ;;
         "AKS-PEER-SNP")
             nix run -L .#terraform -- -chdir=infra/azure-peerpods destroy
+
+            # Clean-up cached image ids.
+            rm -f ${CONTRAST_CACHE_DIR}/image-upload/*.image-id
 
             az group delete --name "${azure_resource_group}_caa_cluster" --yes
         ;;
--- a/justfile
+++ b/justfile
@@ -173,6 +173,9 @@ undeploy:
     fi
 
 upload-image:
+    # Ensure that the resource group exists.
+    az group create  --name "${azure_resource_group}_caa_cluster" --location "$azure_location"
+
     nix run -L .#scripts.upload-image -- --subscription-id="$azure_subscription_id" --location="$azure_location" --resource-group="${azure_resource_group}_caa_cluster"
 
 # Create a CoCo-enabled AKS cluster.
@@ -187,8 +190,7 @@ create platform=default_platform:
             :
         ;;
         "AKS-PEER-SNP")
-            # Ensure that the resource group exists.
-            az group create  --name "${azure_resource_group}_caa_cluster" --location "$azure_location"
+            just upload-image
 
             nix run -L .#terraform -- -chdir=infra/azure-peerpods init
             nix run -L .#terraform -- -chdir=infra/azure-peerpods apply
--- a/packages/scripts.nix
+++ b/packages/scripts.nix
@@ -16,6 +16,82 @@
     text = builtins.readFile ./destroy-coco-aks.sh;
   };
 
+  upload-image = writeShellApplication {
+    name = "upload-image";
+    runtimeInputs = with pkgs; [
+      azure-cli
+      gnused
+      uplosi
+    ];
+    text =
+      let
+        image = pkgs.image-podvm;
+      in
+      ''
+        subscriptionId=""
+        location="GermanyWestCentral"
+        resourceGroup=""
+
+        for i in "$@"; do
+          case $i in
+          --subscription-id=*)
+            subscriptionId="''${i#*=}"
+            shift
+            ;;
+          --location=*)
+            location="''${i#*=}"
+            shift
+            ;;
+          --resource-group=*)
+            resourceGroup="''${i#*=}"
+            shift
+            ;;
+          *)
+            echo "Unknown option $i"
+            exit 1
+            ;;
+          esac
+        done
+
+        set -x
+
+        # Create a unique, semver compatible version.
+        imageVersion=0.0.$(date "+%s")
+
+        # Create an uplosi config.
+        cat <<EOF > uplosi.conf
+        [base]
+        imageVersion = "''${imageVersion}"
+        name = "contrast"
+        provider = "azure"
+
+        [base.azure]
+        subscriptionID = "''${subscriptionId}"
+        location = "''${location}"
+        resourceGroup = "''${resourceGroup}"
+        sharedImageGallery = "''${resourceGroup}_contrast"
+        sharingProfile = "private"
+        EOF
+
+        imageCacheDir="''${CONTRAST_CACHE_DIR}"/image-upload
+        mkdir -p "''${imageCacheDir}"
+
+        cacheFile="''${imageCacheDir}"/${builtins.baseNameOf image}.image-id
+        # Check if th image has been cached.
+        if [[ ! -f "$cacheFile" ]]; then
+          # Upload the image.
+          image_id=$(uplosi upload ${image}/*.raw)
+          echo "$image_id" > "$cacheFile"
+        else
+          # Use the image id in the cache.
+          image_id=$(cat "$cacheFile")
+        fi
+
+        # Store the image id in a terraform variable.
+        echo "image_id = \"''${image_id}\"" > infra/azure-peerpods/image_id.auto.tfvars
+      '';
+  };
+
   generate = writeShellApplication {
     name = "generate";
     runtimeInputs = with pkgs; [
--- a/infra/azure-peerpods/main.tf
+++ b/infra/azure-peerpods/main.tf
@@ -40,9 +40,8 @@ data "azurerm_resource_group" "rg_podvm_image" {
   name = var.image_resource_group_name
 }
 
-resource "azurerm_resource_group" "rg" {
-  name     = local.name
-  location = "germanywestcentral"
+data "azurerm_resource_group" "rg" {
+  name = local.name
 }
 
 resource "azuread_application" "app" {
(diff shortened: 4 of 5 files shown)


packages/uplosi.conf.template
katexochen, line 1: I don't think packages is the right place for a template. Maybe we can just inline it into the script?

packages/uplosi.conf.template
```diff
@@ -0,0 +1,11 @@
+[base]
+imageVersion = "0.0.1"
```
katexochen, line 2: We need a mechanism to update the version on each upload. uplosi has separate image file and -i flag for that.

packages/uplosi.conf.template
```diff
@@ -0,0 +1,11 @@
...
+name = "contrast"
+provider = "azure"
+
+[base.azure]
+subscriptionID = "@@SUBSCRIPTION_ID@@"
+location = "@@LOCATION@@"
+resourceGroup = "@@RESOURCE_GROUP@@"
+sharedImageGallery = "contrast"
+sharingProfile = "private"
```
katexochen, line 11: Did you check this works with private images? Otherwise public is acceptable, too.

flake.nix
katexochen, line 1: flake in uplosi is poorly maintained, just add an override to /overlays with the new version.

justfile
```diff
@@ -189,6 +189,8 @@ create platform=default_platform:
             :
         ;;
         "AKS-PEER-SNP")
+            just upload-image
```
katexochen, line 192: This should be decoupled from create so we don't need to recreate the cluster to iterate on the image.

justfile
```diff
@@ -191,6 +191,11 @@ create platform=default_platform:
         "AKS-PEER-SNP")
             just upload-image
 
+            # Populate Terraform variables.
+            echo "name_prefix = \"$azure_resource_group\"" > infra/azure-peerpods/just.auto.tfvars
```
katexochen, line 195: Maybe we can add a suffix here so this won't collide with the rg used for AKS?

justfile
```diff
@@ -191,6 +191,11 @@ create platform=default_platform:
...
+            echo "image_resource_group_name = \"$azure_resource_group\"" >> infra/azure-peerpods/just.auto.tfvars
```
katexochen, line 196: Not sure tbh, will uplosi create the rg the image should be uploaded to? Otherwise we need to create that for the dev at some point.

katexochen (changes requested):

packages/uplosi.conf.template
Freax13: Yes, I did, I created a cluster with this, and worked. I ran into some error when I tried to create a public gallery, but didn't look much into it.

infra/azure-peerpods/vars.tf
```diff
@@ -9,3 +9,7 @@ variable "image_resource_group_name" {
 variable "subscription_id" {
   type = string
 }
+
+variable "image_id" {
```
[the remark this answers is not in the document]
Freax13: I don't know much about this, but everything seems to work i.e. AFAICT the AKS cluster has all the right permissions to use the image to create VMs. Is that not (yet) expected?

justfile
Freax13: How about we also add the `just upload-image` and `terraform apply` steps right before the `nix run -L .#scripts.deploy-caa` step in node-installer?

We need this step here because Terraform won't work with the image_id variable missing.
Creating the `just.auto.tfvars` config is idempotent and `terraform apply` is as well. I don't know what `terraform init` is doing.

infra/azure-peerpods/vars.tf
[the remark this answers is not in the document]
katexochen: So you create a rg with a private image and AKS can just access it?

justfile
Freax13: We already add a suffix in our Terraform configs, so that'd be a bit redundant:
https://github.com/edgelesssys/contrast/blob/0662a2ee4b80bf22441b5a828d2e0f3a6b598320/infra/azure-peerpods/main.tf#L36
Do you still want me to add a suffix?

justfile
Freax13: No, right, it won't, I forgot about that. I'll fix that.

packages/uplosi.conf.template
katexochen: I think we need a mechanism to detect whether an image upload is needed or not. I thought we'd to this via version, but there might be other optioins.

justfile
katexochen: Depends on how well we can detect whether we need to upload or not.

infra/azure-peerpods/vars.tf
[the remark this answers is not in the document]
Freax13: Unless I messed up somehow and AKS is not actually using the private image, everything seems to work.

I used the following steps to test this;
1. `just create AKS-PEER-SNP`.
2. `az aks get-credentials ...`
3. `just node-installer AKS-PEER-SNP`
4. `kubectl create namespace openssl` + `kubectl apply -f workspace/deployment/` where `workspace/deployment` contains the openssl deployment (before `generate`) in which I manually replaced the runtimeClass with `kata-remote`.

packages/uplosi.conf.template
Freax13: I suppose we could create a file with a name based on `basename $(readlink result)` (e.g. `ay580z91d48zd0irvzfzjph8k4q07sd4-image-podvm-gpu-1-rc1`) and place it `workspace.cache` after uploading. On subsequent runs, we could check if the file already exists and skip the upload.

Freax13: We could put the image id of the uploaded image in that file.

infra/azure-peerpods/vars.tf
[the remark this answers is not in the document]
katexochen: Sounds good. There is `just get-credentials-peerpod` :wink:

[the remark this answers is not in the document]
Freax13: I'm not sure why, but I just got this:
```
  Warning  FailedCreatePodSandBox  16s  kubelet  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to cre
ate shim task: remote hypervisor call failed: rpc error: code = Unknown desc = creating an instance : Creating instance (<nil>): beginning VM creation or updat
e: PUT https://management.azure.com/subscriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/tomcontrast_caa_cluster/providers/Microsoft.Compute/virtu
alMachines/podvm-openssl-backend-64f988dbd4-b562v-6e53dddd                                                                                                     
--------------------------------------------------------------------------------                                                                               
RESPONSE 403: 403 Forbidden                                                                                                                                    
ERROR CODE: LinkedAuthorizationFailed                                                                                                                          
--------------------------------------------------------------------------------                                                                               
{                                                                                                                                                              
  "error": {                                                                                                                                                   
    "code": "LinkedAuthorizationFailed",                                                                                                                       
    "message": "The client 'b96bfc14-d5f0-4982-baf7-094a9675db47' with object id 'b96bfc14-d5f0-4982-baf7-094a9675db47' has permission to perform action 'Micro
soft.Compute/virtualMachines/write' on scope '/subscriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/tomcontrast_caa_cluster/providers/Microsoft.Co
mpute/virtualMachines/podvm-openssl-backend-64f988dbd4-b562v-6e53dddd'; however, it does not have permission to perform action(s) 'Microsoft.Compute/galleries/
images/versions/read' on the linked scope(s) '/subscriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/tomcontrast_caa/providers/Microsoft.Compute/ga
lleries/contrast/images/contrast/versions/0.0.1' (respectively) or the linked scope(s) are invalid."                                                           
  }                                                                                                                                                            
}                                                                                                                                                              
--------------------------------------------------------------------------------                                                                               
```

Seems like you were right after all.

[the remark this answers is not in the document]
Freax13: Hmm, not it's working again.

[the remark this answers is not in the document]
burgerdev: I've seen this error, too, when working with peer pods, and giving explicit read access to the image resolves it. However, I also noticed that it does not appear to break things, which is kind of strange.

justfile
```diff
@@ -348,6 +363,8 @@ rctemplate := '''
 container_registry=""
 # Azure resource group/ resource name. Resource group will be created.
 azure_resource_group=""
+# Azure subscription id.
+azure_subscription_id="0d202bbb-4fa7-4af8-8125-58c269a05435"
```
katexochen, line 367: This is in the wrong section, should go further down as it doesn't need to be changed.

justfile
```diff
@@ -306,6 +316,11 @@ destroy platform=default_platform:
         ;;
         "AKS-PEER-SNP")
             nix run -L .#terraform -- -chdir=infra/azure-peerpods destroy
+
+            # Clean-up cached image ids.
+            rm ${CONTRAST_CACHE_DIR}/*.image-id || true
+
+            nix run -L .#scripts.destroy-coco-aks -- --name="$azure_resource_group"
```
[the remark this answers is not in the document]
Freax13: We do that because we want to destroy the resource group named `$azure_resource_group`. Tthis resource group contains the image gallery. If it doesn't exist yet, we create this resource group shortly before uploading the image: https://github.com/edgelesssys/contrast/blob/105ffe01ef0a2191849d6bdaf360b3b141e40f7f/packages/scripts.nix#L60-L63

justfile
```diff
@@ -306,6 +316,11 @@ destroy platform=default_platform:
         ;;
         "AKS-PEER-SNP")
             nix run -L .#terraform -- -chdir=infra/azure-peerpods destroy
+
+            # Clean-up cached image ids.
+            rm -f ${CONTRAST_CACHE_DIR}/*.image-id
```
katexochen, line 321:
```suggestion
            rm -f ${CONTRAST_CACHE_DIR}/image-upload/*.image-id
```
nit: Maybe we could have this in a separate directory, similar to the kds cache.

justfile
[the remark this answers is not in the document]
katexochen: I think we shouldn't share the aks resource group to also upload images there. I want to be able to destroy my aks setup without deleting the images that I use in a totally different peerpod cluster.

Also, I'd say it is better to inline both group creation and deletion directly into the justfile, so it's easier to understand what is happening. These are one-liners anyway.

packages/scripts.nix
```diff
@@ -16,6 +16,83 @@
...
+      gnused
+      uplosi
+    ];
+    text =
+      let
+        image = pkgs.image-podvm;
+      in
+      ''
+        set -euo pipefail
+
```
katexochen, line 32:
```suggestion
```
set by the builder

packages/scripts.nix
```diff
@@ -16,6 +16,83 @@
...
+            shift
+            ;;
+          *)
+            echo "Unknown option $i"
+            exit 1
+            ;;
+          esac
+        done
+
+        set -x
```
katexochen, line 58: Debugging leftover or intended?

katexochen:
I think the caching isn't correct the way it is currently implemented: If I upload `/nix/store/AAA-image-podvm` and then `/nix/store/BBB-image-podvm`, the cache will contain two files, each named after one of the nix path, both containing the same image reference. If I undo the change from BBB back to AAA, and want to redeploy, the upload will check the cache and find a file for the path AAA with the image reference and assumes it is already uploaded, but under that reference is actually the image from the BBB path.

packages/scripts.nix
```diff
@@ -16,6 +16,83 @@
...
+    ];
+    text =
+      let
+        image = pkgs.image-podvm;
+      in
+      ''
+        set -euo pipefail
+
+        subscriptionId=""
+        location="GermanyWestCentral"
```
katexochen, line 34: we should add location as var to the terraform to prevent location mismatch between cluster and image location.

packages/scripts.nix
Freax13: Neither, I copied parts of the script from `create-coco-aks.sh` and that line was already present there (it's also in `destroy-coco-aks.sh`). I'm not invested in having this, we can keep it or remove it, I'd be fine either way.

Freax13:
> I think the caching isn't correct the way it is currently implemented: If I upload `/nix/store/AAA-image-podvm` and then `/nix/store/BBB-image-podvm`, the cache will contain two files, each named after one of the nix path, both containing the same image reference. If I undo the change from BBB back to AAA, and want to redeploy, the upload will check the cache and find a file for the path AAA with the image reference and assumes it is already uploaded, but under that reference is actually the image from the BBB path.

Good catch, I'll integrate the store hash into the image reference to avoid this.

justfile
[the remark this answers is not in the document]
Freax13: > I think we shouldn't share the aks resource group to also upload images there. I want to be able to destroy my aks setup without deleting the images that I use in a totally different peerpod cluster.

Do you want me to move the image gallery into the same resource group that contains the cluster or do you want the image gallery to be in another, new resource group?

katexochen:
> Good catch, I'll integrate the store hash into the image reference to avoid this.

Not sure that's easily possible. :( Azure's image infra is crap. Image versions are restricted to semver, but having a separate gallery for each image requires additional time for creation during upload.

packages/scripts.nix
```diff
@@ -16,6 +16,83 @@
...
+        [base]
+        imageVersion = "0.0.1"
+        name = "contrast"
+        provider = "azure"
+
+        [base.azure]
+        subscriptionID = "''${subscriptionId}"
+        location = "''${location}"
+        resourceGroup = "''${resourceGroup}"
+        sharedImageGallery = "contrast"
```
Freax13, line 76: Apparently Azure doesn't allow multiple galleries with the same name even if they are in different resource groups:

image

Freax13:
> > Good catch, I'll integrate the store hash into the image reference to avoid this.
> 
> Not sure that's easily possible. :( Azure's image infra is crap. Image versions are restricted to semver, but having a separate gallery for each image requires additional time for creation during upload.

Turns out the limits on the version are pretty lax, so we can just use `0.0.<unix timestamp>` :)

justfile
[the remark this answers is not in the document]
katexochen: I'd say a separate group would be nice. If you aren't working on the image part, you might be happy with the same image for long time. On the other hand I'm not sure how to integrate that into the dev workflow, as it would require additional create/destroy steps for that. :thinking:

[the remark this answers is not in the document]
Freax13: That seems reasonable. Given that this PR has already dragged on for quite a while, are you fine with doing this in a follow-up PR?

katexochen:
static checks need fixing

Freax13:
@burgerdev @3u13r can one of you review this pr?

infra/azure-peerpods/vars.tf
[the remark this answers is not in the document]
burgerdev: Why not just revive the `ra_image` above?

packages/scripts.nix
```diff
@@ -16,6 +16,82 @@
...
+        image = pkgs.image-podvm;
+      in
+      ''
+        subscriptionId=""
+        location="GermanyWestCentral"
+        resourceGroup=""
+
+        for i in "$@"; do
+          case $i in
+          --subscription-id=*)
```
burgerdev, line 37: You could do `az account show -o tsv --query id`, which would avoid the hard-coded Edgeless subscription.

burgerdev (approved):

3u13r (approved):

packages/scripts.nix
Freax13: This doesn't work if the user is part of multiple subscriptions though, does it?

burgerdev: It would show the default subscription. Maybe we should default to`az account show` but allow overriding with an explicit flag for the rare use case?

Freax13: We don't really hard-code this anyway, the value can be changed in `justfile.env`. We only hard-code the default value. Interpolating values into the just variable seems a bit painful, so let's just keep hard-coding the Edgeless subscription.

Freax13:
I have two approvals, but Paul's last review requested changes, so I can't merge this. Given that Paul is on vacation until Wednesday, and has said that he's fine with merging this if there are approvals by other reviewers, I'll temporarily disable the merge requirement to merge this PR.

PR #932, opened 2024-10-15
docs: describe secure persistent volumes

burgerdev:


docs/docs/architecture/secrets.md
```diff
@@ -35,3 +36,64 @@ The workload owner can derive the workload secret themselves, since it's derived
...
+      containers:
+      - name: main
+        image: my.registry/my-image@sha256:0123...
+        command:
+        - /bin/sh
+        - -ec
+        - | # <-- Custom script that mounts the encrypted disk and then calls the original application.
+          device=/dev/csi0
+          if ! cryptsetup isLuks $device; then
+            cryptsetup luksFormat $device /contrast/secrets/workload-secret-seed
```
thomasten, line 63: Could this destroy data if `cryptsetup isLuks` fails for another reason?

docs/docs/architecture/secrets.md
```diff
@@ -35,3 +36,64 @@ The workload owner can derive the workload secret themselves, since it's derived
...
+      resources:
+        requests:
+          storage: 1Gi
+      volumeMode: Block # <-- The requested volume needs to be a raw block device.
+```
+
+:::note
+
+This example assumes that the container image can be modified to include a shell and the `cryptsetup` utility.
+Alternatively, a secure mount can be set up from a sidecar container inside an `emptyDir` mount shared with the main container.
```
thomasten, line 94:
```suggestion
This example assumes that you can modify the container image to include a shell and the `cryptsetup` utility.
Alternatively, you can set up a secure mount from a sidecar container inside an `emptyDir` mount shared with the main container.
```
nit: prefer active voice as per MS style guide

docs/docs/architecture/secrets.md
burgerdev: Yes, it would. Let me add a disclaimer to that.

m1ghtym0 (approved):
Thanks, LGTM!
We're mixing (and probably have been before) workflows into the architecture section.
Maybe we should have a dedicated workflow section as in Constellation at one point?
Not urgent and I'm open to suggestions:-)

burgerdev:
I agree that workflows should not go into the architecture section, but I did not mean to write a workflow here. How about creating follow-up tickets that replace this section with a working example, similar to the Emojivoto-servicemesh experience?

m1ghtym0:
Sounds good!

PR #933, opened 2024-10-15
tools: periodically clean up bare-metal servers

davidweisse:
This adds a cronjob to clean up the runtime classes on the bare-metal servers. All unused runtime classes get deleted and the containerd config is updated, and the corresponding folder in `/opt/edgeless` gets deleted as well.

### DIFF
--- a/packages/cleanup-bm.sh
+++ b/packages/cleanup-bm.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+# Copyright 2024 Edgeless Systems GmbH
+# SPDX-License-Identifier: AGPL-3.0-only
+
+set -euo pipefail
+
+echo "Starting cleanup"
+
+resourcesToCheck=(
+  "pods"
+  "deployments"
+  "statefulsets"
+  "daemonsets"
+  "replicasets"
+  "jobs"
+  "cronjobs"
+)
+
+touch usedRuntimeClasses
+for resource in "${resourcesToCheck[@]}"; do
+  kubectl get "${resource}" --all-namespaces -o jsonpath='{.items[*].spec.runtimeClassName} {.items[*].spec.template.spec.runtimeClassName} {.items[*].spec.jobTemplate.spec.template.spec.runtimeClassName}' 2>/dev/null |
+    tr ' ' '\n' >>usedRuntimeClasses
+done
+
+kubectl get pods --all-namespaces -o jsonpath='{.items[?(@.metadata.annotations.contrast\.edgeless\.systems/pod-role=="contrast-node-installer")].spec.containers[0].args[1]}' |
+  tr ' ' '\n' |
+  grep -o "contrast-cc-.\+" >>usedRuntimeClasses
+sort -u usedRuntimeClasses -o usedRuntimeClasses
+
+mapfile -t unusedRuntimeClasses < <(
+  comm -13 usedRuntimeClasses <(
+    kubectl get runtimeclass -o jsonpath='{.items[*].metadata.name}' |
+      tr ' ' '\n' |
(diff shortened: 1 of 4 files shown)


packages/cleanup-bm.sh
```diff
@@ -0,0 +1,59 @@
...
+# SPDX-License-Identifier: AGPL-3.0-only
+
+set -euo pipefail
+
+echo "Starting cleanup"
+
+mapfile -t runtimeClasses < <(kubectl get runtimeclass -o jsonpath='{.items[*].metadata.name}' | tr ' ' '\n' | grep '^contrast-cc')
+
+declare -a usedRuntimeClasses=()
+declare -a unusedRuntimeClasses=()
```
burgerdev, line 12: I find the logic in this script hard to follow. Have you considered just filling a temp file with runtimeclasses in use, and using `comm(1)` to do the set difference?

packages/cleanup-bm.sh
```diff
@@ -0,0 +1,59 @@
...
+
+resourcesToCheck=(
```
burgerdev, line 14: If we checked for an existing node installer, we could tell whether a runtime is just about to be installed.

packages/cleanup-bm.sh
```diff
@@ -0,0 +1,59 @@
...
+  # Delete unused runtime classes
+  echo "Deleting runtimeclass ${runtimeClass} ..."
+  kubectl delete runtimeclass "${runtimeClass}"
+
+  # Delete unused binaries
+  path="${OPTEDGELESS}/${runtimeClass}"
+  if [ -d "${path}" ]; then
+    echo "Deleting binaries from ${OPTEDGELESS}/${runtimeClass} ..."
+    rm -rf "${path}"
+  fi
```
Freax13, line 53: We also store data in `/var/lib/nydus-snapshotter/`.

packages/containers.nix
```diff
@@ -155,6 +155,19 @@ let
...
       };
     };
+
+    cleanup-bm = dockerTools.buildImage {
+      name = "cleanup-bm";
+      tag = "v0.0.1";
+      copyToRoot = with pkgs; [
+        busybox
+        kubectl
+        dasel
```
Freax13, line 165: Do we need this here or is it enough to specify this in `writeShellApplication`?

packages/cleanup-bm.sh
davidweisse: Is there any way of associating a given node installer with a runtime handler, so we can mark this handler as used?

burgerdev: Given a `RuntimeClass`, you'd need to check the args of the node-installer containers. We could consider making this relationship easier to discover, e.g. by adding a label.

packages/cleanup-bm.sh
```diff
@@ -0,0 +1,58 @@
...
+
+for runtimeClass in "${unusedRuntimeClasses[@]}"; do
+  # Delete unused runtime classes
+  echo "Deleting runtimeclass ${runtimeClass} ..."
+  kubectl delete runtimeclass "${runtimeClass}"
+
+  # Delete unused files
+  if [ -d "${OPTEDGELESS}/${runtimeClass}" ]; then
+    echo "Deleting files from ${OPTEDGELESS}/${runtimeClass} ..."
+    rm -rf "${OPTEDGELESS:?}/${runtimeClass}"
```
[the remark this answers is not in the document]
davidweisse: `nix fmt` seemed to complain about `OPTEDGELESS` possibly expanding to `/`, but in the other cases that would not be a problem.

Freax13 (approved):

burgerdev (approved):

PR #934, opened 2024-10-16
e2e/policy: ensure that modified policy hash is valid hex

Freax13:
The policy hash is hex-encoded and unfortunately, XORing 1 into hex digits doesn't always yield valid hex digits:
```
>>> [chr(ord(hex(i)[2:])^1) for i in range(16)]
['1', '0', '3', '2', '5', '4', '7', '6', '9', '8', '`', 'c', 'b', 'e', 'd', 'g']
```
Instead, decode the hex-encoded hash and XOR 1 into that.

### DIFF
--- a/e2e/policy/policy_test.go
+++ b/e2e/policy/policy_test.go
@@ -7,6 +7,7 @@ package policy
 
 import (
 	"context"
+	"encoding/hex"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


burgerdev (approved):
Thanks, good catch!

PR #935, opened 2024-10-16
erofs: pin to 1.7.1

katexochen:
There is a reproducibility issue in later versions, likely https://git.kernel.org/pub/scm/linux/kernel/git/xiang/erofs-utils.git/commit/?id=0da388cfdc9dcb952c01b0755ab8a1d6d59a5312

### DIFF
--- a/overlays/nixpkgs.nix
+++ b/overlays/nixpkgs.nix
@@ -25,4 +25,16 @@ final: prev: {
       hash = "sha256-MsZ4Bl8sW1dZUB9cYPsaLtc8P8RRx4hafSbNB4vXqi4=";
     };
   });
+
+  # There is a reproducibility issue in later versions,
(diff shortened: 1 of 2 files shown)


overlays/nixpkgs.nix
```diff
@@ -15,4 +15,14 @@ final: prev: {
   azure-cli = prev.azure-cli.override {
     withExtensions = with final.azure-cli.extensions; [ aks-preview ];
   };
+
+  erofs-utils = prev.erofs-utils.overrideAttrs (
```
msanft, line 19: We might want to add a comment here with the PR description. This way it'd be more visible than in the PR / commit. But no strong opinion

msanft (approved):

katexochen:
Superseded by https://github.com/edgelesssys/contrast/pull/964

PR #939, opened 2024-10-17
e2e: add a test for checking AKS upstream

miampf:
Add a test that checks if something in azure upstream is broken by deploying a simple container with the `kata-cc-isolation` runtime class. You can view a successful workflow run [here](https://github.com/edgelesssys/contrast/actions/runs/11386327564/job/31678275073?pr=939).

### DIFF
--- a/.github/workflows/e2e_aks_runtime.yml
+++ b/.github/workflows/e2e_aks_runtime.yml
@@ -9,6 +9,7 @@ on:
       - e2e/aks-runtime/**
 
 jobs:
+  # steps taken from https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt#option-2-step-by-step-installation-instructions
   install-software:
     runs-on: ubuntu-22.04
     steps:
--- a/.github/workflows/e2e_aks_runtime.yml
+++ b/.github/workflows/e2e_aks_runtime.yml
@@ -1,0 +1,25 @@
+name: e2e test aks runtime
+
+on:
+  schedule:
+    - cron: "16 6 * * 6" # 6:16 on Saturdays
+  pull_request:
+    paths:
+      - e2e/aks-runtime/**
+
+jobs:
+  test_matrix:
+    strategy:
+      fail-fast: false
+    name: Test aks runtime
+    uses: ./.github/workflows/e2e.yml
+    with:
+      skip-undeploy: false
+      test-name: aks-runtime
+      platform: AKS-CLH-SNP
+      runner: ubuntu-22.04
+      self-hosted: false
+    secrets: inherit
+    permissions:
+      contents: read
+      packages: write
--- a/.github/workflows/e2e_aks_runtime.yml
+++ b/.github/workflows/e2e_aks_runtime.yml
@@ -43,7 +43,7 @@ jobs:
           echo "SYNC_FIFO_UUID=$sync_uuid" | tee -a "$GITHUB_ENV"
       - name: Build and prepare deployments
         run: |
-          nix run .#just -- coordinator initializer port-forwarder openssl cryptsetup service-mesh-proxy node-installer AKS-CLH-SNP
+          nix shell .#just --command just coordinator initializer port-forwarder openssl cryptsetup service-mesh-proxy node-installer AKS-CLH-SNP
       # steps taken from https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt#option-2-step-by-step-installation-instructions
       - name: Install `az` with extensions
         run: |
--- a/.github/workflows/e2e_aks_runtime.yml
+++ b/.github/workflows/e2e_aks_runtime.yml
@@ -9,15 +9,37 @@ on:
 
 jobs:
   test_matrix:
-    name: Test aks runtime
-    uses: ./.github/workflows/e2e.yml
-    with:
-      skip-undeploy: false
-      test-name: aks-runtime
-      platform: AKS-CLH-SNP
-      runner: ubuntu-22.04
-      self-hosted: false
-    secrets: inherit
-    permissions:
-      contents: read
-      packages: write
+    steps:
+      - name: Install `az` with extensions
+        run: |
+          sudo apt-get update
+          sudo apt-get install apt-transport-https ca-certificates curl gnupg lsb-release
+          sudo mkdir -p /etc/apt/keyrings
+          curl -sLS https://packages.microsoft.com/keys/microsoft.asc |
+            gpg --dearmor | sudo tee /etc/apt/keyrings/microsoft.gpg > /dev/null
+          sudo chmod go+r /etc/apt/keyrings/microsoft.gpg
+          AZ_DIST=$(lsb_release -cs)
+          echo "Types: deb
+          URIs: https://packages.microsoft.com/repos/azure-cli/
+          Suites: ${AZ_DIST}
+          Components: main
+          Architectures: $(dpkg --print-architecture)
+          Signed-by: /etc/apt/keyrings/microsoft.gpg" | sudo tee /etc/apt/sources.list.d/azure-cli.sources
+          sudo apt-get update
+          sudo apt-get install azure-cli
+
+          az extension add --name aks-preview
+          az extension add --name confcom
+      - name: Test aks runtime
+        uses: ./.github/workflows/e2e.yml
+        with:
+          skip-undeploy: false
+          test-name: aks-runtime
+          platform: AKS-CLH-SNP
+          runner: ubuntu-22.04
+          self-hosted: false
+          run-without-nix: true
+        secrets: inherit
+        permissions:
+          contents: read
+          packages: write
--- a/.github/workflows/e2e_aks_runtime.yml
+++ b/.github/workflows/e2e_aks_runtime.yml
@@ -1,6 +1,7 @@
 name: e2e test aks runtime
 
 on:
+  workflow_dispatch:
   schedule:
     - cron: "16 6 * * 6" # 6:16 on Saturdays
   pull_request:
--- a/.github/workflows/e2e_aks_runtime.yml
+++ b/.github/workflows/e2e_aks_runtime.yml
@@ -9,7 +9,8 @@ on:
       - e2e/aks-runtime/**
 
 jobs:
-  test_matrix:
+  install-software:
+    runs-on: ubuntu-22.04
     steps:
       - name: Install `az` with extensions
         run: |
@@ -17,7 +18,7 @@ jobs:
           sudo apt-get install apt-transport-https ca-certificates curl gnupg lsb-release
           sudo mkdir -p /etc/apt/keyrings
           curl -sLS https://packages.microsoft.com/keys/microsoft.asc |
-            gpg --dearmor | sudo tee /etc/apt/keyrings/microsoft.gpg > /dev/null
+          gpg --dearmor | sudo tee /etc/apt/keyrings/microsoft.gpg > /dev/null
           sudo chmod go+r /etc/apt/keyrings/microsoft.gpg
           AZ_DIST=$(lsb_release -cs)
           echo "Types: deb
@@ -31,16 +32,19 @@ jobs:
 
           az extension add --name aks-preview
           az extension add --name confcom
-      - name: Test aks runtime
-        uses: ./.github/workflows/e2e.yml
-        with:
-          skip-undeploy: false
-          test-name: aks-runtime
-          platform: AKS-CLH-SNP
-          runner: ubuntu-22.04
-          self-hosted: false
-          run-without-nix: true
-        secrets: inherit
-        permissions:
-          contents: read
-          packages: write
+
+  test_matrix:
+    name: Test aks runtime
+    needs: install-software
+    secrets: inherit
+    permissions:
+      contents: read
+      packages: write
+    uses: ./.github/workflows/e2e.yml
+    with:
+      skip-undeploy: false
+      test-name: aks-runtime
+      platform: AKS-CLH-SNP
+      runner: ubuntu-22.04
+      self-hosted: false
+      run-without-nix: true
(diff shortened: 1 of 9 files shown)


miampf:
Note that the boilerplate of manually adding the `confcom` extension to the nixpkgs overlay should be removed before merging. This is currently not possible as the `azure-cli` package in `nixpkgs-unstable` is broken.

overlays/nixpkgs.nix
```diff
@@ -1,7 +1,61 @@
...
+        "sha256"
+        "description"
+        "meta"
+      ])
+    );
+
+  confcom = mkAzExtension rec {
+    pname = "confcom";
+    version = "1.0.0";
+    url = "https://azcliprod.blob.core.windows.net/cli-extensions/confcom-${version}-py3-none-any.whl";
```
burgerdev, line 41: The idea of this PR is to detect AKS changes that break the confcom/genpolicy versions that we base our software on, and thus are likely to affect us?

.github/workflows/e2e_aks_runtime.yml
```diff
@@ -0,0 +1,25 @@
...
+  schedule:
+    - cron: "16 6 * * 6" # 6:16 on Saturdays
+  pull_request:
+    paths:
+      - e2e/aks-runtime/**
+
+jobs:
+  test_matrix:
+    strategy:
+      fail-fast: false
```
Freax13, line 13:
```suggestion
```

e2e/aks-runtime/aks_runtime_test.go
```diff
@@ -0,0 +1,141 @@
...
+			WithSelector(kuberesource.LabelSelector().WithMatchLabels(
+				map[string]string{"app.kubernetes.io/name": testContainer},
+			)).
+			WithTemplate(kuberesource.PodTemplateSpec().
+				WithLabels(map[string]string{"app.kubernetes.io/name": testContainer}).
+				WithSpec(kuberesource.PodSpec().
+					WithContainers(kuberesource.Container().
+						WithName(testContainer).
+						WithImage("docker.io/bash@sha256:ce062497c248eb1cf4d32927f8c1780cce158d3ed0658c586a5be7308d583cbb").
+						WithCommand("/usr/local/bin/bash", "-c", "uname -r; while true; do sleep 10; done"),
```
Freax13, line 79:
```suggestion
						WithCommand("/usr/local/bin/bash", "-c", "uname -r; sleep infinity"),
```

Can we use `sleep infinity` here?

e2e/aks-runtime/aks_runtime_test.go
```diff
@@ -0,0 +1,141 @@
...
+	toWrite, err := kuberesource.ResourcesToUnstructured(resources)
+	require.NoError(err)
+
+	// generate policies
+	resourceBytes, err := kuberesource.EncodeUnstructured(toWrite)
+	require.NoError(err)
+	require.NoError(os.WriteFile(path.Join(workdir, "resources.yaml"), resourceBytes, 0o644))
+	require.NoError(az.KataPolicyGen(path.Join(workdir, "resources.yaml")))
+
+	// load in generated resources and patch the runtime handler again
```
Freax13, line 100: Is this comment still accurate? Where do we patch the runtime handler?

overlays/nixpkgs.nix
miampf: The idea of this PR is to make a test for azures own stuff to detect if we can upgrade our own components or if microsoft broke something. The addition of the `confcom` package is just so that I can generate a good policy for kata since our own `genpolicy` doesn't work on the `kata-cc-isolation` runtime class.

e2e/aks-runtime/aks_runtime_test.go
miampf: Good catch, we patch the runtime handler on line 87 but not there anymore.

Freax13 (approved):

overlays/nixpkgs.nix
burgerdev: My concern here is that this tests a certain fixed configuration, whereas Microsoft considers supported whatever is currently available on AKS and the latest release of confcom. If we wanted to show that AKS is working like MS intended, we would need to use their instructions.

On the other hand, it might be useful to test whether the unpatched sources we build on are compatible with `kata-cc-isolation`, but then I'd rather take `.#msft.genpolicy`.

miampf: Since this just deploys a container on AKS it uses their upstream config, and the confcom extension will be updated by myself when a new version comes out. Depending on how often we update our `flake.lock` we _may_ fall behind the newest `confcom` version. An option to solve that would be to keep a `confcom_version.txt` and override the package version in nixpkgs upstream with our own version number, which would bypass whole flake updates.

.github/workflows/e2e_aks_runtime.yml
```diff
@@ -0,0 +1,50 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "16 6 * * 6" # 6:16 on Saturdays
+  pull_request:
+    paths:
+      - e2e/aks-runtime/**
+
+jobs:
+  install-software:
```
burgerdev, line 12: Could you please leave a pointer here to where these instructions come from?

.github/workflows/e2e_aks_runtime.yml
```diff
@@ -0,0 +1,50 @@
...
+          curl -sLS https://packages.microsoft.com/keys/microsoft.asc |
+          gpg --dearmor | sudo tee /etc/apt/keyrings/microsoft.gpg > /dev/null
+          sudo chmod go+r /etc/apt/keyrings/microsoft.gpg
+          AZ_DIST=$(lsb_release -cs)
+          echo "Types: deb
+          URIs: https://packages.microsoft.com/repos/azure-cli/
+          Suites: ${AZ_DIST}
+          Components: main
+          Architectures: $(dpkg --print-architecture)
+          Signed-by: /etc/apt/keyrings/microsoft.gpg" | sudo tee /etc/apt/sources.list.d/azure-cli.sources
```
burgerdev, line 29:
```suggestion
          sudo tee /etc/apt/sources.list.d/azure-cli.sources <<EOF
          Types: deb
          URIs: https://packages.microsoft.com/repos/azure-cli/
          Suites: ${AZ_DIST}
          Components: main
          Architectures: $(dpkg --print-architecture)
          Signed-by: /etc/apt/keyrings/microsoft.gpg
          EOF
```

e2e/aks-runtime/aks_runtime_test.go
```diff
@@ -0,0 +1,141 @@
...
+			WithReplicas(1).
+			WithSelector(kuberesource.LabelSelector().WithMatchLabels(
+				map[string]string{"app.kubernetes.io/name": testContainer},
+			)).
+			WithTemplate(kuberesource.PodTemplateSpec().
+				WithLabels(map[string]string{"app.kubernetes.io/name": testContainer}).
+				WithSpec(kuberesource.PodSpec().
+					WithContainers(kuberesource.Container().
+						WithName(testContainer).
+						WithImage("docker.io/bash@sha256:ce062497c248eb1cf4d32927f8c1780cce158d3ed0658c586a5be7308d583cbb").
```
burgerdev, line 78: Let's use `quay.io` or `ghcr.io`.

e2e/aks-runtime/aks_runtime_test.go
```diff
@@ -0,0 +1,141 @@
...
+		if skipUndeploy {
+			return
+		}
+
+		// delete the deployment
+		deletePolicy := metav1.DeletePropagationForeground
+		require.NoError(c.Client.AppsV1().Deployments(namespace).Delete(context.Background(), testContainer, metav1.DeleteOptions{
+			PropagationPolicy: &deletePolicy,
+		}))
+	})
```
burgerdev, line 122: Why not just delete the namespace?

e2e/aks-runtime/aks_runtime_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	pods, err := c.Client.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
+	require.NoError(err)
+	pod := pods.Items[0] // only one pod was deployed
```
burgerdev, line 126: Consider asserting this.

e2e/internal/az/confcom.go
```diff
@@ -0,0 +1,25 @@
...
+	out, err := exec.Command("az", "confcom", "katapolicygen", "--print-version").Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(out)), nil
+}
+
+// KataPolicyGen executes `az confcom katapolicygen --yaml <resourcePath>`.
+func KataPolicyGen(resourcePath string) error {
+	return exec.Command("az", "confcom", "katapolicygen", "--yaml", resourcePath).Run()
```
burgerdev, line 24: Does this potentially swallow interesting command output?

burgerdev (approved):
lgtm, just a few nits

e2e/internal/az/confcom.go
miampf: Not really, at most it swallows stderr which _could_ be interesting if the command fails. I've redirected the commands stderr output to the stderr output of the OS as a fix.

.github/workflows/e2e_aks_runtime.yml
```diff
@@ -0,0 +1,55 @@
...
+
+          sudo apt-get update
+          sudo apt-get install azure-cli
+
+          az extension add --name aks-preview
+          az extension add --name confcom
+
+  test_matrix:
+    name: Test aks runtime
+    needs: install-software
```
Freax13, line 43: If this is meant to run the e2e test with the az cli installed in `install-software`, this doesn't work. `install-software` runs in another job, in another VM, and any changes there don't carry over to this job. If we want the changes to persist, all steps will have to be part of the same job.

packages/by-name/contrast/package.nix
```diff
@@ -38,7 +46,13 @@ let
       "e2e/workloadsecret"
       "e2e/volumestatefulset"
       "e2e/regression"
+      "e2e/aks-runtime"
     ];
+
+    postInstall = ''
+      wrapProgram $out/bin/aks-runtime.test \
+        --prefix PATH : ${lib.makeBinPath [ azure-cli ]}
+    '';
```
Freax13, line 55: Now that we install the Azure CLI separately, we no longer need this, right? Furthermore, adding azure-cli to the PATH will make the test use that instead of the azure CLI that's already in PATH.

miampf:
New successful run [here](https://github.com/edgelesssys/contrast/actions/runs/11572857857/job/32213724936)

packages/by-name/contrast/package.nix
```diff
@@ -26,7 +27,13 @@ let
...
     tags = [ "e2e" ];
 
-    ldflags = [ "-s" ];
+    nativeBuildInputs = [
+      makeWrapper
+    ];
+
+    ldflags = [
+      "-s"
+    ];
```
Freax13, line 36:
```suggestion
    ldflags = [ "-s" ];
```

.github/workflows/e2e_aks_runtime.yml
```diff
@@ -0,0 +1,104 @@
...
+        uses: ./.github/actions/post_to_teams
+        with:
+          webhook: ${{ secrets.TEAMS_CI_WEBHOOK }}
+          title: "aks-runtime test failed"
+          message: "e2e test aks-runtime failed"
+          additionalFields: '[{"title": "Platform", "value": "AKS-CLH-SNP"}]'
+      - name: Cleanup
+        if: cancelled()
+        run: |
+          kubectl delete ns "$(cat workspace/e2e.namespace)" --timeout 5m
```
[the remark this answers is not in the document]
miampf: This is done in the `t.Cleanup` function in `aks_runtime_test.go`

e2e/aks-runtime/aks_runtime_test.go
```diff
@@ -0,0 +1,142 @@
...
+	t.Cleanup(func() {
+		if skipUndeploy {
+			return
+		}
+
+		// delete the deployment
+		deletePolicy := metav1.DeletePropagationForeground
+		require.NoError(c.Client.CoreV1().Namespaces().Delete(context.Background(), namespace, metav1.DeleteOptions{
+			PropagationPolicy: &deletePolicy,
+		}))
```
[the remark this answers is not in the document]
miampf: Changed this to an error check with a `t.Fatalf` if cleanup fails.

.github/workflows/e2e_aks_runtime.yml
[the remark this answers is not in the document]
burgerdev: But `t.Cleanup` takes the `--skip-undeploy=false` set above into account.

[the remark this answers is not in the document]
burgerdev: Arrgh, sorry, I inverted that in my head. It's clear now: we're forcing the cleanup function with the arg and therefore don't need it in success and error cases.

burgerdev (approved):
LGTM, let's merge it :)

overlays/nixpkgs.nix
```diff
@@ -1,7 +1,61 @@
...
+  # outside of nix would be fetched by the CLI itself from various sources.
+  mkAzExtension =
+    {
+      pname,
+      url,
+      sha256,
+      description,
+      ...
+    }@args:
+    prev.python3.pkgs.buildPythonPackage (
```
katexochen, line 17: This should be final, not prev.

packages/by-name/contrast/package.nix
```diff
@@ -26,8 +26,6 @@ let
 
     tags = [ "e2e" ];
 
-    ldflags = [ "-s" ];
-
```
katexochen, line 30: Why was the stripping removed?

overlays/nixpkgs.nix
```diff
@@ -13,7 +67,10 @@ final: prev: {
 
   # Add the required extensions to the Azure CLI.
   azure-cli = prev.azure-cli.override {
-    withExtensions = with final.azure-cli.extensions; [ aks-preview ];
+    withExtensions = with final.azure-cli.extensions; [
+      aks-preview
+      confcom
```
katexochen, line 72: Is this used anywhere in the end? Looks like we are installing the cli via apt.

PR #943, opened 2024-10-20
runtime-k3s-qemu-snp.yml: node-installer has too little memory

blenessy:
The OOM Killer kicks in when below 600M on my Milan. Setting to 700M to have some slack.

You can reproduce this with [runtime-k3s-qemu-snp.yml v1.1.0][1], which limits memory to 100M memory.

[1]: https://github.com/edgelesssys/contrast/releases/download/v1.1.0/runtime-k3s-qemu-snp.yml

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -158,7 +158,7 @@ func NodeInstaller(namespace string, platform platforms.Platform) (*NodeInstalle
 						WithName("installer").
 						WithImage(nodeInstallerImageURL).
 						WithResources(ResourceRequirements().
-							WithMemoryLimitAndRequest(100),
+							WithMemoryLimitAndRequest(700),
 						).
 						WithSecurityContext(SecurityContext().WithPrivileged(true).SecurityContextApplyConfiguration).
 						WithVolumeMounts(VolumeMount().


Freax13 (approved):
Thank you for your contribution!

Reproducing this wasn't easy.

We used to use a kernel config based on Ubuntu 22.04 on our AMD server. I didn't see any OOMs on that config. However, after I switched our AMD server to a config based on Ubuntu 24.10, I did get OOMs as well. I didn't look much into what changed, but I strongly suspect that either Canonical changed the kernel config for Ubuntu 24.10 or that there are some new config options  that the old Ubuntu 22.04 kernel didn't have and that change the behavior of the OOM killer.
Interestingly, we use a semi-official kernel for TDX based on Ubuntu 24.04, and I also don't get any issues on that server either.

Now that our AMD CI machine is using the 24.10-based config, we should catch this much earlier in the future.

700 MiB (or even just 600 MiB) seems a bit excessive to me, I don't know why the node-installer is using that much memory, but if that's what it takes, that's fine with me.

Again, our CI can't handle outside contributors well, so I tested this manually.

katexochen:
Backporting https://github.com/edgelesssys/contrast/pull/948 instead

PR #945, opened 2024-10-21
infra: better default values

msanft:
Improve the default values in the deployment a little:
- Free-tier clusters are not always available, so this makes the deployment default to using a standard-tier cluster.
- GPU instances (which are the only ones we provide a working image for at the moment) are unavailable in germanywestcentral. Therefore, default to a GPU-enabled region here. This should not have any disadvantages compared to germanywestcentral for non-GPU use.


Freax13 (approved):

burgerdev (changes requested):
Standard tier almost doubles the price of a one-node cluster. Is there no overlap between regions that have GPUs and regions that have free-tier AKS? Can we at least make this configurable, so that peerpod development without need for GPU can choose?

PR #946, opened 2024-10-21
podvm-image: don't mount tmpfs on /usr

msanft:
This is a prerequisite for using peer pods with GPUs, as they require an OCI hook to facilitate GPU attachment to containers, which is expected in `/usr/share` by default. If we mount a tmpfs on `/usr`, the files placed in the initial image through the `contents` attribute of the repart builder will become invisible, so we now only mount it at `/usr/bin` instead.

### DIFF
--- a/packages/nixos/image.nix
+++ b/packages/nixos/image.nix
@@ -45,7 +45,7 @@
           Minimize = "best";
           # We need to ensure that mountpoints are available.
           # TODO (Maybe): This could be done more elegantly with CopyFiles and a skeleton tree in the vcs.
-          MakeDirectories = "/bin /boot /dev /etc /home /lib /lib64 /mnt /nix /opt /proc /root /run /srv /sys /tmp /usr /var";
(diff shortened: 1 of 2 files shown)


burgerdev:
Why do we need a tmpfs on /usr/bin? Do we even need a tmpfs at /usr? 

```
       /usr   This directory is usually mounted from a separate
              partition.  It should hold only shareable, read-only data,
              so that it can be mounted by various machines running
              Linux.
```

msanft:
> Why do we need a tmpfs on /usr/bin? Do we even need a tmpfs at /usr?

Good question! Unfortunately, Nix doesn't adhere to the FHS (or other things that are common sense in other distributions, e.g. /usr being *merged* - i.e. `/bin` linking to `/usr/bin`, and so on). Instead, binaries are always sourced from the Nix store for purity. Hence, `/usr/bin/env`, which is provided as a compatibility wrapper, e.g. to make shebangs of form `!#/usr/bin/env bash` work, is a symlink to the `env` binary in the Nix store. The link is created at boot-time within a so-called *activation script*. Thus `/usr/bin` needs to be writable.

Freax13 (approved):

PR #949, opened 2024-10-22
just: configure platform from jusfile.env file

davidweisse:
The default platform for the justfile is now configured from the `justfile.env` file instead of from within the justfile itself. However, it is still possible to pass different platforms as an argument, e.g., `just node-installer K3s-QEMU-TDX`.

### DIFF
--- a/justfile
+++ b/justfile
@@ -35,7 +35,7 @@ nydus-snapshotter: (push "nydus-snapshotter")
 
 default_cli := "contrast.cli"
 default_deploy_target := "openssl"
-default_platform := "AKS-CLH-SNP"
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -366,6 +366,8 @@ rctemplate := '''
 container_registry=""
 # Azure resource group/ resource name. Resource group will be created.
 azure_resource_group=""
+# Platform to deploy on
+default_platform=""
```
katexochen, line 370:
```suggestion
default_platform="AKS-CLH-SNP"
```

Maybe we can set a sane default?

katexochen (approved):
otherwise lgtm

PR #950, opened 2024-10-24
some fixes to the runtime cleanup

Freax13:
It turns out that we need to fix the containerd state after removing the data folders for the snapshotters, otherwise containers won't start anymore.


packages/scripts.nix
```diff
@@ -497,17 +508,10 @@
       busybox
       kubectl
       dasel
+      cleanup-images
```
katexochen, line 511: You can remove this from the let binding and refer to it as scripts.cleanup-images

.github/workflows/cleanup.yml
katexochen, line 1: can this be done as part of the test?

Freax13: What do you mean?

katexochen: ah never mind, I'm not fully awake yet.

tools/bm-maintenance/deployment_tdx_snp.yml
```diff
@@ -47,7 +47,7 @@ spec:
           serviceAccountName: cleanup-sa
           containers:
             - name: cleanup
-              image: ghcr.io/edgelesssys/contrast/cleanup-bm:v0.0.1
+              image: ghcr.io/edgelesssys/contrast/cleanup-bare-metal:v0.0.1
```
katexochen, line 50: How is the image updated after push? How do we ensure the deployment is running with the latest image?

katexochen: I think you'd have to set PullPolicy to always, otherwise it will just use the cached image.

Freax13: If the image tag is `latest` the default is to always pull the image: https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting.

PR #955, opened 2024-10-28
packages/nixos: pin nvidia driver

msanft:
The `production` nvidia driver in nixpkgs got updated to a version that's incompatible with H100s. This pins the nvidia driver to the specific version needed for H100s by using the upstream `mkDriver` abstraction that builds a driver compatible with the used kernel from exact driver sources.

### DIFF
--- a/packages/nixos/gpu.nix
+++ b/packages/nixos/gpu.nix
@@ -15,7 +15,18 @@ in
   config = lib.mkIf cfg.enable {
     hardware.nvidia = {
       open = true;
-      package = lib.mkDefault config.boot.kernelPackages.nvidiaPackages.production;
+      package = lib.mkDefault (
+        config.boot.kernelPackages.nvidiaPackages.mkDriver {
+          # TODO: Investigate why the latest version breaks
+          # GPU containers.
+          version = "550.90.07";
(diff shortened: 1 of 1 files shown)


Freax13:
> The `production` nvidia driver in nixpkgs got updated to a version that's incompatible with H100s.

How do you know that? Are there any issues for this?

packages/nixos/gpu.nix
```diff
@@ -15,7 +15,16 @@ in
   config = lib.mkIf cfg.enable {
     hardware.nvidia = {
       open = true;
-      package = lib.mkDefault config.boot.kernelPackages.nvidiaPackages.production;
+      package = lib.mkDefault (
+        config.boot.kernelPackages.nvidiaPackages.mkDriver {
+          version = "550.90.07";
```
Freax13, line 20: Can we add a `TODO` to investigate why we can't use the latest version?

Freax13 (approved):
> Things broke when we updated. This might not be true for use with H100s in general, but at least for the CC case. Maybe @derpsteb knows more about the exact driver compatibilities.

IIRC @derpsteb is on vacation for the next couple of weeks. I'm fine downgrading the version for now and delaying investigating whether or not/how we can upgrade.

katexochen (approved):
lgtm beside https://github.com/edgelesssys/contrast/pull/955#discussion_r1819224186

PR #956, opened 2024-10-28
packages: add nvidia OCI hook for peer pods

msanft:
This enables the `podvm-image` package for use in GPU peer pods by doing the following:
- Adding a preconfigured `nvidia-ctk` package for use with peer pods GPU containers.
- Adding a package for the OCI hook required to start peer pods GPU containers.
- Placing the OCI hook in the expected path in the podvm-image if GPU use is configured.

### DIFF
--- a/packages/by-name/mkNixosConfig/package.nix
+++ b/packages/by-name/mkNixosConfig/package.nix
@@ -41,6 +41,7 @@ lib.makeOverridable (
             cloud-api-adaptor
             kernel-podvm-azure
             pause-bundle
+            nvidia-ctk-with-config
             ;
           inherit (outerPkgs.kata) kata-agent;
         })
--- a/packages/by-name/mkNixosConfig/package.nix
+++ b/packages/by-name/mkNixosConfig/package.nix
@@ -41,6 +41,7 @@ lib.makeOverridable (
             cloud-api-adaptor
             kernel-podvm-azure
             pause-bundle
+            nvidia-ctk-oci-hook
             nvidia-ctk-with-config
             ;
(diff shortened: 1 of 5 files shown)


packages/by-name/mkNixosConfig/package.nix
```diff
@@ -35,15 +35,24 @@ lib.makeOverridable (
...
+          {
+            inherit (outerPkgs)
+              azure-no-agent
+              cloud-api-adaptor
+              kernel-podvm-azure
+              pause-bundle
+              ;
+            inherit (outerPkgs.kata) kata-agent;
+          }
+          // lib.optionalAttrs config.contrast.gpu.enable {
```
Freax13, line 49: Do we need to make the packages conditional on the `gpu.enable`? Isn't everything here lazily evaluated anyway?

katexochen:
```
       error: evaluation aborted with the following error message: 'lib.customisation.callPackageWith: Function called without required argument "libnvidia-container-custom" at /nix/store/dvppxran2w1d1jgs7lhrih6012vmd24l-source/packages/by-name/nvidia-ctk-with-config/package.nix:9'
```

I guess this depends on another PR? Wrong base?

packages/by-name/mkNixosConfig/package.nix
Freax13: @msanft Did we resolve this offline? I can't remember.

msanft: Sorry, forgot to push :P

Freax13 (approved):

katexochen (approved):

PR #947, opened 2024-10-22
e2e: add `kata.contrast-node-installer-image` to reproducibility test

miampf:
Also test the `kata.contrast-node-installer-image` package for reproducibility. A successful run can be found [here](https://github.com/edgelesssys/contrast/actions/runs/11386897346).

PR #960, opened 2024-10-29
e2e: fix runtime reproducibility test and also check `kata.contrast-node-installer-image`

miampf:
Follow up of #947. New successful run can be found [here](https://github.com/edgelesssys/contrast/actions/runs/11502616673).

### DIFF
--- a/.github/workflows/e2e_runtime-reproducibility.yml
+++ b/.github/workflows/e2e_runtime-reproducibility.yml
@@ -16,7 +16,7 @@ jobs:
         # is reproducible across individual builds (as the --rebuild flag is used, causing Nix to rebuild the node-installer-image derivation)
         # and across independent builds on Ubuntu 20.04 and 22.04 (which also test the reproducibility of the transitive closure of our packages, as no shared
         # cache is present between the two machines)
-        build-target: ["microsoft.contrast-node-installer-image"]
+        build-target: ["microsoft.contrast-node-installer-image", "kata.contrast-node-installer-image"]
       fail-fast: false
     # Usually we would define the matrix outputs here, but as GitHub Actions don't seem to allow per-combination outputs,
     # we'll write the outputs without defining them here. See https://github.com/orgs/community/discussions/17245#discussioncomment-3814009.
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_runtime-reproducibility.yml
```diff
@@ -64,25 +71,34 @@ jobs:
...
-              for line in f:
-                  matrix, checksum = line.strip().split("=")
-                  if not checksum in seen:
-                      seen[checksum] = []
-                  seen[checksum].append(matrix)
-
+          for file in os.listdir("./checksums"):
+              checksum = pathlib.Path("./checksums", file).read_text()
+              if not checksum in seen:
+                  seen[checksum] = []
```
Freax13, line 94:
```suggestion
          seen = { pathlib.Path("./checksums", file).read_text() for file in os.listdir("./checksums") }
```

miampf: forgot to add info where checksum was seen, added it. List comprehensions like this aren't possible with the new change.

Freax13 (approved):

katexochen:
@miampf could you separate the fix and the additional test into two distinct commits please (and rebase-merge later)?

miampf:
> @miampf could you separate the fix and the additional test into two distinct commits please (and rebase-merge later)?

It will take some time until merge anyways since the test is not failing as expected. When it comes to merging it I will keep this in mind tho :)

miampf:
The test _does_ fail if different values are detected (look [here](https://github.com/edgelesssys/contrast/actions/runs/11611379147), I just randomly generated strings instead of using a hash), however, it does not fail when comparing the images digests (look [here](https://github.com/edgelesssys/contrast/actions/runs/11577111422)), reason being that they are equal. @katexochen, since reproducibility should be broken but the digests are the same, should we compare something other than the image digest? If yes, what do you think would be a good option?

katexochen:
> katexochen, since reproducibility should be broken but the digests are the same, should we compare something other than the image digest? If yes, what do you think would be a good option?

It's fine, the current reproducibility issues are really hard to reproduce, it would be to much work to detect that in a ci job. The test succeeds as this reproduces between the two machine, so there is no blocker to merge this from my perspective.

PR #962, opened 2024-10-29
just: use default platform for get-credentials

davidweisse:
Instead of having multiple commands like `get-credentials-tdxbm` or `get-credentials-snpbm`, the platform can simply be read from the `justfile.env` file.

### DIFF
--- a/justfile
+++ b/justfile
@@ -283,14 +283,30 @@ wait-for-workload target=default_deploy_target:
         ;;
     esac
 
-# Load the kubeconfig from the running AKS cluster.
-get-credentials:
-    nix run -L .#azure-cli -- aks get-credentials \
-        --resource-group "$azure_resource_group" \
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -283,28 +283,36 @@ wait-for-workload target=default_deploy_target:
...
-
-get-credentials-peerpod:
-    nix run -L .#scripts.merge-kube-config -- ./infra/azure-peerpods/kube.conf
-
-# Load the kubeconfig from the CI AKS cluster.
-get-credentials-ci:
-    nix run -L .#azure-cli -- aks get-credentials \
-        --resource-group "contrast-ci" \
-        --name "contrast-ci" \
-        --admin
```
katexochen, line 300: Maybe we can keep this, as it is not a platform?

katexochen (approved):

PR #970, opened 2024-11-04
kata: 3.9.0 -> 3.10.1

fidencio:
Let's update Kata Containers to its latest release


packages/by-name/kata/kata-runtime/0004-genpolicy-enable-sysctl-checks.patch
Freax13, line 1: One patch was dropped during the rebase, so all the following files got a new patch number. This PR only adds the files with the new version numbers, but doesn't remove the old files e.g. [0004-genpolicy-validate-create-sandbox-storages.patch](https://github.com/edgelesssys/contrast/blob/cea2865b23f6594942eefc42da73d95bd9f56de0/packages/by-name/kata/kata-runtime/0004-genpolicy-validate-create-sandbox-storages.patch).

Freax13 (changes requested):

packages/by-name/kata/kata-runtime/0004-genpolicy-enable-sysctl-checks.patch
fidencio: Sorry, that was my bad, should be fixed now.

katexochen (approved):

PR #972, opened 2024-11-04
ci: upload peerpod images during release

Freax13:


packages/scripts.nix
```diff
@@ -91,6 +91,54 @@
...
+        name = "contrast"
+        provider = "azure"
+
+        [base.azure]
+        subscriptionID = "0d202bbb-4fa7-4af8-8125-58c269a05435"
+        location = "GermanyWestCentral"
+        resourceGroup = "contrast-images"
+        sharedImageGallery = "contrast_images"
+        sharingProfile = "community"
+        sharingNamePrefix = "Contrast"
```
fidencio, line 135: I'd recommend adding something like:
```sh
[variant.default.azure]
replicationRegions = [...]
```

And make sure that the other regions where potential customers are testing would also have the image uploaded to them.

Freax13: Good idea! I added the same regions we also use for [constellation](https://github.com/edgelesssys/constellation/blob/54058eed2a6895de2708d90485cf4e9daea3414a/internal/osimage/uplosi/uplosi.conf.in#L6).

katexochen: Why not use the [locations contrast supports](https://docs.edgeless.systems/contrast/getting-started/cluster-setup#create-resource-group)?

packages/scripts.nix
```diff
@@ -92,6 +91,55 @@
...
+        replicationRegions = ["eu-west-1", "eu-west-3", "us-east-1", "us-east-2", "ap-south-1"]
```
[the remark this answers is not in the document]
Freax13: Whoops, no, of course not. Good catch!

fidencio:
lgtm, thanks @Freax13!

fidencio (approved):

3u13r (approved):
Tested that the release workflow overrides an existing images as already done for e.g. the container images.
LGTM

3u13r:
As discussed with @katexochen, blocked until at least https://github.com/edgelesssys/contrast/pull/992 is merged.

PR #977, opened 2024-11-04
just: ensure peerpod vms are cleaned up on undeploy

davidweisse:
This ensures the `cloud-api-adaptor` cleanly deletes the created VMs by waiting until the deletion of all other resources has been completed before deleting the node installer and namespace.

### DIFF
--- a/justfile
+++ b/justfile
@@ -142,14 +142,6 @@ apply-runtime target=default_deploy_target platform=default_platform:
 apply target=default_deploy_target:
     #!/usr/bin/env bash
     set -euo pipefail
-    case {{ target }} in
-        "openssl" | "emojivoto" | "volume-stateful-set")
-            :
-        ;;
-        *)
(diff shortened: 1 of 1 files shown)


justfile
Freax13, line 1: Some other targets use `undeploy` as a sub-target. Please update those references to also pass the target explicitly.

justfile
```diff
@@ -169,6 +169,11 @@ undeploy:
         echo "Namespace $ns does not exist, nothing to undeploy."
         exit 0
     fi
+    if [[ {{ platform }} == "AKS-PEER-SNP" ]]; then
+        kubectl delete --cascade=foreground -f .//{{ workspace_dir }}/deployment
+        kubectl delete namespace $ns
+        exit 0
```
Freax13, line 175: Is there an advantage in delete the namespace here and exiting rather than just falling through to the existing code?

Freax13 (approved):

Freax13 (approved):

PR #980, opened 2024-11-04
ci: check for unclean patches

burgerdev:
The patch workflow we document is unfortunately not yet idempotent across committers: 

1. Commits created by `git am` are recorded with a `Committer: ` tag which depends on who is doing the committing. Since this tag is part of the commit, the commit hash changes and the leading `From <commit>` line of the patch file differs. The committer can't be overridden by a command line arg, afaik.
2. It seems like the number of characters in the `index <hash>..<hash>` lines created by `git format-patch` depends on the git version. 

Overall, the current workflow does not meet its original goal: a patch cherry-picked on the existing stack should not cause a diff in the existing patches.

I suggest the following changes to keep this clean in the future:
* Work around the idempotency by
  * masking the commit hash in the formatted patch and
  * not abbreviating the index hashes.
* Checking for patch cleanliness in a Github action.

### DIFF
--- a/.github/workflows/check_patches.yml
+++ b/.github/workflows/check_patches.yml
@@ -1,0 +1,40 @@
+name: check patches
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - .github/workflows/check_patches.yml
+      - packages/by-name/kata/kata-runtime/package.nix
+      - packages/by-name/kata/kata-runtime/0*.patch
+
+jobs:
+  check-patches:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: contrast
+      - uses: ./contrast/.github/actions/setup_nix
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - id: kataversion
+        working-directory: contrast
+        run: |
+          echo "v=$(nix eval --raw .#kata.kata-runtime.version)" >> "$GITHUB_OUTPUT"
+      - name: Check out Kata
+        run: |
+          git clone --depth 1 -b "${{ steps.kataversion.outputs.v }}" https://github.com/kata-containers/kata-containers.git kata
+      - name: Apply patches
+        working-directory: kata
+        run: |
+          git config --global user.email "actionsbot@example.invalid"
+          git config --global user.name "Actions Bot"
+          git am --no-3way ../contrast/packages/by-name/kata/kata-runtime/0*.patch
+          git format-patch -N --no-signature --zero-commit --full-index -o ../contrast/packages/by-name/kata/kata-runtime/ "${{ steps.kataversion.outputs.v }}"
+      - name: Check for diff
+        working-directory: contrast
+        run: |
+          git diff --exit-code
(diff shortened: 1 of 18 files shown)


burgerdev:
Test is failing due to #981 .

.github/workflows/static.yml
```diff
@@ -68,3 +68,33 @@ jobs:
...
+          echo "v=$(nix eval --raw .#kata.kata-runtime.version)" >> "$GITHUB_OUTPUT"
+      - name: Check out Kata
+        run: |
+          git clone --depth 1 -b "${{ steps.kataversion.outputs.v }}" https://github.com/kata-containers/kata-containers.git kata
+      - name: Apply patches
+        working-directory: kata
+        run: |
+          git config --global user.email "actionsbot@example.invalid"
+          git config --global user.name "Actions Bot"
+          git am --no-3way --whitespace=fix ../contrast/packages/by-name/kata/kata-runtime/*.patch
```
katexochen, line 95: Why is this command so different from what we document in the patching guide?

burgerdev: * `--no-3way`: If the patch workflow is thoroughly followed, 3-way merges should not be necessary. My idea of adding the flag explicitly here was to catch unclean patches in CI. If the existing patches are clean, it does not make a difference to the developer workflow, though, but I can remove the `-3` from the docs to align the two.
* `--whitespace=fix`: This should catch newly introduced whitespace "problems". However, not everybody subscribes to the whitespace dogma of git, so I would not mind removing it. wdyt?

katexochen: I don't really care what we enforce here, but I guess we should align somehow so we don't have the ci failing even everything was done right. Also this isn't using `/0*.patch` to filter out the other patches, is it.

katexochen:
Maybe we can move this into a separate workflow so we can run it with a path filter, so it is only executed in case patches were touched?

.github/workflows/static.yml
burgerdev: Ok, I'm going to
* remove the whitespace check, as its more likely to be annoying than helpful
* change the docs not to use `-3` and explain why

> Also this isn't using /0*.patch to filter out the other patches.

Good catch, thanks!

burgerdev:
> Maybe we can move this into a separate workflow so we can run it with a path filter, so it is only executed in case patches were touched?

... or the `package.nix`, for version changes.

fidencio:
> ... or the `package.nix`, for version changes.

Does it actually matter? I'm considering that:
* When a version bump happens we have 2 scenarios:
  * Patches need to be changed if they don't apply cleanly anymore
    * This would be caught by patches being touched
  * Patches apply cleanly, no changes needed
    *  Then there's no reason to run any of the checks

burgerdev:
> > ... or the `package.nix`, for version changes.
> 
> Does it actually matter? I'm considering that:
> 
> * When a version bump happens we have 2 scenarios:
>   
>   * Patches need to be changed if they don't apply cleanly anymore
>     
>     * This would be caught by patches being touched
>   * Patches apply cleanly, no changes needed
>     
>     * Then there's no reason to run any of the checks

I believe the nix patch machinery applies the patches with the equivalent of `-3`, so a version bump could make the patches unclean, but still applicable and thus not failing the build.

fidencio:
> I believe the nix patch machinery applies the patches with the equivalent of `-3`, so a version bump could make the patches unclean, but still applicable and thus not failing the build.

Fair point. :-)

.github/workflows/static.yml
burgerdev: Addressed.

katexochen (approved):

PR #985, opened 2024-11-07
e2e: improve log collection

miampf:
Improve the log collection to also collect init container logs. This is done by adding a `log-collector` daemon set to every e2e cluster that saves logs from the host node which then are downloaded to the local machine through the updated `get-logs` script.

### DIFF
--- a/packages/scripts.nix
+++ b/packages/scripts.nix
@@ -427,7 +427,10 @@
   # Usage: get-logs $namespaceFile
   get-logs = writeShellApplication {
     name = "get-logs";
-    runtimeInputs = with pkgs; [ kubectl ];
+    runtimeInputs = with pkgs; [ 
+      kubectl
+      ncompress
+    ];
     text = ''
       while ! [[ -s "$1" ]]; do
         sleep 1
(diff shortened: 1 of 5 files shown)


packages/scripts.nix
```diff
@@ -425,30 +425,32 @@
...
+      pod="$(kubectl get pods -o name -n "$namespace" | grep log-collector | cut -c 5-)"
+      mkdir -p ./workspace/logs
+      kubectl wait --for=condition=Ready -n "$namespace" "pod/$pod" 1>/dev/null 2>/dev/null
+      # Download and extract the logs every 3 seconds
+      while true; do
+        kubectl exec -n "$namespace" "$pod" -- /bin/bash -c "rm -f /exported-logs.tar.gz; tar zcvf /exported-logs.tar.gz /export" 1>/dev/null 2>/dev/null
+        rm -f ./workspace/logs/exported-logs.tar.gz
+        kubectl cp -n "$namespace" "$pod:/exported-logs.tar.gz" ./workspace/logs/exported-logs.tar.gz 1>/dev/null 2>/dev/null
+        tar xzvf ./workspace/logs/exported-logs.tar.gz --directory ./workspace/logs 1>/dev/null 2>/dev/null
+        sleep 3
```
burgerdev, line 452: Why do we need to create tarballs continuously? I imagined that this would happen only after the test finished.

burgerdev (commented):
This looks great - I did not think the implementation could be this concise!

packages/scripts.nix
miampf: I did not find a _good_ way to know if the test is finished or not, except maybe making an explicit download script (or giving `get-logs` CLI parameters like `start` and `download`). Also, I believe that this approach is more stable if the namespace with the log collector accidentally gets cleaned up before logs could be downloaded. If you think that this is unnecessary and has to much overhang, I'll gladly go with the "two scripts/two-part script" solution :)

burgerdev: You could handle the TERM signal (or send and intercept a different signal) with logs export.

miampf: I'm not sure if I understand what you mean correctly? I can't imagine a way to (easily) let the remote log collector send a signal to a local process and I think it should not be needed to add extra lines of code to every e2e tests go source just to tell the log collector that it is finished. I _could_ try to queue the current process list, grep the first entry that matches a `*.test` pattern and wait for its PID to stop but that seems a bit overcomplicated?

miampf: I just split the script into two parts now, it seemed like the simplest solution to me :)

miampf:
Run with new script structure and working log collection: https://github.com/edgelesssys/contrast/actions/runs/11912439438?pr=985

packages/scripts.nix
```diff
@@ -424,33 +424,53 @@
...
-          if ! [[ -f "$logfile" ]]; then
-            {
-              touch "$logfile" # prevents creation of to much processes
-              # wait for all containers of the pod to come online, then collect the logs
-              kubectl wait pod --all --for=condition=Ready --timeout="-1s" -n "$namespace" "$pod" 1>/dev/null 2>/dev/null
-              kubectl logs -f --all-containers=true -n "$namespace" "$pod" > "$logfile"
-            } &
-          fi
+
+      if [[ -z "''${1-}" ]]; then
```
burgerdev, line 436:
```suggestion
      if [[ $# < 2 ]]; then
```
... and remove the `-z $2` checks below.

burgerdev (approved):

PR #986, opened 2024-11-08
node-installer: write kata config with less perms

katexochen:


burgerdev (commented):
From https://pkg.go.dev/os#WriteFile:

> If the file does not exist, WriteFile creates it with permissions perm (before umask); otherwise WriteFile truncates it before writing, without changing permissions.

Shouldn't we then use `0o666` for files and `0o777` for directories, to respect the user's `umask`?

nodeinstaller/node-installer.go
```diff
@@ -91,15 +91,15 @@ func run(ctx context.Context, fetcher assetFetcher, platform platforms.Platform,
 		}
 
 		if file.Executable {
-			if err := os.Chmod(filepath.Join(hostMount, targetPath), 0o755); err != nil {
+			if err := os.Chmod(filepath.Join(hostMount, targetPath), 0o777); err != nil {
```
burgerdev, line 94: This is an unintended edit, right?

burgerdev (approved):

PR #988, opened 2024-11-11
packages/nixos: add IMDS setup script

msanft:
Azure needs special care for enabling IMDS within Peerpods. This adds a script to setup IMDS through Proxy ARP (from Peerpods upstream, see https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh), so that all requests to the IMDS from within the pod are routed through an interface that is peered to the Pod VM. Verified to work in 2 distinct Azure peer pods.

It also makes a simplification by excluding `basic.target` in the `wantedBy` list of the Azure readiness report, as it's already contained in the [default dependencies](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Default%20Dependencies).

### DIFF
--- a/packages/by-name/cloud-api-adaptor/package.nix
+++ b/packages/by-name/cloud-api-adaptor/package.nix
@@ -10,6 +10,9 @@
   writeShellApplication,
   gnugrep,
   iptables,
+  iproute2,
+  sysctl,
+  gawk,
   runCommand,
   applyPatches,
   makeWrapper,
@@ -101,6 +104,22 @@ buildGoModule rec {
         "SC2086"
         "SC2153"
       ];
+    };
+
+    setup-nat-for-imds = writeShellApplication {
+      name = "setup-nat-for-imds";
+      runtimeInputs = [
+        iproute2
+        iptables
+        sysctl
+        gawk
+      ];
+      # TODO(burgerdev): generalize for all link-local IPs and investigate routing simplification
+      text = builtins.readFile "${cloud-api-adaptor.src}/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh";
+      meta = {
+        mainProgram = "setup-nat-for-imds";
+        homepage = "https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh";
+      };
     };
   };
 
--- a/packages/nixos/azure.nix
+++ b/packages/nixos/azure.nix
@@ -85,5 +85,25 @@ in
         ExecStart = "${lib.getExe pkgs.azure-no-agent}";
       };
     };
+
+    systemd.services.setup-nat-for-imds = {
+      wantedBy = [ "multi-user.target" ];
+      requires = [ "netns@podns.service" ];
+      wants = [ "network-online.target" ];
+      after = [
+        "network-online.target"
+        "netns@podns.service"
+      ];
+      description = "Setup NAT for IMDS";
+      serviceConfig = {
+        Type = "oneshot";
+        RemainAfterExit = "yes";
+        # TODO(msanft): Find out why just ordering this after network-online.target
+        # isn't sufficient. (Errors with saying that the network is unreachable)
+        Restart = "on-failure";
+        RestartSec = "5s";
+        ExecStart = "${lib.getExe pkgs.cloud-api-adaptor.setup-nat-for-imds}";
+      };
+    };
   };
 }
--- a/packages/test-peerpods.sh
+++ b/packages/test-peerpods.sh
@@ -39,9 +39,17 @@ if [[ $found != true ]]; then
   exit 1
 fi
 
+run_tests() {
+  pod="$(kubectl get pod -l app=alpine -o jsonpath='{.items[0].metadata.name}')"
+
+  # Check IMDS functionality.
+  # -f makes this fail on a 500 status code.
+  kubectl exec "$pod" -- curl -f -i -H "Metadata: true" http://169.254.169.254/metadata/THIM/amd/certification
+}
+
 cleanup() {
-  kubectl delete deploy nginx
-  kubectl wait --for=delete pod --selector=app=nginx --timeout=5m
+  kubectl delete deploy alpine
+  kubectl wait --for=delete pod --selector=app=alpine --timeout=5m
 }
 
 trap cleanup EXIT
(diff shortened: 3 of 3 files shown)


packages/by-name/mkNixosConfig/package.nix
```diff
@@ -42,7 +42,11 @@ lib.makeOverridable (
             kernel-podvm-azure
             pause-bundle
             ;
+
           inherit (outerPkgs.kata) kata-agent;
+          inherit (outerPkgs)
+            peerpod-imds-nat
+            ;
```
katexochen, line 49: put that in the inherit statement above?

packages/by-name/peerpod-imds-nat/setup-nat-for-imds.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+# Copyright 2024 Edgeless Systems GmbH
+# SPDX-License-Identifier: AGPL-3.0-only
```
katexochen, line 3: Should keep upstream copyright. Can we consume from cloud-api-adaptor.src instead of vendoring? Maybe even put it in cloud-api-adaptor.passthru, where we already have some other scripts from the repo.

packages/by-name/cloud-api-adaptor/package.nix
```diff
@@ -86,6 +90,21 @@ buildGoModule rec {
...
+
+    setup-nat-for-imds = writeShellApplication {
+      name = "setup-nat-for-imds";
+      runtimeInputs = [
+        iproute2
+        iptables
+        sysctl
+        gawk
+      ];
+      text = builtins.readFile "${cloud-api-adaptor.src}/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh";
```
burgerdev, line 102:
```suggestion
      # TODO(burgerdev): generalize for all link-local IPs and investigate routing simplification 
      text = builtins.readFile "${cloud-api-adaptor.src}/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh";
```
This script is a bit weird: I wonder if a route `via $host` and the masquerade would be enough. We should reiterate after we can test it.

packages/nixos/azure.nix
```diff
@@ -85,5 +82,21 @@ in
...
+    systemd.services.setup-nat-for-imds = {
+      wantedBy = [ "multi-user.target" ];
+      wants = [ "network-online.target" ];
+      after = [ "network-online.target" ];
+      description = "Setup NAT for IMDS";
+      serviceConfig = {
+        Type = "oneshot";
+        RemainAfterExit = "yes";
+        # TODO: Find out why just ordering this after network-online.target
+        # isn't sufficient. (Errors with saying that the network is unreachable)
```
burgerdev, line 95: Do you know what step produces these errors? One dependency that's not declared is the `podns` existence.

packages/by-name/cloud-api-adaptor/package.nix
msanft: My networking knowledge does not suffice to tell from your comment what you want to do, but I'm happy to add this comment :D

packages/nixos/azure.nix
burgerdev: If the `netns` exists but routing is not set up yet, that would explain it.

msanft: Verifying that hypothesis right now.

msanft: Still fails at the very same line. Ordering is correct though:

```
...
 195ms setup-nat-for-imds.service
 193ms systemd-udevd.service
 190ms systemd-oomd.service
 154ms netns@podns.service
...
```

burgerdev: setup-nat should go after netns, is that actually what your snippet shows?

burgerdev: Actually, having the podns is not enough, it also needs routes set up, which might only happen in the agent-protocol-forwarder.

msanft: Good point, it doesn't! `systemd-analyze blame` does not order by execution but by time the unit took to execute. 

However, the ordering still seems to be correct. `netns@podns.service` executes (exits) 10s before `setup-nat-for-imds.service` starts.

katexochen:
We should add an e2e test step to check imds functionality before merging this.

packages/nixos/azure.nix
```diff
@@ -72,10 +72,7 @@ in
 
     services.udev.extraRules = azure-storage-rules;
     systemd.services.azure-readiness-report = {
-      wantedBy = [
-        "basic.target"
```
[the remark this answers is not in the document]
msanft: No, this is a default dependency which does not need to be explicitly specified.

packages/nixos/azure.nix
msanft: Can you elaborate? This works as-is. I don't know what I should change.

packages/by-name/cloud-api-adaptor/package.nix
```diff
@@ -102,6 +105,22 @@ buildGoModule rec {
...
+        iproute2
+        iptables
+        sysctl
+        gawk
+      ];
+      # TODO(burgerdev): generalize for all link-local IPs and investigate routing simplification
+      text = builtins.readFile "${cloud-api-adaptor.src}/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh";
+      meta = {
+        mainProgram = "setup-nat-for-imds";
+        homepage = "https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh";
```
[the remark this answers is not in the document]
msanft: This breaks the hyperlink. I'd rather have a clickable link to main, as we have below for example. Do you have a strong opinion here?

packages/nixos/azure.nix
```diff
@@ -85,5 +85,25 @@ in
...
+      wants = [ "network-online.target" ];
+      after = [
+        "network-online.target"
+        "netns@podns.service"
+      ];
+      description = "Setup NAT for IMDS";
+      serviceConfig = {
+        Type = "oneshot";
+        RemainAfterExit = "yes";
+        # TODO: Find out why just ordering this after network-online.target
```
katexochen, line 101: TODO is missing responsible person.

katexochen (approved):

PR #989, opened 2024-11-11
packages.OVMF-SNP: patch missing git submodule

katexochen:


katexochen:
e2e test run: https://github.com/edgelesssys/contrast/actions/runs/11795021640

packages/by-name/OVMF-SNP/package.nix
```diff
@@ -8,16 +8,20 @@
...
-edk2.mkDerivation "OvmfPkg/AmdSev/AmdSevX64.dsc" rec {
+edk2.mkDerivation "OvmfPkg/AmdSev/AmdSevX64.dsc" {
   name = "OVMF-SNP";
+
   src = fetchFromGitHub {
     owner = "amdese";
     repo = "ovmf";
-    # https://github.com/AMDESE/ovmf/tree/apic-mmio-fix4
-    rev = "64b3116ed087f8cb026201e56e42efe751e2cf7d";
+    # https://github.com/edgelesssys/ovmf/commits/apic-mmio-fix4-edgeless/
```
msanft, line 17: I don't get this comment. `owner = "amdese";` here? So how do we use the fork?

katexochen: github magic.

msanft: Ok. I would have probably used the upstream src and a patch instead, since if someone tries to find out whether that repository (i.e. the fork) is used anywhere, and doesn't find a direct reference, they might delete it? But I'm fine with that risk for now.

msanft (approved):

PR #991, opened 2024-11-12
packages/buildMicroVM: init

msanft:
This adds a builder for "MicroVM" images which are comprised of only a kernel, its kernel command line, and a root filesystem image. Additionally, this builder has a sanity check to see if the rootfs image contains an ESP, which indicates that an image thats not supposed to be a microVM was built with this builder.

### DIFF
--- a/packages/by-name/buildMicroVM/package.nix
+++ b/packages/by-name/buildMicroVM/package.nix
@@ -1,0 +1,36 @@
+# Copyright 2024 Edgeless Systems GmbH
+# SPDX-License-Identifier: AGPL-3.0-only
+
+# Builds a micro VM image (i.e. rootfs, kernel and kernel cmdline) from a NixOS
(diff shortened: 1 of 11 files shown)


packages/by-name/buildMicroVM/package.nix
```diff
@@ -0,0 +1,36 @@
...
+
+# Builds a micro VM image (i.e. rootfs, kernel and kernel cmdline) from a NixOS
+# configuration. These components can then be booted in a microVM-fashion
+# with QEMU's direct Linux boot feature.
+# See: https://qemu-project.gitlab.io/qemu/system/linuxboot.html
+
+{
+  symlinkJoin,
+  lib,
+  ...
```
katexochen, line 12:
```suggestion
```

msanft:
Decided in-person to not proceed with micro VMs for the time being

PR #999, opened 2024-11-13
scripts: use coordinator rules/settings for bare metal

katexochen:
This fixes a bug where we would publish a coordinator deployment with a policy that can only be deployed in the default namespace.

Bug found by @3u13r

PR #1000, opened 2024-11-13
[release/v1.1] scripts: use coordinator rules/settings for bare metal

edgelessci:
Backport of #999 to `release/v1.1`.

Original description:

---

This fixes a bug where we would publish a coordinator deployment with a policy that can only be deployed in the default namespace.

Bug found by @3u13r


burgerdev (approved):
Congrats on #1000!

PR #1001, opened 2024-11-13
[release/v1.1] node-installer: remove resource limits

katexochen:
There's no strong reason for having resource limits on the node-installer and given the unpredictable nature of the kernel's resource accounting, we should remove the resource limits so that we don't run into unexpected OOMs.

Fixes f5da52d

(cherry picked from commit 7f50e0cc566e630bc30775e44b0a42c31b924360)


3u13r (approved):
LGTM

PR #1005, opened 2024-11-17
e2e: add smoke test for peer pods

burgerdev:


infra/azure-peerpods-iam/main.tf
```diff
@@ -0,0 +1,75 @@
...
+  tenant_id = data.azurerm_subscription.current.tenant_id
+}
+
+locals {
+  name = var.resource_group
+}
+
+data "azurerm_resource_group" "rg" {
+  name = var.resource_group
+}
```
katexochen, line 41: I think we must create the rg here instead of consuming it as data

justfile
```diff
@@ -330,12 +352,12 @@ destroy platform=default_platform:
...
         ;;
         "AKS-PEER-SNP")
-            nix run -L .#terraform -- -chdir=infra/azure-peerpods destroy
+            nix run -L .#terraform -- -chdir=infra/azure-peerpods destroy --auto-approve
 
             # Clean-up cached image ids.
             rm -f ${CONTRAST_CACHE_DIR}/image-upload/*.image-id
 
-            az group delete --name "${azure_resource_group}_caa_cluster" --yes
+            az group delete --name "${azure_resource_group}" --yes
```
katexochen, line 360:
```suggestion
```

infra/azure-peerpods-iam/main.tf
burgerdev: done

infra/azure-peerpods-iam/main.tf
```diff
@@ -0,0 +1,75 @@
...
+provider "azuread" {
+  tenant_id = data.azurerm_subscription.current.tenant_id
+}
+
+locals {
+  name = var.resource_group
+}
+
+resource "azurerm_resource_group" "rg" {
+  name     = var.resource_group
```
katexochen, line 40: We should still add a suffix somewhere to make this azure/peerpod specific. Otherwise rg might collide with aks/nested setup.

Also, upload-image step still expects a the _caa_cluster suffix.

infra/azure-peerpods/vars.tf
```diff
@@ -2,14 +2,26 @@ variable "name_prefix" {
   type = string
 }
 
-variable "image_resource_group_name" {
+variable "resource_group" {
   type = string
 }
```
katexochen, line 7: Maybee we can just derive this from `name_prefix`?

packages/test-peerpods.sh
```diff
@@ -0,0 +1,62 @@
...
+fi
+
+cat >infra/azure-peerpods/just.auto.tfvars <<EOF
+name_prefix = "${azure_resource_group:?}-$RANDOM"
+resource_group = "${azure_resource_group:?}"
+subscription_id = "${azure_subscription_id:?}"
+EOF
+
+nix run -L .#terraform -- -chdir=infra/azure-peerpods init
+nix run -L .#terraform -- -chdir=infra/azure-peerpods apply --auto-approve
```
katexochen, line 25: Could we now run `just create` instead?

katexochen:
otherwise lgtm

infra/azure-peerpods/vars.tf
burgerdev: I wanted the prefix to be randomizable while using a shared group (e.g. `contrast-ci`). Do you think it makes sense to do this within Terraform? It's currently in `test-peerpods.sh`.

infra/azure-peerpods-iam/main.tf
burgerdev: Ok, I added the `_caa_cluster` suffix back. Original reason for using the literal group was so that I can reuse `contrast-ci`, but now that the RG is owned by `create-pre` it does not matter anymore.

infra/azure-peerpods/vars.tf
katexochen: no, we cannot do this in terraform as it should not happen for developers. Maybe we can do this via justfile.env (azure_resource_group)?

packages/test-peerpods.sh
burgerdev: `just create` still includes `just upload-image`, but wanted to be able to use an existing image in the test. If we drop that, we only need to solve the resource group naming.

infra/azure-peerpods/vars.tf
katexochen: Maybe we want to even get rid of `azure_resource_group` and `namespace_suffix` and instead have a single source for naming, like `base_name`.

packages/test-peerpods.sh
katexochen: okay, let's postpone this to a future pr.

PR #1006, opened 2024-11-17
ci: prevent merge of PRs with do-not-merge label

burgerdev:


burgerdev:
* [with label](https://github.com/edgelesssys/contrast/actions/runs/11879587736?pr=1006)
* [without label](https://github.com/edgelesssys/contrast/actions/runs/11879594007?pr=1006)

.github/workflows/require_label.yml
```diff
@@ -22,3 +22,13 @@ jobs:
...
+  do-not-merge:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'do not merge') }}
+    runs-on: ubuntu-22.04
+    steps:
+      - name: prevent merge
+        run: |
+          tee "${GITHUB_STEP_SUMMARY}" <<'EOF'
+          This PR is labeled as `do not merge`, remove the label to make this check pass.
+          EOF
+          exit 1
```
katexochen, line 34:
```suggestion
          echo "::error This PR is labeled as `do not merge`, remove the label to make this check pass."
          exit 1
```

usually github won't link to the summary of the job on failure, maybe add an error message to the log with error prefix.

burgerdev: Hm, I wanted to have the message both in the job logs as well as in the summary, but I think `::error` won't play nice there. Let me rewrite it.

burgerdev:
I'm not going to rerun the e2e tests unless asked to, the PR should be fine now.

katexochen (approved):
actually, I don't think we even need to write to summary, annotation is already shown there anyway:
image

But I'm also fine to merge like this.

burgerdev:
Thanks, we don't need the summary then.

katexochen (approved):

PR #1007, opened 2024-11-18
e2e: default policy test

jmxnzo:
Podvms come with a default policy, which should usually be a "SetPolicy only" policy.  The implemented e2e test ensures that no wrong default policy (i.e. allow-all) is included, by starting a container without any policy annotation and waiting for the Kata runtime event FailedCreatePodSandBox, which indicates that the container start failed as expected.

1.  waitForEvent() was added to deploy.go, following the same pattern as waitFor(), but focusing on watching the eventList and no resource
2. checkIfStartingBlocked() checks if the FailedCreatePodSandBox event with the expected message occurred

### DIFF
--- a/e2e/internal/kubeclient/deploy.go
+++ b/e2e/internal/kubeclient/deploy.go
@@ -33,6 +33,15 @@ const (
 	InitContainersRunning
 )
 
+// WaitEventCondition is an enum type for the possible wait conditions when using `kubeclient.WaitForEvent`.
+type WaitEventCondition int
+
+const (
+	_ WaitEventCondition = iota
+	// StartingBlocked waits until a specific FailedCreatePodSandBox Event is detected which indicates that the container does not start.
+	StartingBlocked
+)
+
 // ResourceWaiter is implemented by resources that can be waited for with WaitFor.
 type ResourceWaiter interface {
 	kind() string
@@ -225,6 +234,32 @@ func (c *Kubeclient) checkIfRunning(ctx context.Context, name string, namespace 
 	return true, nil
 }
 
+// IsStartingBlocked checks whether the FailedCreatePodSandBox Event occurred which indicates that the SetPolicy request is rejected and the Kata Shim fails to start the Pod sandbox.
+func (c *Kubeclient) IsStartingBlocked(name string, namespace string, resource ResourceWaiter, evt watch.Event, startingPoint time.Time) (bool, error) {
+	switch evt.Type {
+	case watch.Error:
+		return false, fmt.Errorf("Watcher of %s %s/%s received an error event", resource.kind(), namespace, name)
+	case watch.Added:
+		fallthrough
+	case watch.Modified:
+		logger := c.log.With("namespace", namespace)
+		event, ok := evt.Object.(*corev1.Event)
+		if !ok {
+			return false, fmt.Errorf("watcher received unexpected type %T", evt.Object)
+		}
+
+		// Expected event: Reason: FailedCreatePodSandBox
+		// TODO(jmxnzo): Add patch to the existing error message in Kata Shim, to specifically allow detecting start-up of containers without policy annotation.
+		if (event.LastTimestamp.After(startingPoint)) && event.Reason == "FailedCreatePodSandBox" {
+			logger.Debug("Pod did not start", "name", name, "reason", event.Reason, "timestamp of failure", event.LastTimestamp.String())
+			return true, nil
+		}
+		return false, nil
+	default:
+		return false, fmt.Errorf("unexpected watch event while waiting for %s %s/%s: type=%s, object=%#v", resource.kind(), namespace, name, evt.Type, evt.Object)
+	}
+}
+
 // WaitFor watches the given resource kind and blocks until the desired number of pods are
 // ready or the context expires (is cancelled or times out).
 func (c *Kubeclient) WaitFor(ctx context.Context, condition WaitCondition, resource ResourceWaiter, namespace, name string) error {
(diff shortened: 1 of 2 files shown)


e2e/internal/kubeclient/deploy.go
```diff
@@ -31,6 +31,8 @@ const (
 	Ready
 	// InitContainersRunning waits until all initial containers of all pods of the resource are running.
 	InitContainersRunning
+	// StartingBlocked waits until a specific FailedCreatePodSandBox Event is detected which indicates that the container does not start.
+	StartingBlocked
```
burgerdev, line 35: The `WaitCondition` type is documented to be tied to `WaitFor`. On the other hand, `WaitForEvent` does not use any of the other conditions. Should this be a separate enum? Or maybe a predicate `func (e Event) bool`?

e2e/internal/kubeclient/deploy.go
```diff
@@ -225,6 +227,39 @@ func (c *Kubeclient) checkIfRunning(ctx context.Context, name string, namespace
 	return true, nil
 }
 
+// Checks whether the FailedCreatePodSandBox Event occurred which indicates that the SetPolicy request is rejected and the Kata Shim fails to start the Pod sandbox.
+func (c *Kubeclient) checkIfStartingBlocked(ctx context.Context, name string, namespace string, resource ResourceWaiter, evt watch.Event, startingPoint time.Time) (bool, error) {
+	switch evt.Type {
+	case watch.Deleted:
+		return false, fmt.Errorf("%s %s/%s was deleted while waiting for it", resource.kind(), namespace, name)
```
burgerdev, line 234: Is this triggered on deletion of the pod, or rather deletion of an event? Not sure whether we should handle it.

e2e/internal/kubeclient/deploy.go
```diff
@@ -225,6 +227,39 @@ func (c *Kubeclient) checkIfRunning(ctx context.Context, name string, namespace
...
+	case watch.Error:
+		return false, fmt.Errorf("Watcher of %s %s/%s received an error event", resource.kind(), namespace, name)
+	case watch.Added:
+		fallthrough
+	case watch.Modified:
+		logger := c.log.With("namespace", namespace)
+		eventList, err := c.Client.CoreV1().Events(namespace).List(ctx, metav1.ListOptions{FieldSelector: "involvedObject.name=" + name, TypeMeta: metav1.TypeMeta{Kind: resource.kind()}})
```
burgerdev, line 241: Why do we need to fetch the event list here, can't we take it from `evt`?

e2e/internal/kubeclient/deploy.go
```diff
@@ -225,6 +227,39 @@ func (c *Kubeclient) checkIfRunning(ctx context.Context, name string, namespace
...
+		if err != nil {
+			return false, err
+		}
+
+		// Expected event:
+		// Type: Warning
+		// Reason: FailedCreatePodSandBox
+		// From: pod/coordinator-0
+		// Message: Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: "UpdateInterfaceRequest: internal error query did not produce any values": unknown
```
burgerdev, line 250: That message is an implementation detail that's likely to change:

* if the Kata runtime decides to send a different agent request first
* if the Kata agent handles/logs policy evaluation slightly different
* if we include a default policy that denies everything, instead of leaving it undefined.

Imho checking for `FailedCreatePodSandbox` should be enough.

e2e/internal/kubeclient/deploy.go
```diff
@@ -370,6 +405,80 @@ loop:
...
 
+// WaitForEvent watches the EventList as long as the Event corresponding to the WaitCondition occurred after calling the function.
+func (c *Kubeclient) WaitForEvent(ctx context.Context, condition WaitCondition, resource ResourceWaiter, namespace, name string) error {
+	// The loop logic of this function is follows the pattern of waitFor(). Because we need to watch the EventList which should not implement the ResourceWaiter interface,
+	// we can not fall back to the resource watcher getter functions. Thus this function reimplements WaitFor() specifically designed to wait for specific ocuring events.
+
+	// StartingPoint is saved right here to ensure filtering only Events that happened after calling the function! The watcher keeps track of all events on the specific resource and
+	// as well calls the checking function when the EventList was changed regardless of the Event of our interest. Therefore avoidance of processing past Events has to be ensured, when
+	// checking the Event occurrence.
+	startingPoint := time.Now()
```
burgerdev, line 416: Is this actually important? We're already watching a specific pod, and changing the policy annotation will schedule a new pod, so I'd expect all events for this pod to qualify.

e2e/nopolicy/nopolicy_test.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// Namespace the tests are executed in.
+var (
+	imageReplacementsFile, namespaceFile, platformStr string
+	skipUndeploy                                      bool
+)
+
+// TestNoPolicy runs e2e test on the example of coordinator pod of the OpenSSL deployment, to ensure that container without policy annotation (default policy) do not start.
+func TestNoPolicy(t *testing.T) {
```
burgerdev, line 36: Does this need to be its own e2e test? Then we'd need to decide when to run it. On the other hand, the unique test logic in here seems to be a couple of lines - maybe we can just add it as a subtest to openssl?

burgerdev (commented):
Please also proof-read your code comments.

e2e/nopolicy/nopolicy_test.go
jmxnzo: Yes i get the point, but don't we require all of the contrast steps to successfully finish prior to the openssl test logic. This conflicts with using no policy annotation, so i would need to adjust the openssl test ?

burgerdev: Actually, now that you mention it I think the test should go into https://github.com/edgelesssys/contrast/blob/main/e2e/policy/policy_test.go.

This test does not depend on the Coordinator etc., scheduling a pod with a single container and the right runtime class should suffice. Could look similar to [this](https://github.com/edgelesssys/contrast/blob/e9964e61853fe5bc2f5226820ddb3f97de2fbce4/e2e/getdents/getdents_test.go#L46), but with only `ct.Init` and `ct.Apply`.

e2e/internal/kubeclient/deploy.go
jmxnzo: Thanks for pointing it out. I was confused with the type handling right here, because the event.Object was type Event and not EventList. Fetching the Event directly from the watcher defo is the better choice and allows to remove the introduced timestamp logic as well!

e2e/internal/kubeclient/deploy.go
jmxnzo: I slightly misinterpreted the event.Object in the watcher pattern, but i got it know:)

jmxnzo: image

e2e/internal/kubeclient/deploy.go
jmxnzo: It's triggered on deletion of an event, my bad!

e2e/internal/kubeclient/deploy.go
jmxnzo: Changed this to seperate enum!

e2e/internal/kubeclient/deploy.go
burgerdev: > The timestamp logic was introduced because the same event is reused in multiple waitForEvent instances.

Why do we need multiple `WaitForEvent` instances?

burgerdev (commented):
Please make sure to also push your changes when you're resolving review comments.

e2e/internal/kubeclient/deploy.go
jmxnzo: The idea of checking for the event occurrence multiple times, was just to get more evidence that the container really don't start using the default policy. Somehow it looks like the watcher also reacts to older events or more precisely includes the current eventList as watch.ADDED events when instantiating the watcher. The timestamp logic only ensures that these watch.ADDED events are not seen as the event occurrence, based on the timestamp comparison.

burgerdev: Did you know you can limit the events with https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions `SendInitialEvents` field?

I'm not really convinced that we need to observe the same failure more than once. If the concern is that it could fail once for an unrelated reason, but then succeed later on, should we improve the error message on the agent side instead and check for it?

jmxnzo: Oh no i didn't know, that actually looks like it bypasses the problem and the introduced timestamp logic can be dropped. 
Yes that was the initial concern, but the test logic like this introduces a lot of timing overhead just to observe the event multiple times. So i totally agree with you, that their might be a better solution for ensurance. Do you mean a specific error message that indicates the success later on?

jmxnzo: For my understanding we would need some message on agent side which determines that the default policy is still used and nevertheless the container was started? But i have no clue where to look at and how to check the policy annotation changes

e2e/internal/kubeclient/deploy.go
```diff
@@ -225,6 +234,31 @@ func (c *Kubeclient) checkIfRunning(ctx context.Context, name string, namespace
 	return true, nil
 }
 
+// Checks whether the FailedCreatePodSandBox Event occurred which indicates that the SetPolicy request is rejected and the Kata Shim fails to start the Pod sandbox.
+func (c *Kubeclient) checkIfStartingBlocked(name string, namespace string, resource ResourceWaiter, evt watch.Event, startingPoint time.Time) (bool, error) {
```
burgerdev, line 238: Consider renaming to `isStartingBlocked`. Also, please check the `static-checks` job results on your PR, it should have complained about the docstring format (must start with the function name).

e2e/policy/policy_test.go
```diff
@@ -58,6 +60,35 @@ func TestPolicy(t *testing.T) {
...
+		}
+		t.Log("Restarting container")
+
+		require.NoError(c.Restart(ctx, kubeclient.Deployment{}, ct.Namespace, coordinator))
+		t.Log("Waiting to ensure container start up failed")
+
+		for range 2 {
+			err := c.WaitForEvent(ctx, kubeclient.StartingBlocked, kubeclient.Pod{}, ct.Namespace, coordinatorPod)
+			require.NoError(err)
+		}
```
burgerdev, line 88: As mentioned in the other thread, I'd just check this once.

burgerdev (approved):
LGTM, but please address outstanding comments.

e2e/internal/kubeclient/deploy.go
burgerdev: This could also be a later addition, I think it's fine to check for the existing error message. The message is from https://github.com/kata-containers/kata-containers/blob/cdaaf708a18da8e5f7e2b9824fa3e43b524893a5/src/agent/src/policy.rs#L37, you would need to add a patch on our side to change it - see https://github.com/edgelesssys/contrast/blob/main/dev-docs/patches.md.

burgerdev:
Take a look at the requirements of https://github.com/edgelesssys/contrast/actions/runs/11934103702/workflow?pr=1007#L18-L24, too. Tests are usually `no changelog`.

e2e/internal/kubeclient/deploy.go
jmxnzo: > Did you know you can limit the events with https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions `SendInitialEvents` field?
> 
> I'm not really convinced that we need to observe the same failure more than once. If the concern is that it could fail once for an unrelated reason, but then succeed later on, should we improve the error message on the agent side instead and check for it?

Using the SendInitialEvents parameter in the request, always rejects the request. It seems like watching an object and using the SendInitialEvents param, can not be combined: https://github.com/kubernetes-client/c/issues/203 . If you don't mind, i would keep the timestamp logic, because it is working perfectly fine.

burgerdev (approved):

jmxnzo:
#1031

PR #1008, opened 2024-11-18
packages/by-name: add peerpod node installer

davidweisse:
This separates the Peerpod node installer from the Kata node installer and removes files that are only necessary for bare metal. The new runtime files are also included in the release.


packages/by-name/peerpod/kata-runtime/package.nix
msanft, line 1: Can't we just use `kata.kata-runtime` here? Same goes for the other components that don't differ between Kata and Peerpods

katexochen:
This needs fundamental rework, closing for now.

PR #1009, opened 2024-11-18
e2e: show debug logs on error

burgerdev:


katexochen:
When will this show? Aren't we normally dropping stdout?

katexochen:
Okay, but shouldn't we also see logs for set/verify? https://github.com/edgelesssys/contrast/actions/runs/11890697328/job/33134201064?pr=1009#step:11:252

burgerdev:
Those did not fail - the one instance I linked is an expected failure during the recovery test.

katexochen:
Makes sense, thanks!

PR #1010, opened 2024-11-18
coordinator: add log messages for handshake failures

burgerdev:
The gRPC stack does not log errors from the `Credentials`, and if the TLS session is not established the error won't be sent back to the client, either. With this commit, the error shows up at least in the server logs.


3u13r (approved):
LGTM

katexochen (approved):

PR #1019, opened 2024-11-19
Add NixOS image for bare-metal Kata

msanft:
This switches the image used in our bare-metal Kata uses (e.g. non-AKS and non-peerpods) to a NixOS image that we build in-tree as a MicroVM image (e.g. separated kernel, initrd, cmdline and rootfs).

I tried to split this up into digestible commits as much as possible, but a large part still has to be done in one commit in order to ensure bisectability through all commits. I also tried to write meaningful commit messages, so please consult these to find out about the specific changes.


packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -144,5 +147,10 @@ buildGoModule rec {
     in
     [ "-skip=^${builtins.concatStringsSep "$|^" skippedTests}$" ];
 
+  passthru.cmdline = {
```
burgerdev, line 150: Why is the cmdline attribute introduced in c07f3782ef78cde7be99346fde85ce0d9ef7a55b, which disables that Kata check?

packages/by-name/boot-microvm/package.nix
```diff
@@ -0,0 +1,36 @@
...
+# SPDX-License-Identifier: AGPL-3.0-only
+
+{
+  writeShellApplication,
+  qemu,
+  OVMF,
+}:
+
+# Usage example:
+# outPath=$(nix build .#kata.kata-image --print-out-paths); nix run .#boot-microvm -- "${outPath}/bzImage" "${outPath}/initrd" "${outPath}/image-podvm-gpu_1-rc1.raw" "$(nix eval --raw .#kata.kata-image.cmdline)"
```
burgerdev, line 11: What do you think about taking the base `$outPath` instead of `$1 $2 $3` and figuring out the kernel, initrd and rootfs automatically? If the cmdline where in the outputs, too, invocation could be very simple...

burgerdev (commented):
Thanks for properly organizing the commits, that indeed improves the review experience! I went through up-to-including 6ded3e8fdecc20d76b7fc84358f4e23454379ad8 for now.

packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -144,5 +147,10 @@ buildGoModule rec {
...
+    default = "tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k cryptomgr.notests net.ifnames=0 pci=lastbus=0 root=/dev/vda1 rootflags=ro rootfstype=erofs console=hvc0 console=hvc1 quiet systemd.show_status=false panic=1 nr_cpus=1 selinux=0 systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket scsi_mod.scan=none";
```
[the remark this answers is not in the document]
msanft: This is not created by us but the default command line used by Kata (hence I moved it to this Nix package). I do not understand the choices as well.

packages/by-name/boot-microvm/package.nix
msanft: I wanted to have a simple utility that's not specific to this image. If we want to tie it to booting this image, i'd put it in it's `passthru`. That might actually be a good idea.

packages/by-name/kata/kata-runtime/package.nix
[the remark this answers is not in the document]
msanft: Describing this in a comment.

packages/by-name/buildVerityMicroVM/package.nix
katexochen, line 1: package comment seems outdated. What part of this is microvm specific at this point? Looks like it's just an image in parts compared to packaging it as uki.

msanft: I think it's still a valid name for a non-bootable image, as you can boot it without a rootfs.

packages/by-name/ociLayerTar/package.nix
```diff
@@ -49,8 +49,11 @@ runCommandLocal "ociLayer"
 
     # Copy files into the tree (./root/)
     for i in ''${!srcs[@]}; do
+        # resolve symlinks
+        src=$(readlink -f ''${srcs[i]})
```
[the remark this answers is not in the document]
katexochen: I can't find any reasoning about this in the PR description. I agree with @burgerdev comment. What symlink isn't working for this fix to be needed?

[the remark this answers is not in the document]
msanft: Commit description, sorry. It's about kata-kernel, which in the NixOS toplevel is a symlink to the bzImage in a separate derivation for the kernel.

packages/by-name/buildVerityMicroVM/package.nix
```diff
@@ -0,0 +1,59 @@
...
+
+# Builds a micro VM image (i.e. rootfs, kernel and kernel cmdline) from a NixOS
+# configuration. These components can then be booted in a microVM-fashion
+# with QEMU's direct Linux boot feature.
+# See: https://qemu-project.gitlab.io/qemu/system/linuxboot.html
+
+{
+  symlinkJoin,
+  lib,
+  ...
```
katexochen, line 12:
```suggestion
```

packages/by-name/ociLayerTar/package.nix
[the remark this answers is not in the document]
burgerdev: Sorry, I don't understand: the "derivation for the kernel" should also be installed in the system, shouldn't it, making the symlink resolve just fine?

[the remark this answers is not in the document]
msanft: This is not about stuff in the system. We pack this into an OCI image layer, which is why I think adding support for resolving symlinks would be too complex compared to what we actually use this for. We select the kernel file from the "top-level" derivation to be included in the OCI layer. But it's just a symlink to a file in the derivation of the kernel. The OCI builder we have just copies the file without checking what it is nor dereferencing it. Thus, this - as of now - is an easy-to-trigger footgun for people using this. As we specify the target location in the layer for each file anyway, I think the proposed solution is just the best we can get with low complexity. I'm very strongly against keeping it as-is, just because the status quo is not sound imo.

nodeinstaller/internal/constants/constants.go
```diff
@@ -64,37 +64,36 @@ func KataRuntimeConfig(baseDir string, platform platforms.Platform, qemuExtraKer
...
+		config.Hypervisor["qemu"]["kernel"] = filepath.Join(baseDir, "share", "kata-kernel")
+		// Replace the kernel params entirely (and don't append) since that's
+		// also what we do when calculating the launch measurement.
+		config.Hypervisor["qemu"]["kernel_params"] = qemuExtraKernelParams
 		if debug {
 			config.Hypervisor["qemu"]["enable_debug"] = true
 		}
-		// Replace the kernel params entirely (and don't append) since that's
-		// also what we do when calculating the launch measurement.
-		config.Hypervisor["qemu"]["kernel_params"] = kernelParams
```
burgerdev, line 77: afaict there is only one line added here. Don't need to change it now, but please try to keep diffs minimal in large PRs.

packages/by-name/image-podvm/package.nix
```diff
@@ -15,5 +15,6 @@ buildVerityUKI (mkNixosConfig {
     debug.enable = withDebug;
     gpu.enable = withGPU;
     azure.enable = withCSP == "azure";
+    peerpods.enable = true;
```
burgerdev, line 18: Note for self: why in this PR?

packages/by-name/kata/contrast-node-installer-image/package.nix
```diff
@@ -17,7 +17,7 @@
   OVMF-SNP,
   OVMF-TDX,
 
-  debugRuntime ? false,
+  debugRuntime ? true,
```
burgerdev, line 20: Revert to false

packages/by-name/kata/contrast-node-installer-image/package.nix
```diff
@@ -116,13 +120,17 @@ let
   kata-container-img = ociLayerTar {
     files = [
       {
-        source = kata.kata-image;
+        source = "${kata.kata-image}/${kata.kata-image.imageFile}";
```
burgerdev, line 123: Could this be referenced directly as `${kata.kata-image.imageFile}`? It's a bit confusing that the path needs to be appended here.

packages/by-name/kata/tdx-launch-digests/package.nix
```diff
@@ -8,28 +8,34 @@
   OVMF-TDX,
   tdx-measure,
 
-  debug ? false,
+  debug ? true,
```
burgerdev, line 11: Revert to false

packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
```diff
@@ -0,0 +1,42 @@
...
+     }
+
++    // For TDX RTMRs to be predictable regardless of VM memory size, we need to
++    // load the initrd to a static address, so no dynamic value ends up in the
++    // mapped kernel image. (as the initrd address does)
++    // As we control the minimum VM memory size in Contrast, we just load the initrd
++    // to the address it gets loaded to for Contrast's minimum VM memory, regardless
++    // of if the VM has more memory, as it would otherwise happen with QEMU's 4Gi
++    // constraint.
++    uint32_t contrast_min_memory = 0x80000000;
```
burgerdev, line 37: I guess that needs to be documented, then. While we're at it, 2GiB seems a bit large, and I remember that we successfully dialed that down for AKS. Is this now a hard minimum?

packages/nixos/debug.nix
```diff
@@ -32,7 +32,9 @@ in
 
     services.getty.autologinUser = "root";
 
-    boot.kernelParams = [ "console=ttyS0" ];
+    # required for local booting, but no boot logs in kata with this
+    boot.kernelParams = lib.optionals config.contrast.azure.enable [ "console=ttyS0" ];
```
burgerdev, line 36: I don't understand: the serial console worked fine on BM, why do we disable it?

packages/nixos/kata.nix
```diff
@@ -1,90 +1,101 @@
...
+      mode = "0755";
+      user = "root";
     };
-  };
 
-  environment.etc."kata-opa/default-policy.rego".source = pkgs.cloud-api-adaptor.default-policy;
+    environment.etc."kata-opa/default-policy.rego".source = pkgs.writeText "default-policy.rego" ''
+      package agent_policy
+      default SetPolicyRequest := true
+    '';
```
burgerdev, line 99: Should we use `${kata.kata-runtime.src}/src/kata-opa/allow-set-policy.rego`?

tools/tdx-measure/rtmr/rtmr.go
```diff
@@ -38,6 +38,9 @@ func (r *Rtmr) Get() [48]byte {
 // Some helpers for common events:
 
 func (r *Rtmr) hashAndExtend(bytes []byte) {
+	// Comment out the following lines for debugging of the event log.
+	// sum := sha512.Sum384(bytes)
+	// fmt.Println(hex.EncodeToString(sum[:]))
```
burgerdev, line 43: Consider removing.

tools/tdx-measure/rtmr/rtmr.go
```diff
@@ -243,35 +246,58 @@ func CalcRtmr0(firmware []byte) ([48]byte, error) {
...
+
+	// TODO(msanft): Find out what goes wrong with the kernel hash calculation.
+	// For now, hard-code the kernel hash from a booted image.
+	// https://github.com/tianocore/edk2/blob/0f3867fa6ef0553e26c42f7d71ff6bdb98429742/OvmfPkg/Tcg/TdTcg2Dxe/TdTcg2Dxe.c#L1381
+	// kernelHash := "51877d12bdb51b50d99b213545042c6bbacde98a27ca52786cae8f4b55d71824d7c7aa3cf75ef8ed23543f2f2828b4a5"
+	// var buffer [48]byte
+	// if _, err := hex.Decode(buffer[:], []byte(kernelHash)); err != nil {
+	// 	panic(err)
+	// }
+	// rtmr.Extend(buffer)
```
burgerdev, line 260: Is that still relevant?

packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
```diff
@@ -0,0 +1,42 @@
...
+Date: Thu, 21 Nov 2024 14:36:23 +0100
+Subject: [PATCH] hw/x86: load initrd to static address
+
+For TDX RTMRs to be predictable regardless of VM memory size, we need to
+load the initrd to a static address, so no dynamic value ends up in the
+mapped kernel image. (as the initrd address does)
+As we control the minimum VM memory size in Contrast, we just load the initrd
+to the address it gets loaded to for Contrast's minimum VM memory, regardless
+of if the VM has more memory, as it would otherwise happen with QEMU's 4Gi
+constraint.
```
burgerdev, line 12: That 4Gi constraint is mentioned a few times, but I believe it's not explained anywhere.

packages/by-name/image-podvm/package.nix
burgerdev: Sorry, that's a leftover - I figured it out while reading :)

packages/nixos/debug.nix
msanft: QEMU, locally, without specifying a plethora of `virtconsole` options, will write to `ttyS0`. Kata reads the guest logs from `hvc*`, which doesn't work anymore when specifying `ttyS0` here. (Even though I think it should, but maybe more than 2 consoles are not supported by the kernel.)

packages/by-name/kata/kata-runtime/package.nix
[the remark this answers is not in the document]
msanft: Many places in the Kata code influence the generation of this value, so adding a single source is hard.

docs/docs/features-limitations.md
```diff
@@ -34,3 +34,10 @@ Currently, this requires inspecting the iptables rules on startup or terminating
 The Contrast Coordinator is a singleton and can't be scaled to more than one instance.
 When this instance's pod is restarted, for example for node maintenance, it needs to be recovered manually.
 In a future release, we plan to support distributed Coordinator instances that can recover automatically.
+
+## Overriding Kata configuration
+
+Kata Containers supports [overriding certain configuration values via Kubernetes annotations](https://github.com/kata-containers/kata-containers/blob/b4da4b5e3b9b21048af9333b071235a57a3e9493/docs/how-to/how-to-set-sandbox-config-kata.md).
+
+It needs to be noted that setting these values is unsupported, and doing so may lead to unexpected
+behaviour, as Contrast isn't tested against all possible configuration combinations.
```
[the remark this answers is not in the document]
msanft: I think we can hardly list things that are expected to work, as we don't test any of these. What do you think?
I'm also fine with specifying that some *may* work, and when they are useful.

I'm not sure about how `agent.kernel_modules` should work exactly, but besides the NVIDIA driver for GPU images, our images don't actually contain loadable modules.

Re `container_annotations`: Yeah, I think we'll want to set these.

packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
msanft: Yeah. We agreed on that minimum in the standup. I should make it more clear that especially this annotation may not be touched (at least in the downwards direction in the annotation docs section)

packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
msanft: It's in the line above the comment, which is why I didn't point it out more here. Will add a direct reference instead.

packages/by-name/kata/kata-runtime/package.nix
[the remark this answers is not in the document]
burgerdev: If there is no source for this, how do we derive it when it stops to work - empirically? Maybe note that in the comment, too.

docs/docs/features-limitations.md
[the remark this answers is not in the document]
burgerdev: Fair enough.

packages/by-name/ociLayerTar/package.nix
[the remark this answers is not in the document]
msanft: I don't like this (very sharp) edge of `ociLayerTar`, but I'll just drop the symlinks then..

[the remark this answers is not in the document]
burgerdev: Yes, I understand - it's entirely possible that the idea of `ociLayerTar` is flawed when combined with nix. We could consider adding the nix store paths to the image and intentionally add only symlinks to the /opt/edgeless tree. But I guess we can figure that out some other time.

packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
msanft: I tried to explain it in more detail now. Please let me know if it's more follow-able now.

burgerdev (approved):
Thanks, lgtm!

PR #1020, opened 2024-11-21
caa: upgrade to v0.10.0

burgerdev:
Changes to the network configuration of Azure pod VMs require a NAT gateway in the subnet. Having that is desirable anyway, because default outbound access is deprecated and scheduled for removal in 2025.

Side note: this bumps peer-pods Kata to 3.9.0, which should make it easier to integrate CAA into the node-installer.

### DIFF
--- a/packages/by-name/cloud-api-adaptor/package.nix
+++ b/packages/by-name/cloud-api-adaptor/package.nix
@@ -9,7 +9,10 @@
   libvirt,
   writeShellApplication,
   gnugrep,
+  iptables,
   runCommand,
+  applyPatches,
+  makeWrapper,
 
   # List of supported cloud providers
   builtinCloudProviders ? [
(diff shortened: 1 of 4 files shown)


packages/by-name/cloud-api-adaptor/package.nix
```diff
@@ -31,19 +32,28 @@ in
 
 buildGoModule rec {
   pname = "cloud-api-adaptor";
-  version = "0.9.0";
+  version = "0.10.0";
+
+  src = applyPatches {
```
msanft, line 37: Should we use `applyPatches` here? This is only useful if there are other packages depending on the source of this one, which I can't think of in the case of CAA?

msanft (commented):
Diff LGTM mostly, but the CI SA needs some augmentation with the necessary permissions, it seems.

packages/by-name/cloud-api-adaptor/package.nix
burgerdev: I need to either apply patches before the `sourceRoot`, or apply patch flags. I find the former easier to reason about, and it allows cherry-picking patches from upstream that modify outside files - is there a downside to using `applyPatches`?

msanft: Not really, it's just not something you see very often in Nix. I agree with your reasoning though and think it makes sense in that case.

msanft (approved):

packages/by-name/cloud-api-adaptor/package.nix
```diff
@@ -3,13 +3,16 @@
 
 {
   lib,
+  pkgs,
```
msanft, line 6: We don't really pass the top-level `pkgs` to functions in our tree. (idk why, tbh)
You'd probably want to pass `iptables` only instead.

packages/by-name/cloud-api-adaptor/package.nix
```diff
@@ -31,23 +34,32 @@ in
...
 
   proxyVendor = true;
-  vendorHash = "sha256-kqzi7jRF3tQ4/yLkJXfZly4EvVKFb400/WXlN0WjYm8=";
+  vendorHash = "sha256-FsckYZAiBfTEp25+dDNqPpB/550NqeEsutWC34s+GmE=";
 
-  nativeBuildInputs = lib.optional withLibvirt pkg-config;
+  nativeBuildInputs = [ makeWrapper ] ++ lib.optional withLibvirt pkg-config;
 
-  buildInputs = lib.optional withLibvirt libvirt;
+  buildInputs = [ pkgs.iptables ] ++ lib.optional withLibvirt libvirt;
```
msanft, line 62: If you reference `iptables` in the `postInstall` below, that's enough to add it to the closure of the package, I think. You shouldn't need it here too.

packages/by-name/cloud-api-adaptor/package.nix
burgerdev: Done

packages/by-name/cloud-api-adaptor/package.nix
burgerdev: Done

burgerdev:
Thanks, but we do have an e2e test for peer pods now, so I don't consider manual testing necessary on a PR.

PR #1021, opened 2024-11-21
e2e: don't require e2e on unrelated PRs

burgerdev:
We run e2e tests with a `pull_request` trigger and an explicit list of ignored paths. This makes it tedious to work on code that can't be excluded easily, like other workflows that don't touch e2e.

Moving to a mix of included and excluded paths should fix this issue. In particular, this commit is intended to trigger e2e on PRs that touch anything that's not in the existing ignorelist and not an unrelated Github workflow.


3u13r (approved):
LGTM

PR #1024, opened 2024-11-22
Add TDX debugging tools and documentation

msanft:
This adds a suite of debugging utilities for low-level TDX problems such as mismatching RTMRs, as well as documentation on how to perform the debugging.

This includes:
- `tdx-tools` in our debug images, which is a Python utility to view RTMRs and their event log.
- A debug switch in `OVMF-TDX`, allowing to debug the TDX virtual firmware.
- A script to launch a TDX QEMU like Kata would, which can then be used with the debug build of OVMF, as Kata doesn't support it.

Additionally, some documentation is added on how to perform debugging on TDX measurements.

### DIFF
--- a/packages/by-name/OVMF-TDX/0001-verify-Hobs-instead-of-measuring-them.patch
+++ b/packages/by-name/OVMF-TDX/0001-verify-Hobs-instead-of-measuring-them.patch
@@ -1,18 +1,20 @@
-From 649d2928a8f6552ba0644e6f9a7f719dfbdabe60 Mon Sep 17 00:00:00 2001
-From: Tom Dohrmann <erbse.13@gmx.de>
-Date: Mon, 23 Sep 2024 10:57:09 +0200
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Moritz Sanft <58110325+msanft@users.noreply.github.com>
+Date: Thu, 21 Nov 2024 09:43:08 +0100
 Subject: [PATCH] verify Hobs instead of measuring them
 
 We don't want the measurement to change depending on the amount of
 memory assigned to the CVM. Instead verify the Hobs to make sure that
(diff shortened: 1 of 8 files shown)


packages/by-name/OVMF-TDX/0001-verify-Hobs-instead-of-measuring-them.patch
```diff
@@ -1,6 +1,6 @@
-From 649d2928a8f6552ba0644e6f9a7f719dfbdabe60 Mon Sep 17 00:00:00 2001
-From: Tom Dohrmann <erbse.13@gmx.de>
-Date: Mon, 23 Sep 2024 10:57:09 +0200
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Moritz Sanft <58110325+msanft@users.noreply.github.com>
```
burgerdev, line 2: That's not supposed to happen, I believe.

burgerdev (approved):
Wow, thanks a lot!

packages/by-name/OVMF-TDX/0001-verify-Hobs-instead-of-measuring-them.patch
msanft: This is the result of recreation the patch in the form we advertise in `dev-docs/patches.md`

burgerdev: `git am` should not replace the author :thinking:

msanft: I had to `git apply --3way`, as `git am --3way` didn't work for some reason I don't remember anymore. Maybe this changed it, but I'll revert the author change manually for now.

PR #1028, opened 2024-11-25
ci: descriptive names for tests

burgerdev:
Our actions tab has a lot of workflows called `e2e test`, making it hard to navigate. 

* Make the names more descriptive.
* Rename the openssl test so that the intent (execute on PR) is clear.


msanft (approved):
I like it, thanks!

PR #1030, opened 2024-11-25
sdk: move verify code into sdk

derpsteb:
# Context
This PR is motivated by Continuum's need to call `contrast verify` from go code. Ideally, it can do this without embedding the Contrast CLI.

After discussing with Markus it seemed like the best way to have a go client on contrast main and marking it as unstable. We figured that this approach has higher chances of resulting in usable code for the first official release. It is also the easiest to work with from Continuum's perspective. Alternatives we talked about: not merging this branch, building something out of tree. 

# Proposed changes
Move the code interacting the coordinator from the CLI into a new module `sdk`. The only noteworthy new piece of code is the struct `GetManifestsResponse`. This is a new type used by the SDK to contain the data from `userapi.GetManifestsRequest`. This is done so we don't have to export `userapi`.

Resolves [AB#4856](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4856)

### DIFF
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -6,20 +6,10 @@ package cmd
 import (
 	"context"
 	_ "embed"
-	"fmt"
-	"log/slog"
 	"os"
-	"path/filepath"
 	"time"
 
 	"github.com/edgelesssys/contrast/cli/telemetry"
-	"github.com/edgelesssys/contrast/internal/atls"
-	"github.com/edgelesssys/contrast/internal/attestation/certcache"
-	"github.com/edgelesssys/contrast/internal/attestation/snp"
-	"github.com/edgelesssys/contrast/internal/attestation/tdx"
-	"github.com/edgelesssys/contrast/internal/fsstore"
-	"github.com/edgelesssys/contrast/internal/logger"
-	"github.com/edgelesssys/contrast/internal/manifest"
 	"github.com/spf13/cobra"
 )
 
@@ -35,7 +25,6 @@ const (
 	rulesFilename        = "rules.rego"
 	layersCacheFilename  = "layers-cache.json"
 	verifyDir            = "verify"
-	cacheDirEnv          = "CONTRAST_CACHE_DIR"
 )
 
 var (
@@ -47,18 +36,6 @@ var (
 	// It is intentionally left empty for dev builds.
 	DefaultCoordinatorPolicyHash = ""
 )
-
-func cachedir(subdir string) (string, error) {
-	dir := os.Getenv(cacheDirEnv)
-	if dir == "" {
-		cachedir, err := os.UserCacheDir()
-		if err != nil {
-			return "", err
-		}
-		dir = filepath.Join(cachedir, "contrast")
-	}
-	return filepath.Join(dir, subdir), nil
-}
 
 func must(err error) {
 	if err != nil {
@@ -80,41 +57,4 @@ func withTelemetry(runFunc func(*cobra.Command, []string) error) func(*cobra.Com
 
 		return cmdErr
 	}
-}
-
-// validatorsFromManifest returns a list of validators corresponding to the reference values in the given manifest.
-func validatorsFromManifest(m *manifest.Manifest, log *slog.Logger, hostData []byte) ([]atls.Validator, error) {
-	kdsDir, err := cachedir("kds")
-	if err != nil {
-		return nil, fmt.Errorf("getting cache dir: %w", err)
-	}
-	log.Debug("Using KDS cache dir", "dir", kdsDir)
-	kdsCache := fsstore.New(kdsDir, log.WithGroup("kds-cache"))
-	kdsGetter := certcache.NewCachedHTTPSGetter(kdsCache, certcache.NeverGCTicker, log.WithGroup("kds-getter"))
-
-	var validators []atls.Validator
-
-	opts, err := m.SNPValidateOpts(kdsGetter)
-	if err != nil {
-		return nil, fmt.Errorf("getting SNP validate options: %w", err)
-	}
-	for _, opt := range opts {
-		opt.ValidateOpts.HostData = hostData
-		validators = append(validators, snp.NewValidator(opt.VerifyOpts, opt.ValidateOpts,
-			logger.NewWithAttrs(logger.NewNamed(log, "validator"), map[string]string{"tee-type": "snp"}),
-		))
-	}
-
-	tdxOpts, err := m.TDXValidateOpts()
-	if err != nil {
-		return nil, fmt.Errorf("generating TDX validation options: %w", err)
-	}
-	var mrConfigID [48]byte
-	copy(mrConfigID[:], hostData)
-	for _, opt := range tdxOpts {
-		opt.TdQuoteBodyOptions.MrConfigID = mrConfigID[:]
-		validators = append(validators, tdx.NewValidator(&tdx.StaticValidateOptsGenerator{Opts: opt}, logger.NewWithAttrs(logger.NewNamed(log, "validator"), map[string]string{"tee-type": "tdx"})))
-	}
-
-	return validators, nil
 }
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -15,7 +15,6 @@ import (
 )
 
 const (
-	cacheDirEnv          = "CONTRAST_CACHE_DIR"
 	coordHashFilename    = "coordinator-policy.sha256"
 	coordRootPEMFilename = "coordinator-root-ca.pem"
 	meshCAPEMFilename    = "mesh-ca.pem"
@@ -27,6 +26,7 @@ const (
 	rulesFilename        = "rules.rego"
 	layersCacheFilename  = "layers-cache.json"
 	verifyDir            = "verify"
+	cacheDirEnv          = "CONTRAST_CACHE_DIR"
 )
 
 var (
@@ -38,6 +38,18 @@ var (
 	// It is intentionally left empty for dev builds.
 	DefaultCoordinatorPolicyHash = ""
 )
+
+func cachedir(subdir string) (string, error) {
+	dir := os.Getenv(cacheDirEnv)
+	if dir == "" {
+		cachedir, err := os.UserCacheDir()
+		if err != nil {
+			return "", err
+		}
+		dir = filepath.Join(cachedir, "contrast")
+	}
+	return filepath.Join(dir, subdir), nil
+}
 
 func must(err error) {
 	if err != nil {
@@ -59,16 +71,4 @@ func withTelemetry(runFunc func(*cobra.Command, []string) error) func(*cobra.Com
 
 		return cmdErr
 	}
-}
-
-func cachedir(subdir string) (string, error) {
-	dir := os.Getenv(cacheDirEnv)
-	if dir == "" {
-		cachedir, err := os.UserCacheDir()
-		if err != nil {
-			return "", err
-		}
-		dir = filepath.Join(cachedir, "contrast")
-	}
-	return filepath.Join(dir, subdir), nil
 }
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	_ "embed"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/edgelesssys/contrast/cli/telemetry"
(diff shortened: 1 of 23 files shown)


sdk/common.go
```diff
@@ -0,0 +1,72 @@
...
+	copy(mrConfigID[:], hostData)
+	for _, opt := range tdxOpts {
+		opt.TdQuoteBodyOptions.MrConfigID = mrConfigID[:]
+		validators = append(validators, tdx.NewValidator(&tdx.StaticValidateOptsGenerator{Opts: opt}, logger.NewWithAttrs(logger.NewNamed(log, "validator"), map[string]string{"tee-type": "tdx"})))
+	}
+
+	return validators, nil
+}
+
+func cachedir(subdir string) (string, error) {
```
burgerdev, line 62: I feel that the caching mechanism is a bit too specific for use in the SDK. Could we make it an input instead?

sdk/common.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/edgelesssys/contrast/internal/logger"
+	"github.com/edgelesssys/contrast/internal/manifest"
+)
+
+const cacheDirEnv = "CONTRAST_CACHE_DIR"
+
+// ValidatorsFromManifest returns a list of validators corresponding to the reference values in the given manifest.
+// Originally an unexported function in the contrast CLI.
+// Can be made unexported again, if we decide to move all userapi calls from the CLI to the SDK.
+func ValidatorsFromManifest(m *manifest.Manifest, log *slog.Logger, hostData []byte) ([]atls.Validator, error) {
```
burgerdev, line 26: `hostData` should be called `coordinatorPolicyChecksum` or something like that.

I also wonder where this would come from. For Contrast releases, we're embedding it during build. But other users of the SDK?

burgerdev, line 26: I have mixed feelings about debug logging in a library. Should it at least be self-configuring?

sdk/verify.go
```diff
@@ -0,0 +1,90 @@
...
+	return GetManifestsResponse{
+		Manifests: resp.Manifests,
+		Policies:  resp.Policies,
+		RootCA:    resp.RootCA,
+		MeshCA:    resp.MeshCA,
+	}, nil
+}
+
+// GetManifestsResponse contains the Coordinator's response to a GetManifests call.
+type GetManifestsResponse struct {
```
burgerdev, line 83: Maybe we should rename this to avoid confusion with the proto API. It's actually more of a `CoordinatorState`. Same goes for the `GetManifests` function.

sdk/common.go
derpsteb: Why so? I often hear this sentiment but haven't gotten around to asking where it comes from. 
Self-configuring means that there is no logger argument and the function instead configures loggers transparently?

I intentionally did not change the function because it could be an SDK-internal function if the gRPC calls from `contrast set` and `contrast recover` were also in the SDK. Not exporting this function would be the better solution imo. However, I did not want to move those two calls into SDK pkg to keep the PR smaller.

derpsteb: changed. see other comment.

sdk/verify.go
```diff
@@ -0,0 +1,90 @@
...
+
+// New returns a Client.
+func New(log *slog.Logger) Client {
+	return Client{
+		log: log,
+	}
+}
+
+// Verify checks if a given manifest is the latest manifest in the given history.
+func Verify(expected []byte, history [][]byte) error {
```
elchead, line 33: nit: makes usage clearer
```suggestion
func Verify(expectedManifest []byte, coordinatorManifests [][]byte) error {
```

sdk/verify.go
```diff
@@ -0,0 +1,90 @@
...
+	currentManifest := history[len(history)-1]
+	if !bytes.Equal(currentManifest, expected) {
+		return fmt.Errorf("active manifest does not match expected manifest")
+	}
+
+	return nil
+}
+
+// GetManifests calls GetManifests on the coordinator's userapi.
+func (c Client) GetManifests(ctx context.Context, manifestBytes []byte, endpoint string, policyHash []byte) (GetManifestsResponse, error) {
```
elchead: the endpoint should be part of the client

sdk/verify.go
elchead, line 33: nit: can you move this up before the client impl?

sdk/verify.go
derpsteb: What do you mean here? It is intended that the sdk wraps the userapi pkg.

elchead:
When importing the package on darwin, I get the following issue:
```
(base) ➜  continuum git:(as/contrast-proxy) ✗ go run ./verifier/main.go 
# github.com/edgelesssys/contrast/internal/attestation/snp
../../go/pkg/mod/github.com/edgelesssys/contrast@v0.0.0-20241126082351-2bd35a1483fe/internal/attestation/snp/issuer.go:109:21: undefined: client.LinuxConfigFsQuoteProvider
../../go/pkg/mod/github.com/edgelesssys/contrast@v0.0.0-20241126082351-2bd35a1483fe/internal/attestation/snp/issuer.go:114:21: undefined: client.LinuxIoctlQuoteProvider
```

This can be easily fixed as proposed here: https://github.com/edgelesssys/contrast/commit/1075e615aa3b9c7f8c1cbf8d6b4221456e3712d3

Could we integrate this in a followup?

burgerdev:
> When importing the package on darwin, I get the following issue:
> 
> ```
> (base) ➜  continuum git:(as/contrast-proxy) ✗ go run ./verifier/main.go 
> # github.com/edgelesssys/contrast/internal/attestation/snp
> ../../go/pkg/mod/github.com/edgelesssys/contrast@v0.0.0-20241126082351-2bd35a1483fe/internal/attestation/snp/issuer.go:109:21: undefined: client.LinuxConfigFsQuoteProvider
> ../../go/pkg/mod/github.com/edgelesssys/contrast@v0.0.0-20241126082351-2bd35a1483fe/internal/attestation/snp/issuer.go:114:21: undefined: client.LinuxIoctlQuoteProvider
> ```
> 
> This can be easily fixed as proposed here: [1075e61](https://github.com/edgelesssys/contrast/commit/1075e615aa3b9c7f8c1cbf8d6b4221456e3712d3)
> 
> Could we integrate this in a followup?

@elchead: would you mind opening a PR for this? But please make sure to keep some symmetry between SNP and TDX.

elchead:
@burgerdev I'd also propose to rename to proto packages to a unique name, as advised in the style guide (https://protobuf.dev/programming-guides/style/#packages). This would avoid the current name clash in Continuum:

```
panic: proto: file "userapi.proto" has a name conflict over userapi.SetManifestResponse
        previously from: "github.com/edgelesssys/continuum/internal/proto/attestation-service/userapi"
        currently from:  "github.com/edgelesssys/contrast/internal/userapi"
See https://protobuf.dev/reference/go/faq#namespace-conflict
```

I suppose that this would be OK for Contrast compatibility because a CLI version is tied to the Contrast deployment version?

elchead (approved):

cli/cmd/common.go
```diff
@@ -82,39 +61,14 @@ func withTelemetry(runFunc func(*cobra.Command, []string) error) func(*cobra.Com
...
-	tdxOpts, err := m.TDXValidateOpts()
-	if err != nil {
-		return nil, fmt.Errorf("generating TDX validation options: %w", err)
-	}
-	var mrConfigID [48]byte
-	copy(mrConfigID[:], hostData)
-	for _, opt := range tdxOpts {
-		opt.TdQuoteBodyOptions.MrConfigID = mrConfigID[:]
-		validators = append(validators, tdx.NewValidator(&tdx.StaticValidateOptsGenerator{Opts: opt}, logger.NewWithAttrs(logger.NewNamed(log, "validator"), map[string]string{"tee-type": "tdx"})))
+func cachedir(subdir string) (string, error) {
```
burgerdev, line 64: Nit: the function moved unnecessarily.

sdk/verify.go
```diff
@@ -0,0 +1,94 @@
...
+
+// New returns a Client.
+func New(log *slog.Logger) Client {
+	return Client{
+		log: log,
+	}
+}
+
+// Verify checks if a given manifest is the latest manifest in the given history.
+func Verify(expectedManifest []byte, manifestHistory [][]byte) error {
```
burgerdev, line 33: This could be a method for `CoordinatorState`, might prevent confusing various byte slices.

burgerdev (commented):
Overall LGTM, thanks. I think @katexochen should take a look, and then we can add tests and doc comments.

sdk/common.go
burgerdev: Mostly because you need to propagate logger structs through your app, even if you did not mean to add logging to begin with. You can also argue that a library should communicate through its arguments and returns, without side effects. This also goes for the cache dir: the `go-sev-guest` library did the right thing by accepting an `HttpsGetter` interface - then the side effects are with the caller.

Otoh I've been annoyed more than once by a library that did not expose logging but logged anyway in a different style.

We could also do a `New()` and `NewWithSLog(*slog.Logger)`.

sdk/verify.go
derpsteb: We talked about this on the phone. Right now I don't see a good argument for putting the endpoint into the client. To me the endpoint seems as much or as little sth that should be provided on each call as e.g. the manifest. It also seemed closer to the CLI behavior. One provides the endpoint on each command, instead of writing it into a config file.

If more people feel like `endpoint` should be a Client property I will change it.

sdk/common.go
3u13r: > Mostly because you need to propagate logger structs through your app, even if you did not mean to add logging to begin with.

I'd very much want the option to configure logging for libraries. If the consumer doesn't want it, they can pass a logger that discards the output. Related discussion: https://go-review.googlesource.com/c/go/+/626486 https://github.com/golang/go/issues/62005

sdk/verify.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/edgelesssys/contrast/internal/userapi"
+)
+
+// Client is used to interact with a Contrast deployment.
+type Client struct {
+	log *slog.Logger
+}
+
+// New returns a Client with a default logger.
+// Configures logging via slog to write to stderr.
```
burgerdev, line 27: I'd prefer a discarding handler as default then, like @3u13r suggested.

burgerdev (approved):
lgtm

sdk/README.md
```diff
@@ -0,0 +1,6 @@
+# Contrast SDK
+
+**Caution:** This SDK is still under active development and not fit for external use yet.
+Please expect breaking changes with new minor versions.
```
burgerdev, line 4: One idea I found while searching for ways to highlight instability of APIs: we could have an expressive build tag on the sdk, like `// +build contrast-unstable-api`, to prevent accidental usage.

packages/kata-debug-shell.sh
derpsteb, line 1: @msanft the linter was complaining about variables not beeing quoted here. you touched these lines the last time. are these changes correct, since there is an actual glob in the config_file path?

sdk/README.md
derpsteb: Asked for another review just to be sure, as the tag has to be added to quite a few places.

packages/kata-debug-shell.sh
burgerdev: I believe @katexochen had a fix for these.

katexochen: should be fixed on main

sdk/README.md
derpsteb: I missed adding the `tags` flag to the `go test` command of the checkPhase in the nix build. Now the tag is only in the sdk's go files. Thanks for pointing it out.

sdk/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The SDK allows writing programs that interact with a Contrast deployment like the CLI does, without relying on the CLI.
+
+# Building
+
+To build the SDK please add the build tag `contrast_unstable_api` to the importing module's files and build commands.
+This tag will be removed once the SDK is officially removed.
```
burgerdev, line 11:
```suggestion
If you decide to use the unstable API and accept the risk of breakage, you need to set the Go build tag `contrast_unstable_api`.
```

burgerdev (approved):
I think we're getting there. Maybe now is a good time to squash the commits?

PR #1031, opened 2024-11-25
e2e: fix policy test dependence on default policy test

jmxnzo:
This fixes a bug introduced into the policy test in PR #1007, because of dependence of the policy test on the default policy test executed prior. Restarting the applied resources cleans up the containers being in FailedToCreatePodSandbox state, which takes place when started with the default policy annotation and thus allows to continue the test workflow with the generated manifest.

### DIFF
--- a/e2e/policy/policy_test.go
+++ b/e2e/policy/policy_test.go
@@ -77,7 +77,7 @@ func TestPolicy(t *testing.T) {
 
 		t.Log("Restarting container")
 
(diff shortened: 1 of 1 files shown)


burgerdev (approved):
LGTM, thanks.

burgerdev:
(removed the `bug fix` label, as it's for user facing bugs and this bug is not user facing)

PR #1033, opened 2024-11-26
cli: pass environment variables to genpolicy

burgerdev:
This allows 

* running `contrast generate` with `HTTP_PROXY` and friends
* using registry credentials from various filesystem paths supported by the `docker_credential` crate (this is not enough for full private registry support!)

### DIFF
--- a/cli/genpolicy/genpolicy.go
+++ b/cli/genpolicy/genpolicy.go
@@ -60,7 +60,13 @@ func (r *Runner) Run(ctx context.Context, yamlPath string, logger *slog.Logger) 
 		"--yaml-file=" + yamlPath,
 	}
 	genpolicy := exec.CommandContext(ctx, r.genpolicy.Path(), args...)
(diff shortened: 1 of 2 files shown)


cli/genpolicy/genpolicy.go
```diff
@@ -74,6 +74,20 @@ func (r *Runner) Run(ctx context.Context, yamlPath string, logger *slog.Logger)
 	return nil
 }
 
+func env() []string {
+	env := []string{"RUST_LOG=info", "RUST_BACKTRACE=1"}
```
3u13r, line 78: I'm wondering if we should just pass `os.Environ()? This way we would also support other config Envs like `HTTP_PROXY` should genpolicy use it at some point.
We can still have default values for `RUST_LOG` and `RUST_BACKTRACE` though.

burgerdev: Done

3u13r (approved):
LGTM

PR #1034, opened 2024-11-26
nixos-img: disable patchelf on executable-marked OS image

jmxnzo:
Prior, we tried to run `patchelf` (automatically, due to the Nix stdenv) on the executable-marked OS image. This is not necessary and costs a considerable investment of time. The PR excludes the automatic patchelf hook from the build.

### DIFF
--- a/packages/by-name/microsoft/kata-igvm/package.nix
+++ b/packages/by-name/microsoft/kata-igvm/package.nix
@@ -57,6 +57,8 @@ let
       '';
     };
 
+    dontPatchELF = true;
+
     meta = {
       description = "The Contrast runtime IGVM file defines the initial state of a pod-VM.";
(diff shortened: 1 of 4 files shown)


packages/by-name/microsoft/kata-igvm/package.nix
```diff
@@ -55,7 +55,10 @@ let
       buildPhase = ''
         igvmmeasure ${igvm} measure -b | dd conv=lcase > $out
       '';
+      dontPatchELF = true;
```
msanft, line 58:
```suggestion
```
This one is superfluous and can be removed

msanft (commented):
LGTM besides the comment.

The check that fails is only formatting. I think you should be able to run `just fmt`, which should fix this.

PR #1035, opened 2024-11-27
caa: use our own image

burgerdev:
Adjust the upstream cloud-api-adaptor DaemonSet to the idiosyncracies of our Nix-built container image, and vice-versa.

* Set the HOME env var so that the app can find the SSH key.
* Add a mount point for `/var/run/netns` because our image does not symlink `/var/run` to `/run`.
* Add a stable symlink to the real entrypoint at the expected location.

Furthermore, add the required iptables binary as a runtime dependency to cloud-api-adaptor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,5 @@ id_rsa*
 kube.conf
 out.env
 infra/**/kustomization.yaml
+infra/**/patch_ds.yaml
 uplosi.conf*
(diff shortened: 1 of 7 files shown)


burgerdev:
Sorry, I'm gonna withdraw this for now because I discovered that it does not work well with the reverted node-installer changes.

PR #1036, opened 2024-11-27
sdk: remove Linux dependencies

elchead:
## Context

It is currently not possible to build the SDK / CLI on platforms other than Linux. One of the issues is that the imported packages include Linux specific code.

These components don't require it, so they shouldn't depend on it to allow for cross-platform usage.

[AB#4959](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4959)

## Proposed changes
- move the issuer code into its own sub-package
- only let the components that require it (coordinator and initializer) import the issuer package

### DIFF
--- a/coordinator/main.go
+++ b/coordinator/main.go
@@ -16,6 +16,7 @@ import (
 	"github.com/edgelesssys/contrast/coordinator/history"
 	"github.com/edgelesssys/contrast/coordinator/internal/authority"
 	"github.com/edgelesssys/contrast/internal/atls"
+	"github.com/edgelesssys/contrast/internal/atls/issuer"
 	"github.com/edgelesssys/contrast/internal/grpc/atlscredentials"
 	"github.com/edgelesssys/contrast/internal/logger"
 	"github.com/edgelesssys/contrast/internal/meshapi"
@@ -154,7 +155,7 @@ func newServerMetrics(reg *prometheus.Registry) *grpcprometheus.ServerMetrics {
 }
 
 func newGRPCServer(serverMetrics *grpcprometheus.ServerMetrics, log *slog.Logger) (*grpc.Server, error) {
-	issuer, err := atls.PlatformIssuer(log)
+	issuer, err := issuer.PlatformIssuer(log)
 	if err != nil {
 		return nil, fmt.Errorf("creating issuer: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


coordinator/main.go
```diff
@@ -154,7 +155,7 @@ func newServerMetrics(reg *prometheus.Registry) *grpcprometheus.ServerMetrics {
 }
 
 func newGRPCServer(serverMetrics *grpcprometheus.ServerMetrics, log *slog.Logger) (*grpc.Server, error) {
-	issuer, err := atls.PlatformIssuer(log)
+	issuer, err := issuer.PlatformIssuer(log)
```
katexochen, line 158: This is now stuttering.

internal/atls/issuer/issuer.go
```diff
@@ -0,0 +1,44 @@
...
+		return tdxissuer.New(
+			logger.NewWithAttrs(logger.NewNamed(log, "issuer"), map[string]string{"tee-type": "tdx"}),
+		), nil
+	default:
+		return nil, fmt.Errorf("unsupported platform: %T", cpuid.CPU)
+	}
+}
+
+// Issuer issues an attestation document.
+type Issuer interface {
```
burgerdev, line 36: The interface is also declared here: https://github.com/edgelesssys/contrast/blob/c75797e083533b282f3c19082e582e02dfd0bb82/internal/atls/atls.go#L88. I think we should only have one - here - which requires inlining the `Getter`.

internal/atls/issuer/issuer_linux.go
```diff
@@ -1,22 +1,23 @@
+//go:build linux
+
 // Copyright 2024 Edgeless Systems GmbH
 // SPDX-License-Identifier: AGPL-3.0-only
```
burgerdev, line 4:
```suggestion
// Copyright 2024 Edgeless Systems GmbH
// SPDX-License-Identifier: AGPL-3.0-only

//go:build linux
```

internal/atls/issuer/issuer_universal.go
```diff
@@ -0,0 +1,17 @@
+//go:build !linux
+
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
```
burgerdev, line 4:
```suggestion
// Copyright 2024 Edgeless Systems GmbH
// SPDX-License-Identifier: AGPL-3.0-only

//go:build !linux
```

burgerdev (approved):
lgtm

PR #1037, opened 2024-11-28
userapi: rename grpc package and change relative proto path

elchead:
As Contrast is migrating towards providing an SDK, the userapi should adhere to best practices and avoid compatibility issues.
As part of integrating the current SDK draft into Continuum, two issues were discovered.

1. The gRPC package name is not globally unique (see https://protobuf.dev/reference/go/faq#namespace-conflict), which leads to conflicts when the third-party project imports other gRPC packages with the same name.
2. The protoc compiler currently takes in a source path of the form "userapi.proto". This leads to import issues when another package imports a proto file with the same path. This happened in the Continuum project:

```
panic: proto: file "userapi.proto" is already registered
                previously from: "github.com/edgelesssys/continuum/internal/proto/attestation-service/userapi"
                currently from:  "github.com/edgelesssys/contrast/internal/userapi"
        See https://protobuf.dev/reference/go/faq#namespace-conflict
``` 

A solution was suggested [here](https://stackoverflow.com/questions/67693170/proto-file-is-already-registered-with-different-packages/78560307#78560307) which ensures that the path names are unlikely to collide. Hence I moved the protoc generation to include the proto file path relative to the root.


burgerdev (approved):
lgtm, breaking the internal gRPC layer is not user-visible right now.

PR #1043, opened 2024-11-29
caa: support peerpods in node-installer (again)

burgerdev:
This reapplies most of the commits that moved CAA into the node-installer. I've split the commits into the clean reapplies and dedicated `fixup` commits where I found problems in the original commits. These include:

* Ensure that the agent config is not skipped while serializing.
* Add missing environment variable and RBAC on the CAA container.
* Include `iptables` in the CAA container image.
* Deploy RBAC resources from the `apply-runtime` target, which is the point when we know the namespace to apply them to.

The last commit adapts the peer-pods e2e test to the prior changes, which means that the peer-pods workflow is not expected to work for the last 3 commits or so. We could squash this PR, but I think I'd prefer to have the history intact.

### DIFF
--- a/.github/workflows/e2e_peerpods.yml
+++ b/.github/workflows/e2e_peerpods.yml
@@ -18,8 +18,17 @@ on:
 jobs:
   test:
     runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+      packages: write
     steps:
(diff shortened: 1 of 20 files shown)


katexochen:
> We could squash this PR, but I think I'd prefer to have the history intact.

You mean you want to keep the fixup commits?

Changes lgtm, didn't test yet.

katexochen (approved):

PR #1044, opened 2024-12-02
cli: genpolicy logging: Add debug log level and repository reference to auth failure

jmxnzo:
The PR includes the debug log level of genpolicy into the Contrast CLI. Thus the resulting mapping is:

Contrast info == Genpolicy warning
Contrast debug == Genpolicy info/debug

As well it patches the existing genpolicy error message in case of an authentication failure, to include the reference to the repository.

### DIFF
--- a/cli/genpolicy/genpolicy.go
+++ b/cli/genpolicy/genpolicy.go
@@ -62,7 +62,7 @@ func (r *Runner) Run(ctx context.Context, yamlPath string, logger *slog.Logger) 
 	genpolicy := exec.CommandContext(ctx, r.genpolicy.Path(), args...)
 	genpolicy.Env = os.Environ()
 	if _, hasRustLog := os.LookupEnv("RUST_LOG"); !hasRustLog {
-		genpolicy.Env = append(genpolicy.Env, "RUST_LOG=info")
+		genpolicy.Env = append(genpolicy.Env, "RUST_LOG=debug")
 	}
 	if _, hasRustBacktrace := os.LookupEnv("RUST_BACKTRACE"); !hasRustBacktrace {
 		genpolicy.Env = append(genpolicy.Env, "RUST_BACKTRACE=1")
--- a/packages/by-name/microsoft/genpolicy/package.nix
+++ b/packages/by-name/microsoft/genpolicy/package.nix
@@ -59,6 +59,13 @@ rustPlatform.buildRustPackage rec {
       # to Microsofts genpolicy.
       # TODO(miampf): remove when picked up by microsoft/kata-containers fork.
       ./0007-genpolicy-support-for-VOLUME-definition-in-container.patch
+
+      # Simple genpolicy logging patch to include the image reference in case of authentication failure
+      # TODO(jmxnzo): remove when authentication failure error logging includes image reference on microsoft/kata-containers fork.
+      # This will be achieved when updating oci_distribution to oci_client crate on microsoft/kata-containers fork.
+      # kata/kata-runtime/0011-genpolicy-bump-oci-distribution-to-v0.12.0.patch introduces this update to kata-containers.
+      # After upstreaming, microsoft/kata-containers fork would need to pick up the changes.
+      ./0008-genpolicy-include-reference-in-logs-when-auth-failure.patch
     ];
   };
 
--- a/cli/genpolicy/logtranslator.go
+++ b/cli/genpolicy/logtranslator.go
@@ -51,7 +51,9 @@ func (l logTranslator) startTranslate() {
 					l.logger.Error(match[3], "position", match[2])
 				case "WARN":
 					l.logger.Warn(match[3], "position", match[2])
-				case "INFO": // prints quite a lot, only show on debug
+				case "INFO":
+					fallthrough // prints quite a lot, only show on debug
+				case "DEBUG":
 					l.logger.Debug(match[3], "position", match[2])
 				}
 			}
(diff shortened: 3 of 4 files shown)


packages/by-name/kata/kata-runtime/0017-Include-reference-in-logs-when-auth-failure.patch
```diff
@@ -0,0 +1,22 @@
...
+diff --git a/src/tools/genpolicy/src/registry.rs b/src/tools/genpolicy/src/registry.rs
+index 9491b2acfb8af1f8c4e4c971a6787c2d803e46ec..5815cff89e4ddcfe94f0bc2c7d44263efda20311 100644
+--- a/src/tools/genpolicy/src/registry.rs
++++ b/src/tools/genpolicy/src/registry.rs
+@@ -159,7 +159,7 @@ impl Container {
+                 })
+             }
+             Err(oci_client::errors::OciDistributionError::AuthenticationFailure(message)) => {
+-                panic!("Container image registry authentication failure ({}). Are docker credentials set-up for current user?", &message);
++                panic!("Container image registry authentication failure ({}) for {:#?}. Are docker credentials set-up for current user?", &message, &reference);
```
burgerdev, line 19: `{:#?}` is probably a debug print that shows the internal structure of https://github.com/youki-dev/oci-spec-rs/blob/e1dded7b5ff2d448d8802e2fe9202c86eb6f6c30/src/distribution/reference.rs#L77. Can we get a readable image reference string here?

packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -93,6 +93,7 @@ buildGoModule rec {
       ./0014-kata-sys-util-remove-obsolete-cgroups-dependency.patch
       ./0015-kata-sys-util-move-json-parsing-to-protocols-crate.patch
       ./0016-protocols-only-build-RLimit-impls-on-Linux.patch
+      ./0017-Include-reference-in-logs-when-auth-failure.patch
```
katexochen, line 96: comment missing, please see last sec in dev-docs/patches.md!

packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -93,6 +93,9 @@ buildGoModule rec {
       ./0014-kata-sys-util-remove-obsolete-cgroups-dependency.patch
       ./0015-kata-sys-util-move-json-parsing-to-protocols-crate.patch
       ./0016-protocols-only-build-RLimit-impls-on-Linux.patch
+
+      # Simple genpolicy logging patch to include the repository reference in case of authentication failure
```
katexochen, line 97: This does not answer all the requested questions.

katexochen:
I guess it should also be possible to upstream that patch?

packages/by-name/microsoft/genpolicy/package.nix
```diff
@@ -57,8 +57,12 @@ rustPlatform.buildRustPackage rec {
       ./0006-genpolicy-support-HostToContainer-mount-propagation.patch
       # This patch is a port of https://github.com/kata-containers/kata-containers/pull/10136/files
       # to Microsofts genpolicy.
-      # TODO(miampf): remove when picked up by microsoft/kata-containers fork.
+      # TODO(miampf):
```
katexochen, line 60: why change this?

packages/by-name/microsoft/genpolicy/package.nix
```diff
@@ -57,8 +57,12 @@ rustPlatform.buildRustPackage rec {
...
       ./0007-genpolicy-support-for-VOLUME-definition-in-container.patch
+
+      # Simple genpolicy logging patch to include the image reference in case of authentication failure
+      # Not merged, TODO(jmxnzo): remove when picked up by microsoft/kata-containers fork
```
katexochen, line 64: How should this be picked up by the microsoft fork if we aren't proposing it upstream? Or what exactly must be picked up so this can be dropped?

cli/genpolicy/genpolicy.go
```diff
@@ -62,7 +62,7 @@ func (r *Runner) Run(ctx context.Context, yamlPath string, logger *slog.Logger)
 	genpolicy := exec.CommandContext(ctx, r.genpolicy.Path(), args...)
 	genpolicy.Env = os.Environ()
 	if _, hasRustLog := os.LookupEnv("RUST_LOG"); !hasRustLog {
-		genpolicy.Env = append(genpolicy.Env, "RUST_LOG=info")
+		genpolicy.Env = append(genpolicy.Env, "RUST_LOG=debug")
```
katexochen, line 65: Do we really need to change the default here?

katexochen (commented):
commit messages should be:

cli/genpolicy: include log-level 'debug' in Contrast logs
microsoft.genpolicy: log image reference on unauthorized access

packages/by-name/microsoft/genpolicy/package.nix
jmxnzo: It was by accident, when copy-pasting. thanks for pointing out!

packages/by-name/microsoft/genpolicy/package.nix
jmxnzo: It was by accident, when copy-pasting. thanks for pointing out!

burgerdev (approved):
lgtm

burgerdev (approved):
Thanks, I think the message is clear now!

katexochen (approved):

PR #1045, opened 2024-12-02
docs: add demo for workload secrets

davidweisse:
This adds a demo application running a MySQL server with an encrypted volume mount, similar to the `volume-tester`. For this, the `cryptsetup` init container uses the workload secret to set up a LUKS partition that is mounted to `/var/lib/mysql`. The demo is also provided as a release artifact as `mysql-demo.yml`.

### DIFF
--- a/docs/docs/examples/mysql.md
+++ b/docs/docs/examples/mysql.md
@@ -1,44 +1,43 @@
-# Confidential emoji voting
+# Encrypted volume mount
 
-<!-- TODO(katexochen): create a screenshot with fixed format -->
+**This tutorial guides you through deploying a simple application with an
+encrypted MySQL database using the Contrast [workload
+secret](../architecture/secrets.md#workload-secrets).**
 
-![screenshot of the emojivoto UI](../_media/emoijvoto.png)
+[MySQL](https://mysql.com) is an open-source database used to organize data into
+tables and quickly retrieve information about its content. All of the data in a
+MySQL database is stored in the `/var/lib/mysql` directory. In this example, we
+use the workload secret to setup an encrypted LUKS mount for the
+`/var/lib/mysql` directory to easily deploy an application with encrypted
+persistent storage using Contrast.
 
-**This tutorial guides you through deploying [emojivoto](https://github.com/BuoyantIO/emojivoto) as a
-confidential Contrast deployment and validating the deployment from a voter's perspective.**
-
-Emojivoto is an example app allowing users to vote for different emojis and view votes
-on a leader board. It has a microservice architecture consisting of a
-web frontend (`web`), a gRPC backend for listing available emojis (`emoji`), and a backend for
-the voting and leader board logic (`voting`). The `vote-bot` simulates user traffic by submitting
-votes to the frontend.
-
-Emojivoto can be seen as a lighthearted example of an app dealing with sensitive data.
-Contrast protects emojivoto in two ways. First, it shields emojivoto as a whole from the infrastructure, for example, Azure.
-Second, it can be configured to also prevent data access even from the administrator of the app. In the case of emojivoto, this gives assurance to users that their votes remain secret.
-
-<!-- TODO(katexochen): recreate in our design -->
-
-![emojivoto components topology](https://raw.githubusercontent.com/BuoyantIO/emojivoto/e490d5789086e75933a474b22f9723fbfa0b29ba/assets/emojivoto-topology.png)
+The resources provided in this demo are designed for educational purposes and
+shouldn't be used in a production environment without proper evaluation. When
+working with persistent storage, regular backups are recommended in order to
+prevent data loss. For confidential applications, please also refer to the
+[security considerations](../architecture/security-considerations.md). Also be
+aware of the differences in security implications of the workload secrets for
+the data owner and the workload owner. For more details, see the [Workload
+Secrets](../architecture/secrets.md#workload-secrets) documentation.
 
 ## Prerequisites
 
 - Installed Contrast CLI
-- A running Kubernetes cluster with support for confidential containers, either on [AKS](../getting-started/cluster-setup.md) or on [bare metal](../getting-started/bare-metal.md).
+- A running Kubernetes cluster with support for confidential containers, either on [AKS](../getting-started/cluster-setup.md) or on [bare metal](../getting-started/bare-metal.md)
 
-## Steps to deploy emojivoto with Contrast
+## Steps to deploy MySQL with Contrast
 
 ### Download the deployment files
 
-The emojivoto deployment files are part of the Contrast release. You can download them by running:
+The MySQL deployment files are part of the Contrast release. You can download them by running:
 
 ```sh
-curl -fLO https://github.com/edgelesssys/contrast/releases/latest/download/emojivoto-demo.yml --create-dirs --output-dir deployment
+curl -fLO https://github.com/edgelesssys/contrast/releases/latest/download/mysql-demo.yml --create-dirs --output-dir deployment
 ```
 
 ### Deploy the Contrast runtime
 
-Contrast depends on a [custom Kubernetes RuntimeClass](../components/runtime.md),
+Contrast depends on a [custom Kubernetes `RuntimeClass`](../components/runtime.md),
 which needs to be installed to the cluster initially.
 This consists of a `RuntimeClass` resource and a `DaemonSet` that performs installation on worker nodes.
 This step is only required once for each version of the runtime.
@@ -65,7 +64,7 @@ kubectl apply -f https://github.com/edgelesssys/contrast/releases/latest/downloa
 ### Deploy the Contrast Coordinator
 
 Deploy the Contrast Coordinator, comprising a single replica deployment and a
-LoadBalancer service, into your cluster:
+`LoadBalancer` service, into your cluster:
 
 <Tabs queryString="platform">
 <TabItem value="aks-clh-snp" label="AKS" default>
@@ -125,11 +124,11 @@ The deployment YAML shipped for this demo is already configured to be used with 
 A [runtime class](../components/runtime) `contrast-cc`
 was added to the pods to signal they should be run as Confidential Containers. During the generation process,
 the Contrast [Initializer](../components/overview.md#the-initializer) will be added as an init container to these
-workloads to facilitate the attestation and certificate pulling before the actual workload is started.
+workloads. It will attest the pod to the Coordinator and fetch the workload certificates and the workload secret.
 
 Further, the deployment YAML is also configured with the Contrast [service mesh](../components/service-mesh.md).
 The configured service mesh proxy provides transparent protection for the communication between
-the different components of emojivoto.
+the MySQL server and client.
 :::
 
 ### Set the manifest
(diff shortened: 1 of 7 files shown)


docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,251 @@
...
+```sh
+kubectl apply -f deployment/
+```
+
+:::note[Persistent workload secrets]
+
+During the initialization process of the workload pod, the Contrast Initializer
+sends an attestation report to the Coordinator and retrieves a randomly
+generated workload secret, seeded with the workload secret ID specified in the
+manifest, and writes it to a `volumeMount`.
```
burgerdev, line 156: I think we should emphasize here that this is a secure-memory-backed mount.

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,251 @@
...
+
+To access the MySQL database from the command line, expose the service on a public IP address via a LoadBalancer service:
+
+```sh
+mysqlIP=$(kubectl get svc mysql-svc -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')
+echo "MySQL is available at the IP  $mysqlIP."
+```
+
+Using the `mysql` command line interface, you can connect to the database as the
+root user with the password specified in the deployment YAML (`password`).
```
burgerdev, line 213: I think this setup does not sound ideal, even for a demo. Could we maybe mention the service mesh here to ensure only Contrast deployments can access the MySQL server?

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,251 @@
...
+:::
+
+## Updating the deployment
+
+Because the workload secret is bound to the workload itself and not to an
+individual pod, once the pod restarts, the `cryptsetup` init container can
+deterministically generate the same key again and open the already partitioned
+LUKS device.
+
+For example, after making changes to the deployment files, the runtime policies
```
burgerdev, line 233: The advice here is not correct. If you change the deployment, you're likely to get a different policy hash, resulting in a different workload secret. This is where the [`WorkloadSecretID`](https://github.com/edgelesssys/contrast/blob/55cb03624be5271088b145fbc4a547cad096afbf/internal/manifest/manifest.go#L36) comes in, which may be, well, underdocumented right now.

docs/docs/examples/mysql.md
burgerdev, line 1: Do we need a disclaimer here, mentioning:

1. Risk of data loss without backups.
2. Workload owner vs data owner concerns.
3. Security concerns with unauthenticated encryption, PBKDF2 defaults, etc.

internal/kuberesource/sets.go
```diff
@@ -627,3 +627,159 @@ sleep inf
 
 	return []any{vss}
 }
+
+// MySQL returns the resources for deploying a MySQL database
+// with an encrypted luks volume using the workload-secret.
+func MySQL() []any {
+	initCommand := `#!/bin/bash
+# cryptsetup complains if there is no /run directory.
+mkdir /run
```
burgerdev, line 636: Consider adding an `emptyDir` mount instead.

internal/kuberesource/sets.go
```diff
@@ -627,3 +627,159 @@ sleep inf
...
+	cryptsetup luksChangeKey --pbkdf-memory=10240 "${device}" --key-file "${tmp_key_path}" "${disk_encryption_key_path}"
+	cryptsetup open "${device}" state -d "${disk_encryption_key_path}"
+	mkfs.ext4 /dev/mapper/state
+	cryptsetup close state
+fi
+
+# No matter if this is the first initialization derive the key (again) and open the device.
+uuid=$(blkid "${device}" -s UUID -o value)
+openssl kdf -keylen 32 -kdfopt digest:SHA2-256 -kdfopt key:$(cat "${workload_secret_path}") \
+	-kdfopt info:${uuid} -binary HKDF | base64 | tr -d '\n' > "${disk_encryption_key_path}"
```
burgerdev, line 666:
```suggestion
	-kdfopt info:${uuid} -binary HKDF | base64 -w0 > "${disk_encryption_key_path}"
```

internal/kuberesource/sets.go
```diff
@@ -627,3 +627,159 @@ sleep inf
...
+cryptsetup luksUUID "${device}"
+cryptsetup open "${device}" state -d "${disk_encryption_key_path}"
+mkdir -p /data
```
burgerdev, line 669: Leftover?

internal/kuberesource/sets.go
```diff
@@ -627,3 +627,159 @@ sleep inf
...
+								).
+								WithResources(ResourceRequirements().
+									WithMemoryLimitAndRequest(100),
+								).
+								WithStartupProbe(
+									Probe().
+										WithFailureThreshold(20).
+										WithPeriodSeconds(5).
+										WithExec(applycorev1.ExecAction().
+											WithCommand("/bin/sh", "-c", "cat /done"),
```
burgerdev, line 720:
```suggestion
											WithCommand("/bin/test", "-f", "/done"),
```
... might be clearer.

burgerdev (commented):
Great!

docs/docs/examples/mysql.md
3u13r: Although, we might want those disclaimers in the general architecture/feature description and likely only refer to them here, right?

internal/kuberesource/sets.go
3u13r: We also already have the same(?) script in https://github.com/edgelesssys/contrast/blob/main/internal/kuberesource/sets.go#L504. We likely want to build a function returning the string (maybe add some templating if needed) and use it in both places.

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,251 @@
...
+They will have to be filled in manually.
+If you don't know the correct values use `ffffffffffffffffffffffffffffffff` and `000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000` respectively and observe the real values in the error messages in the following steps. This should only be done in a secure environment.
+:::
+</TabItem>
+</Tabs>
+
+:::note[Runtime class and Initializer]
+
+The deployment YAML shipped for this demo is already configured to be used with Contrast.
+A [runtime class](https://docs.edgeless.systems/contrast/components/runtime) `contrast-cc-<platform>-<runtime-hash>`
```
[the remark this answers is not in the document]
davidweisse: This should also be patched for the Emojivoto docs.

docs/docs/examples/mysql.md
burgerdev: Sorry, I was mistaken about the workload secret seed. Maybe change 

> the workload secret is bound to the workload itself

to something like

> the workload secret is derived from the resource name

and link to the architecture page for more details. Those details are missing right now, but should be added asap.

internal/kuberesource/sets.go
```diff
@@ -627,3 +627,159 @@ sleep inf
...
+										),
+								).
+								WithRestartPolicy(
+									corev1.ContainerRestartPolicyAlways,
+								),
+						).
+						WithContainers(
+							Container().
+								WithName("sqlite").
```
burgerdev, line 729:
```suggestion
								WithName("mysql").
```

internal/kuberesource/sets.go
```diff
@@ -627,3 +627,159 @@ sleep inf
...
+								WithImage("docker.io/mysql:latest").
```
burgerdev, line 730:
```suggestion
								WithImage("docker.io/library/mysql:latest").
```

internal/kuberesource/sets.go
```diff
@@ -627,3 +627,159 @@ sleep inf
...
+
+# If the device is not already a LUKS device, format it.
+if ! cryptsetup isLuks "${device}"; then
+	# Generate a random key for the first initialization.
+	dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 | tr -d '\n' > "${tmp_key_path}"
+	cryptsetup luksFormat --pbkdf-memory=10240 $device "${tmp_key_path}" </dev/null
+
+	# Now we can get the LUKS UUID and deterministically derive the disk encryption key from the workload secret.
+	uuid=$(blkid "${device}" -s UUID -o value)
+	openssl kdf -keylen 32 -kdfopt digest:SHA2-256 -kdfopt key:$(cat "${workload_secret_path}") \
```
burgerdev, line 655: Can we use a key file instead of passing the key as arg? I just dumped it into logs accidentally by setting `-x`.

davidweisse: `openssl kdf` only supports a key from the command line, but I don't think it's a big problem. I noticed though, that the option `hexkey` is probably more fitting here, since the workload secret is stored in hex.

docs/docs/examples/mysql.md
burgerdev: - (1) is a _safety_ concern that's specific to persistency and (3) is a security concern specific to this demo. I would put at least a small note here to remind people that they're looking at a demo, not a prod reference architecture. 
- (2) is described on the workload secrets docs and should be linked to from here.

internal/kuberesource/sets.go
burgerdev: That's unfortunate, but further drives my point that we should highlight that this is demo code. 

(Not sure whether it was clear in my original comment: my concern is that the secrets end up in logs, which are readable by the infrastructure owner. _This is not the case in your code_, but a simple change like putting `-x` at the top is enough to let it happen, so I'd consider it a bit dangerous)

docs/docs/examples/mysql.md
burgerdev: I like the disclaimer!

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,267 @@
...
+
+The MySQL deployment files are part of the Contrast release. You can download them by running:
+
+```sh
+curl -fLO https://github.com/edgelesssys/contrast/releases/latest/download/mysql-demo.yml --create-dirs --output-dir deployment
+```
+
+### Deploy the Contrast runtime
+
+Contrast depends on a [custom Kubernetes RuntimeClass](../components/runtime.md),
```
burgerdev, line 41:
```suggestion
Contrast depends on a [custom Kubernetes `RuntimeClass`](../components/runtime.md),
```

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,267 @@
...
+```sh
+kubectl apply -f https://github.com/edgelesssys/contrast/releases/latest/download/runtime-k3s-qemu-tdx.yml
+```
+</TabItem>
+</Tabs>
+
+### Deploy the Contrast Coordinator
+
+Deploy the Contrast Coordinator, comprising a single replica deployment and a
+LoadBalancer service, into your cluster:
```
burgerdev, line 68:
```suggestion
`LoadBalancer` service, into your cluster:
```

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,267 @@
...
+
+```sh
+kubectl apply -f deployment/
+```
+
+:::note[Persistent workload secrets]
+
+During the initialization process of the workload pod, the Contrast Initializer
+sends an attestation report to the Coordinator and retrieves a randomly
+generated workload secret, seeded with the workload secret ID specified in the
```
burgerdev, line 164:
```suggestion
sends an attestation report to the Coordinator and receives a workload secret
derived from the Coordinator's secret seed and the workload secret ID specified in the
```

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,267 @@
...
+Using the `mysql` command line interface, you can connect to the database as the
+root user with the password specified in the deployment YAML (`password`).
+
+```sh
+mysql -h $mysqlIP -u root -p
+```
+
+:::warning
+Avoid including the MySQL root user password directly in the Kubernetes
+configuraiton files to prevent exposing sensitive credentials. Use [Kubernetes
```
burgerdev, line 230:
```suggestion
configuration files to prevent exposing sensitive credentials. Use [Kubernetes
```

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,267 @@
...
+Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) in
+production to securely manage and reference the password instead.
```
burgerdev, line 232: Having the password in any yaml, secret or deployment, leaks it to the infrastructure. Without the service mesh, the DB is exposed and the encryption is moot. If we can't force mTLS in MySQL, we should only advertise it with the service mesh!

internal/kuberesource/parts.go
```diff
@@ -473,3 +473,44 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
...
+
+# No matter if this is the first initialization derive the key (again) and open the device.
+uuid=$(blkid "${device}" -s UUID -o value)
+openssl kdf -keylen 32 -kdfopt digest:SHA2-256 -kdfopt hexkey:$(cat "${workload_secret_path}") \
+	-kdfopt info:${uuid} -binary HKDF | base64 -w0 > "${disk_encryption_key_path}" 2>/dev/null
+cryptsetup luksUUID "${device}"
+cryptsetup open "${device}" state -d "${disk_encryption_key_path}"
+mount /dev/mapper/state /state
+touch /done
+sleep inf
```
burgerdev, line 514: Does this container need to stay around? If yes, we should maybe document why. If no, let's make it a pure init container and exit here.

burgerdev (commented):
Overall LGTM, but we need to deal with the password before merging.

m1ghtym0:
Awesome!
I like the idea of having more examples like this in the mid-term future.
One thought: Should we label them so that they are easier to "find" or people know immediately what they are looking at?
* Instead of "Confidential emoji voting" -> "Simple demo app"
* Instead of "Encrypted volume mount" -> "MySQL"

We then might want to have a dedicated "Tasks" section in the docs that goes more in-depth with certain operations such as "Mount encrypted volumes, "recover the coordinator", ...
We have these things hidden in the architecture and components section to some extend.

Let me know what you think:-)

burgerdev:
We could also consider the https://diataxis.fr/ approach for structuring. I'd defer this discussion for the PR here, though.

internal/kuberesource/parts.go
davidweisse: The `cryptsetup` container is responsible for opening the LUKS mount, so if it stops, the encrypted state is closed and the `mysql` container cannot access it. I'll mention it in the docs.

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,260 @@
...
+prevent data loss. For confidential applications, please also refer to the
+[security considerations](../architecture/security-considerations.md). Also be
+aware of the differences in security implications of the workload secrets for
+the data owner and the workload owner. For more details, see the [Workload
+Secrets](../architecture/secrets.md#workload-secrets) documentation.
+
+## Prerequisites
+
+- Installed Contrast CLI
+- A running Kubernetes cluster with support for confidential containers, either on [AKS](../getting-started/cluster-setup.md) or on [bare metal](../getting-started/bare-metal.md).
```
burgerdev, line 26:
```suggestion
- A running Kubernetes cluster with support for confidential containers, either on [AKS](../getting-started/cluster-setup.md) or on [bare metal](../getting-started/bare-metal.md)
```

For consistency, no punctuation for both bullets (which are not full sentences).

docs/docs/examples/mysql.md
```diff
@@ -0,0 +1,260 @@
...
+### Connecting to the application
+
+Other confidential containers can securely connect to the MySQL server via the
+[Service Mesh](../components/service-mesh.md). The configured `mysql-client`
+deployment connects to the MySQL server and inserts test data into a table. To
+view the logs of the `mysql-client` deployment, use the following commands:
+
+```sh
+mysqlClient=$(kubectl get pods -l app.kubernetes.io/name=mysql-client -o=jsonpath='{.items[0].metadata.name}')
+kubectl logs ${mysqlClient}
```
burgerdev, line 225:
```suggestion
kubectl logs -l app.kubernetes.io/name=mysql-client -c mysql-backend
```
should work and is more concise.

burgerdev (approved):
lgtm, only few nits

PR #1046, opened 2024-12-03
e2e: fix log collection download

miampf:
Currently, downloading logs from an e2e cluster may fail since the cluster may be undeployed before the download is initiated. This PR fixes that problem by undeploying the cluster _after_ the logs were downloaded.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -100,6 +100,6 @@ jobs:
           message: "e2e test ${{ inputs.test-name }} failed"
           additionalFields: '[{"title": "Platform", "value": "${{ inputs.platform }}"}]'
       - name: Cleanup
-        if: cancelled() && !inputs.skip-undeploy
+        if: always() && !inputs.skip-undeploy
         run: |
-          kubectl delete ns "$(cat workspace/e2e.namespace)" --timeout 5m
+          kubectl delete ns "$(cat workspace/e2e.namespace)" --timeout 10m
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -79,7 +79,6 @@ jobs:
             --image-replacements workspace/just.containerlookup \
             --namespace-file workspace/e2e.namespace \
             --platform ${{ inputs.platform }} \
-            --skip-undeploy="${{ inputs.skip-undeploy && 'true' || 'false' }}" \
             --namespace-suffix="-ci"
       - name: Download logs
         if: always()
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e.yml
```diff
@@ -79,11 +79,15 @@ jobs:
             --image-replacements workspace/just.containerlookup \
             --namespace-file workspace/e2e.namespace \
             --platform ${{ inputs.platform }} \
-            --skip-undeploy="${{ inputs.skip-undeploy && 'true' || 'false' }}"
+            --skip-undeploy="true"
```
3u13r, line 82: Does this problem also affect https://github.com/edgelesssys/contrast/blob/f50e93b16f7cca03e0c8fd5aa70943567fe1d69c/.github/workflows/e2e_aks_runtime.yml#L90 ? 
If this is the case we need the same change over there. Also then the --skip-undeploy could be removed from the nix/Go side as it's always true (at least if no dev changes it locally in their justfile).

miampf: Oh yeah, I forgot to remove it from the AKS runtime test, will do. I'll also remove skip-undeploy from the go side if you think that nobody needs that control locally

burgerdev (commented):
Do you want to add the `default_platform` for the AKS runtime test in this PR?

miampf:
> Do you want to add the `default_platform` for the AKS runtime test in this PR?

Do you mean to allow selection between AKS-CLH-SNP and AKS-PEER-SNP? If yes, I can do that if you want :)

e2e/aks-runtime/aks_runtime_test.go
```diff
@@ -104,24 +104,10 @@ func TestAKSRuntime(t *testing.T) {
...
-		}
-	})
-
-	pods, err := c.PodsFromDeployment(ctx, namespace, testContainer)
+	pods, err := c.Client.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
 	require.NoError(err)
-	require.Len(pods, 1)
-	pod := pods[0]
+	require.GreaterOrEqual(len(pods.Items), 1)
+	pod := pods.Items[0] // only one pod was deployed
```
burgerdev, line 110: I think this is a bad merge - we deliberately changed the logic here in #1059.

burgerdev (commented):
After addressing the comments, please squash the commits (maybe it makes sense to have one for the flag removal and one for the test fix, but if it does not separate cleanly then 1 commit is fine).

burgerdev (approved):

PR #1050, opened 2024-12-03
kds-cache: add fallback cache for CRLs on request failure

jmxnzo:
This PR adds logic to the cached_client.go to allow caching Certificate Revocation Lists (CRLs). Still the CRLs are always queried from the AMD's certification server (KDS) first.
Only in case of an request failure, the cache is used as fallback to allow continuing. When the cache of the CRL is expired, the HTTPs request failure is returned as an error.

In addition the test cases where adapted to cover the different branches of VCEK and CRL caching.

### DIFF
--- a/internal/attestation/certcache/cached_client.go
+++ b/internal/attestation/certcache/cached_client.go
@@ -44,25 +44,33 @@ func (c *CachedHTTPSGetter) Get(url string) ([]byte, error) {
 	default:
 	}
 
-	// Don't cache CRLs. Unlike VCEKs, these can change over time and the KDS
-	// doesn't rate-limit requests to these.
-	canCache := !crlURL.MatchString(url)
-
-	if canCache {
+	if crlURL.MatchString(url) {
+		// For CRLs always query. When request failure, fallback to cache.
+		c.logger.Debug("Request CRL", "url", url)
+		res, err := c.HTTPSGetter.Get(url)
+		if err == nil {
+			c.cache.Set(url, res)
+			return res, nil
+		}
+		c.logger.Warn("Failed requesting CRL from KDS", "error", err)
 		if cached, ok := c.cache.Get(url); ok {
-			c.logger.Debug("Get cached", "url", url)
+			c.logger.Warn("Falling back to cached CRL", "url", url)
 			return cached, nil
 		}
+		c.logger.Warn("CRL not found in cache", "url", url)
+		return nil, err
 	}
-
-	c.logger.Debug("Get not cached", "url", url)
+	// For VCEK get cache first and request if not present
+	if cached, ok := c.cache.Get(url); ok {
+		c.logger.Debug("Cache hit", "url", url)
+		return cached, nil
+	}
+	c.logger.Debug("Cache miss, requesting", "url", url)
 	res, err := c.HTTPSGetter.Get(url)
 	if err != nil {
 		return nil, err
 	}
-	if canCache {
-		c.cache.Set(url, res)
-	}
+	c.cache.Set(url, res)
 	return res, nil
 }
 
--- a/internal/attestation/certcache/cached_client_test.go
+++ b/internal/attestation/certcache/cached_client_test.go
@@ -4,6 +4,7 @@
 package certcache
 
 import (
+	"errors"
 	"log/slog"
 	"sync"
 	"testing"
@@ -12,6 +13,7 @@ import (
 	"github.com/edgelesssys/contrast/internal/memstore"
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/goleak"
+	"k8s.io/utils/clock"
 	testingclock "k8s.io/utils/clock/testing"
 )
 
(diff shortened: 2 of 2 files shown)


internal/attestation/certcache/cached_client.go
```diff
@@ -44,25 +44,32 @@ func (c *CachedHTTPSGetter) Get(url string) ([]byte, error) {
...
+		if err != nil {
+			if cached, ok := c.cache.Get(url); ok {
+				c.logger.Debug("CRL request failed, fallback to cached CRL", "url", url)
+				return cached, nil
+			}
+			c.logger.Debug("CRL request failed and CRL was not found in cache", "url", url)
+			return nil, err
 		}
+		c.cache.Set(url, res)
+		return res, nil
```
3u13r, line 60: I think we can have less branches if we return early. Also I think we want a Warn or Info if the server could not be reached. 
For example:
```suggestion
		res, err := c.HTTPSGetter.Get(url)
		if err == nil {
			c.cache.Set(url, res)
			return res, nil
		}
		c.logger.Warn("could not reach server", "error", err)
		// Could not reach server, check cache
		if cached, ok := c.cache.Get(url); ok {
			c.logger.Debug("CRL request failed, fallback to cached CRL", "url", url)
			return cached, nil
		}
		c.logger.Debug("CRL request failed and CRL was not found in cache", "url", url)
		return nil, err
		
		
```

internal/attestation/certcache/cached_client_test.go
```diff
@@ -57,46 +57,60 @@ func TestMemcachedHTTPSGetter(t *testing.T) {
...
 
 		_, err := client.Get("foo")
 		assert.Error(err)
-		assert.Equal(1, fakeGetter.hits["foo"])
+		assert.Equal(3, fakeGetter.hits["foo"])
 	})
-	t.Run("Concurrent access", func(t *testing.T) {
+
+	t.Run("Check CRLs are still requested after caching", func(t *testing.T) {
+		fakeGetter.getErr = nil
```
3u13r, line 72: I don't think different t.Run should depend on each other i.e., we only have to set the .getErr to the default value of nil again, because we changed it in the t.Run above. 
If we still want to refactor the struct setup out a bit, we can define in function in this function that returns a default set of the needed structs and then each t.Run does its own modifications on them.

3u13r (changes requested):
That's a great test coverage!

3u13r (approved):
LGTM

internal/attestation/certcache/cached_client.go
```diff
@@ -44,25 +44,33 @@ func (c *CachedHTTPSGetter) Get(url string) ([]byte, error) {
...
-	if canCache {
+	if crlURL.MatchString(url) {
+		// For CRLs always query. When request failure, fallback to cache.
+		c.logger.Debug("Request CRL", "url", url)
+		res, err := c.HTTPSGetter.Get(url)
+		if err == nil {
+			c.cache.Set(url, res)
+			return res, nil
+		}
+		c.logger.Warn("Could not reach KDS", "error", err)
```
katexochen, line 55: This might not be correct, there are also other type of errors beside kds not being reachable.

internal/attestation/certcache/cached_client.go
```diff
@@ -44,25 +44,33 @@ func (c *CachedHTTPSGetter) Get(url string) ([]byte, error) {
...
 		if cached, ok := c.cache.Get(url); ok {
-			c.logger.Debug("Get cached", "url", url)
+			c.logger.Debug("CRL request failed, fallback to cached CRL", "url", url)
 			return cached, nil
 		}
+		c.logger.Debug("CRL request failed and CRL was not found in cache", "url", url)
```
katexochen, line 60:
```suggestion
		c.logger.Debug("CRL not found in cache", "url", url)
```

internal/attestation/certcache/cached_client.go
```diff
@@ -44,25 +44,33 @@ func (c *CachedHTTPSGetter) Get(url string) ([]byte, error) {
...
+		c.logger.Debug("Request CRL", "url", url)
+		res, err := c.HTTPSGetter.Get(url)
+		if err == nil {
+			c.cache.Set(url, res)
+			return res, nil
+		}
+		c.logger.Warn("Could not reach KDS", "error", err)
 		if cached, ok := c.cache.Get(url); ok {
-			c.logger.Debug("Get cached", "url", url)
+			c.logger.Debug("CRL request failed, fallback to cached CRL", "url", url)
```
katexochen, line 57:
```suggestion
			c.logger.Debug("Falling back to cached CRL", "url", url)
```

internal/attestation/certcache/cached_client.go
```diff
@@ -44,25 +44,33 @@ func (c *CachedHTTPSGetter) Get(url string) ([]byte, error) {
...
 			return cached, nil
 		}
+		c.logger.Debug("CRL request failed and CRL was not found in cache", "url", url)
+		return nil, err
 	}
-
-	c.logger.Debug("Get not cached", "url", url)
+	// For VCEK get cache first and request if not present
+	if cached, ok := c.cache.Get(url); ok {
+		c.logger.Debug("Get cached VCEK", "url", url)
```
katexochen, line 65:
```suggestion
		c.logger.Debug("Cache hit", "url", url)
```
Current log statement sounds like we are about to do it, not as it's already done.

katexochen (commented):
Did you verify the CRL expiration date is checked? Could you link the code that does this?

internal/attestation/certcache/cached_client.go
katexochen: Not sure, but maybe we should use Warn level here and in the log statement above. If I'm seeing that the KDS request failed, I'd like to see what path was taken.

jmxnzo:
> Did you verify the CRL expiration date is checked? Could you link the code that does this?

https://github.com/google/go-sev-guest/blob/main/verify/verify.go#L303-L305

internal/attestation/certcache/cached_client.go
```diff
@@ -44,25 +44,33 @@ func (c *CachedHTTPSGetter) Get(url string) ([]byte, error) {
...
+		c.logger.Debug("Request CRL", "url", url)
+		res, err := c.HTTPSGetter.Get(url)
+		if err == nil {
+			c.cache.Set(url, res)
+			return res, nil
+		}
+		c.logger.Warn("Failed requesting CRL from KDS", "error", err)
 		if cached, ok := c.cache.Get(url); ok {
-			c.logger.Debug("Get cached", "url", url)
+			c.logger.Debug("Falling back to cached CRL", "url", url)
```
katexochen, line 57:
```suggestion
			c.logger.Warn("Falling back to cached CRL", "url", url)
```

katexochen (approved):

PR #1051, opened 2024-12-04
packages/kata-image: specify correct fs-type in error message

msanft:
Previously, we used `stat` as a heuristic to detect the filesystem type, which is sub-optimal, as it will report `ext2/ext3` on `ext4` filesystems, leaving the reader of the error message in incertainty about the error. Switch to `df`, which is a better heuristic. \[1\]

\[1\]: https://unix.stackexchange.com/questions/256920/stat-f-show-a-ext4-file-system-type-as-ext2-ext3/274030#274030

### DIFF
--- a/packages/by-name/kata/kata-image/package.nix
+++ b/packages/by-name/kata/kata-image/package.nix
@@ -155,7 +155,7 @@ let
       runHook preBuild
 
       # Check if filesystem is ext.*
-      fstype=$(stat -f -c %T .)
+      fstype=$(df -T . | awk '{print $2}' | tail -n 1)
       if [[ $fstye == "ext4" || $fstype == "ext2/ext3" ]]; then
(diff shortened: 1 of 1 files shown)


packages/by-name/kata/kata-image/package.nix
```diff
@@ -155,7 +155,7 @@ let
       runHook preBuild
 
       # Check if filesystem is ext.*
-      fstype=$(stat -f -c %T .)
+      fstype=$(df -T . | awk '{print $2}' | tail -n 1)
```
burgerdev, line 158: As far as I understand, the issue with stat is that the superblock magic of ext4 is the same as for ext2 and ext3, so it lacks details to understand the diff and nobody bothered to upgrade the type string (or maybe it needs to stay like that for backwards compatibility). For our purposes, this does not really matter, since all ext* are affected. Could we just replace the `$fstype` in the error message with a literal `ext2/ext3/ext4`?

PR #1054, opened 2024-12-05
cli/genpolicy: by default do not include log-level 'debug' in Contras…

jmxnzo:
This PR changes the default logging level of the genpolicy tool back to "info", because re-running the generate step on a deployment with policy annotation results in a runtime error in the Contrast CLI, when matching the genpolicy logging prefix.  The runtime error grounds in the regex used for log translation, see https://github.com/edgelesssys/contrast/blob/main/cli/genpolicy/logtranslator.go/#L36 .

Still the RUST_LOG env variable can manually be set to "debug" to allow parsing the debug logs of genpolicy into the Contrast CLI for better troubleshooting, when no policy annotation is already present in the .yml file. Therefore the translation logic added in #1044 is kept in the Contrast CLI for now.

### DIFF
--- a/cli/genpolicy/genpolicy.go
+++ b/cli/genpolicy/genpolicy.go
@@ -62,7 +62,7 @@ func (r *Runner) Run(ctx context.Context, yamlPath string, logger *slog.Logger) 
 	genpolicy := exec.CommandContext(ctx, r.genpolicy.Path(), args...)
 	genpolicy.Env = os.Environ()
 	if _, hasRustLog := os.LookupEnv("RUST_LOG"); !hasRustLog {
-		genpolicy.Env = append(genpolicy.Env, "RUST_LOG=debug")
+		genpolicy.Env = append(genpolicy.Env, "RUST_LOG=info")
 	}
 	if _, hasRustBacktrace := os.LookupEnv("RUST_BACKTRACE"); !hasRustBacktrace {
 		genpolicy.Env = append(genpolicy.Env, "RUST_BACKTRACE=1")


burgerdev:
I found the bug: the log lines including policies exceed the maximum buffer size of `bufio`: https://pkg.go.dev/bufio#pkg-constants. This leads to an abrupt end of `scanner.Scan` [here](https://github.com/edgelesssys/contrast/blob/9150bce164bcef40807739ef5b42d35d17e9ca30/cli/genpolicy/logtranslator.go#L42), and then nobody reads the tool output and the pipe runs full and blocks. 

If `scanner.Scan()` returns `false`, `scanner.Err()` should be checked. If it's not nil, we should consume the remaining input somehow.

burgerdev (approved):
If there are no objections, I'd like to merge this now in order to make the default settings work now. Let's track and do the bug fix separately.

jmxnzo:
> I found the bug: the log lines including policies exceed the maximum buffer size of `bufio`: https://pkg.go.dev/bufio#pkg-constants. This leads to an abrupt end of `scanner.Scan` [here](https://github.com/edgelesssys/contrast/blob/9150bce164bcef40807739ef5b42d35d17e9ca30/cli/genpolicy/logtranslator.go#L42), and then nobody reads the tool output and the pipe runs full and blocks.
> 
> If `scanner.Scan()` returns `false`, `scanner.Err()` should be checked. If it's not nil, we should consume the remaining input somehow.

Yes that's true, i could also reproduce it yesterday. Another option would be to set the buffer of the scanner to another size (see https://pkg.go.dev/bufio#Scanner.Buffer). I'm not sure about which one is preferable for us, so feel free to suggest!

PR #1055, opened 2024-12-06
ci: suffix ci k8s namespaces with "-ci"

3u13r:
We invoke different e2e tests with the same nix command. This nix command already defines flags that are therefore shared between all current (and future) e2e tests. Let's make it easy to introduce new shared flags while maintaining the possibility that some tests define their own additional flags.

Also introduce a new namespace suffix flag.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -79,7 +79,8 @@ jobs:
             --image-replacements workspace/just.containerlookup \
             --namespace-file workspace/e2e.namespace \
             --platform ${{ inputs.platform }} \
-            --skip-undeploy="${{ inputs.skip-undeploy && 'true' || 'false' }}"
+            --skip-undeploy="${{ inputs.skip-undeploy && 'true' || 'false' }}" \
+            --namespace-suffix="-ci"
       - name: Download logs
         if: always()
         run: |
--- a/e2e/internal/contrasttest/contrasttest.go
+++ b/e2e/internal/contrasttest/contrasttest.go
@@ -12,6 +12,7 @@ import (
 	"crypto/x509"
 	"encoding/hex"
 	"encoding/json"
+	"flag"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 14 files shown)


3u13r:
The e2e test aks runtime check seems to be broken on main: https://github.com/edgelesssys/contrast/actions/runs/12179300514/job/33971147711

3u13r:
In the regression test testregression-2a2a5cbe-ci/mysql-centos-5f9875b787-8blvs reports 
failed to unpack `usr/libexec/mysqld` into `/run/kata-containers/image/layers/sha256_58bde37520d48c550fe5dba6a67740d181f8140ecdf387371bb92a7f6fae126f/usr/libexec/mysqld`            
No space left on device (os error 28)

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -374,14 +398,29 @@ func (ct *ContrastTest) FactorPlatformTimeout(timeout time.Duration) time.Durati
...
 
-	return strings.Join(append(re.FindAllString(strings.ToLower(t.Name()), -1), hex.EncodeToString(buf)), "-")
+	namespaceParts = append(namespaceParts, fmt.Sprintf("%x", buf))
+
+	// Below we join all parts with a dash already, the user usually provides the
+	// suffix with a dash, so we remove any leading dash from the suffix.
+	namespaceSuffix = strings.TrimPrefix(namespaceSuffix, "-")
+
+	// Append the suffix
+	namespaceParts = append(namespaceParts, namespaceSuffix)
```
burgerdev, line 421: If we don't treat the suffix as literal you can't leave it empty, because a trailing dash violates https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names. Can we just append it literally?

.github/workflows/e2e.yml
```diff
@@ -79,7 +79,8 @@ jobs:
             --image-replacements workspace/just.containerlookup \
             --namespace-file workspace/e2e.namespace \
             --platform ${{ inputs.platform }} \
-            --skip-undeploy="${{ inputs.skip-undeploy && 'true' || 'false' }}"
+            --skip-undeploy="${{ inputs.skip-undeploy && 'true' || 'false' }}" \
```
burgerdev, line 82: fyi, @miampf is removing `--skip-undeploy` in https://github.com/edgelesssys/contrast/pull/1046.

burgerdev (approved):

e2e/internal/contrasttest/contrasttest.go
3u13r: done

PR #1056, opened 2024-12-06
platforms: introduce generic bare-metal platform

katexochen:
This introduces a new platform for bare-metal machines. While we already support bare metal with k3s, some smaller thing like paths are really special to k3s. The new platform targets environments where things like containerd are found at their default location and should cover many bare-metal Kubernetes environments.


burgerdev (approved):
lgtm

PR #1057, opened 2024-12-06
nixos/kata: pull in `multi-user.target`

msanft:
Previously, due to setting `systemd.unit=kata-containers.target` on the kernel command line, we didn't pull in units that are `wantedBy=multi-user.target`. This breaks many assumptions of upstream software, including those of `nvidia-persistenced.service`, which was the first component that exhibited a hard breakage upon this unfulfilled assumption. This fixes these assumptions by pulling in `multi-user.target` through `kata-containers.target`.

### DIFF
--- a/packages/nixos/kata.nix
+++ b/packages/nixos/kata.nix
@@ -23,7 +23,6 @@ in
         "basic.target"
         "tmp.mount"
         "kata-agent.service"
-        "multi-user.target"
       ];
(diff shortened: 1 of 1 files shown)


msanft:
Skipping this for now: https://dev.azure.com/Edgeless/Edgeless/_backlogs/backlog/Edgeless Team/Features?workitem=4973

PR #1059, opened 2024-12-09
ci: add default_platform to AKS test

burgerdev:


.github/workflows/e2e_aks_runtime.yml
```diff
@@ -40,6 +41,7 @@ jobs:
           cat <<EOF > justfile.env
           container_registry=${{ env.container_registry }}
           azure_resource_group=${{ env.azure_resource_group }}
+          default_platform=AKS-CLH-SNP
```
3u13r, line 44: We now should be able to remove the platform from the lines 57 and 91

3u13r (changes requested):

.github/workflows/e2e_aks_runtime.yml
burgerdev: The second one is not overridden by the default, but the first one should be.

burgerdev:
@3u13r: Thanks for debugging this! I changed your commit to use the `PodsFromDeployment` built into contrasttest, that makes it a little more concise.

3u13r (approved):
LGTM

PR #1061, opened 2024-12-09
cli/genpolicy: never log already existing policy annotation on 'debug' + handle missing log prefix

jmxnzo:
This PR is a follow up on #1054 and patches the occurred runtime error on the different kata-containers upstream repos, by redacting existing policy annotations when logging. Thus the RUST_LOG env flag can now be set to "debug" and re-running the generate step on a deployment with set policy annotation does not result in a runtime error during translation anymore.
Nevertheless the default log level of genpolicy in the Contrast CLI is kept as 'info' and parsing the 'debug' logs needs to be configured manually by setting the RUST_LOG env to 'debug'. 

In some cases the genpolicy tool does not use the logger to write to stderr, the logtranslator.go additionally was rewritten to continue logging on the latest log level in case of missing log prefix. As well a simple regex was introduced to extract error messages from stderr of genpolicy and translate them to 'ERROR' logging in Contrast CLI, as they are also missing a logging prefix.

Upstream PR: https://github.com/kata-containers/kata-containers/pull/10647


packages/by-name/microsoft/genpolicy/0009-genpolicy-do-not-log-policy-annotation-in-debug.patch
```diff
@@ -0,0 +1,74 @@
...
++        }
++        if let Some(ref labels) = self.labels {
++            debug_struct.field("labels", labels);
++        }
++        if let Some(ref annotations) = self.annotations {
++            // Process annotations: redact values longer than 100 characters
++            let redacted_annotations: BTreeMap<_, _> = annotations
++                .iter()
++                .map(|(key, value)| {
++                    if value.len() > 100 {
```
burgerdev, line 59: `MaxScanTokenSize = 64 * 1024`, so we could allow a bit more here. Kubernetes annotations are often larger. May I suggest 4096?

packages/by-name/microsoft/genpolicy/0009-genpolicy-do-not-log-policy-annotation-in-debug.patch
```diff
@@ -0,0 +1,74 @@
...
++                        (key.clone(), "<redacted annotation>".to_string())
```
burgerdev, line 60: Did you consider replacing everything after the Nth character with `<... truncated ...>`?

packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -102,6 +102,11 @@ buildGoModule rec {
       # No upstream patch available, changes first need to be discussed with Kata maintainers.
       # See https://katacontainers.slack.com/archives/C879ACQ00/p1731928491942299
       ./0017-runtime-allow-initrd-AND-image-to-be-set.patch
+
+      # Simple genpolicy logging redaction of the policy annotation
+      # This avoids printing the entire annotation on log level debug, which resulted in errors of the logtranslator.go
+      # TODO(jmxnzo):
```
burgerdev, line 108: Needs doing

cli/genpolicy/logtranslator.go
```diff
@@ -33,18 +33,40 @@ func (l logTranslator) Write(p []byte) (n int, err error) {
...
+var (
+	genpolicyLogPrefixReg = regexp.MustCompile(`^\[[^\]\s]+\s+(\w+)\s+([^\]\s]+)\] (.*)`)
+	errorMessage          = regexp.MustCompile(`^thread\s+'main'\s+panicked\s+at`)
+)
 
 func (l logTranslator) startTranslate() {
 	go func() {
 		defer close(l.stopDoneC)
 		scanner := bufio.NewScanner(l.r)
+		// Default log level is initially set to 'WARN', only fallback if first line does not match the logging pattern
```
burgerdev, line 45:
```suggestion
		// Default log level is initially set to 'WARN'. This is only relevant if the first line does not match the logging pattern.
```

cli/genpolicy/logtranslator.go
```diff
@@ -33,18 +33,40 @@ func (l logTranslator) Write(p []byte) (n int, err error) {
...
 			if len(match) != 4 {
 				// genpolicy prints some warnings without the logger
-				l.logger.Warn(line)
+				// we continue logging on the latest used log-level
+
+				// Error is written to stderr by genpolicy without using the logger,
+				// simple regex to detect the error message on stderr
+				if errorMessage.MatchString(line) {
+					logLevel = "ERROR"
+				}
```
burgerdev, line 59: This should be tested.

packages/by-name/microsoft/genpolicy/0009-genpolicy-do-not-log-policy-annotation-in-debug.patch
jmxnzo: Great suggestion!

packages/by-name/kata/kata-runtime/package.nix
jmxnzo: That was on purpose, i was waiting to open up the upstream PR. Should be done right now!

burgerdev (approved):
lgtm

PR #1065, opened 2024-12-12
rfc: add platform support RFC

msanft:
This adds an RFC for platform support and representation in Contrast. This is in very much of a bare-bones stage now, so reviews and critique are welcome.

### DIFF
--- a/rfc/008-platform-support.md
+++ b/rfc/008-platform-support.md
@@ -1,0 +1,102 @@
+# RFC 008: Platform Support
+
+## Objective
+
+Find an in-code representation for Contrast platforms that enables conditionals based on platform features,
+rather than full platform definitions.
+
+## Problem statement
+
+Contrast currently distinguishes between platforms based on their distinct features.
+Taking the `AKS-CLH-SNP` platform as an example, this serialized platform string conveys
+the following information:
+
+- Kubernetes engine: `AKS`
+  - (Implicit: Kata upstream / Microsoft fork)
+- Hypervisor: `CLH`
+- Hardware platform: `SNP`
+
+The user picks such a platform through its serialized string (for example `AKS-CLH-SNP`), and the
+code then implements actions based upon that, such as when writing platform-specific Kata
+configuration values in the [node-installer].
+
+To perform these actions in the code, conditionals based on the platform are necessary.
+
+Currently, platforms are represented as an int alias type:
+
+```go
+type Platform int
+```
+
+The individual platforms thus are represented by instantiations of said type:
+
+```go
+AKSCloudHypervisorSNP Platform = 1
+```
+
+Conditionals thus need to work on full platforms, making them look like:
+
+```go
+// TDX platforms only
+if platform == platforms.K3sQEMUTDX || platform == platforms.RKE2QEMUTDX {
+  // ..
+}
+```
+
+This isn't only cumbersome to  write, but also easy to make errors in. For example when forgetting to
+match against a certain platform.
+
+[node-installer]: https://github.com/edgelesssys/contrast/blob/3f39682ea9a383b2557923e257bd065e461b8ee6/nodeinstaller/internal/constants/constants.go#L48
+
+## Proposal
+
+### Structured Representation
+
+Throughout the implementation code, platforms shouldn't be represented as _individual combinations_,
+but rather through a generic, structured type that contains information about their _features_.
+
+So `AKS-CLH-SNP` could become something like this:
+
+```go
+type Platform struct {
+    ContainerdConfigPath  string
+    UnitsToRestart        []string
+    Snapshotter           Snapshotter
+    GPUSupport            bool
+    Debug                 bool
+    Hypervisor            Hypervisor
+    HardwarePlatform      HardwarePlatform
+}
+
+// AKS-CLH-SNP
+foo := Platform{
+    ContainerdConfigPath:   "/etc/containerd/config.tomL",
+    UnitsToRestart:         []string{"containerd.service"},
+    GPUSupport:             false,
+    Debug:                  false,
+    Hypervisor:             CloudHypervisor,
+    HardwarePlatform:       SNP,
+}
+```
+
+This would allow the implementation code (for example the aforementioned switch for [Kata configuration])
+to be reduced to only depend on the information bits it requires.
+
+For example, this:
+
+```go
+case platforms.MetalQEMUTDX, platforms.K3sQEMUTDX, platforms.RKE2QEMUTDX:
+```
+
+Could become this:
+
+```go
+case TDX
+```
+
+This would also enable easy validation of user error, as the individual types can be checked for
+validity at parsing time, making invalid states unrepresentable [^1].
+
+[Kata configuration]: https://github.com/edgelesssys/contrast/blob/3f39682ea9a383b2557923e257bd065e461b8ee6/nodeinstaller/internal/constants/constants.go#L48
+
+[^1]: https://geeklaunch.io/blog/make-invalid-states-unrepresentable/
--- a/tools/vale/styles/config/vocabularies/edgeless/accept.txt
+++ b/tools/vale/styles/config/vocabularies/edgeless/accept.txt
@@ -33,6 +33,7 @@ cyber
 datacenter
 Datadog
 deallocate
+deduplicate
 DEK
 Dockerfile
 Dynatrace
@@ -57,6 +58,7 @@ hyperv
 IAM
 Initializer
 initramfs
+initrd
 Inkscape
 iodepth
 IPSec
@@ -85,6 +87,7 @@ paravisor
 PCR
 plaintext
 podman
+precalculator
 protobuf
 proxied
 QEMU
@@ -94,9 +97,11 @@ Rekor
 resizable
 resourceGroup
 rollout
+RTMR
 runtime
 safeguardable
 SBOM
+serializable
 sigstore
 snapshotter
 SSD
@@ -115,6 +120,7 @@ tmpfs
 TPM
 underspecified
 unencrypted
+unrepresentable
 unspoofable
 untrusted
 updatable
@@ -133,7 +139,3 @@ Xeon
 xsltproc
 Zipkin
 # keep-sorted end
-RTMR
-RTMRs
-precalculator
-initrd


rfc/008-platform-support.md
```diff
@@ -0,0 +1,111 @@
...
+be hard to read for humans at some point.
+
+## Considerations
+
+### Node-Installer Images
+
+As it would be complex and costly to publish node-installer images for all supported platform
+combinations, it might be better to ship a singular one-size-fits-all node installer with support
+for all platforms. This could also be a foundation for heterogeneous deployments that utilize multiple
+platforms (for example GPU- and Non-GPU machines).
```
burgerdev, line 105: The node-installer could figure out a lot of configuration without user interaction - TEE flavour, containerd paths, maybe even hypervisor. On the other hand, outside the runtime the config is only used to decide which genpolicy to use and to populate reference values. I wonder whether we actually need to expose the entire runtime config to the user.

rfc/008-platform-support.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Could become this:
+
+```go
+case TDX
+```
+
+This would also enable easy validation of user error, as the individual types can be checked for
+validity at parsing time, making invalid states unrepresentable [^1].
```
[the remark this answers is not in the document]
msanft: Under the assumption of only validating constructors being used in the program, of course. I do acknowledge that my example is counter-intuitive for that, but it's a simplification, after all.

rfc/008-platform-support.md
```diff
@@ -0,0 +1,111 @@
...
+
+This would reduce implementation complexity, as Contrast could rely on existing YAML / TOML / JSON / etc. parsing
+libraries and no canonical string representation of a platform would need to be implemented.
+
+Additionally, adding platform features would not lead to unreasonably large string representations, which might
+be hard to read for humans at some point.
+
+## Considerations
+
+### Node-Installer Images
```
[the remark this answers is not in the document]
msanft: Runtimes in the sense of Kata runtimes?

rfc/008-platform-support.md
msanft: I think large parts of it have to be exposed. It might be possible that you have an instance that has GPUs available but don't want to run GPU-enabled Contrast on it, for example. 

Also, reference value generation needs almost all values, if I'm not mistaken?

rfc/008-platform-support.md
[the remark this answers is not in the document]
burgerdev: Kubernetes runtimes.

rfc/008-platform-support.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Structured Representation
+
+Throughout the implementation code, platforms shouldn't be represented as _individual combinations_,
+but rather through a generic, structured type that contains information about their _features_.
+
+So `AKS-CLH-SNP` could become something like this:
+
+```go
+type Platform struct {
```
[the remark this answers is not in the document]
burgerdev: with GPU support or without, debug or prod, ...

rfc/008-platform-support.md
burgerdev: You picked the one dimension that I did not list as candidate for automation ;)

It's still an interesting question whether / how we want to deal with mixed images. As a user, how do I tell Contrast to use a GPU image for pod A, but a plain image for pod B? I imagine a config file is not granular enough here. If there is only one image per runtime, how do I mix two runtimes with one coordinator? (I think that's a scenario we might want to support eventually, though)

The reference values need the image measurements, the genpolicy choice and the hardware reference values, where the last two are only a question of AKS-CLH vs. rest.

msanft: > The reference values need the image measurements, the genpolicy choice and the hardware reference values, where the last two are only a question of AKS-CLH vs. rest.

For hardware reference values, I think it's also a matter of TDX vs. SNP (Genoa) vs. SNP (Milan) and so on, right?

About mixed images, how the customer decides to do that is TBD still, I think. I don't have _the_ solution at hand for this. 

About "one image per runtime": I think we could ship all variants in the node-installer if we can make them reasonably small?

burgerdev: We don't fill hardware values for bare metal, so it's just image measurements I think.

tools/vale/styles/config/vocabularies/edgeless/accept.txt
```diff
@@ -137,3 +137,6 @@ RTMR
 RTMRs
 precalculator
 initrd
+deduplicated
```
thomasten, line 140:
```suggestion
precalculator
initrd
deduplicate
```
nit: you only need to add the basic form of a word

tools/vale/styles/config/vocabularies/edgeless/accept.txt
```diff
@@ -137,3 +137,6 @@ RTMR
...
+unrepresentable
+serializable
```
thomasten, line 142: nit: please _keep-sorted_

thomasten (commented):
General ideas LGTM, but I lack knowledge for some of the details.

rfc/008-platform-support.md
```diff
@@ -0,0 +1,111 @@
+# RFC 008: Platform Support
+
+## Objective
+
+Implement support for diverse platforms in Contrast, ensuring that:
+
+1. Users can mix-and-match features throughout all platforms supported by Contrast
```
katexochen, line 7: I don't like how this rfc mixes a user facing feature with a refactoring. For an rfc for a feature, I'd expect far more arguing from the users perspective and UX considerations.

rfc/008-platform-support.md
```diff
@@ -0,0 +1,111 @@
...
+### Structured Representation
+
+Throughout the implementation code, platforms shouldn't be represented as _individual combinations_,
+but rather through a generic, structured type that contains information about their _features_.
+
+So `AKS-CLH-SNP` could become something like this:
+
+```go
+type Platform struct {
+    KubernetesDistribution KubernetesDistribution
```
katexochen, line 47: We must think about what we really need to know. Like, is the distribution relevant, or do we really only need the paths and the systemd unit to restart? Otherwise we are just pushing the problem down a layer.

rfc/008-platform-support.md
```diff
@@ -0,0 +1,111 @@
...
+    Hypervisor Hypervisor
+    HardwarePlatform HardwarePlatform
+}
+
+// AKS-CLH-SNP
+foo := Platform{
+    KubernetesDistribution: AKS,
+    Hypervisor: CloudHypervisor,
+    HardwarePlatform: SNPMilan,
```
katexochen, line 56:
```suggestion
    HardwarePlatform: SNP,
```
Not sure if this is an intended but undocumented change or just a typo. We don't have split platforms for different product lines right now.

rfc/008-platform-support.md
```diff
@@ -0,0 +1,111 @@
...
+Additionally, adding platform features would not lead to unreasonably large string representations, which might
+be hard to read for humans at some point.
+
+## Considerations
+
+### Node-Installer Images
+
+As it would be complex and costly to publish node-installer images for all supported platform
+combinations, it might be better to ship a singular one-size-fits-all node installer with support
+for all platforms. This could also be a foundation for heterogeneous deployments that utilize multiple
```
katexochen, line 104: As as user, I don't want to download tens of gigs of stuff that I don't really need.

rfc/008-platform-support.md
katexochen: > As a user, how do I tell Contrast to use a GPU image for pod A, but a plain image for pod B?

Via the runtime class. We must install different runtimeclasses for such variants.

rfc/008-platform-support.md
katexochen, line 104: Suggestion on how to solve this: Remove all the stuff from the node installer image, give it a list of url/hash pairs. Send the config to the node-installer (via cm or what) and let it download the required components one by one.

tools/vale/styles/config/vocabularies/edgeless/accept.txt
katexochen: and please move the keep-sorted comment to the end of the list again.

rfc/008-platform-support.md
burgerdev: I'd like to qualify that: it's not the user who needs to download it, but the datacenter - a couple gigs might not make a big difference there.

Anyway, I feel like the the unified node installer contradicts the objectives (1) and (2). Maybe we should focus on the tooling to make new node installer variants instead?

burgerdev: This makes installation in an air-gapped / proxied facility less convenient.

rfc/008-platform-support.md
burgerdev: I see that using a diffferent runtimeclass is sufficient, but why is it necessary?

rfc/008-platform-support.md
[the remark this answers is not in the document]
burgerdev: That might help

rfc/008-platform-support.md
katexochen: we can fetch from a file path as fallback and ship a separate image for air gapped, then we cover both with two images.

msanft: I thought about this, but we also have to consider that it brings quite some overhead to development, as we lose the benefit of a self-contained image, and have to build some kind of uploading pipeline again for the components to S3 or so.

katexochen: we could also push/pull the artifacts from a container registry with oras.
I think the dev workflow will be more complex the more variants we have either way, main goal should be to make this convenient for the user while keeping it manageable for devs.

rfc/008-platform-support.md
msanft: It was meant to propose only a refactoring of internal code first, but now shifted towards something that feels like a proposal for a configuration file instead. Let's discuss this in a meeting, and I'll then re-write the RFC to argue more from a user's perspective.

rfc/008-platform-support.md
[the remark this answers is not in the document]
msanft: > A runtime could be used with more than one image (not easily, but possible), but not with more than one hypervisor or snapshotter.

How comes? I can't follow here.

rfc/008-platform-support.md
msanft: I think so. Let's discuss this in the meeting.

rfc/008-platform-support.md
msanft: Also a point for discussion in the meeting, like the comment of @katexochen above.

rfc/008-platform-support.md
msanft: I think both an agnostic monolithic node-installer, as well as a component-based node-installer have their up- and downsides for devs _and_ users. We'll need to think through the individual possibilities here together.

rfc/008-platform-support.md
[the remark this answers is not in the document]
burgerdev: Kubernetes runtimes map to containerd runtimes 1:1, which in turn have a 1:1 mapping to snapshotters. Maybe you could swap the hypervisor with an annotation, though.

katexochen (approved):

burgerdev (approved):

PR #1066, opened 2024-12-12
packages/nixos: minimize image size

msanft:
Previously, our bare-metal Kata image weighed in at around 838MB. This PR reduces its size to about 561MB. It does so by incorporating the following changes:

- Remove Nix (CLI, daemon and store database) from the image closure. Since our image is immutable anyways, we do not need to include this anyways. Removing this saves around 188MB in image size.
- Remove Perl and Python interpreters from the image closure. These are only required for parts Nix' traditional "activation" mechanism, all of which have non-interpreter alternatives available. A tangible example for this is the population of `/etc`, which traditionally was implemented in Perl, but now has an overlayfs-based alternative for available. This saves around 90MB in image size.

Minimizations for the GPU-enabled image are intentionally deferred to be implemented in a follow-up PR.

### DIFF
--- a/packages/nixos/kata.nix
+++ b/packages/nixos/kata.nix
@@ -87,12 +87,8 @@ in
     };
 
     networking.resolvconf.enable = false;
-    systemd.tmpfiles.settings."10-etc-resolvconf"."/etc/resolv.conf".f = {
-      group = "root";
-      mode = "0755";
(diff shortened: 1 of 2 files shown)


burgerdev (approved):
Great job, thanks!

PR #1067, opened 2024-12-13
ci: correctly trigger peerpods e2e on image changes

burgerdev:
Directory paths in the `pull_request` trigger need to include wildcards, otherwise actions will look for a file of that name and never match.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths


msanft (approved):

burgerdev:
Merging this now, peer-pod test failure is expected and addressed in #1072

PR #1077, opened 2024-12-17
scripts: add go-directive-sync, run with generate, sync go directives

katexochen:
In https://github.com/edgelesssys/contrast/commit/791c49f8b568ff11d1e6b5b58224c492db48e0a3, we updated the main modules go directive without at least also updating the go.work file, again. This PR adds a script and runs it as part of generate to prevent this in future.


packages/scripts.nix
```diff
@@ -531,4 +535,32 @@
     ];
     text = builtins.readFile ./test-peerpods.sh;
   };
+
+  # Sync the go directive between go.mod/go.work files (that is, the 'go' statement of these files).
+  # We take take the latest version we find and use that everywhere.
```
burgerdev, line 540:
```suggestion
  # We take the latest version we find and use that everywhere.
```

packages/scripts.nix
```diff
@@ -531,4 +535,32 @@
...
+    name = "go-directive-sync";
+    runtimeInputs = with pkgs; [
+      go
+      findutils
+      coreutils
+    ];
+    text = ''
+      set -euo pipefail
+
+      modFiles=$(find . -regex '.*\(go.mod\|go.work\)$')
```
burgerdev, line 551:
```suggestion
      modFiles=$(find . -regex '.*/go[.]\(mod\|work\)$')
```

packages/scripts.nix
```diff
@@ -531,4 +535,32 @@
...
+
+      goVers=()
+      while IFS= read -r f; do
+        ver=$(grep '^go [0-9]\+\.[0-9]\+\.[0-9]\+$' "$f")
```
burgerdev, line 555:
```suggestion
        ver=$(grep -E '^go [0-9]+[.][0-9]+[.][0-9]+$' "$f")
```

* Use `-E` for unescaped `+` support.
* Nit - I find it harder to reason about backslash-escaped dots than `[.]` because I'm always worried about the shell hijacking the escape.

burgerdev (approved):

PR #1078, opened 2024-12-17
rfc: distributed coordinator

burgerdev:


rfc/009-distributed-coordinator.md
```diff
@@ -0,0 +1,178 @@
...
+
+message RecoverRequest {}
+
+message RecoverResponse {
+  bytes Seed = 1;
+  bytes Salt = 2;
+  bytes RootCAKey = 3;
+  bytes RootCACert = 4;
+  bytes MeshCAKey = 5;
+  bytes MeshCACert = 6;
```
3u13r, line 146: I think we need to explain the security of the HA part a bit more. I think when we allow to directly set the Mesh components of the Coordinator during automatic recovery we loose protection against the Kubernetes admin/workload owner as they can redirect to themself and "provision" a coordinator with the values above. 

The simplest case to think about is one Coordinator needing to be recovered and the workload owner answering the recovery call from this coordinator. The next time the user verifies the deployment, it sees a valid chain of manifests and therefore trusts the new MestCACert. This allows the workload owner to man-in-the-middle the TLS connection from the data owner to the application.


While we have excluded this threat model from our current recovery, I think HA and auto-recovery can be implemented while securing against this threat model as well e.g., via only allowing recovery of coordinator that have the same hashes. Of course, this breaks the upgrade process, but I think we have to drop coordinator upgrades anyway in this threat model.

3u13r (changes requested):

rfc/009-distributed-coordinator.md
burgerdev: Good catch, thanks!

We should aim to support recovery from heterogeneous coordinators if they are explicitly allowed by the manifest. An upgrading workload owner could first set a manifest including new coordinator policies, then deploy new coordinators, then remove old coordinators, then remove old coordinator policies from the manifest. A data owner would need to verify not only the current manifest, but also the history of allowed coordinators.

I think what we need are the following invariants:

* **(A)** A coordinator with current manifest `M` only sends key material to pods that have the `coordinator` role in `M`.
* **(B)** A coordinator with current manifest `M` uses only key material that it generated locally or that was received from a pod with the `coordinator` role in `M`.

(A) should be covered sufficiently by the current proposal text, and we could modify it to achieve (B) as follows:

1. Load the existing latest transition from the store, keeping the signature around but not checking it yet.
2. Fetch the corresponding manifest, but don't set the state yet.
3. Create a validator from the temporary manifest's reference values.
4. Connect to the serving coordinator and validate its reference values.
5. Check that the serving coordinator's policy corresponds to a `coordinator` role in the temp manifest.
6. Receive the `RecoverResponse`.
7. Verify the signature from (1).
8. Initialize the state with received seed, keys, certs and the temp manifest.

rfc/009-distributed-coordinator.md
```diff
@@ -0,0 +1,178 @@
+# RFC 009: Distributed coordinator
+
+## Background
+
+The Contrast Coordinator is a stateful service with a backend storage that can't be shared.
```
msanft, line 5: Is the coordinator state non-sensitive?

burgerdev: It does not need to be CC-secure, if that's what you mean. https://github.com/edgelesssys/contrast/blob/d4892d3258737c4c69be8ce3443760ffc4ac85b0/rfc/004-recovery.md?plain=1#L38-L40

rfc/009-distributed-coordinator.md
3u13r: I think that is a nice summary and solution. Though, I think we could omit the roles from the manifest if we'd need to simplify it, since in the eyes of the data owner all code of all components inside the mesh/deployment must be trusted anyway. I.e., the workload owner should never be able to impersonate any component from the deployment, then all guarantees of shielding the data owner against the workload owner are broken.

This way the verification of the endpoint that recovers the coordinator could be simplified so that the recovering coordinator is first "initialized" by the running coordinator, like any other workload. Then the application logic notices that it was recovered and requests the locally kept keys of the running coordinator. The endpoint that is called here resides behind mTLS client auth. 

The big trade off here, I think, is due to the asynchronous/split nature there is more potential that future changes break the security on the other hand I think the two parts are conceptually quite simple.

burgerdev: > I think that is a nice summary and solution. Though, I think we could omit the roles from the manifest if we'd need to simplify it, since in the eyes of the data owner all code of all components inside the mesh/deployment must be trusted anyway. I.e., the workload owner should never be able to impersonate any component from the deployment, then all guarantees of shielding the data owner against the workload owner are broken.

I think we should still keep roles, for the following reasons:

* Prevent privilege escalation by compromised individual deployments. They may be able to use their own workload secret and authenticate as themselves, but we should not grant them access to the seed by default.
* As you pointed out originally, we need to establish trust in the seed material provided to the recovering coordinator. This is easier if I can rely on said invariants, but I (as a Contrast developer) can only be sure that these are upheld by our Coordinators. Explicitly labeling our Coordinators thus enables the chain of trust.

> This way the verification of the endpoint that recovers the coordinator could be simplified so that the recovering coordinator is first "initialized" by the running coordinator, like any other workload. Then the application logic notices that it was recovered and requests the locally kept keys of the running coordinator. The endpoint that is called here resides behind mTLS client auth.

Just for the record, right now the initializers don't attest the Coordinator and we don't use a server cert for the meshapi. If we were using the root CA cert for the meshapi server, we'd still only be able to verify that

* the root CA key is derived from the seed we receive
* the manifest verifies under the transaction signing key derived from the seed we received

We still don't know whether (B) holds, though. If we add attestation to the coordinator, we could rely on a variant of (A) - "[...] sends the _correct_ key material for its current manifest [...]" and thus know

* the seed matches the manifest currently set at the peer Coordinator

But we're still missing the link between the manifest we're recovering and the seed. I'll try to work that part into the proposal.

rfc/009-distributed-coordinator.md
```diff
@@ -0,0 +1,236 @@
...
+1. Load the latest manifest from persistence without verifying the signature.
+2. Build an aTLS client from the unverified manifest:
+   * Populate ValidateOpts with reference values.
+   * Add a callback that verifies the policy hash with the roles from the unverified manifest.
+   * Add standard client attestation.
+3. Call the serving Coordinator's `Recover` endpoint.
+4. Compare the received manifest with the temporary manifest.
+5. Set up a temporary `SeedEngine` with the received parameters.
+6. Load the manifest history from persistence, verifying the signature with the temporary seed engine.
+7. Compare the loaded manifest with the received manifest.
```
katexochen, line 202: Just to clarify, this means if a coordinator was in recovery state while a manifest was updated, it can't be recovered anymore, right?

rfc/009-distributed-coordinator.md
```diff
@@ -0,0 +1,236 @@
...
+
+The Coordinator uses a generic key-value store interface to read from and write to the persistent store.
+In order to allow distributed read and write requests, we only need to swap the implementation for one that can be used concurrently.
+
+```golang
+type Store interface {
+  Get(key string) ([]byte, error)
+  Set(key string, value []byte) error
+  CompareAndSwap(key string, oldVal, newVal []byte) error
+  Watch(key string, callback func(newVal []byte)) error // <-- New in RFC009.
```
katexochen, line 49: Is the Watch func based on some lower level lib we plan to consume? The interface looks uncommon, watchers usually use channels in Go.

rfc/009-distributed-coordinator.md
burgerdev: If the comparison fails, the recovering Coordinator should start over at (1), so it should eventually succeed.

rfc/009-distributed-coordinator.md
```diff
@@ -0,0 +1,236 @@
...
+[^2]: <https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications>
+
+### Recovery mode
+
+The Coordinator is said to be in *recovery mode* if its state object (containing manifest and seed engine) doesn't correspond to the latest state in persistence.
+Recovery mode can be entered as follows:
+
+* The Coordinator starts up and the history has a latest transition.
+* Receiving a watch event for the latest manifest, and the latest manifest isn't the current one.
+* Syncing the latest manifest during API calls and discovering a new manifest.
```
thomasten, line 96: Syncing means getting latest from the store?

rfc/009-distributed-coordinator.md
```diff
@@ -0,0 +1,236 @@
...
+
+Recovery mode exits after the state has been updated by:
+
+* A successful [Peer recovery](#peer-recovery).
+* A successful recovery through the `userapi.Recover` RPC.
+* A successful initial `userapi.SetManifest` RPC.
```
thomasten, line 102:
```suggestion
```
Is this a leftover from a previous version of this RFC?

rfc/009-distributed-coordinator.md
```diff
@@ -0,0 +1,236 @@
...
+During `generate`, we append the value of the `contrast.edgeless.systems/pod-role` annotation to the policy entry.
+If there is no coordinator among the resources, we add a policy entry for the embedded coordinator policy.
+
+### Peer recovery
+
+The peer recovery process is attempted by Coordinators that are in [Recovery mode](#recovery-mode):
+
+1. Once after startup, to speed up initialization.
+2. Once as callback to the `Watch` event, .
+3. Periodically from a goroutine, to reconcile missed watch events.
```
thomasten, line 160:
```suggestion
1. Once after entering recovery mode.
2. Periodically from a goroutine, to reconcile missed watch events.
```
suggest referring to the enter conditions because these overlap

rfc/009-distributed-coordinator.md
```diff
@@ -0,0 +1,236 @@
...
+6. Load the manifest history from persistence, verifying the signature with the temporary seed engine.
+7. Compare the loaded manifest with the received manifest.
+   If they match, update the state with the temporary seed engine and the fetched history.
+
+If the recovery was successful, the client Coordinator leaves the peer recovery process.
+Otherwise, it continues with the next available peer, or fails the process if none are left.
+
+## Open issues
+
+* TODO(burgerdev): inconsistent state if `userapi.Recover` is called while a recovered coordinator exists. Fix candidates:
```
thomasten, line 210: Can you give an example of such an inconsistent state?

rfc/010-distributed-coordinator.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Persistency changes
+
+The Coordinator uses a generic key-value store interface to read from and write to the persistent store.
+In order to allow distributed read and write requests, we only need to swap the implementation for one that can be used concurrently.
+
+```golang
+type Store interface {
+  Get(key string) ([]byte, error)
+  Set(key string, value []byte) error
```
msanft, line 47: If only `CompareAndSwap` is atomic, do we have a race risk with `Get`/`Set` here?

rfc/010-distributed-coordinator.md
```diff
@@ -0,0 +1,249 @@
...
+2. The Coordinator policy can't be embedded into the Coordinator during build, and deriving it at runtime is inconvenient.
+
+Thus, we introduce a new field to the manifest that specifies roles available to the verified identity.
+For now, the only known role is `coordinator`, but this could be extended in the future (for example: delegate CAs).
+
+```golang
+type PolicyEntry struct {
+  SANs             []string
+  WorkloadSecretID string `json:",omitempty"`
+  Roles            []string `json:",omitempty"` // <-- New in RFC009.
```
msanft, line 152:
```suggestion
  Roles            []string `json:",omitempty"` // <-- New in RFC009.
```
I know this is probably only an example, but we'll want to type the roles, I think?

burgerdev: I don't understand the suggested diff, but I think you meant to say roles should be an enum array, not a string array? If only Contrast is the audience for these I'd agree, but we could also make them available to users as certificate extensions, in which case we should probably not restrict ourselves here.

rfc/010-distributed-coordinator.md
burgerdev: These store primitives are used for different purposes:

* `Get`: snapshot of the current state of a piece of data without any freshness guarantees
* `Set`: store content-addressed data (policies, manifests)
* `CompareAndSwap`: atomically advance the latest transition (i.e. active manifest)

We implicitly expect the store to be race free for concurrent Get/Set calls (no partial or corrupt data).

All of this has been defined originally for recovery persistence: https://github.com/edgelesssys/contrast/blob/main/rfc/004-recovery.md#storage-interface.

msanft (approved):
Thanks for writing this up so thoroughly and well-explained!

To the best of my (arguably limited) knowledge on this part of Contrast, this looks good.

thomasten (approved):
Thanks, looks solid to me! (modulo my incomplete knowledge on some Contrast details)
nit: there are still some mentions of RFC009 in the text

katexochen (approved):

rfc/010-distributed-coordinator.md
```diff
@@ -0,0 +1,249 @@
...
+  bytes MeshCACert = 6;
+  bytes LatestManifest = 7;
+}
+```
+
+When this method is called, the serving Coordinator:
+
+1. Provides an attestation statement to the client (this isn't the case for the mesh API right now).
+2. Queries the current state and attaches it to the context (this already happens automatically for all `meshapi` calls).
+3. Verifies that the client identity is allowed to recover by the state's manifest (has the `coordinator` role).
```
3u13r, line 192:
```suggestion
3. Verifies that the client identity is allowed to be recovered by the state's manifest (has the `coordinator` role).
```

3u13r (approved):
LGTM

PR #1082, opened 2024-12-18
attestation.snp: reflect dependency of validators on productLine in verify.Options

jmxnzo:
Our reference values of SNP attestation have a hard dependency on the productLine used during attestation. This includes the derivation of trustedRoots, as well as the trustedMeasurements.  By design, we start a validator configured with specific verifyOpts for each reference value. 
This led to the error "VCEK could not be verified by any trusted roots", because validators configured with the wrong productLine tried to fulfill the attestation verification. Therefore this PR adds the explicit setting of the productLine in verifyOpts, to reflect the dependency of our reference values on the productLine. As expected we then run into the error below prior to validation, which reveals that a validator is configured for the wrong productLine:
 `time=2024-12-18T10:46:20.815Z level=ERROR msg="Validation failed" mesh-authority.validator.tee-type=snp mesh-authority.validator.nonce=6881501f40cebdb1492a87316e5a11cbb1bb9dfc2939240a9df098dacdab8b6a mesh-authority.validator.error="verifying report: expected product name SEV_PRODUCT_MILAN, got SEV_PRODUCT_GENOA"
`

Logging considerations were moved to #1095

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -214,6 +214,11 @@ func (m *Manifest) SNPValidateOpts(kdsGetter trust.HTTPSGetter) ([]ValidatorOpti
 		}
 
 		verifyOpts := verify.DefaultOptions()
+		// Setting the productLine explicitly, because of full dependence of trustedMeasurements and derivation of trustedRoots on productLine.
+		verifyOpts.Product, err = kds.ParseProductLine(string(refVal.ProductName))
+		if err != nil {
+			return nil, fmt.Errorf("SNP reference values: %w", err)
+		}
 		verifyOpts.TrustedRoots, err = trustedRoots(refVal.ProductName)
 		if err != nil {
 			return nil, fmt.Errorf("determine trusted roots: %w", err)


internal/attestation/snp/validator.go
```diff
@@ -58,10 +58,13 @@ func (v *Validator) OID() asn1.ObjectIdentifier {
 func (v *Validator) Validate(attDocRaw []byte, nonce []byte, peerPublicKey []byte) (err error) {
 	v.logger.Info("Validate called", "nonce", hex.EncodeToString(nonce))
 	defer func() {
+		// Note: We tolerate Validate() to fail for validators without matching configuration. Only 1 of n validators has to succeed.
+		// Thus the error handling of the validators takes place in the calling function verifyEmbeddedReport().
```
burgerdev, line 62: Please make `verifyEmbeddedReport` log the combined error at ERROR level. Otherwise we end up in the situation again that motivated #1027, where connection attempts are silently failing.

internal/attestation/snp/validator.go
```diff
@@ -58,10 +58,13 @@ func (v *Validator) OID() asn1.ObjectIdentifier {
...
+		// Do not log any errors in the subroutines and use info to still provide insights of the validation processes.
 		if err != nil {
-			v.logger.Error("Validation failed", "error", err)
+			v.logger.Info("Validation failed", "nonce", hex.EncodeToString(nonce), "error", err)
 		} else {
-			v.logger.Info("Validation successful")
+			v.logger.Info("Validation successful", "nonce", hex.EncodeToString(nonce))
```
burgerdev, line 67: Do we need to log the nonce? If validation fails because of it, wouldn't we see that in the corresponding error?

internal/manifest/manifest.go
```diff
@@ -281,6 +286,8 @@ func trustedRoots(productName ProductName) (map[string][]*trust.AMDRootCerts, er
 	return trustedRoots, nil
 }
 
+
+
```
burgerdev, line 290:
```suggestion
```

internal/attestation/snp/validator.go
burgerdev: Discussed in private: the errors are reproted as expected.

jmxnzo: I double checked the logging mechanisms and it works fine on the coordinator as well as on the local attestation validation. However it might be reasonable to add better formatting to the joined error message.

internal/attestation/snp/validator.go
jmxnzo: The current error message didn't contain nonce logging. I added logging the nonce as identifier to the atls connection and the oid as identifier of the validator into the our current logging. 
On log-level DEBUG, the single validation failures are logged as well. Otherwise just the successful validation call is logged with the corresponding nonce and OID.

burgerdev (approved):

edgelessci:
- #1097

PR #1087, opened 2024-12-19
packages/nixos: write Kata agent output to console and journal

msanft:
Previously, we only wrote Kata agent output to the console / TTY (which is also what it does upstream [^1]). This makes the output appear in the Kata runtime logs on the Host via the `vmconsole=` field. However, when debugging from within the pod VM, it's helpful to also have the logs in there so that they can be searched in with `journalctl`. This makes it so the output of the agent is written to both the console _and_ the journal.

[^1]: https://github.com/kata-containers/kata-containers/blob/3.10.1/src/agent/kata-agent.service.in

### DIFF
--- a/packages/nixos/kata.nix
+++ b/packages/nixos/kata.nix
@@ -50,7 +50,8 @@ in
       after = [ "systemd-tmpfiles-setup.service" ]; # Not upstream, but required for /etc/resolv.conf bind mount.
       serviceConfig = {
         Type = "exec"; # Not upstream.
-        StandardOutput = "tty";
+        StandardOutput = "journal+console";
+        StandardError = "inherit";
         ExecStart = "${lib.getExe pkgs.kata-agent}";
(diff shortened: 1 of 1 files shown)


katexochen:
what does the `[1/x]` mean? How is this related to the other PRs you just opened?

msanft:
> what does the [1/x] mean? How is this related to the other PRs you just opened?

My idea was to mark each PR that originates from the `burgerdev/gpu` branch, so that e.g. @burgerdev, who created (part of) the patches can match against that.

katexochen:
If things depend on each other, please base dependent branches on the PR they depend on. Otherwise I don't see how this is helpful.

msanft:
> If things depend on each other, please base dependent branches on the PR they depend on. Otherwise I don't see how this is helpful.

They don't depend on each other, so I've just removed the mark.

PR #1088, opened 2024-12-19
packages/kata-kernel: add GPU variant

msanft:
This adds a kernel build with GPU support to the `kata.kata-kernel` package.

### DIFF
--- a/packages/by-name/kata/kata-kernel-uvm/package.nix
+++ b/packages/by-name/kata/kata-kernel-uvm/package.nix
@@ -2,11 +2,13 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
 {
+  lib,
   fetchurl,
   linuxManualConfig,
   stdenvNoCC,
   fetchzip,
   kata,
+  withGPU ? false,
   ... # Required for invocation through `linuxPackagesFor`, which calls this with the `features` argument.
 }:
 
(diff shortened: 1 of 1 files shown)


packages/by-name/kata/kata-kernel-uvm/package.nix
```diff
@@ -2,11 +2,13 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
 {
+  lib,
   fetchurl,
   linuxManualConfig,
   stdenvNoCC,
   fetchzip,
   kata,
+  withGPU ? false,
```
burgerdev, line 11: Why do we need a separate kernel for GPUs?

msanft: This is aligned with how Kata handles this upstream. They build minimal kernels for each use-case, which is desirable from both a hardening perspective as well as the storage footprint.

burgerdev: Is the diff actually that big?

packages/by-name/kata/kata-kernel-uvm/package.nix
```diff
@@ -20,28 +22,51 @@ let
...
-        --replace-fail '# CONFIG_NLS_ISO8859_1 is not set' 'CONFIG_NLS_ISO8859_1=y' \
-        --replace-fail '# CONFIG_ATA is not set' 'CONFIG_ATA=y'
+            # Enable dm-init, so that we can use `dm-mod.create`.
+            cat <<EOF >> $config
+            CONFIG_MD=y
+            CONFIG_BLK_DEV_DM=y
+            CONFIG_DM_INIT=y
+            CONFIG_DM_CRYPT=y
+            CONFIG_DM_VERITY=y
+            CONFIG_DM_INTEGRITY=y
```
burgerdev, line 38: Don't we want these configs for the regular image, too?

packages/by-name/kata/kata-kernel-uvm/package.nix
```diff
@@ -20,28 +22,51 @@ let
...
-        --replace-fail '# CONFIG_TMPFS_XATTR is not set' 'CONFIG_TMPFS_XATTR=y' \
-        --replace-fail '# CONFIG_EFIVAR_FS is not set' 'CONFIG_EFIVAR_FS=y' \
-        --replace-fail '# CONFIG_RD_ZSTD is not set' 'CONFIG_RD_ZSTD=y' \
-        --replace-fail '# CONFIG_VFAT_FS is not se' 'CONFIG_VFAT_FS=y' \
-        --replace-fail '# CONFIG_NLS_CODEPAGE_437 is not set' 'CONFIG_NLS_CODEPAGE_437=y' \
-        --replace-fail '# CONFIG_NLS_ISO8859_1 is not set' 'CONFIG_NLS_ISO8859_1=y' \
-        --replace-fail '# CONFIG_ATA is not set' 'CONFIG_ATA=y'
+            # Enable dm-init, so that we can use `dm-mod.create`.
+            cat <<EOF >> $config
+            CONFIG_MD=y
```
burgerdev, line 33: Is this needed, or is `CONFIG_BLK_DEV_DM` enough?

packages/by-name/kata/kata-kernel-uvm/package.nix
```diff
@@ -20,28 +22,51 @@ let
...
+        else
+          ''
+            config=$(find . -regex '.*/config-[0-9.-]+-confidential')
 
-      echo "CONFIG_ATA_PIIX=y" >> $config
-    '';
+            # 1. Our kernel build is independent of any initrd.
+            # 2. Enable dm-init, so that we can use `dm-mod.create`.
+            substituteInPlace $config \
+              --replace-fail 'CONFIG_INITRAMFS_SOURCE="initramfs.cpio.gz"' 'CONFIG_INITRAMFS_SOURCE=""' \
```
burgerdev, line 48: Why is this not needed for the GPU image?

packages/by-name/kata/kata-kernel-uvm/package.nix
msanft: I wouldn't know why we should diverge from it, at least. 

And for the size of the diff: I can get you exact numbers, but I wouldn't expect it to be small. Consider driver-disablement (e.g. nouveau), PCI support, etc.

packages/by-name/kata/kata-kernel-uvm/package.nix
msanft: The GPU config doesn't have this set in the first place, so we can't replace it.

packages/by-name/kata/kata-kernel-uvm/package.nix
msanft: Looking at the diff, we _could_ use the GPU config for both. The real diff is about 20 lines. (not counting moved options, etc.)

However, I'd still prefer to keep close to upstream, as this means that we don't have to keep up with the changes to both files on each Kata release. Do you have a strong opinion @burgerdev @katexochen ?

burgerdev: I think what I missed is that the base config files are different for the two kernels - sorry.

packages/by-name/kata/kata-kernel-uvm/package.nix
msanft: These are from [upstream](https://github.com/kata-containers/kata-containers/blob/2c6126d3ab708e480b5aad1e7f7adbe22ffaa539/tools/packaging/kernel/configs/fragments/common/confidential_containers/cryptsetup.conf#L2). I'll add a comment indicating that.

burgerdev (approved):

PR #1091, opened 2024-12-19
Add {Metal,K3s}-QEMU-SNP-GPU platforms

msanft:
This adds a new `K3s-QEMU-SNP-GPU` platform that can be used for testing GPU-specific features on our on-prem infrastructure that uses this setup. This intentionally does not yet add any specific actions done if that platform is selected.

### DIFF
--- a/nodeinstaller/node-installer_test.go
+++ b/nodeinstaller/node-installer_test.go
@@ -36,15 +36,15 @@ func TestPatchContainerdConfig(t *testing.T) {
 			platform: platforms.AKSCloudHypervisorSNP,
 			expected: expectedConfAKSCLHSNP,
 		},
-		"BareMetalQEMUTDX": {
+		"K3sQEMUTDX": {
 			platform: platforms.K3sQEMUTDX,
 			expected: expectedConfBareMetalQEMUTDX,
 		},
-		"BareMetalQEMUSNP": {
+		"K3sQEMUSNP": {
 			platform: platforms.K3sQEMUSNP,
 			expected: expectedConfBareMetalQEMUSNP,
 		},
-		"BareMetalQEMUSNPGPU": {
+		"K3sQEMUSNPGPU": {
 			platform: platforms.K3sQEMUSNPGPU,
 			expected: expectedConfBareMetalQEMUSNPGPU,
 		},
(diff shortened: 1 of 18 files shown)


katexochen:
I thinks this should also add `Metal-QEMU-SNP-GPU` at the same time, which I expect to be needed.

nodeinstaller/node-installer_test.go
```diff
@@ -42,6 +42,10 @@ func TestPatchContainerdConfig(t *testing.T) {
 			platform: platforms.K3sQEMUSNP,
 			expected: expectedConfBareMetalQEMUSNP,
 		},
+		"BareMetalQEMUSNPGPU": {
```
katexochen, line 45:
```suggestion
		"K3sQEMUSNPGPU": {
```

msanft:
> I thinks this should also add `Metal-QEMU-SNP-GPU` at the same time, which I expect to be needed.

My initial thought was to split these to keep a minimal diff, but I also think combining these will make sense. Will do so!

katexochen:
Please also update .github/workflows/release.yml L301 so the runtime.yml is published for these platforms.

msanft:
> Please also update .github/workflows/release.yml L301 so the runtime.yml is published for these platforms.
Done :)

katexochen:
We also need to update packages/contrast-releases.nix

katexochen:
Release fails on this state: https://github.com/edgelesssys/contrast/actions/runs/12580504016/job/35062540645

katexochen:
opened https://github.com/edgelesssys/contrast/pull/1114

katexochen:
still failing: https://github.com/edgelesssys/contrast/actions/runs/12581815191/job/35066296786

packages/scripts.nix
```diff
@@ -259,7 +259,7 @@
           cp ${pkgs.microsoft.genpolicy.settings-coordinator}/genpolicy-settings.json .
           ${pkgs.microsoft.genpolicy}/bin/genpolicy < "$tmpdir/coordinator_base.yml"
         ;;
-        "metal-qemu-snp"|"k3s-qemu-snp"|"metal-qemu-tdx"|"k3s-qemu-tdx"|"rke2-qemu-tdx")
+        "metal-qemu-snp"|"k3s-qemu-snp"|"k3s-qemu-snp-gpu"|"metal-qemu-tdx"|"k3s-qemu-tdx"|"rke2-qemu-tdx")
```
katexochen, line 262: metal-qemu-snp-gpu is missing here.

katexochen (approved):

PR #1092, opened 2024-12-19
nixos: minimize GPU image

msanft:
This reduces the size of the GPU-enabled image to 2.2G, from over 3G previously. It does so by de-duplicating NVIDIA drivers and disabling unnecessary NVIDIA tools like `nvidia-settings`.

### DIFF
--- a/packages/nixos/gpu.nix
+++ b/packages/nixos/gpu.nix
@@ -10,6 +10,67 @@
 
 let
   cfg = config.contrast.gpu;
+
+  nvidiaPackage =
+    (
+      (config.boot.kernelPackages.nvidiaPackages.mkDriver {
+        # TODO(msanft): Investigate why the latest version breaks GPU containers.
+        version = "550.90.07";
+        sha256_64bit = "sha256-Uaz1edWpiE9XOh0/Ui5/r6XnhB4iqc7AtLvq4xsLlzM=";
+        sha256_aarch64 = "sha256-uJa3auRlMHr8WyacQL2MyyeebqfT7K6VU0qR7LGXFXI=";
+        openSha256 = "sha256-VLmh7eH0xhEu/AK+Osb9vtqAFni+lx84P/bo4ZgCqj8=";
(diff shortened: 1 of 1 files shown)


packages/nixos/gpu.nix
```diff
@@ -10,6 +10,68 @@
...
+            inherit (nvidiaPackage) version makeFlags;
+            src = oldAttrs.src // {
+              rev = nvidiaPackage.version;
+            };
+
+            postFixup = ''
+              # Save a copy of persistenced for mounting in containers
+              mkdir $out/origBin
+              cp $out/{bin,origBin}/nvidia-persistenced
+              patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2 $out/origBin/nvidia-persistenced
```
burgerdev, line 63: What's the interpreter before this?

packages/nixos/gpu.nix
```diff
@@ -10,6 +10,68 @@
...
+        # TODO(msanft): Clarify with upstream why that is the case.
+        passthru = oldAttrs.passthru // {
+          persistenced = oldAttrs.passthru.persistenced.overrideAttrs (oldAttrs: {
+            inherit (nvidiaPackage) version makeFlags;
+            src = oldAttrs.src // {
+              rev = nvidiaPackage.version;
+            };
+
+            postFixup = ''
+              # Save a copy of persistenced for mounting in containers
```
burgerdev, line 60: Why do we need the origBin?

packages/nixos/gpu.nix
msanft: This is copied from the upstream source.

packages/nixos/gpu.nix
msanft: We probably don't. But I would like to not diverge from the upstream source. Other packages in nixpkgs might expect this to be here.

packages/nixos/gpu.nix
burgerdev: That was not clear to me.

burgerdev (approved):

PR #1095, opened 2024-12-19
attestation: add name to Validator as unique identifier

jmxnzo:
This PR is a follow-up on #1027 and aims to enhance the logging of the different validation processes on the coordinator to reach better readability of the logs. 
-  Name was introduced as unique identifier in Validator struct and mapping to manifest.json, following the naming convention:
`<attestation>-<reference-values-index>[-<product>]`
The index corresponds to the position of the reference value separated by attestation type in manifest.json.
-  It was ensured that if all validators fail, the joined error message is printed and the formatting was reworked for readability. 
-  Single validation failures are printed on DEBUG only instead of ERROR level using the validator name and nonce.

### DIFF
--- a/coordinator/internal/authority/credentials.go
+++ b/coordinator/internal/authority/credentials.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/edgelesssys/contrast/internal/atls"
@@ -86,10 +87,11 @@ func (c *Credentials) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.A
 		return nil, nil, fmt.Errorf("generating SNP validation options: %w", err)
 	}
 
-	for _, opt := range opts {
+	for i, opt := range opts {
+		name := fmt.Sprintf("snp-%d-%s", i, strings.TrimPrefix(opt.VerifyOpts.Product.Name.String(), "SEV_PRODUCT_"))
 		validator := snp.NewValidatorWithReportSetter(opt.VerifyOpts, opt.ValidateOpts,
-			logger.NewWithAttrs(logger.NewNamed(c.logger, "validator"), map[string]string{"tee-type": "snp"}),
-			&authInfo)
+			logger.NewWithAttrs(logger.NewNamed(c.logger, "validator"), map[string]string{"reference-values": name}),
+			&authInfo, name)
 		validators = append(validators, validator)
 	}
 
@@ -98,9 +100,10 @@ func (c *Credentials) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.A
 		log.Error("Could not generate TDX validation options", "error", err)
 		return nil, nil, fmt.Errorf("generating TDX validation options: %w", err)
 	}
-	for _, opt := range tdxOpts {
+	for i, opt := range tdxOpts {
+		name := fmt.Sprintf("tdx-%d", i)
 		validators = append(validators, tdx.NewValidatorWithReportSetter(&tdx.StaticValidateOptsGenerator{Opts: opt},
-			logger.NewWithAttrs(logger.NewNamed(c.logger, "validator"), map[string]string{"tee-type": "tdx"}), &authInfo))
+			logger.NewWithAttrs(logger.NewNamed(c.logger, "validator"), map[string]string{"reference-values": name}), &authInfo, name))
 	}
 
 	serverCfg, err := atls.CreateAttestationServerTLSConfig(c.issuer, validators, c.attestationFailuresCounter)
--- a/internal/atls/atls.go
+++ b/internal/atls/atls.go
@@ -15,6 +15,7 @@ import (
 	"crypto/x509/pkix"
 	"encoding/asn1"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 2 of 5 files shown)


jmxnzo:
Wait for merge of #1082

internal/atls/atls.go
```diff
@@ -248,7 +250,13 @@ func verifyEmbeddedReport(validators []Validator, cert *x509.Certificate, peerPu
 				return nil
 			}
 			// Otherwise, we'll keep track of the error and continue with the next validator.
-			retErr = errors.Join(retErr, fmt.Errorf("validator %s failed: %w", validator.OID(), validationErr))
+			// The joined error should reveal the atls nonce once to maintain readability. Because the error is only revealed if all validators fail,
+			// we can implicitly include the nonce in the first validator error and the concatenate the other errors.
```
burgerdev, line 254: I think it would make more sense to first collect the errors with `Join`, and then wrap them with a message that contains the nonce before returning from this function.

internal/attestation/snp/validator.go
```diff
@@ -26,42 +26,45 @@ type Validator struct {
...
 		verifyOpts:   VerifyOpts,
 		validateOpts: ValidateOpts,
 		reportSetter: reportSetter,
 		logger:       log,
+		name:         name,
 	}
 }
 
-// OID returns the OID of the validator.
+// OID returns the root OID for the raw SNP report extension used by the validator.
```
burgerdev, line 55: Why `root`? And, if you are changing the docstring here, please also change it for TDX.

sdk/common.go
```diff
@@ -31,10 +33,11 @@ func ValidatorsFromManifest(kdsDir string, m *manifest.Manifest, log *slog.Logge
 	if err != nil {
 		return nil, fmt.Errorf("getting SNP validate options: %w", err)
 	}
-	for _, opt := range opts {
+	for i, opt := range opts {
 		opt.ValidateOpts.HostData = coordinatorPolicyChecksum
+		name := "snp-" + strconv.Itoa(i) + "-" + strings.TrimPrefix(opt.VerifyOpts.Product.Name.String(), "SEV_PRODUCT_")
```
burgerdev, line 38: `fmt.Sprintf` as above

coordinator/internal/authority/credentials.go
```diff
@@ -86,10 +88,11 @@ func (c *Credentials) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.A
 		return nil, nil, fmt.Errorf("generating SNP validation options: %w", err)
 	}
 
-	for _, opt := range opts {
+	for i, opt := range opts {
+		name := "snp-" + strconv.Itoa(i) + "-" + strings.TrimPrefix(opt.VerifyOpts.Product.Name.String(), "SEV_PRODUCT_")
 		validator := snp.NewValidatorWithReportSetter(opt.VerifyOpts, opt.ValidateOpts,
 			logger.NewWithAttrs(logger.NewNamed(c.logger, "validator"), map[string]string{"tee-type": "snp"}),
```
burgerdev, line 94: Now that we have a name in the constructor, we could remove the attrs here and add them there?

internal/attestation/snp/validator.go
jmxnzo: https://github.com/edgelesssys/contrast/blob/main/internal/oid/oid.go

burgerdev: Hm. I'd say it's technically a node, but not the root node.

coordinator/internal/authority/credentials.go
burgerdev: So, I'd say replace `map[string]string{"tee-type": "snp"}` with `map[string]string{"reference-values": name}`

internal/atls/atls.go
```diff
@@ -248,7 +250,7 @@ func verifyEmbeddedReport(validators []Validator, cert *x509.Certificate, peerPu
 				return nil
 			}
 			// Otherwise, we'll keep track of the error and continue with the next validator.
-			retErr = errors.Join(retErr, fmt.Errorf("validator %s failed: %w", validator.OID(), validationErr))
+			retErr = errors.Join(retErr, fmt.Errorf(" validator %s failed: %w ||", validator.String(), validationErr))
```
katexochen, line 253:
```suggestion
			retErr = errors.Join(retErr, fmt.Errorf(" validator %s failed: %w", validator.String(), validationErr))
```

internal/atls/atls.go
```diff
@@ -260,6 +262,11 @@ func verifyEmbeddedReport(validators []Validator, cert *x509.Certificate, peerPu
 		return ErrNoMatchingValidators
 	}
 
+	// The joined error should reveal the atls nonce once to maintain readability.
+	if retErr != nil {
+		retErr = fmt.Errorf("AtlsConnectionNonce: %s || %w", hex.EncodeToString(nonce), retErr)
```
katexochen, line 267:
```suggestion
		retErr = fmt.Errorf("with AtlsConnectionNonce %s: %w", hex.EncodeToString(nonce), retErr)
```

internal/atls/atls.go
jmxnzo: Imo, the double dash serves better readability of the joined error, because we have no specific formatting implemented for now. But i can drop this, if it seems unnecessary.

katexochen: errors are already joined using newlines: https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/errors/join.go;l=52-54
adding custom formatting here isn't idiomatic

PR #1103, opened 2024-12-23
nodeinstaller: add nydus-pull container

davidweisse:
This adds the `nydus-pull` container, which is running inside the Node Installer as a workaround for the `containerd` bug described in https://github.com/containerd/containerd/issues/8674. When using different snapshotters, such as the `nydus-snapshotter`, it may happen that some layers were previously pulled using K8s default snapshotter which can then cause the pod to run into this error.

The `nydus-pull` container watches for all pod events and tries to manually pull failing images that are likely to be caused by this bug. We saw this previously in the regression E2E tests on bare metal, specifically with the [redis](https://github.com/edgelesssys/contrast/blob/main/e2e/regression/testdata/redis.yml) container.

A successful run of the regression test can be found here: https://github.com/edgelesssys/contrast/actions/runs/12468508667
First, the `redis` container gets a `CreateContainerError`, which is then caught by the `nydus-pull` container. On the next backoff, the container starts normally.


nyduspull/main.go
katexochen, line 1: This should be moved into the tools directory and ideally have its own go module.

burgerdev (approved):
lgtm % comments

packages/by-name/contrast/package.nix
```diff
@@ -174,6 +174,7 @@ buildGoModule rec {
           fileset.unions [
             (path.append root "service-mesh")
             (path.append root "tools/tdx-measure")
+            (path.append root "tools/nydus-pull")
```
katexochen, line 177: maybe we can exclude tools entirely here?

packages/by-name/nydus-pull/package.nix
```diff
@@ -0,0 +1,35 @@
...
+
+  proxyVendor = true;
+  vendorHash = "sha256-UFrQPzK2/yWRI5kwGwMVfl3MGBKWAtQOCnpNOYapvQs=";
+
+  subPackages = [ "." ];
+
+  CGO_ENABLED = 0;
+  ldflags = [
+    "-s"
+    "-w"
```
katexochen, line 20:
```suggestion
```

tools/nydus-pull/main.go
```diff
@@ -0,0 +1,95 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+var version = "0.0.0-dev"
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Println("Usage: nydus-pull <runtime-handler>")
+		os.Exit(1)
```
katexochen, line 28: use log.Fatal instead.

burgerdev (approved):

tools/nydus-pull/main.go
```diff
@@ -0,0 +1,94 @@
...
+func main() {
+	if len(os.Args) != 2 {
+		log.Fatal("Usage: nydus-pull <runtime-handler>")
+	}
+
+	nodeName := os.Getenv("NODE_NAME")
+
+	if err := run(context.Background(), os.Args[1], nodeName); err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		os.Exit(1)
```
katexochen, line 34: here log.fatal, too.

katexochen (approved):
otherwise lgtm

PR #1104, opened 2024-12-23
e2e: allow usage of different PatchManifestFunc + add test for equal WorkloadSecretID 

jmxnzo:
This PR introduces the patchManifest() method in contrasttest.go to allow executing PatchManifestFunc on the current manifest, which obtain different modifications. 

- type PatchManifestFunc was introduced to allow using modification functions on the current manifest.
- existing patchReferenceValues (implements patchManifestFunc) was reduced to the modifcations of the reference values.
- patchWorkloadSecretID (implements patchManifestFunc)  was added to allow overwriting a certain workload secret.
- a workload secret test case was added, to check that setting the same WorkloadSecretID in two different deployments will result in the identical contrast-secrets.

### DIFF
--- a/e2e/internal/contrasttest/contrasttest.go
+++ b/e2e/internal/contrasttest/contrasttest.go
@@ -179,52 +179,61 @@ func (ct *ContrastTest) Generate(t *testing.T) {
 	require.NoError(err)
 	require.NotEmpty(hash, "expected apply to fill coordinator policy hash")
 
-	ct.patchReferenceValues(t, ct.Platform)
+	ct.PatchManifest(t, patchReferenceValues(ct.Platform))
 }
 
-// patchReferenceValues modifies the manifest to contain multiple reference values for testing
-// cases with multiple validators, as well as filling in bare-metal SNP-specific values.
-func (ct *ContrastTest) patchReferenceValues(t *testing.T, platform platforms.Platform) {
+// PatchManifestFunc defines a function type allowing the given manifest to be modified.
+type PatchManifestFunc func(manifest.Manifest) manifest.Manifest
+
+// PatchManifest modifies the current manifest by executing a provided PatchManifestFunc on it.
+func (ct *ContrastTest) PatchManifest(t *testing.T, patchFn PatchManifestFunc) {
 	manifestBytes, err := os.ReadFile(ct.WorkDir + "/manifest.json")
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


e2e/internal/contrasttest/contrasttest.go
```diff
@@ -179,52 +179,62 @@ func (ct *ContrastTest) Generate(t *testing.T) {
...
 
-	ct.patchReferenceValues(t, ct.Platform)
+	ct.PatchManifest(t, patchReferenceValues(ct.Platform))
 }
 
-// patchReferenceValues modifies the manifest to contain multiple reference values for testing
-// cases with multiple validators, as well as filling in bare-metal SNP-specific values.
-func (ct *ContrastTest) patchReferenceValues(t *testing.T, platform platforms.Platform) {
+// PatchManifestFunc defines a variadic function type allowing the current manifest to be modified,
+// based on the active platform and additional variadic arguments for extended customization.
```
burgerdev, line 186: Comment is outdated.

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -179,52 +179,62 @@ func (ct *ContrastTest) Generate(t *testing.T) {
...
+			// Duplicate the reference values to test multiple validators by having at least 2.
+			m.ReferenceValues.SNP = append(m.ReferenceValues.SNP, m.ReferenceValues.SNP[len(m.ReferenceValues.SNP)-1])
+
+			// Make the last set of reference values invalid by changing the SVNs.
+			m.ReferenceValues.SNP[len(m.ReferenceValues.SNP)-1].MinimumTCB = manifest.SNPTCB{
+				BootloaderVersion: toPtr(manifest.SVN(255)),
+				TEEVersion:        toPtr(manifest.SVN(255)),
+				SNPVersion:        toPtr(manifest.SVN(255)),
+				MicrocodeVersion:  toPtr(manifest.SVN(255)),
+			}
```
burgerdev, line 216: Not related to this PR, but I think this function does too many surprising things - not sure why we don't just unit test for the multiple validators case.

burgerdev (approved):

e2e/internal/contrasttest/contrasttest.go
jmxnzo: #1115 will rework the unit testing cases for the multiple validator case.

PR #1113, opened 2024-12-31
contrast-releases: fix runtime.yml conditional and demo shell

katexochen:
Fix in https://github.com/edgelesssys/contrast/commit/61b5f35286fbb679d4abe29510b718b12389ccb9 was broken.

Also the shell wasn't working anymore as intended, as runtime/coordinator weren't installed anymore.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -73,12 +73,14 @@
               toDemoShell =
                 version: contrast-release:
                 lib.nameValuePair "demo-${version}" (
-                  pkgs.mkShell {
+                  pkgs.mkShellNoCC {
                     packages = [ contrast-release ];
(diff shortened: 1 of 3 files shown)


msanft (approved):
Verified that the shell doesn't contain the coordinator / runtime specs before this fix, while it does contain them afterwards.

About the fix in `contrast-releases.nix`, I don't understand how it was wrong before exactly.

katexochen:
> About the fix in `contrast-releases.nix`, I don't understand how it was wrong before exactly.

It wouldn't include the runtime*/coordinator*yml for other platforms than the new metal ones.

PR #1114, opened 2025-01-02
ci: fix platform case statements

katexochen:
Looks like this was copied from justfile without syntax adjustment.

### DIFF
--- a/packages/scripts.nix
+++ b/packages/scripts.nix
@@ -253,7 +253,7 @@
 
       pushd "$tmpdir" >/dev/null
 
-      case $platform in
+      case "$platform" in
         "aks-clh-snp")
           cp ${pkgs.microsoft.genpolicy.rules-coordinator}/genpolicy-rules.rego rules.rego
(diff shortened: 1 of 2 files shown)


burgerdev (approved):

packages/scripts.nix
```diff
@@ -265,7 +265,7 @@
           ${pkgs.kata.genpolicy}/bin/genpolicy < "$tmpdir/coordinator_base.yml"
         ;;
         *)
-          echo "Unsupported platform: {{ platform }}" >&2
+          echo "Unsupported platform: $platform" >&2
```
msanft, line 268:
```suggestion
          echo "Unsupported platform: ${platform}" >&2
```
nit

msanft (approved):

packages/scripts.nix
katexochen: That's not working here as this is an inline string in nix.

edgelessci:
- #1116

PR #1115, opened 2025-01-02
e2e: use real reference values on bare-metal runners

jmxnzo:
Until now we are using the generated reference values in our e2e tests on bare metal, which contain placeholder null values. This PR ensures, that the reference values of our e2e test correctly reflect the bare-metal runners (tdx & snp).

- A ConfigMap, persistently storing the missing reference values of our bare-metal runners was added to our e2e cluster. Thus reachable by any workflow ran on the e2e cluster.
- Reworked e2e testing logic of patchReferenceValues() as stated in [#1104](https://github.com/edgelesssys/contrast/pull/1104#discussion_r1900627363) by moving the invalidation of reference values into an own function

### DIFF
--- a/e2e/internal/contrasttest/contrasttest.go
+++ b/e2e/internal/contrasttest/contrasttest.go
@@ -30,6 +30,7 @@ import (
 	ksync "github.com/katexochen/sync/api/client"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // Flags contains the parsed Flags for the test.
@@ -178,8 +179,10 @@ func (ct *ContrastTest) Generate(t *testing.T) {
 	hash, err := os.ReadFile(path.Join(ct.WorkDir, "coordinator-policy.sha256"))
 	require.NoError(err)
 	require.NotEmpty(hash, "expected apply to fill coordinator policy hash")
-
-	ct.PatchManifest(t, patchReferenceValues(ct.Platform))
+	patchManifestFunc, err := patchReferenceValues(ct.Kubeclient, ct.Platform)
+	require.NoError(err)
+	ct.PatchManifest(t, patchManifestFunc)
+	ct.PatchManifest(t, addInvalidReferenceValues(ct.Platform))
 }
 
 // PatchManifestFunc defines a function type allowing the given manifest to be modified.
@@ -197,12 +200,11 @@ func (ct *ContrastTest) PatchManifest(t *testing.T, patchFn PatchManifestFunc) {
 	require.NoError(t, os.WriteFile(ct.WorkDir+"/manifest.json", manifestBytes, 0o644))
 }
 
-// patchReferenceValues returns a PatchManifestFunc which modifies a manifest to contain multiple reference values for testing
-// cases with multiple validators, as well as filling in bare-metal SNP-specific values.
-func patchReferenceValues(platform platforms.Platform) PatchManifestFunc {
+// addInvalidReferenceValues returns a PatchManifestFunc which adds a fresh, invalid entry to the specified reference values.
+func addInvalidReferenceValues(platform platforms.Platform) PatchManifestFunc {
 	return func(m manifest.Manifest) manifest.Manifest {
 		switch platform {
-		case platforms.AKSCloudHypervisorSNP:
+		case platforms.MetalQEMUSNP, platforms.MetalQEMUSNPGPU, platforms.K3sQEMUSNP, platforms.K3sQEMUSNPGPU, platforms.AKSCloudHypervisorSNP:
 			// Duplicate the reference values to test multiple validators by having at least 2.
 			m.ReferenceValues.SNP = append(m.ReferenceValues.SNP, m.ReferenceValues.SNP[len(m.ReferenceValues.SNP)-1])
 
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -79,7 +79,8 @@ jobs:
             --image-replacements workspace/just.containerlookup \
             --namespace-file workspace/e2e.namespace \
             --platform ${{ inputs.platform }} \
-            --namespace-suffix="-ci"
+            --namespace-suffix="-ci" \
+            --minimum-tcb-file ./e2e/internal/contrasttest/testdata/snp-minimum-tcb.json \
```
burgerdev, line 83:
```suggestion
            --minimum-tcb-file ./e2e/internal/contrasttest/testdata/snp-minimum-tcb.json
```

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -199,7 +202,21 @@ func (ct *ContrastTest) PatchManifest(t *testing.T, patchFn PatchManifestFunc) {
 
 // patchReferenceValues returns a PatchManifestFunc which modifies a manifest to contain multiple reference values for testing
 // cases with multiple validators, as well as filling in bare-metal SNP-specific values.
-func patchReferenceValues(platform platforms.Platform) PatchManifestFunc {
+func patchReferenceValues(platform platforms.Platform, tcbSpecificationFile string) (PatchManifestFunc, error) {
+	var baremetalRefVal manifest.ReferenceValues
+	if platform == platforms.K3sQEMUSNP || platform == platforms.MetalQEMUSNP {
```
burgerdev, line 207: We need this for TDX, too (`MinimumTeeTcbSvn`, `MrSeam`). Looks like just adding the platforms here would be sufficient, though.

e2e/internal/contrasttest/testdata/snp-minimum-tcb.json
burgerdev, line 1: I don't think this should be a file under version control, because the test runner platform can change underneath, which would require backporting changes to this file to release branches. How about using [GitHub Action Vars](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-configuration-variables-for-multiple-workflows)?

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -199,7 +202,21 @@ func (ct *ContrastTest) PatchManifest(t *testing.T, patchFn PatchManifestFunc) {
...
+		// load minimumTCB specification for bare metal runners from path tcbSpecificationFile
+		tcbSpecs, err := os.ReadFile(tcbSpecificationFile)
```
burgerdev, line 209: Can we make the parameter optional? That would allow running the tests from an environment where we don't have the overrides file.

jmxnzo: An option would be to set the default minimumTCB values to zero in first place, as a fallback if the tcbSpecificationFile is not provided. Basically keeping the test logic from prior and adding the new logic in case the tcbSpecificationFile is set.

e2e/internal/contrasttest/contrasttest.go
jmxnzo: They were already hardcoded into the test logic, i will move this to use the Github Action Vars as well.  https://github.com/edgelesssys/contrast/blob/d457d4a01ade77b6b1cbf5fefdae475286b80ae9/e2e/internal/contrasttest/contrasttest.go#L251-L256

e2e/internal/contrasttest/testdata/snp-minimum-tcb.json
jmxnzo: Using GitHub Actions Vars resulted in either tolerating our just workflow to fail or having the an identical fallback to the Github Action Vars. We decided to keep a ConfigMap on the e2e cluster, which keeps the missing values of our baremetal runners.

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -213,27 +214,69 @@ func patchReferenceValues(platform platforms.Platform) PatchManifestFunc {
...
+	if platform != platforms.AKSCloudHypervisorSNP {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+		tcbSpecs, err := k.GetConfigMapData(ctx, "default", "bm-tcb-specs", "tcb-specs.json")
+		if err != nil {
+			return nil, fmt.Errorf("Failed getting ConfigMap: bm-tcb-specs with error: %w", err)
+		}
+		err = json.Unmarshal(tcbSpecs, &baremetalRefVal)
+		if err != nil {
+			return nil, fmt.Errorf("Failed unmarshaling minimum tcb specification from ConfigMap: bm-tcb-specs with error: %w", err)
```
burgerdev, line 243:
```suggestion
			return nil, fmt.Errorf("unmarshaling reference values: %w", err)
```

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -213,27 +214,69 @@ func patchReferenceValues(platform platforms.Platform) PatchManifestFunc {
...
+// based on the 'bm-tcb-specs' ConfigMap persistently stored in the 'default' namespace.
+func patchReferenceValues(k *kubeclient.Kubeclient, platform platforms.Platform) (PatchManifestFunc, error) {
+	var baremetalRefVal manifest.ReferenceValues
+	// ConfigMap bm-tcb-specs will only exist on baremetal instances.
+	if platform != platforms.AKSCloudHypervisorSNP {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+		tcbSpecs, err := k.GetConfigMapData(ctx, "default", "bm-tcb-specs", "tcb-specs.json")
+		if err != nil {
+			return nil, fmt.Errorf("Failed getting ConfigMap: bm-tcb-specs with error: %w", err)
```
burgerdev, line 239:
```suggestion
			return nil, fmt.Errorf("getting ConfigMap bm-tcb-specs: %w", err)
```

e2e/internal/kubeclient/kubeclient.go
```diff
@@ -221,3 +221,13 @@ func (c *Kubeclient) logContainerStatus(pod corev1.Pod) {
 		}
 	}
 }
+
+// GetConfigMapData returns the byte representation of the value stored under the provided key in the data section of the ConfigMap.
+func (c *Kubeclient) GetConfigMapData(ctx context.Context, namespace, name, key string) ([]byte, error) {
```
burgerdev, line 226: I don't think this needs to be a function - it's only two lines effectively, and the error handling is now duplicated here and at the caller. The logging is also not strictly needed, considering that the function returns an error and the test won't poroceed anyway.

burgerdev (approved):

PR #1118, opened 2025-01-02
ci/release: allow setting main ref for testing

katexochen:
Some workflows are executed from main branch, which makes the workflow difficult to test. This PR adds the ability to set the ref explicitly so it can be pointed to the same branch when testing the workflow.


msanft (approved):

edgelessci:
- #1119

PR #1122, opened 2025-01-07
generate: add flag to skip service mesh injection

katexochen:
This allows to keep existing service mesh parts including modifications.


davidweisse (commented):
Can't we just remove the annotation in the YAML which is responsible for injecting the service mesh? Any existing service mesh container should remain unchanged on `generate`.

katexochen:
> Can't we just remove the annotation in the YAML which is responsible for injecting the service mesh? Any existing service mesh container should remain unchanged on `generate`.

I think we can still add this flag for consistency and faster toggle. wdyt @burgerdev?

burgerdev (approved):
Not a common use case, but on the other hand I don't see a downside.

PR #1134, opened 2025-01-10
treewide: pass CGO_ENABLED via env

msanft:
Since https://github.com/NixOS/nixpkgs/commit/905dc8d978b38b0439905cb5cd1faf79163e1f14, it is necessary to pass the Go build options, such as `CGO_ENABLED` via the environment. So proactively comply with that and pass them via the environment before the deprecation window ends.

### DIFF
--- a/packages/by-name/contrast/package.nix
+++ b/packages/by-name/contrast/package.nix
@@ -200,7 +200,7 @@ buildGoModule rec {
     install -D ${kata.genpolicy.settings-dev}/genpolicy-settings.json cli/genpolicy/assets/genpolicy-settings-kata.json
   '';
 
-  CGO_ENABLED = 0;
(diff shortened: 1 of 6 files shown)


packages/by-name/contrast/package.nix
```diff
@@ -211,7 +211,7 @@ buildGoModule rec {
   tags = [ "contrast_unstable_api" ];
 
   preCheck = ''
-    export CGO_ENABLED=1
+    export env.CGO_ENABLED=1
```
katexochen, line 214:
```suggestion
    export CGO_ENABLED=1
```

msanft:
Causes an overlap right now that will only be removed with a nixpkgs update, so closing this for now.

PR #1135, opened 2025-01-10
Add bare-metal GPU runtime class

msanft:
This adds a runtime class for the local just-based deployments as well as the release artifacts that corresponds to the GPU-enabled runtime for Contrast on bare-metal platforms.


packages/by-name/kata/contrast-node-installer-image/package.nix
```diff
@@ -38,79 +45,82 @@ let
...
+                url = "file:///opt/edgeless/snp/share/qemu/efi-virtio.rom";
+                path = "/opt/edgeless/@@runtimeName@@/snp/share/qemu/efi-virtio.rom";
+              }
+              {
+                url = "file:///bin/nydus-overlayfs";
+                path = "/opt/edgeless/@@runtimeName@@/bin/nydus-overlayfs";
+                executable = true;
+              }
+            ]
+            ++ lib.optionals (!withGPU) [
```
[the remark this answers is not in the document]
msanft: Yeah. To be honest, looking at it again, these binaries are tiny anyway. We should just include them regardless and have a little less complexity here.

packages/by-name/kata/contrast-node-installer-image/package.nix
```diff
@@ -91,6 +88,20 @@ let
...
+            }
+            {
+              url = "file:///opt/edgeless/tdx/share/OVMF.fd";
+              path = "/opt/edgeless/@@runtimeName@@/tdx/share/OVMF.fd";
+            }
+            {
+              url = "file:///opt/edgeless/tdx/bin/qemu-system-x86_64";
+              path = "/opt/edgeless/@@runtimeName@@/tdx/bin/qemu-system-x86_64";
+              executable = true;
+            }
```
burgerdev, line 104: It looks like this is just reordering items - is that intentional?

burgerdev (approved):
lgtm, besides the diff in `contrast-node-install.json`.

packages/by-name/kata/contrast-node-installer-image/package.nix
msanft: No, this is a byproduct of splitting out the TDX-specific stuff and merging it again. Will adjust

packages/containers.nix
```diff
@@ -198,5 +198,9 @@ containers
   push-node-installer-kata =
     pushOCIDir "push-node-installer-kata" pkgs.kata.contrast-node-installer-image
       "v${pkgs.contrast.version}";
+  push-node-installer-kata-gpu = pushOCIDir "push-node-installer-kata-gpu" (
+    pkgs.kata.contrast-node-installer-image.override
+    { withGPU = true; }
```
katexochen, line 203:
```suggestion
    pkgs.kata.contrast-node-installer-image.gpu
```
Can't we use the passedthru gpu attribute here?

katexochen (approved):

PR #1136, opened 2025-01-10
e2e: add GPU test

msanft:
This adds an E2E test for GPU use on Contrast.
It currently runs on the GPU-enabled bare-metal SNP runner.

The test currently only verifies that the GPU is available via
nvidia-smi, which also verifies that driver and CUDA work correctly.

### DIFF
--- a/.github/workflows/e2e_manual.yml
+++ b/.github/workflows/e2e_manual.yml
@@ -10,6 +10,7 @@ on:
         options:
           - genpolicy
           - getdents
+          - gpu
           - openssl
           - policy
           - regression
@@ -24,6 +25,7 @@ on:
         options:
           - AKS-CLH-SNP
           - K3s-QEMU-SNP
+          - K3s-QEMU-SNP-GPU
           - K3s-QEMU-TDX
       skip-undeploy:
         description: "Skip undeploy"
--- a/.github/workflows/e2e_nightly.yml
+++ b/.github/workflows/e2e_nightly.yml
@@ -19,7 +19,16 @@ jobs:
           - name: K3s-QEMU-TDX
             runner: TDX
             self-hosted: true
+          - name: K3s-QEMU-SNP-GPU
+            runner: SNP
+            self-hosted: true
         test-name: [servicemesh, openssl, policy, workloadsecret, volumestatefulset]
+        include:
+          - platform:
+              name: K3s-QEMU-SNP-GPU
+              runner: SNP
+              self-hosted: true
+            test-name: [gpu]
       fail-fast: false
     name: "${{ matrix.platform.name }}"
     uses: ./.github/workflows/e2e.yml
--- a/e2e/gpu/testdata/gpu-pod.yaml
+++ b/e2e/gpu/testdata/gpu-pod.yaml
@@ -1,0 +1,25 @@
+# TODO(msanft): Move this to internal/kuberesource/sets.go as soon as genpolicy
+# support for GPU pods is added.
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gpu-pod
+  namespace: "@@REPLACE_NAMESPACE@@"
+  annotations:
+    # Allow-all policy
+    # TODO(msanft): Generate a policy dynamically once we support policy generation for GPU pods.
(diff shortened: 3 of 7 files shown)


e2e/gpu/testdata/gpu-pod.yaml
katexochen, line 1: Please add a todo to move this into resourcegen as soon as genpolicy support is there.

e2e/gpu/testdata/gpu-pod.yaml
```diff
@@ -0,0 +1,23 @@
...
+    io.katacontainers.config.hypervisor.default_memory: "15258"
+    cdi.k8s.io/gpu: "nvidia.com/pgpu=0"
+spec:
+  runtimeClassName: "@@REPLACE_RUNTIME@@"
+  restartPolicy: OnFailure
+  containers:
+    - name: vllm
+      image: docker.io/library/ubuntu:latest
```
burgerdev, line 17: Please don't use docker.io images for e2e - these are often rate-limited and lead to sporadic test failures. quay.io is a good alternative, or ghcr.io. And we should be using a stable tag.

.github/workflows/e2e_on_pull_request.yml
```diff
@@ -25,6 +25,9 @@ jobs:
           - name: K3s-QEMU-SNP
             runner: SNP
             self-hosted: true
+          - name: K3s-QEMU-SNP-GPU
+            runner: SNP
+            self-hosted: true
```
burgerdev, line 30: Do we need to run this test on every PR? Contention for the 1 GPU could be quite high.

flake.nix
```diff
@@ -8,7 +8,7 @@
       # see https://github.com/renovatebot/renovate/issues/29721 and
       # https://github.com/renovatebot/renovate/blob/743fed0ec6ca5810e274571c83fa6d4f5213d4e7/lib/modules/manager/nix/extract.ts#L6.
       # We must keep the following string in the file for renovate to work: "github:NixOS/nixpkgs/nixpkgs-unstable"
-      url = "github:NixOS/nixpkgs?ref=nixos-unstable";
+      url = "github:msanft/nixpkgs/msanft/libnvidia-container/driver-bins";
```
burgerdev, line 11: The comment above does not apply anymore, and it would be good to have a note that says how we get back to unstable.

msanft: This commit is to be removed, see the message

.github/workflows/e2e_on_pull_request.yml
msanft: I also thought about giving this a more fine-grained trigger. But that's something you and @katexochen will need to decide.

e2e/gpu/testdata/gpu-pod.yaml
msanft: Do you have another base image recommendation that has official sources on ghcr or quay? I'm definitely with you on this, but neither of the registries seem to have an official ubuntu image. We could be using any other base image though.

e2e/gpu/testdata/gpu-pod.yaml
katexochen: Checkout openssl e2e test/just workflow and internal/kuberesource/sets.go.
Regression test is somewhat special, no good orientation.

e2e/gpu/testdata/gpu-pod.yaml
katexochen: You could also just push that ubuntu image to our own reg.

burgerdev: Yes, that's what we did for a couple others: https://github.com/edgelesssys/contrast/blob/662af0c011bd68bb09b78f8bf96427d84250a0b1/e2e/regression/testdata/redis.yml#L18.

.github/workflows/e2e_on_pull_request.yml
msanft: Decided in-person with @katexochen to not run this on PRs due to the waiting time.

.github/workflows/e2e_manual.yml
```diff
@@ -10,6 +10,7 @@ on:
         options:
           - genpolicy
           - getdents
+          - gpu
```
burgerdev, line 13: Should this also go into nightly and/or regression test workflows?

burgerdev (approved):

.github/workflows/e2e_manual.yml
msanft: @katexochen do you have a stance on that? I think there should be no problem running it on nightly.

katexochen: yes, run it nightly

.github/workflows/e2e_nightly.yml
```diff
@@ -34,6 +34,28 @@ jobs:
...
       packages: write
 
+  gpu-test-matrix:
+    strategy:
+      matrix:
+        platform:
+          - name: K3s-QEMU-SNP-GPU
+            runner: SNP
+            self-hosted: true
+        test-name: [servicemesh, openssl, policy, workloadsecret, volumestatefulset, gpu]
```
katexochen, line 44: Do we really need to run all those tests for GPU?

msanft: I would like to also run tests that exercise service mesh functionality, etc. on the GPU runtime class.

katexochen:
commit message should be `e2e: add GPU test`

.github/workflows/e2e_nightly.yml
```diff
@@ -34,6 +34,28 @@ jobs:
       contents: read
       packages: write
 
+  gpu-test-matrix:
```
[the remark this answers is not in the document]
msanft: Because the GPU test can't run on non-GPU hardware, and because I *also* want the other tests to run on GPU hardware / runtime class.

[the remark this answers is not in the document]
katexochen: That can be done in a single matrix,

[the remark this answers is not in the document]
msanft: Should we add an extra-tests field to the `platform` matrix? How would that look like?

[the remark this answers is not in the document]
katexochen: See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow or constellation. If in doubt, drop everything and use only fully specified `include`.

katexochen (approved):

PR #1139, opened 2025-01-10
microsoft.genpolicy: support volume mounting

jmxnzo:
This PR adds a patch to support volume mounting for the genpolicy binary of the microsoft fork, which is necessary for volumestatefulset and mysql deployments. Updating the release of the microsoft fork in https://github.com/edgelesssys/contrast/commit/a3496ccfaac275f8272fa91fd8b8303081b6bc39 resulted in introducing the required change.

### DIFF
--- a/packages/by-name/microsoft/genpolicy/package.nix
+++ b/packages/by-name/microsoft/genpolicy/package.nix
@@ -71,6 +71,12 @@ rustPlatform.buildRustPackage rec {
       # This avoids printing the entire annotation on log level debug, which resulted in errors of the logtranslator.go
       # TODO(jmxnzo): remove when https://github.com/kata-containers/kata-containers/pull/10647 is picked up by microsoft/kata-containers fork
       ./0009-genpolicy-do-not-log-policy-annotation-in-debug.patch
+      # Patches the RootfsPropagation check in allow_create_container_input to allow setting up bidirectional volumes, which need to propagate their changes to a
+      # volume mounted on the root filesystem and possibly shared across multiple containers on the host.
+      # RootfsPropagation describes the mapping to mount propagations: https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation
+      # It reflects genpolicy-support-mount-propagation-and-ro-mounts.patch on upstream kata.genpolicy, but drops the patched propagation mode
+      # derivation, because it was already built in to the microsoft fork.
+      ./0010-genpolicy-support-mount-propagation-and-ro-mounts.patch
     ];
   };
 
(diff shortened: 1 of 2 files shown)


packages/by-name/microsoft/genpolicy/package.nix
```diff
@@ -71,6 +71,10 @@ rustPlatform.buildRustPackage rec {
       # This avoids printing the entire annotation on log level debug, which resulted in errors of the logtranslator.go
       # TODO(jmxnzo): remove when https://github.com/kata-containers/kata-containers/pull/10647 is picked up by microsoft/kata-containers fork
       ./0009-genpolicy-do-not-log-policy-annotation-in-debug.patch
+      # Patches the RootfsPropagation check in allow_create_container_input to allow setting up volumes. It reflects
```
katexochen, line 74: How exactly are RootfsPropagation and volume setup connected?

packages/by-name/microsoft/genpolicy/package.nix
```diff
@@ -71,6 +71,10 @@ rustPlatform.buildRustPackage rec {
...
+      # 0012-genpolicy-support-mount-propagation-and-ro-mounts.patch on upstream kata repo, but drops the patched propagation mode
```
katexochen, line 75:
```suggestion
      # genpolicy-support-mount-propagation-and-ro-mounts.patch on kata.genpolicy, but drops the patched propagation mode
```
patch numbers change over time.

packages/by-name/microsoft/genpolicy/package.nix
```diff
@@ -71,6 +71,10 @@ rustPlatform.buildRustPackage rec {
...
+      # derivation, because it was already built in to the microsoft/fork.
```
katexochen, line 76:
```suggestion
      # derivation, because it was already built in to the microsoft fork.
```

packages/by-name/microsoft/genpolicy/package.nix
burgerdev: The second part of the suggestion was not applied, but should be.

packages/by-name/microsoft/genpolicy/package.nix
burgerdev: Volume propagation allows creating a mount point in one container (== mount namespace) and accessing it in another. Setting the container's [volumeMount to `bidirectional` ](https://github.com/edgelesssys/contrast/blob/662af0c011bd68bb09b78f8bf96427d84250a0b1/internal/kuberesource/sets.go#L536) makes containerd add the `RootfsPropagation` to the pod. I'm not sure if it's strictly necessary - I'd have expected only the mount point to need propagation - but it should also not harm because the container rootfs is not mounted elsewhere.

That being said, I still owe an upstream issue and PR for this - will start with that once I have some cycles.

burgerdev (approved):

PR #1140, opened 2025-01-10
initializer/cryptsetup: rework bash entrypoint

jmxnzo:
Similarly to #1161 , this PR serves as a preliminary step toward moving cryptsetup to a subcommand of the initializer binary. Before translating the Bash entrypoint (used in the initializer binary after merging #1161) into Go code, the script was reworked to allow a cleaner integration of the volume encryption.
There were a few peculiarities in the cryptsetup LUKS standardization that have not been fully addressed in the current version of the entrypoint:
-  **key** is equivalent to **passphrase** (see the LUKS EXTENSION section in the [cryptsetup manual](https://man7.org/linux/man-pages/man8/cryptsetup.8.html)).
-  When using the `--key-file` flag with cryptsetup, the file's contents are unexpectedly treated as a passphrase rather than a cryptographic key.  
- Using the keyfile as passphrase won't result in higher entropy of the encryption key. As a result, we can directly use the workload secret and UUID without performing key derivation, because there is no straightforward way to avoid the built-in PBKDF for LUKS in cryptsetup.

e2e/volumestatefulset: https://github.com/edgelesssys/contrast/actions/runs/12764474415/job/35576714135

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -557,32 +557,20 @@ set -e
 device="/dev/csi0"
 # workload_secret_path is the path to the Contrast workload secret.
 workload_secret_path="/contrast/secrets/workload-secret-seed"
-# tmp_key_path is the path to a temporary key file.
-tmp_key_path="/dev/shm/key"
-# disk_encryption_key_path is the path to the disk encryption key.
-disk_encryption_key_path="/dev/shm/disk-key"
 
-# If the device is not already a LUKS device, format it.
 if ! cryptsetup isLuks "${device}"; then
-	# Generate a random key for the first initialization.
-	dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 -w0 > "${tmp_key_path}"
-	cryptsetup luksFormat --pbkdf-memory=10240 $device "${tmp_key_path}" </dev/null
+	# cryptsetup derives the encryption key of the master key using PBKDF2 based on
+	# the workloadSecret as passphrase and a random 32 byte salt (stored in LUKS header)
+	# which ensures uniqueness of encryption key per disk.
+	cryptsetup luksFormat --pbkdf-memory=10240 $device "${workload_secret_path}" </dev/null
+	cryptsetup open "${device}" state -d "${workload_secret_path}"
 
-	uuid=$(blkid "${device}" -s UUID -o value)
-	openssl kdf -keylen 32 -kdfopt digest:SHA2-256 -kdfopt hexkey:$(cat "${workload_secret_path}") \
-		-kdfopt info:${uuid} -binary HKDF | base64 -w0 > "${disk_encryption_key_path}" 2>/dev/null
-	cryptsetup luksChangeKey --pbkdf-memory=10240 "${device}" --key-file "${tmp_key_path}" "${disk_encryption_key_path}"
-	cryptsetup open "${device}" state -d "${disk_encryption_key_path}"
+	# Create the ext4 filesystem on the mapper device.
 	mkfs.ext4 /dev/mapper/state
-	cryptsetup close state
+else
+	cryptsetup open "${device}" state -d "${workload_secret_path}"
 fi
 
-# No matter if this is the first initialization derive the key (again) and open the device.
-uuid=$(blkid "${device}" -s UUID -o value)
-openssl kdf -keylen 32 -kdfopt digest:SHA2-256 -kdfopt hexkey:$(cat "${workload_secret_path}") \
-	-kdfopt info:${uuid} -binary HKDF | base64 -w0 > "${disk_encryption_key_path}" 2>/dev/null
-cryptsetup luksUUID "${device}"
-cryptsetup open "${device}" state -d "${disk_encryption_key_path}"
 mount /dev/mapper/state /state
 touch /done
 sleep inf


internal/kuberesource/parts.go
```diff
@@ -553,36 +553,44 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
...
-	dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 -w0 > "${tmp_key_path}"
+	# First init requires formatting to LUKS to allow getting the uuid of LUKS block device.
+	echo "init_passphrase" > "${tmp_key_path}"
 	cryptsetup luksFormat --pbkdf-memory=10240 $device "${tmp_key_path}" </dev/null
 
-	uuid=$(blkid "${device}" -s UUID -o value)
-	openssl kdf -keylen 32 -kdfopt digest:SHA2-256 -kdfopt hexkey:$(cat "${workload_secret_path}") \
-		-kdfopt info:${uuid} -binary HKDF | base64 -w0 > "${disk_encryption_key_path}" 2>/dev/null
+	# Generate passphrase based on workload secret and uuid of LUKS device.
+	echo "$(cat "${workload_secret_path}")$(cryptsetup luksUUID "${device}")" > "${disk_encryption_key_path}"
```
burgerdev, line 577:
```suggestion
	cat "${workload_secret_path}" > "${disk_encryption_key_path}"
	cryptsetup luksUUID "${device}" >> "${disk_encryption_key_path}"
```
... is a bit easier to read and does not need to care about `echo` peculiarities. Applies to the other if branch, too.

burgerdev (commented):
lgtm, but I'd like to get a second opinion from @3u13r.

internal/kuberesource/parts.go
```diff
@@ -553,36 +553,44 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
...
-	# Generate a random key for the first initialization.
-	dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 -w0 > "${tmp_key_path}"
+	# First init requires formatting to LUKS to allow getting the uuid of LUKS block device.
+	echo "init_passphrase" > "${tmp_key_path}"
 	cryptsetup luksFormat --pbkdf-memory=10240 $device "${tmp_key_path}" </dev/null
 
-	uuid=$(blkid "${device}" -s UUID -o value)
-	openssl kdf -keylen 32 -kdfopt digest:SHA2-256 -kdfopt hexkey:$(cat "${workload_secret_path}") \
-		-kdfopt info:${uuid} -binary HKDF | base64 -w0 > "${disk_encryption_key_path}" 2>/dev/null
+	# Generate passphrase based on workload secret and uuid of LUKS device.
```
burgerdev, line 576: Prior discussion for reference: https://github.com/edgelesssys/contrast/pull/1153#discussion_r1916087727

jmxnzo:
As described in [Figure 4: pseudo code for key creation](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/LUKS-standard/on-disk-format.pdf ) each PBKDF2 derived encryption key used to encrypt the AFSplit of the master key, is derived by the passphrase and a 32 byte salt from a random source.
For a key slot, all parameters how to decrypt its key material with a given user password are
stored in the phdr (f.e. salt, iteration depth) with the corresponding encrypted master key stripes. Thus the random salt fulfills the same function as including the UUID into our passphrase and similarly adds the uniqueness of the disk to the encryption key. Therefore to maintain readability, we dropped the logic of doing the first key initialization to get the UUID and rekeying afterwards. @3u13r Please give feedback on this code change again!

internal/kuberesource/parts.go
```diff
@@ -552,37 +552,28 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
 // container to setup an encrypted LUKS mount.
 func CryptsetupInitCommand() string {
 	return `#!/bin/bash
-set -e
+set -ex
+
+# Regarding https://man7.org/linux/man-pages/man8/cryptsetup.8.html
+# passphrase <=> key in this document, not an actual cryptographic key
```
3u13r, line 558: I think the intend of the comment is not clear. Please be more descriptive in what you are trying to communicate here.

internal/kuberesource/parts.go
```diff
@@ -552,37 +552,28 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
 // container to setup an encrypted LUKS mount.
 func CryptsetupInitCommand() string {
 	return `#!/bin/bash
-set -e
+set -ex
```
3u13r, line 555: I assume that the addition of -x is intentional, since it is now safe to do so, since we don't print the passphrase anymore.
Though we have to be careful, if the script changes the next time, should we include a comment that describes why this is safe in this case, if we decide to keep it?

3u13r (changes requested):

internal/kuberesource/parts.go
jmxnzo: I added a comment for description. We'll keep the init command transitionally, allthough #1153 will add a subcommand to the initializer creating the encrypted volume and thus reduce importance for this.

3u13r (approved):
LGTM

PR #1142, opened 2025-01-13
service-mesh: test readiness with exec probe

burgerdev:
The startup probe for the mesh proxy is redirected onto the same port, causing Envoy to open connections to itself recursively. Unfortunately, this is not visible in the Envoy logs - I only discovered this condition while working on a custom proxy implementation.

Things that are still unclear:

* There is some race condition that causes the redirection loop to only occur ~50% of the time.
* Switching on debug logging makes this condition much rarer.
* Exempting the TPROXY ports with iptables should make the SO_ORIGINAL_DST lookup fail, but the loop happens nonetheless.
* The issue reproduces with opening and holding a single connection, while opening another, but not with just one connection.

Tests:
* [x] https://github.com/edgelesssys/contrast/actions?query=branch%3Aburgerdev%2Fmesh-probe+event%3Aworkflow_dispatch++

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -469,8 +469,8 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
 			WithInitialDelaySeconds(1).
 			WithPeriodSeconds(5).
 			WithFailureThreshold(5).
-			WithTCPSocket(TCPSocketAction().
-				WithPort(intstr.FromInt(15006))),
+			WithExec(ExecAction().
+				WithCommand("test", "-f", "/ready")),
 		).
 		WithArgs(
 			"-l", "debug",
--- a/internal/kuberesource/wrappers.go
+++ b/internal/kuberesource/wrappers.go
@@ -336,6 +336,11 @@ func TCPSocketAction() *applycorev1.TCPSocketActionApplyConfiguration {
 	return applycorev1.TCPSocketAction()
 }
 
+// ExecAction creates a new ExecActionApplyConfiguration.
+func ExecAction() *applycorev1.ExecActionApplyConfiguration {
+	return applycorev1.ExecAction()
+}
+
 // RuntimeClassConfig wraps applypodsv1.RuntimeClassApplyConfiguration for a runtime class.
 type RuntimeClassConfig struct {
 	*applynodev1.RuntimeClassApplyConfiguration
--- a/packages/containers.nix
+++ b/packages/containers.nix
@@ -108,13 +108,14 @@ let
       name = "service-mesh-proxy";
       tag = "v${pkgs.service-mesh.version}";
       copyToRoot = with pkgs; [
+        busybox
         envoy
         iptables-legacy
       ];
       config = {
         # Use Entrypoint so we can append arguments.
         Entrypoint = [ "${pkgs.service-mesh}/bin/service-mesh" ];
-        Env = [ "PATH=/bin" ]; # This is only here for policy generation.
+        Env = [ "PATH=/bin" ];
       };
     };
 
--- a/service-mesh/main.go
+++ b/service-mesh/main.go
@@ -59,6 +59,11 @@ func run() (retErr error) {
 		return fmt.Errorf("failed to set up iptables rules: %w", err)
 	}
 
+	// Signal readiness for startup probe.
+	if err := os.WriteFile("/ready", nil, 0o644); err != nil {
+		return err
+	}
+
 	// execute the envoy binary
 	envoyBin, err := exec.LookPath("envoy")
 	if err != nil {


packages/containers.nix
```diff
@@ -108,13 +108,14 @@ let
...
       copyToRoot = with pkgs; [
+        busybox
         envoy
         iptables-legacy
       ];
       config = {
         # Use Entrypoint so we can append arguments.
         Entrypoint = [ "${pkgs.service-mesh}/bin/service-mesh" ];
-        Env = [ "PATH=/bin" ]; # This is only here for policy generation.
+        Env = [ "PATH=/bin" ];
```
msanft, line 118: nit, as it was here before already. But one could theoretically use [`buildEnv`](https://noogle.dev/f/pkgs/buildEnv) here.

service-mesh/main.go
```diff
@@ -59,6 +59,11 @@ func run() (retErr error) {
 		return fmt.Errorf("failed to set up iptables rules: %w", err)
 	}
 
+	// Signal readiness for startup probe.
```
3u13r, line 62: We signal readiness, before we are actually ready here, could we use: https://www.envoyproxy.io/docs/envoy/latest/operations/admin.html#get--ready in the yaml as an HTTP probe?

3u13r (changes requested):

internal/kuberesource/parts.go
```diff
@@ -469,8 +469,8 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
 			WithInitialDelaySeconds(1).
 			WithPeriodSeconds(5).
 			WithFailureThreshold(5).
-			WithTCPSocket(TCPSocketAction().
-				WithPort(intstr.FromInt(15006))),
+			WithExec(ExecAction().
+				WithCommand("test", "-f", "/ready")),
```
msanft, line 473: I'm fine with having the file-based solution, but is there any reasoning behind using this and not e.g. a `wait -S SIGCONT`? PID propagation to the Go process?

service-mesh/main.go
burgerdev: We don't want to expose the admin interface by default, do we? I don't think the probe works with localhost.

packages/containers.nix
burgerdev: How would `buildEnv` help here?

msanft: It would automagically give you an image with the correct PATH. See [this example](https://github.com/NixOS/nixpkgs/blob/fa39fa603cde60b27fa1a8b01c24b0be45362a6f/pkgs/build-support/docker/examples.nix#L86-L94)

internal/kuberesource/parts.go
burgerdev: What I'm trying to do here is to tell k8s that my container is ready. There's only a limited amount of ways to do so, and using the network probes does not work (in fact, is the bug I'm trying to solve). This leaves me with the exec probe.

I don't really understand what you're suggesting:

1. What is `wait -S`?
2. Who should call it?

msanft: I understand what you're doing with the exec probe, but I think that we're misusing the existence of the `/ready` file as a "poor man's IPC" here, while "real" (signal-based) IPC would come to my mind much more naturally here. I'm fine with the file-based solution, as the risk of races etc. would seem very low in this case. I just wondered about the file-based solution here, as it seemed a little unconventional to me.

To answer your questions:

1. `wait(1p)` waits for the specified signal
2. The execProbe, which would halt until the signal is received and then exit successfully

service-mesh/main.go
3u13r: Yeah exposing it locally might be fine, but it's generally added complexity.

packages/containers.nix
burgerdev: Afaiu `buildEnv` can help set symlinks, but does it also automagically set the `Env` image configuration? We need that, otherwise policy generation fails.

internal/kuberesource/parts.go
burgerdev: 1. Do you have a reference for this usage of wait? I can't find it in `man 1p wait`, nor in the POSIX spec.
2. I don't think probes are meant to be blocking.

service-mesh/main.go
burgerdev: What we could do: 

1. Always expose the admin server on `127/8`.
2. Add an HTTP server to the proxy that forwards only `/ready`.

internal/kuberesource/parts.go
msanft: Sorry for the confusion here. I got `trap` and `wait` scrambled around, but I swear there was some thing to wait for an explicit signal. Let's just go with the file-based approach here

internal/kuberesource/wrappers.go
```diff
@@ -336,6 +336,11 @@ func TCPSocketAction() *applycorev1.TCPSocketActionApplyConfiguration {
 	return applycorev1.TCPSocketAction()
 }
 
+// ExecAction creates a new ExecActionApplyConfiguration.
+func ExecAction() *applycorev1.ExecActionApplyConfiguration {
```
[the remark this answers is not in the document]
burgerdev: Consistency with the other probe types (above). We could decide not to wrap any of them, but maybe in another PR.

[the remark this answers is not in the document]
3u13r: Yeah, the same points apply for the other probe. For (some) other structs above we at least wrap them in our own struct and/or make pointer types out of struct types.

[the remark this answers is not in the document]
katexochen: Wrapped mostly for better readability at the call site.

3u13r (approved):
LGTM

katexochen (approved):

edgelessci:
- #1144

PR #1145, opened 2025-01-13
Post 1.3.0 release updates to main

edgelessci:
Updating main as part of the 1.3.0 release.


katexochen:
Manually deleted the GPU artifacts which won't be part of this release.

PR #1147, opened 2025-01-14
just: match both yaml and yml

msanft:
This fixes a little papercut I ran into when working with the `.custom` directory, where I wondered why my files weren't included. Turned out we only match against `.yml` while I used `.yaml`. This makes it so both endings are valid and included.


burgerdev (approved):

burgerdev (changes requested):
Wait, doesn't this result in a glob failure if there's either no `yml` or no `yaml`?

burgerdev:
That fails for me:
```
$ touch a.yaml
$ ls *.{yml,yaml}
ls: cannot access '*.yml': No such file or directory
 a.yaml
$ echo $?
2
```

msanft:
Oh, i missed the _either_. Do you have a suggestion on how to glob this instead? `*.y{a,}ml`?

burgerdev:
We could set `shopt -s nullglob` ([docs](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html)).

msanft:
Tested this with `yml` files only, and with both `yml` and `yaml`.

katexochen:
Please don't overengineer this. We use .yml by convention.

PR #1151, opened 2025-01-14
e2e: prevent download of streamed log directory

miampf:
This fixes [this error](https://github.com/edgelesssys/contrast/actions/runs/12703367171/job/35411042848#step:12:131). The error occurred because data was streamed into the export directory while it was made ready for download. The fix just copies the `export` directory to `export-no-stream`, which is then downloaded instead. This ensures that streaming is never interrupted and logs are never missed.


katexochen:
commit message `fix(e2e)` -> `e2e`

miampf:
> commit message `fix(e2e)` -> `e2e`

Renamed the commits, will squash and merge anyways.

burgerdev (approved):

PR #1152, opened 2025-01-14
resourcegen: add unique suffix to nodeinstaller clusterrole binding

davidweisse:
When running two Contrast deployments, the `ClusterRoleBinding` of one node installer gets overridden by the other, since they have the same name: `nodeinstaller-clusterrole-binding`. This results in one of the `nydus-pull` containers in the node installer to fail due to missing permissions.

Since `ClusterRoleBinding`s are not namespaced, and `RoleBinding`s limit the container's access to its own namespace, this changes the name of the `ClusterRoleBinding` to `nodeinstaller-<platform-hash>-<namespace>`. By doing this, we achieve two things:
1. Two Contrast deployments with the same runtime hash can be deployed in different namespaces simultaneously.
2. Two Contrast deployments with different runtime hashes can be deployed simultaneously.

### DIFF
--- a/internal/kuberesource/mutators.go
+++ b/internal/kuberesource/mutators.go
@@ -295,6 +295,9 @@ func PatchNamespaces(resources []any, namespace string) []any {
 		case *applycorev1.ServiceAccountApplyConfiguration:
 			r.Namespace = nsPtr
 		case *applyrbacv1.ClusterRoleBindingApplyConfiguration:
+			if namespace != "" {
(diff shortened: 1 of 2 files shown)


burgerdev:
#1156 adopts the namespace suffix for the role binding and additionally qualifies the resource names.

katexochen:
superseded by https://github.com/edgelesssys/contrast/pull/1156

PR #1153, opened 2025-01-14
initializer: add cryptsetup subcommand

jmxnzo:
This PR translates the bash entrypoint currently used for setting up an encrypted volume with cryptsetup to a  `cryptsetup` subcommand in the initializer binary, which takes the device path and the required mount point as flags. 
- adds wrapper functions of the following commands of the cryptsetup executable: cryptsetup open, isLuks, luksFormat
-  setupEncryptedMount function reimplements the current bash entrypoint of #1140 as a go function
-  introduces new `cryptsetup`  Cobra subcommand to the initializer command, wrapping the setupEncryptedMount function -- follow-up on #1157 
-  const workloadSecretPath is shared between initializer main.go and mount.go

The shared code of mounting mechanisms between the coordinator and initializer was moved to internal/mount in #1163 .

e2e volumestatefulset: https://github.com/edgelesssys/contrast/actions/runs/13008716710/job/36281340998

### DIFF
--- a/docs/docs/architecture/secrets.md
+++ b/docs/docs/architecture/secrets.md
@@ -40,7 +40,13 @@ If the data owner and the workload owner is the same entity, then they can safel
 ### Secure persistence
 
 Remember that persistent volumes from the cloud provider are untrusted.
-Using the workload secret, applications can set up trusted storage on top of untrusted block devices.
+Using the built-in `cryptsetup` subcommand of the initializer, applications can set up trusted storage on top of untrusted block devices based on the workload secret.
+Functionally the initializer will act as a sidecar container which serves to set up a secure mount inside an `emptyDir` mount shared with the main container.
+
+#### Usage `cryptsetup` subcommand
+
+The `cryptsetup` subcommand takes two arguments `cryptsetup -d [device-path] -m [mount-point]`, to set up a LUKS-encrypted volume at `device-path` and mount that volume at `mount-point`.
+
 The following, slightly abbreviated resource outlines how this could be realized:
 
 :::warning
@@ -59,48 +65,55 @@ spec:
   template:
     spec:
       containers:
-      - name: main
-        image: my.registry/my-image@sha256:0123...
-        command:
-        - /bin/sh
-        - -ec
-        - | # <-- Custom script that mounts the encrypted disk and then calls the original application.
-          device=/dev/csi0
-          if ! cryptsetup isLuks $device; then
-            cryptsetup luksFormat $device /contrast/secrets/workload-secret-seed
-            cryptsetup open $device state -d /contrast/secrets/workload-secret-seed
-            mkfs.ext4 /dev/mapper/state
-            cryptsetup close state
-          fi
-          cryptsetup open $device state -d /contrast/secrets/workload-secret-seed
-          /path/to/original/app
-        name: volume-tester
-        volumeDevices:
-        - name: state
-          devicePath: /dev/csi0
-        securityContext:
-          privileged: true # <-- This is necessary for mounting devices.
+        - name: main
+          image: my.registry/my-image@sha256:0123... # <-- Original application requiring encrypted disk.
+          volumeMounts:
+            - mountPath: /state
+              mountPropagation: HostToContainer
+              name: share
+      initContainers:
+        - args:
+            - cryptsetup # <-- cryptsetup subcommand provided as args to the initializer binary.
+            - "--device-path"
+            - /dev/csi0
+            - "--mount-point"
+            - /state
+          image: "ghcr.io/edgelesssys/contrast/initializer:latest"
+          name: encrypted-volume-initializer
+          resources:
+            limits:
+              memory: 100Mi
+            requests:
+              memory: 100Mi
+          restartPolicy: Always
+          securityContext:
+            privileged: true # <-- This is necessary for mounting devices.
+          startupProbe:
+            exec:
+              command:
+                - /bin/test
+                - "-f"
+                - /done
+            failureThreshold: 20
+            periodSeconds: 5
+          volumeDevices:
+            - devicePath: /dev/csi0
+              name: state
+          volumeMounts:
+            - mountPath: /state
+              mountPropagation: Bidirectional
+              name: share
       runtimeClassName: contrast-cc
   volumeClaimTemplates:
-  - apiVersion: v1
-    kind: PersistentVolumeClaim
-    metadata:
-      name: state
-    spec:
-      accessModes:
-      - ReadWriteOnce
-      resources:
-        requests:
-          storage: 1Gi
-      volumeMode: Block # <-- The requested volume needs to be a raw block device.
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
+        name: state
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 1Gi
+        volumeMode: Block # <-- The requested volume needs to be a raw block device.
 ```
-
-:::note
-
-This example assumes that you can modify the container image to include a shell and the `cryptsetup` utility.
-Alternatively, you can set up a secure mount from a sidecar container inside an `emptyDir` mount shared with the main container.
-The Contrast end-to-end tests include [an example] of this type of mount.
-
-[an example]: https://github.com/edgelesssys/contrast/blob/0662a2e/internal/kuberesource/sets.go#L504
-
-:::
--- a/initializer/mount.go
+++ b/initializer/mount.go
@@ -1,0 +1,144 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package main
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/exec"
+	"os/signal"
+	"syscall"
+
+	"github.com/edgelesssys/contrast/internal/logger"
+	"github.com/edgelesssys/contrast/internal/mount"
+	"github.com/spf13/cobra"
+)
+
+// parsedCryptsetupFlags holds configuration for mounting a LUKS encrypted device.
+type parsedCryptsetupFlags struct {
+	devicePath       string
+	mappingName      string
+	volumeMountPoint string
+}
+
+// NewCryptsetupCmd creates a Cobra subcommand that mounts an encrypted LUKS volume.
+func NewCryptsetupCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "cryptsetup -d [device-path] -m [mount-point]",
+		Short: "cryptsetup on block device at [device-path] with decrypted mapper device at [mount-point]",
+		Long: `Set up an LUKS encrypted VolumeMount on the provided VolumeDevice
+		located at the specified [device-path] and mount the decrypted mapper
+		device to the provided [mount-point].
+
+		In certain deployments, we require a persistent volume claim configured
+		as block storage to be encrypted by the initializer binary.
+		Therefore we expose the defined PVC as a block VolumeDevice to our
+		initializer container. This allows the initializer to setup the
+		encryption on the block device located at [device-path] using cryptsetup
+		with the current workload secret as passphrase.
+
+		The mapped decrypted block device can then be shared with other containers
+		on the pod by setting up a shared VolumeMount on the specified [mount-point],
+		where the mapper device will be mounted to.`,
+		RunE: setupEncryptedMount,
+	}
+	cmd.Flags().StringP("device-path", "d", "", "path to the volume device to be encrypted")
+	cmd.Flags().StringP("mount-point", "m", "", "mount point of decrypted mapper device")
+	must(cmd.MarkFlagRequired("device-path"))
+	must(cmd.MarkFlagRequired("mount-point"))
+
+	return cmd
+}
+
+// parseCryptsetupFlags returns struct of type parsedCryptsetupFlags, representing the provided flags to the subcommand, which is then used to setup the encrypted volume mount.
+func parseCryptsetupFlags(cmd *cobra.Command) (*parsedCryptsetupFlags, error) {
+	devicePath, err := cmd.Flags().GetString("device-path")
+	if err != nil {
+		return nil, err
+	}
+	mountPoint, err := cmd.Flags().GetString("mount-point")
+	if err != nil {
+		return nil, err
+	}
+	mapperHash := sha256.Sum256([]byte(devicePath + mountPoint))
+	mappingName := hex.EncodeToString(mapperHash[:8])
+	return &parsedCryptsetupFlags{
+		devicePath:       devicePath,
+		mappingName:      mappingName,
+		volumeMountPoint: mountPoint,
+	}, nil
+}
+
+// setupEncryptedMount sets up an encrypted LUKS volume mount using the device path and mount point, provided to the setupEncryptedMount subcommand.
+func setupEncryptedMount(cmd *cobra.Command, _ []string) error {
+	// Register channel waiting for SIGTERM signal
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM, syscall.SIGINT)
+	logger, err := logger.Default()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: creating logger: %v\n", err)
+		return err
+	}
+	parsedCryptsetupFlags, err := parseCryptsetupFlags(cmd)
+	if err != nil {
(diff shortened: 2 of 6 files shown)


initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+	if !isLuks(ctx, logger, csiDevicePath) {
+		// TODO(jmxnzo) might just use stdin instead for the initial passphrase generation
+		if err := createInitPassphrase(tmpKeyPath); err != nil {
+			return err
+		}
+		logger.Info("formatting csi device to LUKS with initial passphrase)")
+		if err := luksFormat(ctx, csiDevicePath, tmpKeyPath); err != nil {
+			return err
+		}
+
```
burgerdev, line 44: We should consider what happens if the container is terminated in between formatting and changing the key. At least add a TODO for now.

initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+		if err := openEncryptedDevice(ctx, csiDevicePath, diskEncryptionKeyPath, mapperDeviceMountPoint); err != nil {
+			return err
+		}
+	} else {
+		if err := setEncryptionPassphrase(ctx, csiDevicePath, workloadSecretPath, diskEncryptionKeyPath); err != nil {
+			return err
+		}
+		if err := openEncryptedDevice(ctx, csiDevicePath, diskEncryptionKeyPath, mapperDeviceMountPoint); err != nil {
+			return err
+		}
```
burgerdev: `openEncryptedDevice` is common to both branches, right?

packages/containers.nix
```diff
@@ -62,7 +62,7 @@ let
           busybox
           cryptsetup
           e2fsprogs # mkfs.ext4
-          mount
```
burgerdev, line 65: Where is it added for the first time? Does it lead to problems adding it here? Since we directly use mount, we might as well directly specify the dependency.

cli/cmd/generate.go
burgerdev, line 1: As we discussed privately, I don' think we should start with _doing the cryptsetup in the initializer_ just yet. Rather, drop 48f5be5 and just use the subcommand where we currently use the script.

internal/kuberesource/sets.go
```diff
@@ -573,47 +573,6 @@ func MySQL() []any {
...
-										WithFailureThreshold(20).
-										WithPeriodSeconds(5).
-										WithExec(applycorev1.ExecAction().
-											WithCommand("/bin/test", "-f", "/done"),
-										),
-								).
-								WithRestartPolicy(
-									corev1.ContainerRestartPolicyAlways,
-								),
-						).
```
burgerdev, line 657: This should be kept for now, see other comments.

initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+var errNotIdentified = errors.New("blkid did not identify the device")
+
+func blkid(ctx context.Context, devName string) (*blk, error) {
+	cmd := exec.CommandContext(ctx, "blkid", "-o", "export", devName)
+	out, err := cmd.CombinedOutput()
+	var exitErr *exec.ExitError
+	if errors.As(err, &exitErr) && exitErr.ExitCode() == 2 {
+		// See man page, sec return code.
```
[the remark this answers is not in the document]
jmxnzo: Currently reused the code of https://github.com/edgelesssys/contrast/blob/main/coordinator/mount.go 
As described in description, this will be moved and reworked.

initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+			b.UUID = value
+		case "TYPE":
+			b.Type = value
+		case "BLOCK_SIZE":
+			b.BlockSize, err = strconv.Atoi(value)
+			if err != nil {
+				return nil, fmt.Errorf("parsing BLOCK_SIZE of blkid output %q: %w", value, err)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
jmxnzo: Currently reused the code of https://github.com/edgelesssys/contrast/blob/main/coordinator/mount.go 
As described in description, this will be moved and reworked.

initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+	UUID      string
+	BlockSize int
+	Type      string
+}
+
+var errNotIdentified = errors.New("blkid did not identify the device")
+
+func blkid(ctx context.Context, devName string) (*blk, error) {
+	cmd := exec.CommandContext(ctx, "blkid", "-o", "export", devName)
+	out, err := cmd.CombinedOutput()
```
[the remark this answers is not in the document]
jmxnzo: Currently reused the code of https://github.com/edgelesssys/contrast/blob/main/coordinator/mount.go 
As described in description, this will be moved and reworked.

packages/containers.nix
jmxnzo: It is added twice in busybox and util-linux. But the mount pkgs above refers to mount-util-linux: https://search.nixos.org/packages?channel=24.11&show=mount&from=0&size=50&sort=relevance&type=packages&query=mount . For me it looked like we just duplicated it.

initializer/mount.go
jmxnzo: First branch was missing the formatting of the decrypted mapper device after opening.

initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+	} else {
+		if err := setEncryptionPassphrase(ctx, csiDevicePath, workloadSecretPath, diskEncryptionKeyPath); err != nil {
+			return err
+		}
+		if err := openEncryptedDevice(ctx, csiDevicePath, diskEncryptionKeyPath, mapperDeviceMountPoint); err != nil {
+			return err
+		}
+	}
+	luksUUID(ctx, csiDevicePath)
```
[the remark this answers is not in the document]
jmxnzo: We did print the LUKS UUID out once in the cryptsetup bash entrypoint. But i'm not really sure why, any reason we require this? @davidweisse

initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+	if err := setupMount(ctx, logger, "/dev/mapper/"+mapperDeviceMountPoint, "/"+mapperDeviceMountPoint); err != nil {
+		return err
+	}
+	file, err := os.Create("/done")
+	if err != nil {
+		return fmt.Errorf("Creating startup probe done directory:%w", err)
+	}
+	defer file.Close()
+
+	time.Sleep(10*time.Minute)
```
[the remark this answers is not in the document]
jmxnzo: I added the SIGTERM logic, i think @davidweisse just moved the cryptsetup init command to https://github.com/edgelesssys/contrast/blob/main/internal/kuberesource/parts.go and the code was added with the volumestatefulset in https://github.com/edgelesssys/contrast/commit/125a228590842f23d1d9fb0732e782d058d3ad7c by @3u13r . Any suggestions?

initializer/mount.go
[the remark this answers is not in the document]
davidweisse: Only used for debugging. I don't think we need this.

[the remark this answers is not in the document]
3u13r: > I don't think we need this.

I also cannot think of a use-case.

initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+const (
+	// TODO(jmxnzo) don't hardcode these, add struct EncryptedMount as parameter to setupEncryptedMount function
+	// TODO follow naming convention that /dev/mapper/{x} is remounted to /{x}
+	// csiDevicePath is the path to the block device to be encrypted.
+	csiDevicePath = "/dev/csi0"
```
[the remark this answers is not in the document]
jmxnzo: Seems like a good choice to me! But we would need to derive the name of the new mapped device from those two, to allow having multiple encrypted volume mounts, because the name of mapping needs to be unique.

initializer/mount.go
```diff
@@ -0,0 +1,283 @@
...
+	_, err = file.WriteString("init_passphrase")
+	if err != nil {
+		fmt.Println("Error writing to file:", err)
+		return nil
+	}
+	return nil
+}
+
+// setEncryptionPassphrase writes the UUID of the device and the current workload secret to the diskEncryptionKeyPath used as passphrase.
+func setEncryptionPassphrase(ctx context.Context, devName, workloadSecretPath, diskEncryptionKeyPath string) error {
```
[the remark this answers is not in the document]
jmxnzo: LUKS pbkdf2 does not take the UUID as input for key derivation, as described in standardization https://gitlab.com/cryptsetup/cryptsetup/-/wikis/LUKS-standard/on-disk-format.pdf . At the moment we set the passphrase which is input to pbkdf2  as the concatenation of workloadsecret+LUKS UUID,

initializer/mount.go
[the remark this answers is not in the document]
jmxnzo: #1163

initializer/mount.go
```diff
@@ -0,0 +1,246 @@
...
+	if err := setupMount(ctx, logger, "/dev/mapper/"+luksVolume.mappingName, luksVolume.volumeMountPoint); err != nil {
+		return err
+	}
+
+	if err := os.WriteFile("/done", []byte(""), 0o644); err != nil {
+		return fmt.Errorf("Creating startup probe done directory:%w", err)
+	}
+	// Waits for SIGTERM signal and then properly release the resources.
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM, syscall.SIGINT)
```
3u13r, line 118: I think we should register the channel at least at the top of this function as to not miss any signals. We could also even decide to cancel the context then, but this can be done in a follow-up.

initializer/mount.go
```diff
@@ -0,0 +1,246 @@
...
+func mount(ctx context.Context, devName, mountPoint string) error {
+	if err := os.MkdirAll(mountPoint, 0o755); err != nil {
+		return fmt.Errorf("mkdir: %w", err)
+	}
+	cmd := exec.CommandContext(ctx, "mount", devName, mountPoint)
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("mount: %w, output: %q", err, out)
+	}
```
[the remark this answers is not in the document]
jmxnzo: https://github.com/edgelesssys/contrast/pull/1163

initializer/mount.go
```diff
@@ -0,0 +1,146 @@
...
+		the encryption passphrase is derived from the UUID of the LUKS formatted
+		block device and the current workload secret.
+
+		The mapped decrypted block device can then be shared with other containers
+		on the pod by setting up a shared VolumeMount on the specified [mount-point],
+		where the mapper device will be mounted to.`,
+		RunE: setupEncryptedMount,
+	}
+	cmd.Flags().StringP("device-path", "d", "/dev/csi0", "path to the volume device to be encrypted")
+	cmd.Flags().StringP("mount-point", "m", "/state", "mount point of decrypted mapper device")
```
burgerdev, line 59:
```suggestion
	cmd.Flags().StringP("device-path", "d", "", "path to the volume device to be encrypted")
	cmd.Flags().StringP("mount-point", "m", "", "mount point of decrypted mapper device")
```

The flags are marked required anyway, so we don't need to keep arbitrary defaults here.

initializer/mount.go
```diff
@@ -0,0 +1,146 @@
...
+func must(err error) {
+	if err != nil {
+		panic(err)
+	}
+}
+
+// NewSetupEncryptedMountCmd creates a Cobra subcommand of the initializer to set up the specified encrypted LUKS volume.
+func NewSetupEncryptedMountCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "setupEncryptedMount -d [device-path] -m [mount-point]",
```
burgerdev, line 39: Can we make the subcommand name a bit simpler, e.g. `cryptsetup` or just `luks`? That would also elegantly circumnavigate the style question of snake/kebab/camel-case.

initializer/mount.go
```diff
@@ -0,0 +1,146 @@
...
+func parseSetupEncryptedMountFlags(cmd *cobra.Command) (*luksVolume, error) {
+	devicePath, err := cmd.Flags().GetString("device-path")
+	if err != nil {
+		return nil, err
+	}
+	mountPoint, err := cmd.Flags().GetString("mount-point")
+	if err != nil {
+		return nil, err
+	}
+	hash := md5.Sum([]byte(devicePath + mountPoint))
```
burgerdev, line 76: This is fine, but I already see code scanners etc. complaining about the use of md5. Any objections to just using sha256?

internal/kuberesource/sets.go
```diff
@@ -525,7 +525,8 @@ func VolumeStatefulSet() []any {
 							Container().
 								WithName("volume-tester-init").
 								WithImage("ghcr.io/edgelesssys/contrast/initializer:latest").
-								WithCommand("/bin/sh", "-c", CryptsetupInitCommand()).
+								WithEnv(NewEnvVar("COORDINATOR_HOST", "coordinator")).
```
burgerdev, line 528: This is not needed because we're running the subcommand only.

internal/kuberesource/sets.go
```diff
@@ -525,7 +525,8 @@ func VolumeStatefulSet() []any {
 							Container().
 								WithName("volume-tester-init").
 								WithImage("ghcr.io/edgelesssys/contrast/initializer:latest").
-								WithCommand("/bin/sh", "-c", CryptsetupInitCommand()).
```
burgerdev, line 528: Can we replace all `CryptsetupInitCommand` with this subcmd and then remove the script from the codebase?

internal/kuberesource/parts.go
```diff
@@ -553,32 +553,3 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
 			"-l", "debug",
 		)
 }
-
-// CryptsetupInitCommand returns the init command for the cryptsetup
```
burgerdev, line 557: Now that we provide a container image and command, you can replace the documentation at https://github.com/edgelesssys/contrast/blob/6791f9e5c244889e8dedb67edb8cae33a6318985/docs/docs/architecture/secrets.md?plain=1#L101-L104 and instead explain how to do this with the initializer image.

initializer/mount.go
```diff
@@ -0,0 +1,145 @@
...
+	"os/signal"
+	"syscall"
+
+	"github.com/edgelesssys/contrast/internal/logger"
+	"github.com/edgelesssys/contrast/internal/mount"
+	"github.com/spf13/cobra"
+)
+
+// luksVolume holds configuration for mounting a LUKS encrypted device.
+type luksVolume struct {
```
3u13r, line 23:
```suggestion
type parsedCryptsetupFlags struct {
```

This is more in line with the naming in the user-facing CLI and also clearer in that this struct is thightly coupled with the CLI command and its flags.

3u13r (changes requested):

docs/docs/architecture/secrets.md
```diff
@@ -59,47 +64,66 @@ spec:
...
+            storage: 1Gi
+        volumeMode: Block # <-- The requested volume needs to be a raw block device.
 ```
 
 :::note
 
-This example assumes that you can modify the container image to include a shell and the `cryptsetup` utility.
-Alternatively, you can set up a secure mount from a sidecar container inside an `emptyDir` mount shared with the main container.
-The Contrast end-to-end tests include [an example] of this type of mount.
+For more detailed information, the Contrast end-to-end tests include [an example] of a deployment with an encrypted volume mount.
```
burgerdev, line 126: The link below is not a good example anymore. We could just remove this paragraph altogether.

docs/docs/architecture/secrets.md
```diff
@@ -59,47 +64,66 @@ spec:
...
+              mountPropagation: HostToContainer
+              name: share
+      initContainers:
+        - args:
+            - cryptsetup # <-- cryptsetup subcommand provided as args to the initializer binary.
+            - "--device-path"
+            - /dev/csi0
+            - "--mount-point"
+            - /state
+          image: "ghcr.io/jmxnzo/contrast/initializer@sha256:3e47b53ee79e64b1994270b7b5ef152dbb5fd8319534bbfbee00441bf4bb5e61"
```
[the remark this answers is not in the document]
jmxnzo: It does trigger the image replacement as expected!

docs/docs/architecture/secrets.md
```diff
@@ -59,47 +64,66 @@ spec:
...
+                - "-f"
+                - /done
+            failureThreshold: 20
+            periodSeconds: 5
+          volumeDevices:
+            - devicePath: /dev/csi0
+              name: state
+          volumeMounts:
+            - mountPath: /run
+              name: run
```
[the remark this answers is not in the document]
burgerdev: Can we fix this in the image instead?

burgerdev (approved):

PR #1156, opened 2025-01-15
runtime: allow installation of multiple Contrast runtimes side-by-side

burgerdev:
The Contrast runtime used to have a static name, and only one runtime could be installed into the `kube-system` namespace. We're changing the naming scheme in this PR: resources of the Contrast runtime (the node installer and its RBAC config) are now named like the runtime handler, which allows installing more than one version.

**User-visible consequence**: the runtime `DaemonSet` is not called `contrast-node-installer` anymore, but something like `contrast-cc-aks-clh-snp-c52b1454`.


katexochen (approved):
If this should keep the `changelog` lable it needs a PR description.

PR #1157, opened 2025-01-15
initializer: add Cobra for subcommands

jmxnzo:
As part of #1153, we would like to have the possibility to add subcommands to the initializer binary. Therefore this PR integrates the [Cobra library](https://github.com/spf13/cobra) to the initializer, facilitating the creation of subcommands in the initializer in the future.
There is no behavioral change of the initializer and the command to start the initializer is kept equal.

### DIFF
--- a/initializer/main.go
+++ b/initializer/main.go
@@ -22,15 +22,36 @@ import (
 	"github.com/edgelesssys/contrast/internal/grpc/dialer"
 	"github.com/edgelesssys/contrast/internal/logger"
 	"github.com/edgelesssys/contrast/internal/meshapi"
+
+	"github.com/edgelesssys/contrast/internal/constants"
+	"github.com/spf13/cobra"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


initializer/main.go
```diff
@@ -22,15 +22,34 @@ import (
...
+}
+
+func newRootCmd() *cobra.Command {
+	root := &cobra.Command{
+		Use:          "initializer",
+		Short:        "initializer",
+		RunE:         run,
+		SilenceUsage: true,
+	}
+	root.InitDefaultVersionFlag()
```
burgerdev, line 48: > If c.Version is empty, it will do nothing

https://pkg.go.dev/github.com/spf13/cobra#Command.InitDefaultVersionFlag

You should be able to pass `constants.Version` though.

burgerdev (approved):

burgerdev (approved):

burgerdev (commented):
Please squash the commits.

PR #1159, opened 2025-01-16
ci: simplify if statements in workflows

katexochen:


.github/workflows/e2e.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       packages: write
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - if: ${{ !inputs.self-hosted }}
+      - if: (!inputs.self-hosted)
```
msanft, line 36: Do we need the parentheses here?

msanft (approved):
I roughly have in mind that bracing these like we did before was necessary for some reason. Do you still remember? Generally, if the CI continues to work as expected (which it seems to do based on the tests run on this PR), I'm fine with the change.

.github/workflows/e2e.yml
katexochen: yes, otherwise it's interpreted as yaml foo

katexochen:
> I roughly have in mind that bracing these like we did before was necessary for some reason. Do you still remember? Generally, if the CI continues to work as expected (which it seems to do based on the tests run on this PR), I'm fine with the change.

If is always interpreted, so braces aren't needed.

msanft:
> If is always interpreted, so braces aren't needed.

Thanks! Good to merge for me then.

PR #1160, opened 2025-01-16
Add policy support for GPU containers

msanft:
This adds patches to support GPU containers in genpolicy, and incorporates those into our GPU E2E test.


packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
```diff
@@ -0,0 +1,177 @@
...
+   print("state_allows: saving to state key =", key, "value =", value)
+-  path := get_state_path(key) 
++  path := get_state_path(key)
+   action := {
+     "op": "add",
+-    "path": path, 
++    "path": path,
+     "value": value,
+   }
+ }
```
burgerdev, line 64: Consider removing the formatting change from this commit.

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
```diff
@@ -0,0 +1,177 @@
...
+@@ -199,13 +199,17 @@ allow_anno(p_oci, i_oci) {
+ }
+ allow_anno(p_oci, i_oci) {
+     print("allow_anno 2: p Annotations =", p_oci.Annotations)
++    print("allow_anno 2: p DynamicAnnotations =", p_oci.DynamicAnnotations)
```
burgerdev, line 69: Does this already fail if there is no `DynamicAnnotations` on `p_oci`?

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
```diff
@@ -0,0 +1,177 @@
...
+         allow_anno_key(i_key, p_oci)
++        allow_dynamic_anno_value(i_key, i_annotations, p_oci)
+     }
+ 
+     print("allow_anno 2: true")
+@@ -227,6 +231,49 @@ allow_anno_key(i_key, p_oci) {
+     print("allow_anno_key 2: true")
+ }
+ 
++allow_anno_key(i_key, p_oci) {
```
burgerdev, line 88: Did you consider adding the value to `allow_anno_key`? That way, you could write the check as

```rego
allow_anno(p_oci, i_oci) {
  # [...]
  every i_key, i_val in i_annotations {
    allow_anno_keyval(i_key, i_val, p_oci)
  }
}

allow_anno_keyval(i_key, i_val, p_oci) {
 # [...] Existing check
}

allow_anno_keyval(i_key, i_val, p_oci) {
  some p_key, p_val in p_oci.DynamicAnnotations
  regex.match(p_key, i_key)
  regex.match(p_val, i_val)
}
```

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
```diff
@@ -0,0 +1,177 @@
...
+@@ -96,6 +96,13 @@ pub struct KataSpec {
+     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+     pub Annotations: BTreeMap<String, String>,
+ 
++    /// DynamicAnnotations contains arbitrary metadata for the container.
++    /// It is different to `Annotations` in that it allows dynamic keys *and*
++    /// values, and that they are checked for *all* keys, whereas `Annotations`
++    /// only allows dynamic values, and only checks them for certain keys at all.
++    #[serde(default, skip_serializing_if = "Option::is_none")]
++    pub DynamicAnnotations: Option<BTreeMap<String, String>>,
```
burgerdev, line 156: The `KataSpec` struct is in fact an OCI container spec. Can we pack this into `PolicyData.common` instead?

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
msanft: No, this will be `undefined` by default.

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
msanft: I think this would make rebasing harder. What benefits do you see compared to the current implementation?

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
msanft: Do you see it fit there? I figured that this would be a container-scoped setting, and thus put it there. Otherwise we'd also allow it for the pause container.

burgerdev: But these are not sourced from the individual containers, but copied from the settings file - which is what the `common` field is for.

msanft: I meant the `pause_container` / `other_container` sections in `genpolicy-settings.json`. But I'm fine with moving it to `common`, if you see it fit better.

burgerdev: This is not yet resolved, is it?

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
```diff
@@ -0,0 +1,149 @@
...
+         "CopyFileRequest": [
+diff --git a/src/tools/genpolicy/rules.rego b/src/tools/genpolicy/rules.rego
+index 43cb19a56fe8ea5833708f0639c9e85ddd884cb3..271df2aebe05bd4bbd7aa396be24fb6fee0668bf 100644
+--- a/src/tools/genpolicy/rules.rego
++++ b/src/tools/genpolicy/rules.rego
+@@ -199,26 +199,31 @@ allow_anno(p_oci, i_oci) {
+ }
+ allow_anno(p_oci, i_oci) {
+     print("allow_anno 2: p Annotations =", p_oci.Annotations)
++    p_dynamic_annotations := policy_data.common.dynamic_annotations
```
burgerdev, line 51: Instead of plumbing this through all of the predicates, I'd just access the global where you actually need it.

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
```diff
@@ -0,0 +1,149 @@
...
+     print("allow_anno 2: i Annotations =", i_oci.Annotations)
+ 
+-    i_keys := object.keys(i_oci.Annotations)
++    i_keys := object.keys(i_annotations)
+     print("allow_anno 2: i keys =", i_keys)
+ 
+     every i_key in i_keys {
+-        allow_anno_key(i_key, p_oci)
++        allow_anno_key(i_key, p_oci, p_dynamic_annotations)
++        allow_dynamic_anno_value(i_key, i_annotations, p_dynamic_annotations)
```
burgerdev, line 64: Why is this not

```rego
allow_dynamic_anno_value(i_key, i_value)
```

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
msanft: It's in `common` now. Unless I misunderstand you, I think that should resolve it?

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
burgerdev: The current implementation is probably wrong, looking at the negative check in `allow_dynamic_anno_value 2` which is only necessary because you're checking keys and values separately.

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
```diff
@@ -0,0 +1,149 @@
...
++    print("allow_anno_key 3: true")
++}
++
++
++# Account for containers without dynamic annotations
++# at all..
++allow_dynamic_anno_value(i_key, i_annotations, p_dynamic_annotations) {
++    print("allow_dynamic_anno_value 1: i key =", i_key)
++
++    not p_dynamic_annotations
```
burgerdev, line 102: This is more or less included in `allow_dynamic_anno_value 2`.

msanft: In my testing, it seemed that `every` can't work with `p_dynamic_annotations` being `undefined`. Hence I had to add this.

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
msanft: Do you mean?
```rego
allow_dynamic_anno_value(i_key, i_value, p_dynamic_annotations)
```
I don't know how this would be implemented without passing the dynamic annotations.

msanft: Ah, makes sense in conjunction with the above comment.

msanft:
> The current implementation is probably wrong, looking at the negative check in allow_dynamic_anno_value 2 which is only necessary because you're checking keys and values separately.

@burgerdev can you elaborate on that? What's wrong exactly?

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
burgerdev: Sorry, I misread the diff.

packages/by-name/kata/kata-runtime/0020-genpolicy-support-dynamic-annotations.patch
msanft: Can you elaborate on that? What's wrong exactly?

burgerdev: I also don't subscribe to the rebasing argument: you will have one line of diff in `allow_anno`, currently you have two. If you implement it as 

```rego
allow_anno_keyval(i_key, i_val) {
  allow_anno_key(i_key)
}

allow_anno_keyval(i_key, i_val) {
  # lookup in common.DynamicAnnotations ...
}
```

you don't need to touch any of the other existing predicates.

burgerdev: > Can you elaborate on that? What's wrong exactly?

The list of dynamic annotations should be additive here, i.e. adding a new entry to dynamic annotations should never make my policy stricter. Suppose I added this to the settings:

```json
"dynamic_annotations": {"^.*$": "^$"}
```

Upstream currently allows all annotations that are explicitly added to the pod, but with this setting these would all be rejected because the key matches but the value doesn't.

msanft: > Upstream currently allows all annotations that are explicitly added to the pod, but with this setting these would all be rejected because the key matches but the value doesn't.

But this would also be the case for your suggestion, wouldn't it?

burgerdev: No, because the annotation would be accepted by the existing rule, which would be an alternative to the new rule.

msanft:
[GPU E2E Test](https://github.com/edgelesssys/contrast/actions/runs/12928958338/job/36057361668)

burgerdev (approved):
Thanks, looks great now!

msanft:
[GPU E2E Test](https://github.com/edgelesssys/contrast/actions/runs/12945738258)

katexochen (approved):

PR #1165, opened 2025-01-17
packages.tdx-tools: noble-24.04 -> 3.1

katexochen:
https://github.com/canonical/tdx/releases

### DIFF
--- a/packages/by-name/tdx-tools/package.nix
+++ b/packages/by-name/tdx-tools/package.nix
@@ -2,20 +2,21 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
 {
+  lib,
   python3Packages,
   fetchFromGitHub,
 }:
 
 python3Packages.buildPythonApplication rec {
   pname = "tdx-tools";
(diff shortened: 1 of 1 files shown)


packages/by-name/tdx-tools/package.nix
```diff
@@ -27,4 +28,12 @@ python3Packages.buildPythonApplication rec {
...
     cd tests/lib/tdx-tools
   '';
+
+  meta = {
+    homepage = "https://github.com/canonical/tdx";
+    license = with lib.licenses; [
+      gpl3Only
+      asl20
+    ];
+  };
```
msanft, line 38: General question: Should we set this for packages we aim to package within Contrast only?

msanft (approved):

packages/by-name/tdx-tools/package.nix
katexochen: Yes, I find this helpful. You can click on the homepage link and license info is important for us, too. No need to add other meta fields like maintainers.

PR #1171, opened 2025-01-21
service-mesh: blackhole traffic destined for the TROXY port

3u13r:


service-mesh/config.go
```diff
@@ -9,22 +9,32 @@ import (
...
 	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
 var loopbackCIDR = netip.MustParsePrefix("127.0.0.1/8")
 
+const (
+	blackHoleClusterName = "BlackHoleCluster"
+	ingressClusterName   = "ingressCluster"
```
burgerdev, line 35: Nit: we should capitalize consistently

service-mesh/config.go
```diff
@@ -397,3 +412,129 @@ func downstreamTLSTransportSocket(requireClientCertificate bool) (*envoyCoreV3.T
...
+		listenPort := listener.Address.GetSocketAddress().GetPortValue()
+		if listenPort == 0 {
+			return fmt.Errorf("listener %s listens on port 0", listener.Name)
+		}
+		if listenPort != 15006 && listenPort != 15007 {
+			// listener is none of the ingress listeners
+			continue
+		}
+		if len(listener.FilterChains) != 1 {
+			return fmt.Errorf("listener %s doesn't has exactly one existing listener", listener.Name)
```
burgerdev, line 434:
```suggestion
			return fmt.Errorf("listener %s doesn't have exactly one existing listener", listener.Name)
```

service-mesh/config.go
```diff
@@ -397,3 +412,129 @@ func downstreamTLSTransportSocket(requireClientCertificate bool) (*envoyCoreV3.T
...
+	// Add BlackHole matching to all listeners
+	for _, listener := range config.StaticResources.Listeners {
+		if listener.FilterChainMatcher != nil {
+			return fmt.Errorf("listener %s already has a filterChainMatcher", listener.Name)
+		}
+		listenPort := listener.Address.GetSocketAddress().GetPortValue()
+		if listenPort == 0 {
+			return fmt.Errorf("listener %s listens on port 0", listener.Name)
+		}
+		if listenPort != 15006 && listenPort != 15007 {
```
burgerdev, line 429: Can we make these arguments and use constants where we're calling `addBlackHoleToConfig`? That'd allow easier customization later on - I've been wondering lately when we'll have the first user that wants these ports for their app :)

service-mesh/config.go
```diff
@@ -397,3 +412,129 @@ func downstreamTLSTransportSocket(requireClientCertificate bool) (*envoyCoreV3.T
...
+		listener.FilterChains = append(listener.FilterChains, bhFilter)
+	}
+	return nil
+}
+
+// Blackhole traffic that arrives on the original destination listerners,
+// which original port is the envoy itself, i.e. traffic that was not redirected
+// to the envoy via the TROXY iptables rule. Such traffic would lead to a
+// traffic storm since envoy would connect to the original destination
+// i.e. itself again.
```
burgerdev, line 455: Please add a sentence that explains how the traffic is blackholed - is it because there is no upstream in the cluster?

service-mesh/config_test.go
```diff
@@ -0,0 +1,46 @@
...
+			require.NoError(err)
+			fmt.Println(string(configJSON))
+			assert.JSONEq(string(defaultEnvoyConfig), string(configJSON))
+		})
+	}
+}
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
```
burgerdev, line 46: Did you introduce this for a specific reason? This binary looks pretty serial...

3u13r: No, I thought we would just add this per convention for new tests, since e.g., https://github.com/edgelesssys/contrast/blob/5a6afdd360c22a30fbe336b1d56148623e33ec40/internal/fsstore/fsstore_test.go#L17 also looks pretty serial. It looks like at least Malte was of the same opinion, since all the times where this was also "unnecessarily" included are tests he wrote. But I don't have any (strong) opinion about this, does anyone else?

burgerdev: I don't, was just curious.

service-mesh/config.go
3u13r: I hope I've answered your question in the added sentences. If not, please unresolve.

burgerdev (approved):

PR #1176, opened 2025-01-23
ci: cleanup azure ci resources 

katexochen:
run: https://github.com/edgelesssys/contrast/actions/runs/12926242566/job/36048840852

### DIFF
--- a/packages/by-name/azure-ci-rg-cleanup/cleanup.sh
+++ b/packages/by-name/azure-ci-rg-cleanup/cleanup.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+# Copyright 2025 Edgeless Systems GmbH
+# SPDX-License-Identifier: AGPL-3.0-only
+
+set -euo pipefail
+
+declare -a resources
+
+keepResources=(
+  "/subscriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/contrast-ci/providers/Microsoft.ContainerService/managedClusters/contrast-ci"
+  "/subscriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/contrast-ci/providers/Microsoft.Compute/galleries/contrast_ci_contrast"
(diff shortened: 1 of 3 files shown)


packages/by-name/azure-ci-rg-cleanup/cleanup.sh
```diff
@@ -0,0 +1,52 @@
...
+
+declare -a resources
+
+keepResources=(
+  "/subscriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/contrast-ci/providers/Microsoft.ContainerService/managedClusters/contrast-ci"
+  "/subscriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/contrast-ci/providers/Microsoft.Compute/galleries/contrast_ci_contrast"
+  "/subscriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/contrast-ci/providers/Microsoft.Compute/galleries/contrast_ci_contrast/images/contrast"
+)
+
+while IFS= read -r resourceID; do
```
burgerdev, line 15: I think this block might be easier to read if it were structured like

```bash
all_az_resources() {
  az resource list ... | jq -r ...
}

for resourceID in $(all_az_resources); do
...
```

burgerdev (approved):

packages/by-name/azure-ci-rg-cleanup/cleanup.sh
katexochen: in general, I prefer the way it is currently done. While it doesn't make a different in this case, the current way is much more robust: for example, it handles values containing whitespace correctly, while the for loop has to rely on the fact that the output of `all_az_resources` isn't quoted.

burgerdev: Yes, agreed to the general case. So we keep it for consistency with other places where we _do_ depend on whitespace? sgtm

PR #1179, opened 2025-01-23
packages/qemu-tdx-static: 8.2.2 -> 9.0.2

msanft:
This updates the QEMU patchset for TDX to 9.0.2, which is the latest upstream version.

### DIFF
--- a/packages/by-name/qemu-tdx-static/package.nix
+++ b/packages/by-name/qemu-tdx-static/package.nix
@@ -11,8 +11,8 @@
 }:
 let
   tdxPatches = fetchzip {
-    url = "https://launchpadlibrarian.net/744102817/qemu_8.2.2+ds-0ubuntu2+tdx1.0.debian.tar.xz";
-    hash = "sha256-ByvNvdGeYJq5tBh8eONU8drQpg1yWolLTf8yoM2VTik=";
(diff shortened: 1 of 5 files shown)


packages/by-name/qemu-tdx-static/package.nix
```diff
@@ -24,11 +24,11 @@ in
   hostCpuTargets = [ "x86_64-softmmu" ];
 })).overrideAttrs
   (previousAttrs: rec {
-    version = "8.2.2";
```
katexochen, line 27: Can we unpin this?

packages/by-name/qemu-tdx-static/package.nix
```diff
@@ -24,11 +24,11 @@ in
...
+    version = "9.0.2";
 
     src = fetchurl {
       url = "https://download.qemu.org/qemu-${version}.tar.xz";
-      hash = "sha256-hHNGwbgsGlSyw49u29hVSe3rF0MLfU09oSYg4pYrxPM=";
+      hash = "sha256-qMP1lq7Olto7AMr7dLqvoNFFFer7jtHuP39cLQ6/ArY=";
     };
```
katexochen, line 32: why can't we use upsteam src here?

packages/by-name/qemu-tdx-static/package.nix
msanft: Not really, as the patches are version-specific, and nixpkgs upstream is already at `9.2.0`

msanft:
Postponed. The 9.0.2 version doesn't support TDX for some reason. I don't know why they have it in their TDX repo, but it's literally missing all TDX patches when you check it.

PR #1186, opened 2025-01-24
nodeinstaller: compare with the template file content

burgerdev:
This should fix the nightly test failures that fail waiting for the coordinator to come up, explained in detail in the code comments.


katexochen (approved):
great, thanks!

PR #1189, opened 2025-01-27
e2e/runtime-reproducibility: fix teams notifications

katexochen:
This should have used `github.ref_name == 'main'` instead of `github.ref`, which is `refs/heads/main` on main.
But even better, we can check for the `event_name` to be `schedule`, so we can also filter out `workflow_dispatch` events on main.


3u13r (approved):
LGTM

PR #1191, opened 2025-01-27
internal/testkeys: use fixed keys in unit tests

davidweisse:
We can speed up unit tests, and make them more deterministic by using fixed keys, instead of generating them at runtime. A number of fixed keys are stored in the `internal/testkeys` package in Base64 encoded PKCS8 format, which can be used in unit tests with the `testkeys.New` function.


internal/testkeys/testkeys.go
```diff
@@ -0,0 +1,54 @@
...
+package testkeys
+
+import (
+	"crypto"
+	"crypto/x509"
+	"encoding/base64"
+	"fmt"
+)
+
+type EncodedKeyString string
```
burgerdev, line 13: What's the idea behind this type def?

internal/testkeys/testkeys.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, err
+	}
+	key, err := x509.ParsePKCS8PrivateKey(keyBytes)
+	if err != nil {
+		return nil, err
+	}
+	if _, ok := key.(*T); !ok {
+		return nil, fmt.Errorf("invalid key type: %T", key)
+	}
+	return key.(*T), nil
```
burgerdev, line 53:
```suggestion
	typedKey, ok := key.(*T); 
	if !ok {
		return nil, fmt.Errorf("invalid key type: %T", key)
	}
	return typedKey, nil
```

internal/testkeys/testkeys.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func New[T crypto.PrivateKey](keyString EncodedKeyString) (*T, error) {
```
burgerdev, line 41: I think this would be easier to use if the signature was 

```go
func New[T crypto.PrivateKey](t *testing.T, keyString EncodedKeyString) *T
```

and the function would fail the test if the key does not decode.

internal/testkeys/testkeys.go
burgerdev, line 1: In addition to the `New` function, we could define 

```go
func ECDSA() *ecdsa.PrivateKey
func RSA() *rsa.privateKey
```

for the cases where we just need some key and don't care about the details. wdyt?

internal/testkeys/testkeys.go
burgerdev: Nevermind, I get now that this says "I'm really sure that this is a PEM-encoded key and not a random other string".

burgerdev (approved):

burgerdev (approved):

PR #1195, opened 2025-01-29
docs: guidance for pod memory allocation

burgerdev:


docs/docs/deployment.md
```diff
@@ -128,6 +128,40 @@ spec: # v1.PodSpec
...
+Once you determined the memory requirements of your application, add a resource section to the pod specification:
+
+```yaml
+spec: # v1.PodSpec
+  containers:
+  - name: my-container
+    image: my-image:latest
+    resources:
+      requests:
+        memory: 50Mi
```
burgerdev, line 158: Document that request should be equal to limit.

docs/docs/troubleshooting.md
```diff
@@ -164,3 +164,16 @@ contrast version v0.X.0
     image versions:       ghcr.io/edgelesssys/contrast/coordinator@sha256:...
                           ghcr.io/edgelesssys/contrast/initializer@sha256:...
 ```
+
+## VM runs out of memory
+
+Since pod VMs are statically sized, it's easier to run out of memory due to misconfigurations.
+Setting the right memory limits is even more important on bare metal, where the image layers need to be stored in the guest memory, too.
+If the guest pull fails due to lack of memory, you might see an error message like this:
```
katexochen, line 172: Would turn that sentence around: if you see the error, it might be failing cos of memory.

docs/docs/deployment.md
```diff
@@ -128,6 +128,40 @@ spec: # v1.PodSpec
   runtimeClassName: contrast-cc
 ```
 
+### Pod resources
+
+Contrast workloads are deployed as one confidential virtual machine (CVM) per pod.
+In order to configure the CVM resources correctly, Contrast workloads require a stricter specification of pod resources compared to standard [Kubernetes resource management].
+
+The CVM memory is calculated from the sum of the memory limits of the containers and a static overhead for daemons.
+InitContainers aren't taken into account.
```
katexochen, line 137: From reading these two sentences, it's still not clear to me how this is working. What daemons are we talking about? And what does the initContainer thing mean for me as a user?

m1ghtym0 (commented):
Agree with Paul's comments, otherwise lgtm:-)

docs/docs/troubleshooting.md
burgerdev: Done

docs/docs/deployment.md
burgerdev: I tried to explain things in more detail, ptal.

docs/docs/deployment.md
burgerdev: Done

m1ghtym0 (approved):

PR #1196, opened 2025-01-29
nodeinstaller: align default_memory and overhead to actual usage

burgerdev:
This PR corrects the calculation of the `RuntimeClass` overhead.

First, a bit of background on Kubernetes resource accounting. The total memory footprint of a pod is the sum of the individual container requests/limits and the overhead from the PodSpec or the overhead from the RuntimeClass. This total is accounted towards the node memory commitment, so we should aim to get it right in order to avoid overcommitting and idle resources. 

Ignoring the pod memory requests for now, we have three sources that add to the total memory used by a pod:

* Size of the VM, which is the sum of
  * the Kata runtime setting `default_memory`
  * the individual container limits
* VM-related processes on the host (i.e. Kata shim, qemu/CLH)

This implies immediately that the **RuntimeClass overhead is the sum of `default_memory` and the host overhead**. 

Empirically (see below), we observe that not all memory allotted to a VM is actually usable by the containers, because there are daemons running in the VM and the Linux kernel reserves some memory for itself. This guest overhead needs to be accounted for in the `default_memory` setting. It follows that **`default_memory` needs to be (at least) the sum of the memory used by daemons and the memory that is not available to userland**.

Rounding up the numbers and optimizing for <2GiB containers, we arrive at <56MiB for daemons, <200MiB for kernel and <50MiB for host overhead. Thus:

* `default_memory = 256Mi`
* `overhead = 320Mi`

In order to properly account for the resources in k8s, we need to configure Kata to use the pod cgroups. This is accomplished by setting `sandbox_cgroup_only = true`. The concepts are explained in more detail at https://github.com/kata-containers/kata-containers/blob/f9bbe4e4396cde04ce8162aead1d12b4f3e7bc96/docs/design/host-cgroups.md. 

---

### Memory usage and overhead survey 

I launched pods with varying memory limits and recorded the overhead both inside and outside the VM. Most notable findings:

* The footprint of guest components unrelated to the workload (`free -m used`) grows a little bit with the VM, but still fits into 70MiB even for large workloads.
* The portion of the VM unavailable to the workload (`Overhead VM`) scales linearily with VM size.
* The overhead outside the VM (`Overhead host`) is constant and on the order of 50MiB worst case.

#### AKS-CLH-SNP

|default_memory [MiB] |Container limit [MiB]|Theoretical VM mem [MiB]|free -m total [MiB]|free -m used [MiB]|cgroup memory.current [B]|Overhead VM [MiB]|Overhead host [MiB]|
|--------------|---------------|--------------------------|-------------|------------|------------|-----------|-------------|
|256           |50             |306                       |197          |27          |335077376   |109        |14           |
|256           |150            |406                       |295          |26          |432263168   |111        |6            |
|256           |500            |756                       |640          |30          |801509376   |116        |8            |
|256           |1024           |1280                      |1090         |32          |1353519104  |190        |11           |
|256           |2048           |2304                      |1965         |31          |2431700992  |339        |15           |
|256           |4096           |4352                      |3716         |47          |4587950080  |636        |23           |

#### K3s-QEMU-TDX

|default_memory [MiB] |Container limit [MiB]|Theoretical VM mem [MiB]|free -m total [MiB]|free -m used [MiB]|cgroup memory.current [B]|Overhead VM [MiB]|Overhead host [MiB]|
|--------------|---------------|--------------------------|-------------|------------|------------|-----------|-------------|
|256           |50             |306                       |207          |49          |355393536   |99         |33           |
|256           |150            |406                       |305          |49          |466104320   |101        |39           |
|256           |500            |756                       |650          |52          |841797632   |106        |47           |
|256           |1024           |1280                      |1100         |53          |1387155456  |180        |43           |
|256           |2048           |2304                      |1976         |62          |2462515200  |328        |44           |
|256           |4096           |4352                      |3727         |74          |?           |625        |?     |

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -25,10 +25,13 @@ func ContrastRuntimeClass(platform platforms.Platform) (*RuntimeClassConfig, err
 		return nil, fmt.Errorf("getting default runtime handler: %w", err)
 	}
 
+	// Consists of the default VM memory, 70MiB for the Kata shim and 100MiB for qemu overhead.
+	memoryOverhead := platforms.DefaultMemoryInMegaBytes(platform) + 170
+
 	r := RuntimeClass(runtimeHandler).
 		WithHandler(runtimeHandler).
 		WithLabels(map[string]string{"addonmanager.kubernetes.io/mode": "Reconcile"}).
-		WithOverhead(Overhead(corev1.ResourceList{"memory": resource.MustParse("1152Mi")}))
+		WithOverhead(Overhead(corev1.ResourceList{"memory": *resource.NewQuantity(int64(memoryOverhead)*1024*1024, resource.BinarySI)}))
 
 	if platform == platforms.AKSCloudHypervisorSNP {
 		r.WithScheduling(Scheduling(map[string]string{"kubernetes.azure.com/kata-cc-isolation": "true"}))
@@ -381,7 +384,7 @@ func Coordinator(namespace string) *CoordinatorConfig {
 									WithPort(intstr.FromInt(1313))),
 							).
 							WithResources(ResourceRequirements().
-								WithMemoryLimitAndRequest(100),
+								WithMemoryLimitAndRequest(200),
 							),
 					),
 				),
--- a/nodeinstaller/internal/constants/configuration-qemu-tdx.toml
+++ b/nodeinstaller/internal/constants/configuration-qemu-tdx.toml
@@ -18,10 +18,6 @@ cpu_features="-vmx-rdseed-exit,pmu=off"
 default_vcpus = 1
 default_maxvcpus = 0
 default_bridges = 1
-# On TDX, when lowering this, the patch:
-# packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
-# needs to be updated accordingly.
-default_memory = 2048
 default_maxmemory = 0
 disable_block_device_use = false
 shared_fs = "virtio-9p"
@@ -55,7 +51,7 @@ dial_timeout = 60
 enable_debug = false
 internetworking_model="tcfilter"
 disable_guest_seccomp=true
-sandbox_cgroup_only=false
+sandbox_cgroup_only=true
 static_sandbox_resource_mgmt=true
 sandbox_bind_mounts=[]
 vfio_mode="guest-kernel"
--- a/packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
+++ b/packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
@@ -12,76 +12,24 @@ which depend not only on the size of the initrd, but also on the memory space
 of the guest, this is not viable for Contrast's reference-value-based attestation
 approach.
 
-As we control the minimum VM memory size in Contrast, we just load the initrd
-to the address it gets loaded to for Contrast's minimum VM memory (2Gi), regardless
-of if the VM has more memory.
-
-QEMU, by default, does a similar thing.
-Consider the below line (cited from above):
-
-   `initrd_max = x86ms->below_4g_mem_size - acpi_data_size - 1;`
-
-This adds an artifical upper bound of where the initrd can be loaded to, as
-the calculation is based on the VM memory (below_4g_mem_size), but capped at 4Gi.
-This means, the initrd, regardless of guest memory size, will always be loaded
-at address 0x100000000 (4Gi) max (minus ACPI data size).
-
-Essentially, overwriting this to 0x80000000 (2Gi), we create an artificial lower *and*
-upper bound (set to Contrast minimum TDX VM memory size).
-This means that the initrd will *always* be loaded at 0x80000000 (2Gi), minus ACPI
-data size. The difference to QEMU's setting is, that we *fix* the address, rather than
-setting *only* an upper bound.
-
-This way, we get the initrd to *always* be loaded at a static address.
+The actual value of the static address does not matter, since OVMF fills the
+corresponding Linux header field with the correct address after measuring the
+kernel.
 
 Signed-off-by: Moritz Sanft <58110325+msanft@users.noreply.github.com>
 ---
- hw/i386/x86.c | 35 +++++++++++++++++++++++++++++++++++
- 1 file changed, 35 insertions(+)
+ hw/i386/x86.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
 
 diff --git a/hw/i386/x86.c b/hw/i386/x86.c
-index 504575abfa98bc25e498e219a2d58d8d31e5feaa..0763462c16f4106d0aa6a46c2b9c360e36ae3e96 100644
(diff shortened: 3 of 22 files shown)


katexochen:
> I launched pods with varying memory limits and recorded the overhead both inside and outside the VM. Most notable findings:

Over what period did you conduct your observation?

burgerdev:
> > I launched pods with varying memory limits and recorded the overhead both inside and outside the VM. Most notable findings:
> 
> Over what period did you conduct your observation?

Started the pod with a minimal container and observed when that container was ready. I did not watch the memory of an individual pod over time.

nodeinstaller/internal/constants/configuration-qemu-tdx.toml
```diff
@@ -18,10 +18,6 @@ cpu_features="-vmx-rdseed-exit,pmu=off"
 default_vcpus = 1
 default_maxvcpus = 0
 default_bridges = 1
-# On TDX, when lowering this, the patch:
-# packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
-# needs to be updated accordingly.
-default_memory = 2048
```
burgerdev, line 24: @msanft: Changing this does not seem to require the anticipated change in the qemu patch. Do you know why that's the case?

msanft: Changing this as in removing it?

I think the [default value](https://github.com/kata-containers/kata-containers/blob/eb1cf792de5b62a6d1897602018bb67db965850d/src/runtime/pkg/katautils/config-settings.go.in#L64) is 2048, which would explain this.

burgerdev: Take a look at `platforms.go`, I'm changing the default to 512.

internal/kuberesource/parts.go
```diff
@@ -25,10 +25,13 @@ func ContrastRuntimeClass(platform platforms.Platform) (*RuntimeClassConfig, err
 		return nil, fmt.Errorf("getting default runtime handler: %w", err)
 	}
 
+	// Consists of the default VM memory, 70MiB for the Kata shim and 100MiB for qemu overhead.
```
msanft, line 28:
```suggestion
	// Consists of the default VM memory, 70MiB for the Kata shim and 100MiB for kernel overhead.
```

nit: I think this is more correct.

msanft: Also, shouldn't we address this in `DefaultMemoryInMegaBytes` directly, as it applies for all containers?

msanft (commented):
Thanks for doing this!

internal/kuberesource/parts.go
msanft: Ah, I get the semantics now. I think we could add to the comment of `DefaultMemoryInMegaBytes` that it corresponds to Kata's `default_memory`.

nodeinstaller/internal/constants/configuration-qemu-tdx.toml
msanft: I'm not exactly sure. I would expect it not to work. Did you check what memory argument is being passed to QEMU?

burgerdev: Yes: the expected sum of 512 and the pod limit.

internal/kuberesource/parts.go
burgerdev: No, this is actually for the host overhead qemu has (total qemu usage minus quest VM size).

nodeinstaller/internal/constants/configuration-qemu-tdx.toml
msanft: Which should - in almost all cases - amount to something < 2Gi, right?

burgerdev: Yes. Should I check whether verification fails for large VMs?

internal/kuberesource/parts.go
msanft, line 28: Out of curiosity: Does each pod get its own shim?

burgerdev: Yes, and because of the `sandbox_cgroup_only = true` change it's now being accounted towards pod memory, not in a Kata overhead group.

nodeinstaller/internal/constants/configuration-qemu-tdx.toml
msanft: It should only fail for small VMs, really, as we try to load the initrd to an address larger than the maximum VM memory. Unless I miss some QEMU sanity check anywhere, this should break.

burgerdev: Well, at least the openssl test does not fail on TDX with this change.

msanft: I see, and that's what confuses me here. 
This will need some thorough investigation, but I'm not yet sure when I have the capacity for it.

msanft (approved):

burgerdev:
Regression tests on bare metal should pass now: https://github.com/edgelesssys/contrast/actions/runs/13303897274.

Started a nightly run to find other problems: https://github.com/edgelesssys/contrast/actions/runs/13303882063.

packages/by-name/qemu-tdx-static/0004-hw-x86-load-initrd-to-static-address.patch
```diff
@@ -43,42 +25,10 @@ diff --git a/hw/i386/x86.c b/hw/i386/x86.c
...
-+     * upper bound (set to Contrast minimum TDX VM memory size).
-+     * This means that the initrd will *always* be loaded at 0x80000000 (2Gi), minus ACPI
-+     * data size. The difference to QEMU's setting is, that we *fix* the address, rather than
-+     * setting *only* an upper bound.
-+     *
-+     * This way, we get the initrd to *always* be loaded at a static address.
-+     */
 +    uint32_t contrast_min_memory = 0x80000000; // 2Gi
 +    initrd_max = contrast_min_memory - acpi_data_size - 1;
 +
```
msanft, line 34: Should we make this more coherent with the (new) comment by actually setting it to a static address?

burgerdev: Are you referring to the `acpi_data_size`? That' a constant afaict: https://github.com/qemu/qemu/blob/495de0fd82d8bb2d7035f82d9869cfeb48de2f9e/hw/i386/pc.c#L288 (or 0 maybe).

msanft: No, I mean setting `initrd_max` (or `initrd_addr` directly) to something like `0xFFFFF` or some other constant. I'm fine with keeping it as-is too, but having something being called `contrast_min_memory` even though there is no minimum requirement "anymore" bugs me a little here.

burgerdev: Done

msanft (approved):

PR #1199, opened 2025-01-31
transitengineapi: add http handler enc/dec

jmxnzo:
This PR adds basic http handling functionality for the encrypt and decrypt endpoints of the transit engine API, allowing the auto-unsealing process for user-managed Vaults. The PR introduces the following points:

- add GetState() of authority, to always ensure operating on current manifest
- underlying cryptographic functionality for AES-GCM 256
- http-handler functions for encrypt and decrypt endpoints
- message-level implementation of encryption/decryption request and response
- key derivation based on workloadSecretID specified as URL parameter
- support for associatedData request parameter of Vault

### ToDo's
- authorize transit engine requests to name parameter with client cert policy hash
- implement Vault json error bodies

### DIFF
--- a/coordinator/internal/transitengineapi/crypto.go
+++ b/coordinator/internal/transitengineapi/crypto.go
@@ -1,0 +1,97 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package transitengine
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/edgelesssys/contrast/internal/crypto"
+)
+
+// ciphertextContainer describes a base64-encoded ciphertext prepended with the nonce and specified key version.
+type ciphertextContainer struct {
+	nonce      []byte
+	ciphertext []byte
+	keyVersion uint32
+}
+
+// symmetricEncryptRaw returns a ciphertextContainer, based on the encryption key and associatedData handed in.
+func symmetricEncryptRaw(encKey, plaintext []byte, associatedData []byte) (ciphertextContainer, error) {
+	aesCipher, err := aes.NewCipher(encKey)
+	if err != nil {
+		return ciphertextContainer{}, err
+	}
+	gcm, err := cipher.NewGCM(aesCipher)
+	if err != nil {
+		return ciphertextContainer{}, err
+	}
+	nonce, err := crypto.GenerateRandomBytes(gcm.NonceSize())
+	if err != nil {
+		return ciphertextContainer{}, err
+	}
+	ciphertext := gcm.Seal(nil, nonce, plaintext, associatedData)
+	return ciphertextContainer{
+		nonce:      nonce,
+		ciphertext: ciphertext,
+	}, nil
+}
+
+// symmetricDecryptRaw extracts the nonce and returns the decrypted ciphertext based on encryption keys handed in,
+// if the associatedData is valid.
+func symmetricDecryptRaw(decKey []byte, ciphertextContainer ciphertextContainer, associatedData []byte) ([]byte, error) {
+	aesCipher, err := aes.NewCipher(decKey)
+	if err != nil {
+		return nil, err
+	}
+	gcm, err := cipher.NewGCM(aesCipher)
+	if err != nil {
+		return nil, err
+	}
+
+	plaintext, err := gcm.Open(nil, ciphertextContainer.nonce, ciphertextContainer.ciphertext, associatedData)
+	if err != nil {
+		return nil, err
+	}
+	return plaintext, nil
+}
+
+// UnmarshalJSON umarshalls a json string to a ciphertextContainer holding the version prefix,
+// decoded base64 nonce and ciphertext.
+func (c *ciphertextContainer) UnmarshalJSON(data []byte) error {
+	var encoded string
+	if err := json.Unmarshal(data, &encoded); err != nil {
+		return err
+	}
+	// Split "vault:vX:base64" format
+	parts := strings.SplitN(encoded, ":", 3)
+	if len(parts) < 3 {
+		return fmt.Errorf("invalid ciphertext format")
+	}
+	version, err := extractVersion(parts[1])
+	if err != nil {
+		return fmt.Errorf("ciphertext version: %w", err)
+	}
+	c.keyVersion = version
+	fullCiphertext, err := base64.StdEncoding.DecodeString(parts[2])
+	if err != nil {
+		return fmt.Errorf("decoding ciphertext: %w", err)
+	}
+	c.nonce = fullCiphertext[:aesGCMNonceSize]
+	c.ciphertext = fullCiphertext[aesGCMNonceSize:]
+	return nil
+}
+
+// MarshalJSON marshalls a ciphertextContainer to a json string.
+func (c ciphertextContainer) MarshalJSON() ([]byte, error) {
+	fullCiphertext := append(c.nonce, c.ciphertext...)
+	encodedfullCiphertext := base64.StdEncoding.EncodeToString(fullCiphertext)
+	// Convert to "vault:vX:base64" format
+	versioned := fmt.Sprintf("vault:v%d:%s", c.keyVersion, encodedfullCiphertext)
+	return json.Marshal(versioned)
+}
--- a/coordinator/internal/transitengineapi/crypto_test.go
+++ b/coordinator/internal/transitengineapi/crypto_test.go
@@ -1,0 +1,70 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package transitengine
+
+import (
+	"encoding/base64"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCryptoCyclic(t *testing.T) {
+	testCases := map[string]struct {
+		key            string
+		plaintext      string
+		associatedData string
+	}{
+		"plaintext 1": {
+			key:            "H0ltoqZhwB/h1/HqIPWr5rC9Xt/sGkJ3UDU10vYplGw=",
+			plaintext:      "rWk4fnCShUgfxjMShRHS5Q==",
+			associatedData: "VKpPXX5vwsZUPdFTQOTIdg==",
+		},
+		"plaintext 2": {
+			key:            "RV30FnNmnE2b8hUAL4P89tsqF/346nODNTXdYMrOuA0=",
+			plaintext:      "ZSZ/iGeN22qnQrvkogjvVnZG/3zhCboO/DC8Gl/GAFBRXYMh4112FA==",
+			associatedData: "c0JDPPZH0yrHVrl+x8e9JE0wgxrjCBk9Z2Vpi72+Fvy2wUMVPZ2i/A==",
+		},
+		"plaintext 3": {
+			key:            "vMYJT/IsiWCOX5/8cAcP7eEM7G4ZRwMNQtauHORAXY4=",
+			plaintext:      "X3pjwqhM4wMM/OFTpbZsOECwtZ4TZs0ZCptD/kg=",
+			associatedData: "f6kf/DPKlb4ig6JJ7Ls0a+EupP3QSrmkj/bTkIuaW6tm+7z2Ugd4cg==",
+		},
+		"plaintext 4": {
+			key:            "BlWwHNYdKgELBFXuS1BIkccaRY6DHPRmIHWAr6gje+Q=",
+			plaintext:      "XXjgvMHQ3CASYjqHhq1i5rplfNHS",
+			associatedData: "23HOhX2Pqm7qWwa8X/lSIQKRQ+s=",
+		},
+	}
+
+	t.Run("cyclic crypto test", func(t *testing.T) {
+		for name, tc := range testCases {
+			t.Run(name, func(t *testing.T) {
+				var ciphertextContainer ciphertextContainer
+				var recPlaintextBytes []byte
+
+				keyBytes, err := base64.StdEncoding.DecodeString(tc.key)
+				require.NoError(t, err)
+				plaintextBytes, err := base64.StdEncoding.DecodeString(tc.plaintext)
+				require.NoError(t, err)
+				associatedDataBytes, err := base64.StdEncoding.DecodeString(tc.associatedData)
+				require.NoError(t, err)
+				t.Run("encrypt", func(t *testing.T) {
+					require := require.New(t)
+					var err error
+					ciphertextContainer, err = symmetricEncryptRaw(keyBytes, plaintextBytes, associatedDataBytes)
+					require.NoError(err, "encryption")
+				})
+
+				t.Run("decrypt", func(t *testing.T) {
+					require := require.New(t)
+					var err error
+					recPlaintextBytes, err = symmetricDecryptRaw(keyBytes, ciphertextContainer, associatedDataBytes)
+					require.NoError(err, "decryption")
+				})
+				require.Equal(t, plaintextBytes, recPlaintextBytes)
+			})
+		}
+	})
+}
--- a/coordinator/internal/transitengineapi/transitengineapi.go
+++ b/coordinator/internal/transitengineapi/transitengineapi.go
@@ -1,0 +1,194 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// transitengine provides all functionality related to the transit engine API endpoints: decrypt and encrypt.
+// It is organized in a layered approach, keeping http request processing separated from the underlying crypto
+// business logic(crypto.go).
+package transitengine
+
+import (
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/edgelesssys/contrast/coordinator/internal/authority"
+)
+
+const (
+	// aesGCMNonceSize specifies the default nonce size in bytes used in AES GCM, used during parsing of ciphertextContainer.
+	aesGCMNonceSize = 12
+	// aesGCMKeySize specifies the default key size in bytes to use AES-256 GCM.
+	aesGCMKeySize = 32
+)
+
+type (
+	// encryptionRequest holds the request-specific plaintext and currently supported, optional query parameters: associatedData and keyVersion.
+	encryptionRequest struct {
+		Plaintext      []byte `json:"plaintext"`
+		KeyVersion     uint32 `json:"key_version"`
+		AssociatedData []byte `json:"associated_data,omitempty"`
+	}
+	// decryptionRequest holds the request-specific ciphertextContainer and currently supported, optional query parameters: associatedData.
+	decryptionRequest struct {
+		CiphertextContainer ciphertextContainer `json:"ciphertext"`
+		AssociatedData      []byte              `json:"associated_data,omitempty"`
+	}
+	// encryptionResponse holds the response-specific ciphertextContainer.
+	encryptionResponse struct {
+		Ciphertext ciphertextContainer `json:"ciphertext"`
+	}
+	// decryptionResponse holds the response-specific plaintext.
+	decryptionResponse struct {
+		Plaintext []byte `json:"plaintext"`
+	}
+)
+
+type stateAuthority interface {
+	GetState() (*authority.State, error)
+}
+
+// NewTransitEngineAPI sets up the transit engine API with a provided seedEngineAuthority.
+func NewTransitEngineAPI(authority stateAuthority, _ *slog.Logger) *http.ServeMux {
+	mux := http.NewServeMux()
+
+	// 'name' wildcard is kept to reflect existing transit engine API specifications:
+	// https://openbao.org/api-docs/secret/transit/#encrypt-data
+	// name <=> workloadSecretID, which should be used for the key derivation.
+	mux.Handle("/v1/transit/encrypt/{name}", getEncryptHandler(authority))
+	mux.Handle("/v1/transit/decrypt/{name}", getDecryptHandler(authority))
+
+	return mux
+}
+
+func getEncryptHandler(authority stateAuthority) http.HandlerFunc {
+	// TODO(jmxnzo): Implement Vault json error bodies
+	return func(w http.ResponseWriter, r *http.Request) {
+		workloadSecretID := r.PathValue("name")
+		if workloadSecretID == "" {
+			http.Error(w, "Invalid URL format", http.StatusBadRequest)
+			return
+		}
+		var encReq encryptionRequest
+		if err := parseRequest(r, &encReq); err != nil {
+			http.Error(w, fmt.Sprintf("parsing encryption request: %v", err), http.StatusBadRequest)
+			return
+		}
+		key, err := deriveEncryptionKey(authority, fmt.Sprintf("%d_%s", encReq.KeyVersion, workloadSecretID))
+		if err != nil {
+			http.Error(w, fmt.Sprintf("key derivation: %v", err), http.StatusInternalServerError)
+			return
+		}
+		ciphertextContainer, err := symmetricEncryptRaw(key, encReq.Plaintext, encReq.AssociatedData)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("encrypting: %v", err), http.StatusInternalServerError)
+			return
+		}
+		ciphertextContainer.keyVersion = encReq.KeyVersion
+		var encResp encryptionResponse
+		encResp.Ciphertext = ciphertextContainer
+		if err = writeJSONResponse(w, encResp); err != nil {
+			http.Error(w, fmt.Sprintf("writing response: %v", err), http.StatusInternalServerError)
+			return
+		}
+	}
+}
+
+func getDecryptHandler(authority stateAuthority) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		workloadSecretID := r.PathValue("name")
+		if workloadSecretID == "" {
+			http.Error(w, "Invalid URL format", http.StatusBadRequest)
+			return
+		}
+		var decReq decryptionRequest
+		if err := parseRequest(r, &decReq); err != nil {
+			http.Error(w, fmt.Sprintf("parsing decryption request: %v", err), http.StatusBadRequest)
+			return
+		}
+		key, err := deriveEncryptionKey(authority, fmt.Sprintf("%d_%s", decReq.CiphertextContainer.keyVersion, workloadSecretID))
+		if err != nil {
+			http.Error(w, fmt.Sprintf("key derivation: %v", err), http.StatusInternalServerError)
+			return
+		}
+		plaintext, err := symmetricDecryptRaw(key, decReq.CiphertextContainer, decReq.AssociatedData)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("decrypting: %v", err), http.StatusInternalServerError)
+			return
+		}
+		var decResp decryptionResponse
+		decResp.Plaintext = plaintext
+		if err = writeJSONResponse(w, decResp); err != nil {
+			http.Error(w, fmt.Sprintf("writing response: %v", err), http.StatusInternalServerError)
+			return
+		}
+	}
+}
+
+// deriveEncryptionKey derives the workload secret used as the encryption key by receiving the seedengine of the current state.
+func deriveEncryptionKey(authority stateAuthority, workloadSecretID string) ([]byte, error) {
+	state, err := authority.GetState()
+	if err != nil {
+		return nil, err
+	}
+	// TODO(jmxnzo): authentication of client certs <-> parsed workloadSecretID.
+	derivedWorkloadSecret, err := state.SeedEngine.DeriveWorkloadSecret(workloadSecretID)
+	if err != nil {
+		return nil, err
+	}
+	if len(derivedWorkloadSecret) < aesGCMKeySize {
+		return nil, fmt.Errorf("derived key too small, expected key length: %d", aesGCMKeySize)
+	}
+	return derivedWorkloadSecret[:aesGCMKeySize], nil
+}
+
+// writeJSONResponse wraps any payload inside a "data" object and sends it as an HTTP response.
+func writeJSONResponse(w http.ResponseWriter, payload any) error {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	response := map[string]any{
+		"data": payload,
+	}
+	return json.NewEncoder(w).Encode(response)
+}
+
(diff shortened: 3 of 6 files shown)


coordinator/internal/seedengine/seedengine.go
```diff
@@ -19,6 +19,11 @@ import (
 	"golang.org/x/crypto/hkdf"
 )
 
+// GetterSeedEngine describes a getter function to return a SeedEngine.
+type GetterSeedEngine interface {
```
3u13r, line 23: Interfaces should be defined in the package they are used.

coordinator/internal/transitengine/crypto.go
```diff
@@ -0,0 +1,146 @@
...
+	additionalData []byte
+}
+
+type encryptionMap struct {
+	Plaintext     string `json:"plaintext"`
+	Ciphertext    string `json:"ciphertext"`
+	symmetricOpts symmetricOpts
+}
+
+// Note: In case only err is returned the resulting httpError will fallbback to InternalServerError status code 500.
```
3u13r, line 34: Could you elaborate a bit more? Since this is not a function documentation, this floats a bit between the functions/code? Is is related to the function below?

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+			if errors.As(err, &httpError) {
+				status = httpError.HTTPStatus()
+				if status == 0 {
+					status = http.StatusInternalServerError
+				}
+			} else {
+				status = http.StatusInternalServerError
+			}
+			http.Error(w, err.Error(), status)
```
3u13r, line 52: Since we also have `w` in the cryptFunc we can write the status code there. Though I'd like to go in the direction that the cryptFunc is a pure function that doesn't know about being in a http request, i.e., remove `w` and `r` from the arguments. Though I haven't found a good solution for that.

coordinator/internal/transitengine/crypto.go
```diff
@@ -0,0 +1,146 @@
...
+	defer r.Body.Close()
+
+	var encryptionMapping encryptionMap
+	if err := json.NewDecoder(r.Body).Decode(&encryptionMapping); err != nil {
+		return nil, httpError{error: err, status: http.StatusBadRequest}
+	}
+
+	encryptionMapping.symmetricOpts = symmetricOpts{
+		convergent:     false,
+		additionalData: []byte("test"),
```
3u13r, line 143: Left over from testing?

coordinator/internal/transitengine/crypto.go
```diff
@@ -0,0 +1,146 @@
...
+	return nil
+}
+
+// symmetricEnc returns the encrypted plaintext based on the symmetric options and encryption key handed in.
+func symmetricEnc(encKey, plaintext []byte, _ symmetricOpts) ([]byte, error) {
+	aesCipher, err := aes.NewCipher(encKey)
+	if err != nil {
+		return nil, err
+	}
+	gcm, err := cipher.NewGCM(aesCipher)
```
3u13r, line 86: Using the raw Go crypto libs feels a bit sharp. I think that tink doesn't have good support for this use-case. Have we checked that there isn't a good library for this?

coordinator/internal/transitengine/crypto.go
```diff
@@ -0,0 +1,146 @@
...
+	encryptionMapping, err := parseEncryptionRequest(r)
+	if err != nil {
+		return err
+	}
+	ciphertext := strings.TrimPrefix(encryptionMapping.Ciphertext, "vault:v1:")
+	ciphertextBytes, err := b64.StdEncoding.DecodeString(ciphertext)
+	if err != nil {
+		return httpError{error: err, status: http.StatusBadRequest}
+	}
+	encryptionMapping.symmetricOpts.nonce = ciphertextBytes[:12]
```
3u13r, line 67: Those inline numbers should definitely be constants, since
1. The name documents the usage
2. this makes sure that we keep using the same number in multiple places.

coordinator/internal/transitengine/crypto_test.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package transitengine
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	r "math/rand"
```
3u13r, line 10: Why naming the import here? Also single letter variables are usually reserved for either function receivers are variables with very limited scope

coordinator/internal/transitengine/crypto_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+					encRespBody := getResponseBody(t, handleEncrypt, *req, key)
+					data, ok := encRespBody["data"].(map[string]any)
+					require.True(t, ok)
+					ciphertext, ok = data["ciphertext"].(string)
+					require.True(t, ok)
+				})
+				t.Run("decryption request handling", func(t *testing.T) {
+					decryptReqBody, err := createReqBodyJSON("ciphertext", ciphertext)
+					require.NoError(t, err)
```
3u13r, line 47: Instead of using `require.NoError(t, err)` we call `require := require.New(t)` in the first lines of a new test and then call `require.NoError(err)`.

coordinator/internal/transitengine/crypto_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/edgelesssys/contrast/internal/crypto"
+
+	"github.com/stretchr/testify/require"
+
+	b64 "encoding/base64"
+)
+
+func TestCryptoAPICyclic(t *testing.T) {
+	t.Run("encrypt-decrypt handler", func(t *testing.T) {
+		for range 10 {
```
3u13r, line 24: What is the idea of calling the test 10 times? You can also run go tests multiple times by specifying `-count 1000`. Generally, using randomness in tests is also an anti-pattern, instead create crypto related things (keys etc) beforehand and use them as constants.

3u13r (changes requested):

coordinator/internal/transitengine/transitengine.go
jmxnzo: I changed the implemented entire pattern here like you suggested. I dropped `w` and `r` for the encryption function and reduced to only using encryptionMap as param.  Also i separated request parsing , encryption operation and response. Therefore customWrapperHandler can set the httpError respectively. Thx for pointing out the pattern @3u13r

coordinator/internal/transitengine/crypto.go
jmxnzo: Will be further implemented soon, to allow usage of additionalData flag.

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/edgelesssys/contrast/coordinator/internal/seedengine"
+)
+
+// NewTransitEngineAPI starts a transit engine API listening on the given port and is initialized with a seedengine getter for key derivation.
+func NewTransitEngineAPI(port int, seedEngineGetter seedengine.GetterSeedEngine, logger *slog.Logger) error {
+	mux := http.NewServeMux()
+	// TODO(jmxnzo): add routing based on inline URL name
+	mux.Handle("/v1/transit/encrypt/autounseal", customWrapperHandler(seedEngineGetter, logger, handleEncrypt))
+	mux.Handle("/v1/transit/decrypt/autounseal", customWrapperHandler(seedEngineGetter, logger, handleDecrypt))
```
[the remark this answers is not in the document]
burgerdev: Sorry, this was a leftover comment from my last review.

coordinator/main.go
```diff
@@ -129,6 +130,17 @@ func run() (retErr error) {
...
+		mux := transitengine.NewTransitEngineAPI(meshAuth, logger)
+		logger.Info("Transit Engine API initialized")
+		port := 8200
+		fmt.Printf("Serving transit engine API on port %d\n", port)
+		if err := http.ListenAndServe(fmt.Sprintf(":%d", port), mux); err != nil {
+			logger.Error("Failed to start transit engine API", "err", err)
+		}
+		return nil
+	})
+
```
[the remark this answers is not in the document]
jmxnzo: Of course, it was intended for testing purposes only and i'll drop this when squashing everything!

3u13r (changes requested):

coordinator/internal/transitengine/crypto.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"crypto/aes"
+	"crypto/cipher"
+
+	"github.com/edgelesssys/contrast/internal/crypto"
+)
+
+// symmetricEncryptRaw returns a ciphertextContainer, based on the encryption key and associatedData handed in.
+func symmetricEncryptRaw(encKey, plaintext b64Plaintext, associatedData []byte) (ciphertextContainer, error) {
+	aesCipher, err := aes.NewCipher(encKey)
```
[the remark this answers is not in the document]
burgerdev: Sorry, looks like I mixed that up - I thought nonce sizes vary with block sizes. If 96bit nonces are standard, we should definitely go with AES256.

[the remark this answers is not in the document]
jmxnzo: Based on the key size at derivation, we now use AES-256 instead with the 96 bit nonces.

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,314 @@
...
+		version:    version,
+		nonce:      fullCiphertext[:aesGCMNonceSize],
+		ciphertext: fullCiphertext[aesGCMNonceSize:],
+	}, nil
+}
+
+// MarshalJSON wraps b64Plaintext in an object with the key "plaintext".
+func (p b64Plaintext) MarshalJSON() ([]byte, error) {
+	return json.Marshal(map[string]string{
+		"plaintext": b64.StdEncoding.EncodeToString(p),
```
burgerdev, line 228: The `encryptionRequest` already maps `plaintext: b64Plaintext`. Does this now marshal as?

```json
{ "plaintext": {"plaintext": "EAWEUng1BEDFQ"}}
```
I guess it does not matter simply because we're only unmarshalling?

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	e.plaintext = plaintext
+	if encAssociatedData, exists := obj["associated_data"]; exists {
+		e.associatedData, err = decodeBase64(encAssociatedData)
+		if err != nil {
+			return fmt.Errorf("decoding associated_data: %w", err)
+		}
+	}
+	return nil
+}
```
burgerdev, line 266: I think we can get rid of the custom `(Un)MarshalJSON`, considering that JSON unmarshalling into `[]byte` does the base64 conversion automatically (cf. https://pkg.go.dev/encoding/json#Marshal). So, define request and response objects, marshal/unmarshal with the standard library and do verification of the obtained objects afterwards (if necessary).

The `symmetricEncryptRaw` signature would also make more sense if it did not refer to base64 at all.

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,314 @@
...
+	if encAssociatedData, exists := obj["associated_data"]; exists {
+		d.associatedData, err = decodeBase64(encAssociatedData)
+		if err != nil {
+			return fmt.Errorf("decoding associated_data: %w", err)
+		}
+	}
+	return nil
+}
+
+// decodeBase64 is a helper function, ensuring that the b64 encoded string is not empty and returning the base64 decoding.
```
burgerdev, line 292: Why do we need the non-empty check?

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,117 @@
...
+		{"Plaintext": "iZo0IezTmQ6Ms1GJUbbY4nrsRydO31b6xuqlJwi+R9xLt1K9uaI8ZiuInXc92qpulYNaAWAiBmNNghKM0dpAPdSXwc93+YCT1Zm2i1cuW7H6Uz5tL7E=", "Ciphertext": "vault:v1:qqjtmjUKfzoAb3VeUgWjbRDdYu3K/cdmi7sEVcKPRiOdbY5OyuQrNQHtYZ/mje8hcPyHnmgDJEDOpjLhQgUG6yoYamqksut//lv7DDbKYbzroro5BRiCqQjhfqnhmna79maV5okq8zI5YZBoSSn36ivu"},
+		{"Plaintext": "FLiOXWBy8oVETtiNzfw0rbMgCfa3DVSfKL4GhR86EcluUe78nLiDxt0HtP5vTwaaz7mvLXu2nOtsdlz8kYY1YrZCLLNlYzjm/vYe++CcH/x+fDKepJem5le0BCsdog==", "Ciphertext": "vault:v1:0qp8vsJ+JFf5m5HekxQeUw0+gj/NdoDcmy7ExSw0G7PB1RBQ80T+TUMjSvmmgu02eQ2oCKEkfFMolfNt1zq+sZkLQuTLpbW8p+Vd8ALPGdyyD20MIb2ez6dm9nMM4jiXL2FkfARuHcHoY3/LCBQVLE+kkJLAdwze/Z4="},
+		{"Plaintext": "Xqv9SzcpNK99JF1I7xRAJ0FOkzka", "Ciphertext": "vault:v1:N431rZS6bcuGDDJ8Jh93yvih4oc5wHGtz02M2vFQ5IioIlZFqfv29nDImWNaZUW7Zw=="},
+		{"Plaintext": "6HQ035OxE30=", "Ciphertext": "vault:v1:ZUuU9vMiCa3CE7PvmYlJYhHQOJm2/Rk6xuUA7DSJu4ExttPl"},
+		{"Plaintext": "yqUBQzznRjbXMxhQkwo5q2Az3/6nvgRQ86uffx8ZqT7rufplfhJz+xDfvi4EOw==", "Ciphertext": "vault:v1:fpjOfUK8hZEk6DmcH715zdyTMAqyW9ymAMxQFgoMR2Q5639NYCA3rbrR4OKfGwCqO7UrBg3LeFksYBPhiO4pbX6dHXGYIfjDdgo="},
```
burgerdev, line 30: This seems to be testing only a narrow part of the API (missing: version, AD), but that part exhaustively. May I suggest to do tests with different plaintexts in `crypto.go` for the low-level function and focusing on the HTTP API here?

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/edgelesssys/contrast/coordinator/internal/authority"
+	"github.com/edgelesssys/contrast/coordinator/internal/seedengine"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCryptoAPICyclic(t *testing.T) {
+	data := []map[string]string{
```
burgerdev, line 21: This should be a 

```go
[]struct{
  name string
  plaintext string
}
```

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,117 @@
...
+		{"Plaintext": "vpIhKQhFuGwLv5B/XLYr960uZQ==", "Ciphertext": "vault:v1:d01M+wZYaG9LFnuc18s8oh6PuVFw3+7DBX4LkXXQ6d64DvmQt6qwjj1MHmA88UE="},
```
burgerdev, line 22: Afaict, `Ciphertext` is unused - and rightfully so, since this API is not deterministic.

coordinator/main.go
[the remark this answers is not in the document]
burgerdev: Let's leave the comment open until it's actually resolved.

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,314 @@
...
+		if strVersion == "" || workloadSecretID == "" {
+			http.Error(w, "Invalid URL format", http.StatusBadRequest)
+			return
+		}
+		version, err := extractVersion(strVersion)
+		if err != nil {
+			http.Error(w, fmt.Sprint("URL version: %w", err), http.StatusBadRequest)
+			return
+		}
+		key, err := deriveEncryptionKey(authority, workloadSecretID+strVersion)
```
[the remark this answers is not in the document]
jmxnzo: Thanks for pointing out and the thorough review! I changed this to your suggestion. StrVersion implicitly always was ensured to match `^v(\d+)$`, therefore i added '_' as seperator and changed the order to mitigate the described attack possibilities.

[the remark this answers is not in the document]
jmxnzo: But as described in the other comments, i dropped the strVersion completely as it does not specify the `key_version` as expected and use the `key_version` value instead.

coordinator/internal/transitengine/transitengine.go
jmxnzo: I got rid of most of the custom (Un)MarshalJSON logic, as you advised. But for the ciphertextContainer we rely on the (Un)marshalling functions, because of the translation of the json string with format "vault:vX:base64"  to the ciphertext container holding the separated fields. 
`symmetricEncryptRaw` operates on []byte again like prior.

coordinator/internal/transitengine/transitengine.go
jmxnzo: I thought we do not want to allow requests with empty plaintext or ciphertext, but dropped this limitation now.

coordinator/internal/transitengine/crypto.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type (
+	// ciphertextContainer describes a base64-encoded ciphertext prepended with the nonce and specified key version.
+	ciphertextContainer struct {
+		nonce      []byte
+		ciphertext []byte
+		keyVersion uint
+	}
+)
```
burgerdev, line 20: Needs gofmt.

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,244 @@
...
+	c.ciphertext = fullCiphertext[aesGCMNonceSize:]
+	return nil
+}
+
+// MarshalJSON marshalls a ciphertextContainer to a json string.
+func (c ciphertextContainer) MarshalJSON() ([]byte, error) {
+	encodedNonce := b64.StdEncoding.EncodeToString(c.nonce)
+	encodedCiphertext := b64.StdEncoding.EncodeToString(c.ciphertext)
+	// Convert to "vault:vX:base64" format
+	versioned := fmt.Sprintf("vault:v%d:%s%s", c.keyVersion, encodedNonce, encodedCiphertext)
```
burgerdev, line 226: What happens if `keyVersion==0`, considering that Vault key versions start with 1?

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,244 @@
...
+		if !strings.HasPrefix(contentType, "application/json") {
+			return fmt.Errorf("invalid content-type: %s (want application/json)", contentType)
+		}
+	}
+	return nil
+}
+
+// UnmarshalJSON umarshalls a json string to a ciphertextContainer holding the version prefix,
+// decoded base64 nonce and ciphertext.
+func (c *ciphertextContainer) UnmarshalJSON(data []byte) error {
```
burgerdev, line 197: `ciphertextContainer` methods should go to `crypto.go`.

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,244 @@
...
+	defer r.Body.Close()
+	if err := validateContentType(r); err != nil {
+		return decryptionRequest{}, err
+	}
+	var decReq decryptionRequest
+	if err := json.NewDecoder(r.Body).Decode(&decReq); err != nil {
+		return decryptionRequest{}, err
+	}
+	return decReq, nil
+}
```
burgerdev, line 183: I think this could be simplified like this:

```go
func parseRequest(r *http.Request, into *any) error {
  // ...
}
// At the call site:
var decReq decryptionRequest
parseRequest(r, &decReq)
```

coordinator/internal/transitengine/transitengine.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	c.nonce = fullCiphertext[:aesGCMNonceSize]
+	c.ciphertext = fullCiphertext[aesGCMNonceSize:]
+	return nil
+}
+
+// MarshalJSON marshalls a ciphertextContainer to a json string.
+func (c ciphertextContainer) MarshalJSON() ([]byte, error) {
+	encodedNonce := b64.StdEncoding.EncodeToString(c.nonce)
+	encodedCiphertext := b64.StdEncoding.EncodeToString(c.ciphertext)
```
burgerdev, line 224: The encoding would be invalid if the nonce size were not divisible by 4. Consider appending first and then encoding for correctness, although it is unlikely to ever become an issue.

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,167 @@
...
+			Version:        2000,
+			expStatus:      200,
+		},
+		"f_empty_name": {
+			Plaintext:      "vpIh",
+			name:           "",
+			AssociatedData: "vpIh",
+			expStatus:      404,
+		},
+		"f_name_with_backsl": {
```
burgerdev, line 49: Nit: that's a forward slash

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,167 @@
...
+					res := rec.Result()
+					require.Equal(entry.expStatus, res.StatusCode)
+
+					if res.StatusCode != http.StatusOK {
+						return
+					}
+					encRespBody, err := parseResponseBody(res)
+					require.NoError(err)
+					data, _ := encRespBody["data"].(map[string]any)
+					ciphertext, _ = data["ciphertext"].(string)
```
burgerdev, line 100: The test should verify that these casts are ok.

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,167 @@
...
+					require.NoError(err)
+					req := httptest.NewRequest(http.MethodPut, "/v1/transit/encrypt/"+entry.name, bytes.NewReader(jsonBody))
+					req.Header.Set("Content-Type", "application/json")
+					rec := httptest.NewRecorder()
+					mux.ServeHTTP(rec, req)
+					res := rec.Result()
+					require.Equal(entry.expStatus, res.StatusCode)
+
+					if res.StatusCode != http.StatusOK {
+						return
```
burgerdev, line 95: Please leave a TODO to implement Vault JSON error bodies: https://developer.hashicorp.com/vault/api-docs#error-response. Does not need to happen in this PR.

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,167 @@
...
+					mux.ServeHTTP(rec, decryptReq)
+					res := rec.Result()
+					require.Equal(entry.expStatus, res.StatusCode)
+					if res.StatusCode != http.StatusOK {
+						return
+					}
+					decRespBody, err := parseResponseBody(res)
+					require.NoError(err)
+					data, _ := decRespBody["data"].(map[string]any)
+					receivedPlaintext, _ = data["plaintext"].(string)
```
burgerdev, line 123: Same.

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,167 @@
...
+					require.Equal(entry.Plaintext, receivedPlaintext, "Unexpected received plaintext after cycling handler functions")
+				})
+			})
+		}
+	})
+}
+
+func parseResponseBody(res *http.Response) (map[string]any, error) {
```
burgerdev, line 131: If I were to inline this function, it would boil down to

```diff
- decRespBody, err := parseResponseBody(res)
- require.NoError(err)
+ defer res.Body.Close()
+ respBody, err := io.ReadAll(res.Body)
+ require.NoError(err)
+ var decRespBody map[string]any
+ require.NoError(json.Unmarshal(respBody, &decRespBody))
```

I'd prefer the inlined version because it does not interrupt my reading flow (need to figure out what `parseResponseBody` does).

burgerdev: Actually, you can make it even simpler with 

```go
require.NoError(json.NewDecoder(respBody).Decode(&decRespBody))
```

coordinator/internal/transitengine/transitengine.go
jmxnzo: Regarding https://developer.hashicorp.com/vault/api-docs/secret/transit#encrypt-data default of key_version is 0, i looked it up to be compatible.

coordinator/internal/transitengine/transitengine.go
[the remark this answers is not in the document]
jmxnzo: Can i mark this as resolved this with the updated separation logic?

coordinator/internal/transitengine/transitengine.go
jmxnzo: The nonce is fixed to 12 bytes(thus divisible by 4), but i agree it is cleaner this way and adapted the suggestion.

coordinator/internal/transitengine/transitengine.go
burgerdev: Are you referring to [`key_version (int: 0)`](https://developer.hashicorp.com/vault/api-docs/secret/transit#key_version)? I took that to mean

> The default for `key_version` is 0, which means "use latest key".

which would be consistent with the example responses in the docs.

coordinator/internal/transitengine/transitengine_test.go
```diff
@@ -0,0 +1,164 @@
...
+						return
+					}
+
+					var decRespBody map[string]any
+					require.NoError(json.NewDecoder(res.Body).Decode(&decRespBody))
+
+					data, exist := decRespBody["data"].(map[string]any)
+					require.True(exist)
+					receivedPlaintext, exist = data["plaintext"].(string)
+					require.True(exist)
```
burgerdev, line 133: These are not existence checks, but type cast checks. If the expected fields are not found, this will panic. You could avoid the type cast by assigning the expected type to begin with:

```
var decRespBody map[string]map[string]string
```

burgerdev (commented):
We're closing in!

coordinator/internal/transitengine/transitengine.go
jmxnzo: I'll add that exception to the documentation of the transit engine API. Because we don't really manage the version keys, this fallback isn't well-suiting for our endpoint specification. Thanks for pointing that out!

burgerdev (approved):
lgtm, thanks a lot!

coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -0,0 +1,191 @@
...
+		if workloadSecretID == "" {
+			http.Error(w, "Invalid URL format", http.StatusBadRequest)
+			return
+		}
+		var encReq encryptionRequest
+		if err := parseRequest(r, &encReq); err != nil {
+			http.Error(w, fmt.Sprintf("parsing encryption request: %v", err), http.StatusBadRequest)
+			return
+		}
+		key, err := deriveEncryptionKey(authority, fmt.Sprintf("%d_", encReq.KeyVersion)+workloadSecretID)
```
burgerdev, line 77:
```suggestion
		key, err := deriveEncryptionKey(authority, fmt.Sprintf("%d_%s", encReq.KeyVersion, workloadSecretID))
```

coordinator/internal/transitengineapi/crypto.go
```diff
@@ -0,0 +1,97 @@
...
+	"strings"
+
+	"github.com/edgelesssys/contrast/internal/crypto"
+)
+
+// ciphertextContainer describes a base64-encoded ciphertext prepended with the nonce and specified key version.
+type ciphertextContainer struct {
+	nonce      []byte
+	ciphertext []byte
+	keyVersion uint
```
3u13r, line 21: Having a uint (which is 64 bit on my machine, but doesn't have to be), makes the parsing more complicated. To also avoid the size ambiguity can we use a usint64 here?

3u13r (changes requested):

coordinator/internal/transitengineapi/transitengineapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"testing"
+
+	"github.com/edgelesssys/contrast/coordinator/internal/authority"
+	"github.com/edgelesssys/contrast/coordinator/internal/seedengine"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTransitAPICyclic(t *testing.T) {
+	type testData struct {
+		name           string
```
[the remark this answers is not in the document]
jmxnzo: https://developer.hashicorp.com/vault/api-docs/secret/transit#encrypt-data name parameter is reused to reflect the existing transit API implementations. I agree that our architecture has a different use as name refers to the workloadsecretID and is used as key name, but i would prefer to keep the reflection right here.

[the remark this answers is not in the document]
3u13r: Ok, can you add references to this in code?, for example in the line where the URL paths are defined?

coordinator/internal/transitengineapi/crypto.go
jmxnzo: Changed it to uint32, that should be sufficient for the key versioning.

coordinator/internal/transitengineapi/transitengineapi_test.go
```diff
@@ -0,0 +1,167 @@
...
+			AssociatedData: "vpIh",
+			expStatus:      404,
+		},
+		"f bad request nob64plain": {
+			Plaintext:      "thi$$hoU_ld+*notwork",
+			name:           "autounseal",
+			AssociatedData: "vpIh",
+			expStatus:      400,
+		},
+		"f bad request nob64AD": {
```
3u13r, line 61: `nob64AD` the AD stands for additional data, right? I think we should spell this out and more generally be more verbose i.e. name this "failed bad request no base64 additional data"

3u13r (changes requested):

3u13r (changes requested):

coordinator/internal/transitengineapi/crypto_test.go
```diff
@@ -0,0 +1,72 @@
...
+		},
+		"plaintext 4": {
+			key:            "BlWwHNYdKgELBFXuS1BIkccaRY6DHPRmIHWAr6gje+Q=",
+			plaintext:      "XXjgvMHQ3CASYjqHhq1i5rplfNHS",
+			associatedData: "23HOhX2Pqm7qWwa8X/lSIQKRQ+s=",
+		},
+	}
+
+	t.Run("cyclic crypto test", func(t *testing.T) {
+		for testName, entry := range encryptionInputTable {
```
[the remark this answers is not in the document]
jmxnzo: I didn't know that we have a naming convention for that. I adapted it in crypto test as well. Thx!

3u13r (approved):
LGTM,
I'd assume that you should be able to fix the CI by rebasing and `just codegen`.

PR #1206, opened 2025-02-03
nodeinstaller: ignore absence of containerd config template

burgerdev:
Commit ff6bfe1 introduced an `os.ReadFile` operation for the containerd config template used by k3s, assuming that the template file would always be present. However, a fresh installation of k3s does not come with a template file:
https://docs.k3s.io/advanced#configuring-containerd.

On a fresh k3s installation, the nodeinstaller will not find the template file, (falsely) assume that there is no config and use an embedded default, which happens to be incompatible with k3s.

Going forward, the nodeinstaller will treat an absent template like an empty template.

### DIFF
--- a/nodeinstaller/node-installer.go
+++ b/nodeinstaller/node-installer.go
@@ -7,8 +7,10 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


msanft:
Diff looks reasonable to me. Did you test it by removing the template file on e.g. discovery?

burgerdev:
I tested it on a VM with a new k3s installation.

msanft (approved):

edgelessci:
Successfully created backport PR for `release/v1.4`:
- #1207

katexochen:
I think this actually changed with newer k3s versions. I remember it would always create that template, that's also the reason we had to build the handling for it.

burgerdev:
I tested with 1.31.5 and 1.30.5, both don't create the template on first install.

PR #1209, opened 2025-02-04
cli/generate: add flag for output file

davidweisse:
This adds an output flag `--output` (`-o`) to `contrast generate` to specify an output file, instead of working in-place. Because a lot of operations during `contrast generate` are done in-place, the simplest solution is to just operate on a temporary copy of the deployment files, and combine the generated YAML into the specified output file.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"log/slog"
 	"os"
@@ -72,6 +73,7 @@ subcommands.`,
 	cmd.Flags().String("image-replacements", "", "path to image replacements file")
 	cmd.Flags().Bool("skip-initializer", false, "skip injection of Contrast Initializer")
 	cmd.Flags().Bool("skip-service-mesh", false, "skip injection of Contrast service mesh sidecar")
+	cmd.Flags().StringP("output", "o", "", "output file for generated YAML")
 	must(cmd.Flags().MarkHidden("image-replacements"))
 	must(cmd.MarkFlagFilename("policy", "rego"))
 	must(cmd.MarkFlagFilename("settings", "json"))
@@ -94,6 +96,14 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	paths, err := findGenerateTargets(args, log)
 	if err != nil {
 		return err
+	}
+	if flags.outputFile != "" {
+		tmpDir, newPaths, err := getTmpPaths(paths)
+		if err != nil {
+			return fmt.Errorf("get temporary paths: %w", err)
+		}
+		defer os.RemoveAll(tmpDir)
+		paths = newPaths
 	}
 
 	// generate a manifest by checking if a manifest exists and using that,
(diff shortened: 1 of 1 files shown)


cli/cmd/generate.go
```diff
@@ -95,6 +97,17 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	if flags.outputFile != "" {
+		if fi, err := os.Stat(flags.outputFile); err == nil && fi.IsDir() {
+			return fmt.Errorf("output file %s is a directory", flags.outputFile)
+		}
```
burgerdev, line 103: Do we need this check? I'd expect a similar error when trying to `os.WriteFile` a directory. If you're concerned about first doing the work and only then discovering a problem with the file, we could also open it here, or even in the flag parser.

cli/cmd/generate.go
```diff
@@ -374,6 +398,43 @@ func runtimeClassNamePatcher(handler string) func(*applycorev1.PodSpecApplyConfi
...
+	}
+	var newPaths []string
+	for _, path := range paths {
+		in, err := os.Open(path)
+		if err != nil {
+			return tmpDir, nil, fmt.Errorf("open %s: %w", path, err)
+		}
+		out, err := os.Create(filepath.Join(tmpDir, filepath.Base(path)))
+		if err != nil {
+			return tmpDir, nil, fmt.Errorf("create %s: %w", path, err)
```
burgerdev, line 414: The `path` argument is not correct here.

cli/cmd/generate.go
```diff
@@ -374,6 +398,43 @@ func runtimeClassNamePatcher(handler string) func(*applycorev1.PodSpecApplyConfi
...
+		}
+		if _, err := io.Copy(out, in); err != nil {
+			return tmpDir, nil, fmt.Errorf("copy %s: %w", path, err)
+		}
+		in.Close()
+		out.Close()
```
burgerdev, line 420: You should also close these on error - maybe it'd be easier to define a `copyFile` function, or wrap the loop body in a closure.

cli/cmd/generate.go
```diff
@@ -374,6 +398,43 @@ func runtimeClassNamePatcher(handler string) func(*applycorev1.PodSpecApplyConfi
...
+}
+
+func getCombinedYAML(paths []string) ([]byte, error) {
+	var combinedYAML []byte
+	for _, path := range paths {
+		data, err := os.ReadFile(path)
+		if err != nil {
+			return nil, fmt.Errorf("read %s: %w", path, err)
+		}
+		combinedYAML = append(combinedYAML, data...)
```
burgerdev, line 433: Are we sure that the yaml documents are properly separated? Does this rely on the yaml rendering from genpolicy?

davidweisse: Yes, it does. Right now, after genpolicy every resource gets prepended with a `---` separator, so appending the files together is no problem. If we don't want to be dependent on that, we could parse each YAML and separate it manually.

cli/cmd/generate.go
davidweisse: That was the idea. If we open/create the file at the beginning however, the output file will still exist, even if we run into an error. So I thought this would be a save alternative.

cli/cmd/generate.go
burgerdev: Nah, it's probably fine, but please leave a small note that this is what we expect here.

burgerdev:
Thanks for checking! I think log messages on stdout are wrong(TM) and should go to stderr. Let's fix this separately, though.

burgerdev (approved):

PR #1213, opened 2025-02-05
coordinator: handle signals

burgerdev:
The TERM and INT signal are usually masked for PID 1, which means that the Coordinator ignores these signals and the container runtime needs to send a KILL signal to terminate it. This results in 30 wasted seconds between the pod deletion request and the deletion itself, which is most annoying in interactive development workflows. Thus, we handle the signal by cancelling the main context, which in turn results in a graceful listener shutdown.

### DIFF
--- a/coordinator/main.go
+++ b/coordinator/main.go
@@ -11,6 +11,8 @@ import (
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


3u13r (approved):
LGTM

PR #1216, opened 2025-02-07
packages/kata-runtime: allow arbitrary CDI annotations

msanft:
As discussed via Teams, there is no sense in checking CDI annotations if the agent doesn't care about them anyway. This allows arbitrary CDI annotations in the policy.


msanft:
[GPU E2E Test](https://github.com/edgelesssys/contrast/actions/runs/13198353025)

burgerdev:
There is a runtime function [`constrainGRPCSpec`](https://github.com/kata-containers/kata-containers/blob/eb1cf792de5b62a6d1897602018bb67db965850d/src/runtime/virtcontainers/kata_agent.go#L982) - if we removed annotations there we would be sure that they are not used by the agent and could keep the current annotation verification.

packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -118,21 +118,27 @@ buildGoModule rec {
       # we currently use.
       # TODO(msanft): Get native CDI working, which will allow us to drop this patch / undo the revert.
       # See https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/5061
-      ./0018-agent-remove-CDI-support.patch
+      # ./0018-agent-remove-CDI-support.patch
```
burgerdev, line 121: I think we should drop the block.

burgerdev (approved):

PR #1222, opened 2025-02-11
docs: document incompatibility with non-CC GPUs

msanft:
This documents the requirement of CC-enabled GPUs in Contrast, which was
previously not mentioned in the docs at all.

### DIFF
--- a/docs/docs/getting-started/bare-metal.md
+++ b/docs/docs/getting-started/bare-metal.md
@@ -39,23 +39,23 @@ Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inoti
 <TabItem value="amd" label="AMD SEV-SNP">
 To enable GPU usage on a Contrast cluster, some conditions need to be fulfilled for *each cluster node* that should host GPU workloads:
 
-1. Ensure that CC-capable GPUs are available on the machine.
+1. Ensure that GPUs supporting confidential computing (CC) are available on the machine.
 
    ```sh
    lspci -nnk | grep '3D controller' -A3
-   ``
+   ```
 
-   which should show a [CC-capable](https://www.nvidia.com/en-us/data-center/solutions/confidential-computing/) GPU like the NVIDIA H100:
+   This should show a [CC-capable](https://www.nvidia.com/en-us/data-center/solutions/confidential-computing/) GPU like the NVIDIA H100:
 
-   ```console-out
+   ```shell-session
    41:00.0 3D controller [0302]: NVIDIA Corporation GH100 [H100 PCIe] [10de:2331] (rev a1)
       Subsystem: NVIDIA Corporation GH100 [H100 PCIe] [10de:1626]
       Kernel driver in use: vfio-pci
       Kernel modules: nvidiafb, nouveau
    ```
 
-   :::warning
-   Non-CC GPUs aren't supported in Contrast.
+   :::info
+   Contrast doesn't support non-CC GPUs.
    :::
 
 2. You must activate the IOMMU. You can check by running:
--- a/docs/docs/getting-started/bare-metal.md
+++ b/docs/docs/getting-started/bare-metal.md
@@ -39,7 +39,26 @@ Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inoti
 <TabItem value="amd" label="AMD SEV-SNP">
 To enable GPU usage on a Contrast cluster, some conditions need to be fulfilled for *each cluster node* that should host GPU workloads:
 
-1. You must activate the IOMMU. You can check by running:
+1. Ensure that CC-capable GPUs are available on the machine.
+
+   ```sh
+   lspci -nnk | grep '3D controller' -A3
+   ``
+
+   which should show a [CC-capable](https://www.nvidia.com/en-us/data-center/solutions/confidential-computing/) GPU like the NVIDIA H100:
+
+   ```console-out
+   41:00.0 3D controller [0302]: NVIDIA Corporation GH100 [H100 PCIe] [10de:2331] (rev a1)
+      Subsystem: NVIDIA Corporation GH100 [H100 PCIe] [10de:1626]
+      Kernel driver in use: vfio-pci
+      Kernel modules: nvidiafb, nouveau
+   ```
+
+   :::warning
+   Non-CC GPUs aren't supported in Contrast.
+   :::
+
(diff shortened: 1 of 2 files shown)


docs/docs/getting-started/bare-metal.md
```diff
@@ -39,21 +39,40 @@ Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inoti
...
+   which should show a [CC-capable](https://www.nvidia.com/en-us/data-center/solutions/confidential-computing/) GPU like the NVIDIA H100:
+
+   ```console-out
+   41:00.0 3D controller [0302]: NVIDIA Corporation GH100 [H100 PCIe] [10de:2331] (rev a1)
+      Subsystem: NVIDIA Corporation GH100 [H100 PCIe] [10de:1626]
+      Kernel driver in use: vfio-pci
+      Kernel modules: nvidiafb, nouveau
+   ```
+
+   :::warning
```
3u13r, line 57: I'd reserve warning for the severity of security problems or similar. `note` or `info` sound important enough.

3u13r (changes requested):

docs/docs/getting-started/bare-metal.md
```diff
@@ -39,21 +39,40 @@ Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inoti
...
+1. Ensure that CC-capable GPUs are available on the machine.
+
+   ```sh
+   lspci -nnk | grep '3D controller' -A3
+   ``
+
+   which should show a [CC-capable](https://www.nvidia.com/en-us/data-center/solutions/confidential-computing/) GPU like the NVIDIA H100:
+
+   ```console-out
+   41:00.0 3D controller [0302]: NVIDIA Corporation GH100 [H100 PCIe] [10de:2331] (rev a1)
```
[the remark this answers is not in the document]
msanft: Id like to have an example of how correct output for a certain GPU would look like. People who do the setup might not have used lspci before.

packages/nixos/gpu.nix
```diff
@@ -123,7 +123,6 @@ in
     # instead.
 
     # Configure the persistenced for use with CC GPUs (e.g. H100).
-    # TODO(msanft): This needs to be adjusted for non-CC-GPUs.
```
burgerdev, line 126: I'd say we should at least leave a note here along the lines of
* This setup does not work for normal GPUs.
* Here are resources for how it _might be possible_ to change this for normal GPUs.

docs/docs/getting-started/bare-metal.md
```diff
@@ -39,21 +39,40 @@ Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inoti
 <TabItem value="amd" label="AMD SEV-SNP">
 To enable GPU usage on a Contrast cluster, some conditions need to be fulfilled for *each cluster node* that should host GPU workloads:
 
-1. You must activate the IOMMU. You can check by running:
+1. Ensure that CC-capable GPUs are available on the machine.
+
+   ```sh
+   lspci -nnk | grep '3D controller' -A3
+   ``
```
thomasten, line 46:
````suggestion
   ```
````

File: docs/docs/getting-started/bare-metal.md [hunk 5]
```diff
@@ -39,21 +39,40 @@ Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inoti
...
+
+   which should show a [CC-capable](https://www.nvidia.com/en-us/data-center/solutions/confidential-computing/) GPU like the NVIDIA H100:
+
+   ```console-out
```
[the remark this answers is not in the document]
thomasten: All three aren't supported by Docusaurus, but `shell-session` is. (https://docusaurus.io/docs/markdown-features/code-blocks#supported-languages)

docs/docs/getting-started/bare-metal.md
```diff
@@ -39,21 +39,40 @@ Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inoti
...
+   41:00.0 3D controller [0302]: NVIDIA Corporation GH100 [H100 PCIe] [10de:2331] (rev a1)
+      Subsystem: NVIDIA Corporation GH100 [H100 PCIe] [10de:1626]
+      Kernel driver in use: vfio-pci
+      Kernel modules: nvidiafb, nouveau
+   ```
+
+   :::info
+   Non-CC GPUs aren't supported in Contrast.
```
thomasten, line 58:
```suggestion
   Contrast doesn't support non-CC GPUs.
```
nit: prefer active voice

docs/docs/getting-started/bare-metal.md
```diff
@@ -39,21 +39,40 @@ Increase the `user.max_inotify_instances` sysctl limit by adding `user.max_inoti
 <TabItem value="amd" label="AMD SEV-SNP">
 To enable GPU usage on a Contrast cluster, some conditions need to be fulfilled for *each cluster node* that should host GPU workloads:
 
-1. You must activate the IOMMU. You can check by running:
+1. Ensure that CC-capable GPUs are available on the machine.
+
+   ```sh
+   lspci -nnk | grep '3D controller' -A3
```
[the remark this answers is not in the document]
msanft: A plain `grep -i nvidia` would also match on the lines below, so the `-A3` would include some clutter.

docs/docs/getting-started/bare-metal.md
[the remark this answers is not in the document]
burgerdev: Together with the command prompt it would be a shell-session, but as is? Prism supports `log`, which may be appropriate, but I'm really looking for just `verbatim` and the lib disappoints me.

packages/nixos/gpu.nix
```diff
@@ -123,8 +123,10 @@ in
...
     # See: https://docs.nvidia.com/cc-deployment-guide-snp.pdf (Page 23 & 24)
+    # Note that the current setup does not work with non-CC-GPUs. (For most use-cases)
```
burgerdev, line 127:
```suggestion
    # Note that the current setup does not support all use cases with non-CC-GPUs.
```

The point is that not all APIs are working and thus the generic use case is unsupported, right?

burgerdev (approved):

thomasten (approved):

3u13r (approved):
LGTM

PR #1223, opened 2025-02-11
cli/generate: automatic cryptsetup configuration

davidweisse:
This implements [RFC 009](https://github.com/edgelesssys/contrast/blob/main/rfc/009-secure-pv-injection.md) and automatically adds the necessary configuration to set up an encrypted mount using the `contrast.edgeless.systems/secure-pv` annotation, which translates to a configured Initializer on `contrast generate`.

A successful run for the `volumestatefulset` test can be found [here](https://github.com/edgelesssys/contrast/actions/runs/13265880625).

### DIFF
--- a/initializer/main.go
+++ b/initializer/main.go
@@ -15,6 +15,8 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/edgelesssys/contrast/internal/atls"
@@ -164,6 +166,21 @@ func run(cmd *cobra.Command, _ []string) (retErr error) {
 		}
 	}
 
-	log.Info("Initializer done")
-	return nil
+	cryptsetupDevicePath := os.Getenv("CRYPTSETUP_DEVICE")
+	if cryptsetupDevicePath == "" {
+		log.Info("Initializer done")
+		return nil
+	}
+
+	log.Info("Setting up encrypted mount")
+
+	flags := &cryptsetupFlags{
+		devicePath:       cryptsetupDevicePath,
+		volumeMountPoint: "/state",
+	}
+
+	ctx, cancel := signal.NotifyContext(ctx, syscall.SIGTERM, syscall.SIGINT)
+	defer cancel()
+
+	return setupEncryptedMount(ctx, log, flags)
 }
--- a/internal/kuberesource/mutators.go
+++ b/internal/kuberesource/mutators.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
 	applyappsv1 "k8s.io/client-go/applyconfigurations/apps/v1"
 	applybatchv1 "k8s.io/client-go/applyconfigurations/batch/v1"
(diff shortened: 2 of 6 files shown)


internal/kuberesource/mutators.go
```diff
@@ -152,6 +210,30 @@ func AddServiceMesh(
...
+	case *applyappsv1.StatefulSetApplyConfiguration:
+		if r.Spec != nil && r.Spec.VolumeClaimTemplates != nil {
+			for _, volumeClaim := range r.Spec.VolumeClaimTemplates {
+				if volumeClaim.Name != nil && *volumeClaim.Name == volumeName {
+					return nil
+				}
+			}
+		}
+	}
+	return fmt.Errorf("device %s not found", volumeName)
```
burgerdev, line 234:
```suggestion
	return fmt.Errorf("device %q not found", volumeName)
```

internal/kuberesource/mutators.go
```diff
@@ -152,6 +210,30 @@ func AddServiceMesh(
...
 }
 
+func checkIfDeviceExists(resource any, spec *applycorev1.PodSpecApplyConfiguration, volumeName string) error {
+	// Check for existing volume with unique name.
+	for _, volume := range spec.Volumes {
+		if volume.Name != nil && *volume.Name == volumeName {
+			if volume.EmptyDir != nil {
+				return fmt.Errorf("device %s cannot be of type EmptyDir", *volume.Name)
+			}
+			return nil
```
burgerdev, line 220: We might be able to exclude more invalid configurations here, because there are not many volume types that can be block devices.

initializer/main.go
```diff
@@ -164,6 +166,20 @@ func run(cmd *cobra.Command, _ []string) (retErr error) {
...
+	}
+
+	log.Info("Setting up encrypted mount")
+
+	flags := &cryptsetupFlags{
+		devicePath:       cryptsetupDevicePath,
+		volumeMountPoint: "/state",
+	}
+
+	ctx, _ = signal.NotifyContext(ctx, syscall.SIGTERM, syscall.SIGINT)
```
burgerdev, line 182:
```suggestion
	ctx, cancel := signal.NotifyContext(ctx, syscall.SIGTERM, syscall.SIGINT)
	defer cancel()
```
I think you're supposed to defer that function, although it probably doesn't matter a lot considering that we're exiting anyway.

internal/kuberesource/mutators.go
```diff
@@ -152,6 +210,30 @@ func AddServiceMesh(
 	return res, retErr
 }
 
+func checkIfDeviceExists(resource any, spec *applycorev1.PodSpecApplyConfiguration, volumeName string) error {
+	// Check for existing volume with unique name.
+	for _, volume := range spec.Volumes {
+		if volume.Name != nil && *volume.Name == volumeName {
+			if volume.PersistentVolumeClaim == nil && volume.ISCSI == nil {
+				return fmt.Errorf("device %q must reference a PVC or iSCSI", volumeName)
```
burgerdev, line 218:
```suggestion
				return fmt.Errorf("volume %q must reference a PVC or iSCSI device", volumeName)
```

burgerdev (approved):
Great!

burgerdev (approved):

PR #1224, opened 2025-02-12
runtime: derive handler from version

burgerdev:
Our runtime handler suffix used to be derived from the guest VM image only, missing the following Contrast components that influence runtime behaviour or attestation:

* node installer
* Kata configuration files
* embedded coordinator policy

While this is not a security issue, it may cause problems when installing multiple Contrast versions side by side. For example, 1.4.0 and 1.4.1 have the same runtime handler name.

This PR resolves the issue at least for release versions, by mixing the version into the runtime handler suffix generation input. While we could also make the version itself part of the suffix, the runtime handler name is already long enough to bring the node installer name close to the limits for k8s objects.

### DIFF
--- a/packages/by-name/kata/contrast-node-installer-image/package.nix
+++ b/packages/by-name/kata/contrast-node-installer-image/package.nix
@@ -222,6 +222,15 @@ let
     ];
   };
 
+  version = ociLayerTar {
+    files = [
+      {
+        source = ../../../../version.txt;
+        destination = "/usr/share/misc/contrast/version.txt";
+      }
+    ];
+  };
+
   layers = [
     installer-config
     kata-container-img
@@ -231,6 +240,7 @@ let
     qemu-tdx
     kata-runtime
     nydus
+    version
   ];
 
   manifest = ociImageManifest {
--- a/packages/by-name/microsoft/contrast-node-installer-image/package.nix
+++ b/packages/by-name/microsoft/contrast-node-installer-image/package.nix
@@ -93,11 +93,21 @@ let
     ];
   };
 
+  version = ociLayerTar {
+    files = [
+      {
+        source = ../../../../version.txt;
+        destination = "/usr/share/misc/contrast/version.txt";
+      }
+    ];
+  };
+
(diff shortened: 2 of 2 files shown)


katexochen:
> node installer

not possible, as the runtime hash is embedded into the node-installer, see https://github.com/edgelesssys/contrast/blob/main/packages/by-name/kata/contrast-node-installer-image/package.nix#L262

> Kata configuration files

not possible for similar reason. We generate these, and might even generate them based on the host arch in the future.

> embedded coordinator policy

This should be doable if we really want it.

I think including the version is a good solution for the release, not sure about dev workflow.

packages/by-name/contrast/package.nix
```diff
@@ -51,7 +51,12 @@ let
     let
       runtimeHandler =
         platform: hashFile:
-        "contrast-cc-${platform}-${builtins.substring 0 8 (builtins.readFile hashFile)}";
+        let
+          runtimeHash = builtins.readFile hashFile;
+          inherit (contrast) version;
+          runtimeVersionHash = builtins.hashString "sha256" "${version}${runtimeHash}";
```
katexochen, line 57: I don't really like doing this here, pretty hard to understand the hash if it is extended here and there. Can we add the version file as layer into the node-installer-image and include it in the hash creation in https://github.com/edgelesssys/contrast/blob/main/packages/by-name/kata/contrast-node-installer-image/package.nix#L261-L264 ?

katexochen:
> > node installer
> 
> not possible, as the runtime hash is embedded into the node-installer, see https://github.com/edgelesssys/contrast/blob/main/packages/by-name/kata/contrast-node-installer-image/package.nix#L262

Never mind, this IS possible: if we build the node-installer twice, once without the runtime-name, then hash the output, then again with runtime name.

packages/by-name/contrast/package.nix
burgerdev: Added a layer containing the version file.

katexochen (approved):

PR #1225, opened 2025-02-12
ci: cancel old workflow runs on push

burgerdev:


burgerdev:
New push cancels old run: https://github.com/edgelesssys/contrast/actions/runs/13284368504/job/37089611778.

.github/workflows/e2e_on_pull_request.yml
```diff
@@ -7,13 +7,18 @@ on:
...
+      - ".github/workflows/e2e_on_pull_request.yml"
       - "!dev-docs/**"
       - "!docs/**"
       - "!rfc/**"
       - "!tools/asciinema/**"
       - "!tools/vale/**"
 
+# Cancel old runs of this workflow for this pull request if new commits are pushed.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
```
3u13r, line 19:
```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```
This looks more in line with what other people are doing.

3u13r (approved):

.github/workflows/e2e_on_pull_request.yml
burgerdev: I chose PR number because I did not want to cancel non-PR jobs (e.g. on push), but on the other hand we only have that one trigger anyway.

3u13r: Ah, this is also something I saw. I think we could implement that idea also, despite the workflow only running on PRs anyway. Official GitHub example: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-only-cancel-in-progress-jobs-on-specific-branches. Also both solutions of https://stackoverflow.com/questions/68418857/how-to-cancel-existing-runs-when-a-new-push-happens-on-github-actions-but-only seem like good solutions.

burgerdev: Since `github.ref` is `refs/pull/<pr_number>/merge` anyway, I'd say we keep your suggestion.

PR #1230, opened 2025-02-13
cli: don't panic on absent pod spec annotations

burgerdev:
The CLI fails with a segmentation fault if a StatefulSet without annotations is among the pod resources. This PR adds a nil check for this situation.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -355,7 +355,7 @@ func injectInitializer(resources []any) error {
 	for _, resource := range resources {
 		switch r := resource.(type) {
 		case *applyappsv1.StatefulSetApplyConfiguration:
-			if r.Spec != nil && r.Spec.Template != nil &&
+			if r.Spec != nil && r.Spec.Template != nil && r.Spec.Template.Annotations != nil &&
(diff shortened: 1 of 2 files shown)


cli/cmd/generate.go
katexochen, line 1: Can we have a test for this?

cli/cmd/generate.go
```diff
@@ -355,7 +355,7 @@ func injectInitializer(resources []any) error {
 	for _, resource := range resources {
 		switch r := resource.(type) {
 		case *applyappsv1.StatefulSetApplyConfiguration:
-			if r.Spec != nil && r.Spec.Template != nil &&
+			if r.Spec != nil && r.Spec.Template != nil && r.Spec.Template.Annotations != nil &&
```
3u13r, line 358: Should we add a test for this to make sure that we don't regress? I'm fine with doing this in a follow up PR if we want to still release today.

katexochen (approved):

3u13r (approved):
LGTM

burgerdev:
As discussed offline, going to add a test in a follow-up PR.

PR #1227, opened 2025-02-12
e2e: allow for delays in negative policy test

burgerdev:

PR #1233, opened 2025-02-16
e2e: use correct testing context in policy test

burgerdev:
This is a follow-up fix for #1227, where I accidentally reused the `require` object in the closure instead of allocating a new one. That caused the test to only execute the `Eventually` block once.

### DIFF
--- a/e2e/policy/policy_test.go
+++ b/e2e/policy/policy_test.go
@@ -167,7 +167,7 @@ func TestPolicy(t *testing.T) {
 		// connect to the Coordinator. Thus, retry for some time until an attestation failure happens.
 		require.EventuallyWithT(func(t *assert.CollectT) {
 			newFailures := getFailures(ctx, t, ct)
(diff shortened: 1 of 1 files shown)


katexochen:
why the switch from require to assert?

burgerdev:
It does not make a difference for the single assertion, but this way I can keep the outer `require` object as is and don't need to rename stuff or add a lot of `t, `s.

Maybe the `require := require.New(t)` pattern is just not good?

katexochen (approved):

katexochen:
> It does not make a difference for the single assertion, but this way I can keep the outer `require` object as is and don't need to rename stuff or add a lot of `t, `s.

makes sense

> Maybe the `require := require.New(t)` pattern is just not good?

don't know, it's convenient.

burgerdev:
I thought about changing the pattern to 

```go

import stretchrreq "github.com/stretchr/testify/require"
require := stretchrreq.New(t)
```

but it doesn't nicely roll off the tongue either.

PR #1235, opened 2025-02-17
e2e: improve workloadsecrettest timeout behavior

burgerdev:
We've seen an increased amount of nightly failures for the workload secret test since #1104 introduced a new subtest. This PR aims to improve the situation by:

* using the test's timeout exclusively for the restart operation (the `ct.Set` command has its own timeout)
* restarting the pods in parallel and only waiting serially

While looking at the test, I saw that `FactorPlatformTimeout` has an explicit list of slow platforms. I think we should not silently default to the shorter timeout, because it's easy to forget to update this switch statement when a new platform is introduced.


jmxnzo (approved):
LGTM!

PR #1237, opened 2025-02-18
Expand coordinator store interface

davidweisse:
This adds the `Watch()` function to the `Store` interface of the Coordinator. This function returns a channel which receives updates to the store, and a cancel function which can be used to stop watching.

The `NewStore()` method is separated and brought to a new file, so it can be controlled via a `enterprise` build tag.

### DIFF
--- a/coordinator/history/history.go
+++ b/coordinator/history/history.go
@@ -251,4 +251,9 @@ type Store interface {
 	// If the current value is not equal to oldVal, an error must be returned. The comparison must
 	// treat a nil slice the same as an empty slice.
 	CompareAndSwap(key string, oldVal, newVal []byte) error
+
+	// Watch watches for changes to the value of key.
+	//
+	// If the value of key changes, the new value is sent on the channel.
+	Watch(key string) (ch <-chan []byte, cancel func(), err error)
 }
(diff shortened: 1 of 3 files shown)


coordinator/history/history.go
```diff
@@ -251,4 +246,9 @@ type Store interface {
 	// If the current value is not equal to oldVal, an error must be returned. The comparison must
 	// treat a nil slice the same as an empty slice.
 	CompareAndSwap(key string, oldVal, newVal []byte) error
+
+	// Watch watches for changes to the value of key.
+	//
+	// If the value of key changes, the new value is sent on the channel.
+	Watch(key string) (ch <-chan []byte, cancel func(), err error)
```
katexochen, line 253: Why did we drop the semantics of watch returning a watcher?

davidweisse: That's the simpler solution, since we would need a new `watcher` package outside of the Coordinator to isolate the Interface and avoid circular dependencies. But if you prefer that, I can change it.

katexochen: ok, just wanted to know.

burgerdev (approved):

PR #1238, opened 2025-02-18
attestation: get product from attestation instead of report

3u13r:
Non v3 Reports don't have the eax bits set in the report. We already have the product in the attestation, just use that one.


burgerdev (approved):
Nice workaround. Would it be feasible to test this?

katexochen (approved):

3u13r:
> Would it be feasible to test this?

Yeah, we thought about testing the whole caching feature e2e by redirecting the go-sev-guest lib to an invalid kds endpoint on the client side. Though the endpoint is not yet configurable upstream. Since this feature "broke" once already, we might want to build this eventually(tm).

edgelessci:
Successfully created backport PR for `release/v1.5`:
- #1240

PR #1239, opened 2025-02-18
grpc: retry connecting to coordinator on EOF

burgerdev:
If the gRPC connection is terminated before the TLS handshake takes place, the ultimate cause reported by the gRPC library is EOF. In this situation, we should not abort the request, but retry.


katexochen (approved):

edgelessci:
Successfully created backport PR for `release/v1.5`:
- #1241

PR #1243, opened 2025-02-19
kata.kata-runtime: 3.13.0 -> 3.14.0

katexochen:
https://github.com/kata-containers/kata-containers/releases/tag/3.14.0


packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -11,19 +11,18 @@
...
       repo = "kata-containers";
       rev = version;
-      hash = "sha256-xBEK+Tczc4MVnETx5sV9sb5/myxLeP7YDDigTroN4Lg=";
+      hash = "sha256-OcKsD9y1kb4IgYyCsqW3aHMB4pDFlIWfAUzoUJVe8pQ=";
     };
 
     patches = [
-      ./0001-govmm-Directly-pass-the-firwmare-using-bios-with-SNP.patch
-      ./0002-emulate-CPU-model-that-most-closely-matches-the-host.patch
+      ./0001-emulate-CPU-model-that-most-closely-matches-the-host.patch
```
katexochen, line 25: Implemented upstream in https://github.com/kata-containers/kata-containers/commit/c1ca49a66c16e1a1e817591e1bba085c8fd0ffe6

packages/by-name/kata/kata-kernel-uvm/package.nix
```diff
@@ -59,7 +59,6 @@ let
           --replace-fail 'CONFIG_MODULE_SIG=y' 'CONFIG_MODULE_SIG=n' \
           --replace-fail '# CONFIG_DMIID is not set' 'CONFIG_DMIID=y' \
           --replace-fail '# CONFIG_TMPFS_POSIX_ACL is not set' 'CONFIG_TMPFS_POSIX_ACL=y' \
-          --replace-fail '# CONFIG_TMPFS_XATTR is not set' 'CONFIG_TMPFS_XATTR=y' \
```
3u13r, line 62: This doesn't belong in this commit, right?

3u13r (approved):
Only one nit. Reviewed mostly the diff of the diffs. Since a lot pf the patches on top come from @burgerdev, he should have a look too, I think.

packages/by-name/kata/kata-kernel-uvm/package.nix
katexochen: It does. It was added upstream in https://github.com/kata-containers/kata-containers/pull/10790, so --replace-fail will fail after the update as the line we are trying to replace is gone.

3u13r: But it is part of the commit named "kuberesource: increase memory limit for openssl". I get that it's needed in the PR and needed because of the update, but this should be either part of the upgrade commit (for this to be bisectable) or a separate PR, or what am I missing?

katexochen: True. <3

burgerdev (approved):
Thanks a lot, lgtm!

PR #1244, opened 2025-02-21
cli: propagate last error from setLoop on timeout

burgerdev:
The retry loop we use for setting a manifest at the Coordinator used to have a context with deadline. If a retriable, but persistent error occured, it would retry until the context expired and then return the context error. This error is usually the least interesting - rather, we want to see the last error that the wrapped set call returned. Thus, this commit changes the returned error to include both the reason for the retry loop quitting as well as the last inner error seen.

If the retry loop expiration is controlled by the input context, and that context is also passed to the wrapped call, it is very likely that we will see the context expiration error as last inner error. To avoid this, we don't attach a deadline to the outer context, but configure the retrier with an explicit deadline. The aTLS handshake already has a timeout of 30s, and we add a 5s timeout to the wrapped call to account for storage sync and update.

This commit also reworks the retry_test to make it easier to follow. The stubDoer now owns the list of errors and the clock, which it increments during the Do call. Rewriting the test uncovered a bug, where a retry could be triggered although the context already expired, if the ticker channel had a value and select chose that over the context done channel.


internal/retry/retry.go
```diff
@@ -14,44 +15,66 @@ import (
...
+		case <-timer.C():
+			return fmt.Errorf("retry deadline exceeded: last error: %w", err)
+		case <-ctx.Done():
+			return fmt.Errorf("%w: last error: %w", ctx.Err(), err)
+		default:
 		}
 
 		select {
+		case <-timer.C():
+			return err
```
davidweisse, line 75:
```suggestion
			return fmt.Errorf("retry deadline exceeded: last error: %w", err)
```

internal/retry/retry.go
```diff
@@ -14,44 +15,66 @@ import (
...
 		case <-ctx.Done():
-			return ctx.Err()
+			return err
```
davidweisse, line 77:
```suggestion
			return fmt.Errorf("%w: last error: %w", ctx.Err(), err)
```

davidweisse (approved):

cli/cmd/set.go
```diff
@@ -265,6 +265,9 @@ type setDoer struct {
 }
 
 func (d *setDoer) Do(ctx context.Context) error {
+	// 30s hard-coded in aTLS + 5s for Coordinator storage sync.
+	ctx, cancel := context.WithTimeout(ctx, 35*time.Second)
```
3u13r, line 269: If we want a greater timeout, then we might want to define those in a central place. Also I think 35 might be a bit on the low side, since we set the minimal timeout to 30 seconds, it might very well be higher.

internal/retry/retry.go
```diff
@@ -14,44 +15,66 @@ import (
...
+// deadline passed).
 func (r *IntervalRetrier) Do(ctx context.Context) error {
 	ticker := r.clock.NewTicker(r.interval)
 	defer ticker.Stop()
 
+	timer := r.clock.NewTimer(r.deadline.Sub(r.clock.Now()))
+	defer timer.Stop()
+
 	for {
 		err := r.doer.Do(ctx)
```
3u13r, line 54: > If the retry loop expiration is controlled by the input context, and
that context is also passed to the wrapped call, it is very likely that
we will see the context expiration error as last inner error. To avoid
this, we don't attach a deadline to the outer context, but configure the
retrier with an explicit deadline. 


The problem is that when this ctx times out, we get a unintesting error, right? 
Would it be simpler/also feasible to simply save the lastErr in the IntervalRetrier and when we Do returns an error, we check if the context is done and then return both ctx.Err and r.lastErr (or we just always return both)?

3u13r (changes requested):

internal/retry/retry.go
burgerdev: That's what I'm doing below, isn't it?

```go
		case <-timer.C():
			return fmt.Errorf("retry deadline exceeded: last error: %w", err)
		case <-ctx.Done():
			return fmt.Errorf("%w: last error: %w", ctx.Err(), err)
```

Just from the returned error of the doer, I can't tell whether this was caused by the deadline externally imposed on me, the IntervalRetrier, or by a more strict deadline internal to the doer. Also, we'd have to track the last two errors, because the last error might just be the context error that I'm already aware of.

3u13r: My idea is to not have another timeout/ctx I want to think about. The context I pass to `retrier.Do` should be the context of the retrier itself. If I want the retrier to end after a certain amount I pass a `context.WithTimeout()`. 

I think the following things always return the interesting error:

1. Do returns err
2. if err is not retriable, return the error and the lastErr (this will become important later)
3. if err is retriable, check if ctx is done, if so return err and lastErr
4. save err as lastErr

if the inner error is a deadline exceeded because the retrier ctx expired then we always add the error before this deadline exceeded.

3u13r: > I'm not entirely happy with the most recent error being replaced by the last error on context expiration, 

You mean because if the .Do returns a context deadline exceeded, the error is retriable but in most cases uninsteresting?
Maybe we can also write a error filter function that we give to the retrier like we do with the retriable(err error) function.

Also I think that using only the context is an abstraction that makes all the invocations of the retrier easier, since on the invocation site, the calling function might themselves not know if the context has a timeout.

internal/retry/retry.go
```diff
@@ -49,9 +52,21 @@ func (r *IntervalRetrier) Do(ctx context.Context) error {
...
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
+			if lastErr == nil {
+				// The context already expired after the first attempt, so we just return the error we have.
+				lastErr = err
+			}
+			return fmt.Errorf("retries failed before %w: last err: %w", ctx.Err(), lastErr)
+		default:
+			lastErr = err
```
3u13r, line 64: My idea was to set this e.g. in `case <-ticker.C():` below and change `return fmt.Errorf("retries failed before %w: last err: %w", ctx.Err(), lastErr)` to `return fmt.Errorf("retries failed before %w: err: %w last err: %w", ctx.Err(), err, lastErr)` This way we also should return all interesting error, right?

3u13r (changes requested):

internal/retry/retry.go
burgerdev: I also thought of simply `errors.Join`ing everything, but at the end of the day this error is directly user visible so I'd at least try to reduce the amount of repetition. In the case of `ctx.Done`, what's the chance that the last error is anything other than `ctx.Err` (potentially wrapped)?

internal/retry/retry.go
burgerdev: Sorry, there may have been a misunderstanding: I was arguing in favour of your suggestion, using only the context for controlling the retry loop.

internal/retry/retry.go
3u13r: If ctx is done then I expect the err to the a wrapped instance of the ctx error. The lastErr is the more interesting error, I think, since it shows why the requests until that point weren't successful.

3u13r (approved):
LGTM

PR #1245, opened 2025-02-21
docs: troubleshooting CLI connection errors

burgerdev:
The error messages will only be propagated as described here after #1244 is merged.

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -177,3 +177,20 @@ LAST SEEN   TYPE      REASON      OBJECT                            MESSAGE
 ```
 
 This error can be resolved by increasing the memory limit of the containers, see the [Workload deployment](deployment.md#pod-resources) guide.
+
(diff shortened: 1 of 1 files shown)


m1ghtym0 (approved):
Thanks, very useful!
Would make sense to me to backport this into the 1.5 released docs before merging, so it's immediatly visible to people trying out Contrast right now.

burgerdev:
I'd rather not cherry-pick #1244 into a patch release, and without it the error messages are not visible. Since there is not much left on the 1.6 milestone, I'd rather try to release that soon.

m1ghtym0:
> I'd rather not cherry-pick #1244 into a patch release, and without it the error messages are not visible. Since there is not much left on the 1.6 milestone, I'd rather try to release that soon.

Sorry, I missed the fact with the error messages. Just checked the preview and didn't read the PR message carefully enough.
There is no need to port to the last release, then.

PR #1246, opened 2025-02-24
kata.kata-agent: disable lto

katexochen:


3u13r:
This is for fast compile times https://doc.rust-lang.org/cargo/reference/profiles.html#lto ?

3u13r:
Should we document this in code? We might want to have a release build at some point which enables this right?

katexochen:
I've added a comment. I can't think of a task the agent does that would require such level of optimization.

3u13r (approved):
LGTM

katexochen:
> Afaiu, LTO is used by Kata to drive down the agent size. Last time I checked, this amounted to 8MiB/30%, which is low enough that I'm willing to pay that price.

so, should we drop this pr then? Or add a release override?

PR #1247, opened 2025-02-24
kata.genpolicy: fail when layer can't be processed

katexochen:


burgerdev (approved):
lgtm. The error on SNP is expected, I guess, but the error on TDX is a bit strange, given that you based on your memory limit fix.

PR #1248, opened 2025-02-25
atls: forward context to validators

burgerdev:
Validators often need to request resources from the internet, for example CRLs. If these resources are unavailable, validation might time out on the TLS handshake level without much information about the root cause.

This commit allows the atls package to forward the context of the TLS handshake to the individual validators. Missing here and to be added in future commits:

- Creating a context with appropriate timeout in atlscredentials.
- Forwarding the context to the HTTPSGetters for SNP/TDX.

### DIFF
--- a/internal/atls/atls.go
+++ b/internal/atls/atls.go
@@ -58,13 +58,12 @@ func CreateAttestationServerTLSConfig(issuer Issuer, validators []Validator, att
 
 // CreateAttestationClientTLSConfig creates a tls.Config object that verifies a certificate with an embedded attestation document.
 //
-// ATTENTION: The tls.Config ensures freshness of the server's attestation only for the first connection it is used for.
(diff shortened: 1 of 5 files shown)


3u13r:
Neat. Regarding `// TODO(burgerdev): equip HTTPSGetter with context.` it would likely be best if one could just pass the ctx directly to the .Verify() call, right? Passing the context around the library directly to our cached Getter is a bit weird. Is this something we want to build upstream?

3u13r (approved):

burgerdev:
> Neat. Regarding `// TODO(burgerdev): equip HTTPSGetter with context.` it would likely be best if one could just pass the ctx directly to the .Verify() call, right? Passing the context around the library directly to our cached Getter is a bit weird. Is this something we want to build upstream?

Yes, that thought came to me, too. Maybe an additional `VerifyContext` and a type cast into `ContextHTTPSGetter` or something like that.

PR #1251, opened 2025-02-25
cli: fix panic on set without workload owner key

thomasten:
`contrast set` panics if no workload owner key is provided. This is because `clientConnection.privKey` has the value `(*ecdsa.PrivateKey)(nil)`, which is not equal to nil. (The panic happens later in `publicKey()` when the underlying value nil is used.) We can fix this either by assigning nil in `set.go`, or by checking the underlying value in `atls.go`. So only one of the two changes of this PR is needed, but I'm not sure what is the preferred one.

### DIFF
--- a/cli/cmd/set.go
+++ b/cli/cmd/set.go
@@ -8,6 +8,7 @@ import (
 	"crypto/ecdsa"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -82,7 +83,9 @@ func runSet(cmd *cobra.Command, args []string) error {
 	}
 
 	workloadOwnerKey, err := loadWorkloadOwnerKey(flags.workloadOwnerKeyPath, &m, log)
-	if err != nil {
+	if errors.Is(err, os.ErrNotExist) {
+		workloadOwnerKey = nil
+	} else if err != nil {
 		return fmt.Errorf("loading workload owner key: %w", err)
 	}
 
@@ -109,9 +112,15 @@ func runSet(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("getting validators: %w", err)
 	}
-	dialer := dialer.NewWithKey(atls.NoIssuer, validators, atls.NoMetrics, &net.Dialer{}, workloadOwnerKey)
 
-	conn, err := dialer.Dial(cmd.Context(), flags.coordinator)
+	var dialr *dialer.Dialer
+	if workloadOwnerKey == nil {
+		dialr = dialer.New(atls.NoIssuer, validators, atls.NoMetrics, &net.Dialer{})
+	} else {
+		dialr = dialer.NewWithKey(atls.NoIssuer, validators, atls.NoMetrics, &net.Dialer{}, workloadOwnerKey)
+	}
+
+	conn, err := dialr.Dial(cmd.Context(), flags.coordinator)
 	if err != nil {
 		return fmt.Errorf("failed to dial coordinator: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


katexochen:
We should have a unit test to ensure this is done correct in the future.

internal/atls/atls.go
```diff
@@ -347,7 +348,7 @@ func (c *clientConnection) verify(rawCerts [][]byte, verifiedChains [][]*x509.Ce
 func (c *clientConnection) getCertificate(cri *tls.CertificateRequestInfo) (*tls.Certificate, error) {
 	priv := c.privKey
 
-	if priv == nil {
+	if priv == nil || reflect.ValueOf(priv).IsNil() {
```
katexochen, line 351:
```suggestion
	if priv == nil || priv != (*ecdsa.PrivateKey)(nil) {
```
we don't want reflect, I think we could compare withe the typed nil pointer here instead?

cli/cmd/set.go
```diff
@@ -231,7 +231,7 @@ func policyMapToBytesList(m []deployment) [][]byte {
 	return policies
 }
 
-func loadWorkloadOwnerKey(path string, manifst *manifest.Manifest, log *slog.Logger) (*ecdsa.PrivateKey, error) {
+func loadWorkloadOwnerKey(path string, manifst *manifest.Manifest, log *slog.Logger) (crypto.PrivateKey, error) {
```
3u13r, line 234: I think the underlying error here is returning `nil, nil` and then using the non-error value as if nothing happened.
I think this function should simply return the error of `os.ReadFile()` (Note that we currently also continue on all errors that are not a IsNotExists error). Then in the caller, we check for `os.IsNotExist()` and the use a nil value without type as the privateKey.

3u13r (changes requested):

katexochen:
Please rebase on main to fix the CI issue.

internal/atls/atls.go
thomasten: Yes, but that wouldn't cover other key types of course. But I think I'd prefer that too.

cli/cmd/set.go
thomasten, line 1: > We should have a unit test to ensure this is done correct in the future.

 Not sure how I would add a unit test for the CLI, but I can probably add one for atls if we decide to fix it there.

cli/cmd/set.go
thomasten: Oh yes, at least there's an error check missing.
Regarding your suggestion, I think both yours and the current implementation can make sense. (For the latter, nil is a valid workload owner key semantically for the set command.) I don't have a preference. @katexochen?

katexochen: I agree that we should fix the error and work with that. Seems to be the least error prone to do.

3u13r (approved):
LGTM, though, I'm not sure if we shouldn't _also_ fix the other side in a follow up. I personally like quite defensible written functions which are crypto-related. But we'd likely need to check more places where we receive an pointer argument.

cli/cmd/set.go
```diff
@@ -109,9 +112,15 @@ func runSet(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("getting validators: %w", err)
 	}
-	dialer := dialer.NewWithKey(atls.NoIssuer, validators, atls.NoMetrics, &net.Dialer{}, workloadOwnerKey)
 
-	conn, err := dialer.Dial(cmd.Context(), flags.coordinator)
+	var dialr *dialer.Dialer
+	if workloadOwnerKey == nil {
```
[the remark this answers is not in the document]
thomasten: workloadOwnerKey is used a second time later in the code, so one may argue that connecting the block might make this less intuitive. I don't have a strong opinion on this though.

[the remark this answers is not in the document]
burgerdev: I missed that. Please proceed :)

burgerdev (approved):

PR #1253, opened 2025-02-26
enterprise: add config map backed history store

davidweisse:
This implements the Store interface using Kubernetes Config Maps in preparation for the distributed Coordinator. For now, this can be tested by manually setting the `enterprise` build tag during the build process.

### DIFF
--- a/enterprise/coordinator/history/configmapstore.go
+++ b/enterprise/coordinator/history/configmapstore.go
@@ -1,0 +1,230 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//go:build enterprise
+
+package history
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+)
+
+// timeout is the timeout for Kubernetes API calls.
+const timeout = 10 * time.Second
+
+var keyRe = regexp.MustCompile(`^[a-zA-Z0-9-]+/[a-zA-Z0-9-]+$`)
+
+// ConfigMapStore is a Store implementation backed by Kubernetes Config Maps.
+type ConfigMapStore struct {
+	client    kubernetes.Interface
+	namespace string
+	uid       types.UID
+	logger    *slog.Logger
+}
+
+// NewConfigMapStore creates a new instance backed by Kubernetes Config Maps.
+func NewConfigMapStore(client kubernetes.Interface, namespace string, log *slog.Logger) (*ConfigMapStore, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	self, err := client.CoreV1().Pods(namespace).Get(ctx, os.Getenv("HOSTNAME"), metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	ownerRefs := self.GetOwnerReferences()
+	var statefulset metav1.OwnerReference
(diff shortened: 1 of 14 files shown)


enterprise/coordinator/history/configmapstore.go
```diff
@@ -0,0 +1,171 @@
...
+					UID:        s.uid,
+				},
+			},
+		},
+		BinaryData: map[string][]byte{filepath.Base(key): value},
+	}
+}
+
+func objectName(key string) string {
+	return fmt.Sprintf("contrast-store-%s-%s", filepath.Dir(key), filepath.Base(key))
```
burgerdev, line 170: This assumes that paths always have one level of nesting, and that there's no overlap between dir names (`foo/bar-baz` vs. `foo-bar/baz`,). These are probably fine assumptions, but we should at least document the constraints on the interface and enforce them here. "Keys must consist of two alphanumeric identifiers separated by forward slash".

enterprise/coordinator/history/configmapstore.go
```diff
@@ -0,0 +1,171 @@
...
+	return err
+}
+
+// CompareAndSwap updates the key to newVal if its current value is oldVal.
+func (s *ConfigMapStore) CompareAndSwap(key string, oldVal, newVal []byte) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	cm, err := s.client.CoreV1().ConfigMaps(s.namespace).Get(ctx, objectName(key), metav1.GetOptions{})
+	if err != nil && !(errors.IsNotFound(err) && len(oldVal) == 0) {
```
burgerdev, line 87: We should compare `oldVal == nil` to distinguish empty value from absent value.

enterprise/coordinator/history/configmapstore.go
```diff
@@ -0,0 +1,171 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package history
```
burgerdev, line 4: As discussed privately, this package should be enterprise-tagged.

enterprise/coordinator/history/configmapstore.go
davidweisse: A `nil` latest transition currently gets marshalled into an empty slice beforehand (https://github.com/edgelesssys/contrast/blob/main/coordinator/history/history.go#L199), so we can't distinguish anyways.

burgerdev: Oh, I see. Latest transitions are by definition not empty, so that's all fine with the current usage pattern. However, the semantics of the store interface still bother me, but we can address that separately.

enterprise/coordinator/history/configmapstore.go
burgerdev: > Keys must consist of two alphanumeric identifiers separated by forward slash

... would be a way to avoid that (alphanumeric == `[a-zA-Z0-9]`).

burgerdev: We should enforce this going forward.

enterprise/coordinator/history/configmapstore.go
```diff
@@ -0,0 +1,225 @@
...
+		uid:       statefulset.UID,
+		logger:    log,
+	}, nil
+}
+
+// Get the value for key.
+func (s *ConfigMapStore) Get(key string) ([]byte, error) {
+	if !keyRe.MatchString(key) {
+		return nil, fmt.Errorf("invalid key %q", key)
+	}
```
burgerdev, line 76: It could make sense to merge `objectName` and the check, since you're destructuring the string the same way in both functions. And the resulting function could have a nice little unit test.

burgerdev (approved):

enterprise/coordinator/history/configmapstore.go
```diff
@@ -0,0 +1,234 @@
...
+type ConfigMapStore struct {
+	client    kubernetes.Interface
+	namespace string
+	uid       types.UID
+	logger    *slog.Logger
+}
+
+// NewConfigMapStore creates a new instance backed by Kubernetes Config Maps.
+func NewConfigMapStore(client kubernetes.Interface, log *slog.Logger) (*ConfigMapStore, error) {
+	namespace, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
burgerdev, line 44: Let's make the namespace an input, that way you should be able to test with `NewConfigMapStore` instead of creating the struct directly.

burgerdev (approved):
lgtm

burgerdev (approved):

PR #1255, opened 2025-02-27
atls: add handshake timeout to aTLS servers and clients

burgerdev:
This is factored out from #1244. We want to explicitly configure timeouts on the server side that are smaller than on the client side, otherwise the client will only ever see a connection timeout, which can mean a lot of things. If the server times out, the client will now receive a TLS error 80, which can then be diagnosed with the documentation from #1245. 

Contrary to the documentation, I found that `MinConnectionTimeout` is rather a `ConnectionTimeout` that will be applied to the TCP+TLS connection.

### DIFF
--- a/coordinator/internal/authority/credentials.go
+++ b/coordinator/internal/authority/credentials.go
@@ -5,6 +5,7 @@ package authority
 
 import (
 	"context"
+	"crypto/tls"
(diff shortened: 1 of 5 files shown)


3u13r (approved):
LGTM

PR #1232, opened 2025-02-13
SNP anonymous IDBlock signature on bare metal

3u13r:
Start adding an IDBlock (and IDAuthBlock) for bare-metal QEMU Contrast deployments. 
There is no verification yet as the verifier does not differentiate between qemu bare-metal and IGVM-based images and igvm idblocks are not implemented in this PR.

PR #1258, opened 2025-02-28
nodeinstaller: embed SNP IDBlock for all platforms

katexochen:
This fixed the GPU platforms, which were broken after https://github.com/edgelesssys/contrast/pull/1232 as we did use the non-gpu IDBlock on SNP GPU platforms, too.


katexochen:
gpu test https://github.com/edgelesssys/contrast/actions/runs/13583240981

PR #1261, opened 2025-02-28
attestation: forward context to KDS getter

burgerdev:
Implement `trust.ContextHTTPSGetter` for `CachedHTTPSGetter` in order to use https://github.com/google/go-sev-guest/pull/156. This is the last step towards ensuring server timeouts are smaller than client timeouts _for SNP_.


katexochen (approved):
Thanks!

PR #1270, opened 2025-03-05
enterprise: add peer discovery package

davidweisse:
This PR adds the `peerdiscovery` enterprise package, which provides a `GetPeers` function, which returns a list of IPs of ready Coordinators to be used for peer recovery.

### DIFF
--- a/enterprise/coordinator/peerdiscovery/peerdiscovery.go
+++ b/enterprise/coordinator/peerdiscovery/peerdiscovery.go
@@ -1,0 +1,47 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//go:build enterprise
+
+package peerdiscovery
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


burgerdev (changes requested):
@davidweisse pointed out to me that we actually don't need a local cache, but can fetch the list when needed from the k8s api. This was a bad suggestion from RFC 010 to begin with, I suppose. 

As discussed privately, let's have a package with a 
```go
package peerdiscovery
func GetReadyCoordinators(ctx context.Context, k *kubernetes.Interface, namespace string) ([]string, error)
```
that just does the listing and filtering.

enterprise/coordinator/peerdiscovery/peerdiscovery.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+)
+
+// GetPeers returns a list of Coordinator IPs that are ready to be used for peer recovery.
+func GetPeers(ctx context.Context, client kubernetes.Interface, namespace string) ([]string, error) {
```
burgerdev, line 20: Please add some tests.

burgerdev (approved):
Awesome, thanks!

PR #1271, opened 2025-03-05
cli: only print to stdout if it is a tty

davidweisse:
With this PR, user-friendly messages during CLI commands like `✔️  Generated workload policy annotations` are only printed if a terminal is connected. This way, the output of, e.g., `contrast generate` can be piped directly to `kubectl apply`:

`contrast generate -o /dev/stdout | kubectl apply -f -`


burgerdev (approved):
Nice, thanks!

PR #1272, opened 2025-03-05
coordinator: move meshapi server to authority; add test for NewMeshCert

burgerdev:
This PR moves the mesh API handler to the authority package and adds a test for `NewMeshCert`. Some of the test code will be factored out when we add tests for the `Recover` method from RFC 010.

### DIFF
--- a/coordinator/internal/authority/authority.go
+++ b/coordinator/internal/authority/authority.go
@@ -15,6 +15,7 @@ import (
 	"github.com/edgelesssys/contrast/coordinator/internal/seedengine"
 	"github.com/edgelesssys/contrast/internal/ca"
 	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/meshapi"
 	"github.com/edgelesssys/contrast/internal/userapi"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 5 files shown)


3u13r (approved):
Code itself LGTM, though we are moving more and more into the direction of MarbleRun's "Core" struct, holding most of the state and also implementing most of the logic. If possible I'd like to have a second look by @katexochen on this decision. Is it time for an architecture diagram?

coordinator/internal/authority/authority.go
```diff
@@ -35,6 +36,7 @@ type Authority struct {
 	metrics metrics
 
 	userapi.UnimplementedUserAPIServer
+	meshapi.UnimplementedMeshAPIServer
```
3u13r, line 39: The comments above 
> // Authority manages the manifest state of Contrast.

and 

> // state holds all required configuration to serve requests from userapi.

are outdated.

3u13r (changes requested):

katexochen:
> Code itself LGTM, though we are moving more and more into the direction of MarbleRun's "Core" struct, holding most of the state and also implementing most of the logic. If possible I'd like to have a second look by katexochen on this decision. Is it time for an architecture diagram?

I agree, I'd love to see a some more structure/concept and an architectural diagram on the Coordinator as part of RFC 010. Doesn't have to be in this PR, but it has become a challenge to understand the code base as we dropped some layers of abstraction and moved/merged parts in many PRs.

No hard opinion on the monolith, but if we go monolith, it should be a clean one. ;)

burgerdev:
Abandoning in favour of #1291.

PR #1274, opened 2025-03-06
e2e: increase timeout for scaling deployment in workloadsecrettest

burgerdev:
This test has been failing frequently on bare metal because the pod takes longer than a minute to start up, which is probably due to slow guest pulls. Increasing the bare timeout to a minute results in a 2 minute timeout after scaling.


3u13r (approved):
LGTM

PR #1282, opened 2025-03-10
manifest: join validation errors and return late

davidweisse:
The manifest validation errors are joined by `errors.Join` to return all failures during validation, not just the first. The error is then printed in a tree-like structure by the new `validationError` error type.

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -10,6 +10,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/edgelesssys/contrast/internal/constants"
 	"github.com/google/go-sev-guest/abi"
@@ -41,18 +42,18 @@ type PolicyEntry struct {
 
 // Validate checks the validity of a policy entry given its policy hash.
 func (e PolicyEntry) Validate(policyHash HexString) error {
+	var errs []error
 	if _, err := policyHash.Bytes(); err != nil {
-		return fmt.Errorf("decoding policy hash %q: %w", policyHash, err)
-	}
-	if len(policyHash) != hex.EncodedLen(sha256.Size) {
-		return fmt.Errorf("policy hash %s has invalid length: %d (expected %d)", policyHash, len(policyHash), hex.EncodedLen(sha256.Size))
+		errs = append(errs, fmt.Errorf("decoding policy hash %q: %w", policyHash, err))
+	} else if len(policyHash) != hex.EncodedLen(sha256.Size) {
+		errs = append(errs, fmt.Errorf("invalid policy hash length: %d (expected %d)", len(policyHash), hex.EncodedLen(sha256.Size)))
 	}
 
 	if err := e.Role.Validate(); err != nil {
-		return fmt.Errorf("validating role %s: %w", e.Role, err)
+		errs = append(errs, newValidationError("Role", err))
 	}
 
-	return nil
+	return errors.Join(errs...)
 }
 
 // HexStrings is a slice of HexString.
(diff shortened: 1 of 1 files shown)


internal/manifest/manifest.go
```diff
@@ -140,87 +142,139 @@ func validateHexString(value HexString, expectedNumBytes int) error {
...
+}
+
+// validationError wraps an error with a prefix message.
+//
+// The error will be printed indented below the message.
+// Errors joined by errors.Join will be printed on separate lines with the same indentation.
+type validationError struct {
+	msg string
+	err error
+}
```
katexochen, line 241: While I like the idea of giving some positional information for the user, I don't quite think this is a good implementation. The location information that is passed in this pr is imprecise and doesn't add much to the original error. Something like a full json path or line/col would be more helpful (but also more complex). Maybe the default formatting of Go is good for now.

internal/manifest/manifest.go
```diff
@@ -140,87 +142,139 @@ func validateHexString(value HexString, expectedNumBytes int) error {
...
+
+func newValidationError(msg string, err error) *validationError {
+	return &validationError{msg: msg, err: err}
+}
+
+func (e *validationError) Error() string {
+	return e.formatError(0)
+}
+
+func (e *validationError) formatError(indent int) string {
```
katexochen, line 251: Is this ever called with indent != 0?

internal/manifest/manifest.go
```diff
@@ -112,22 +113,23 @@ func (r Role) Validate() error {
...
 		}
 	}
 
 	if len(r.SNP)+len(r.TDX) == 0 {
-		return fmt.Errorf("reference values in manifest cannot be empty. Is the chosen platform supported?")
+		errs = append(errs, fmt.Errorf("reference values in manifest cannot be empty. Is the chosen platform supported?"))
 	}
 
-	return nil
+	return errors.Join(errs...)
```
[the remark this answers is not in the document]
davidweisse: `errors.Join` always wraps the given errors in a new error, so calling `Unwrap()` would reveal only the last error.

internal/manifest/manifest.go
davidweisse: This is recursively called for all validation errors if they are nested, e.g., when validating the reference values.

internal/manifest/manifest.go
davidweisse: My concern was that the default formatting might not be very readable. If we want to change the output later to something more structured, that's fine by me, but I'm not sure how difficult that will be.

internal/manifest/manifest.go
```diff
@@ -140,87 +142,139 @@ func validateHexString(value HexString, expectedNumBytes int) error {
...
+	var sb strings.Builder
+	sb.WriteString(prefix + e.msg + ":")
+
+	if e.err == nil {
+		return sb.String()
+	}
+
+	var errs []error
+	// Interface is implemented by errors returned by errors.Join.
+	if unwrapErr, ok := e.err.(interface{ Unwrap() []error }); ok {
```
burgerdev, line 262: There's also `Unwrap() error` which we'd need to check, see https://pkg.go.dev/errors#pkg-overview.

internal/manifest/manifest.go
```diff
@@ -140,87 +142,139 @@ func validateHexString(value HexString, expectedNumBytes int) error {
...
+}
+
+func (e *validationError) Error() string {
+	return e.formatError(0)
+}
+
+func (e *validationError) formatError(indent int) string {
+	prefix := strings.Repeat("  ", indent)
+	var sb strings.Builder
+	sb.WriteString(prefix + e.msg + ":")
```
burgerdev, line 254:
```suggestion
	var sb strings.Builder
	sb.WriteString(strings.Repeat("  ", indent))
	sb.WriteString(e.msg)
	sb.WriteString(":")
```

No strong feelings about this, but since we're using `strings.Builder` anyway we might as well get rid of the `+` concat.

internal/manifest/manifest.go
burgerdev: Line/col will be hard to implement using the standard library. For JSONPath, however, we can attach the current context to the `validationError` and then build a path from it while traversing the tree.

internal/manifest/manifest.go
```diff
@@ -140,87 +142,146 @@ func validateHexString(value HexString, expectedNumBytes int) error {
...
+
+func (e *validationError) Error() string {
+	return e.formatError(e.path)
+}
+
+func (e *validationError) formatError(path string) string {
+	var sb strings.Builder
+	for i, err := range e.errs {
+		var ve *validationError
+		if errors.As(err, &ve) {
```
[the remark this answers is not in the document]
davidweisse: We can flatten out any joined errors in `newValidationError()`. This also solves the indexing problem above, as we can now return joined errors without a JSONPath and can just call `newValidationError(fmt.Sprintf("Policies[%q]", policyHash), err)`. I don't think an intermediate wrapping error is intended and am not sure how that would look like. The only time a validation error is wrapped in a normal error would be in the initial call to `manifest.Validate()`, in which case the validation error will be printed last.

internal/manifest/manifest.go
```diff
@@ -140,87 +142,157 @@ func validateHexString(value HexString, expectedNumBytes int) error {
...
+		return []error{ve}
+	}
+	if wrapped, ok := err.(interface{ Unwrap() []error }); ok {
+		for _, err := range wrapped.Unwrap() {
+			errs = append(errs, flatten(err)...)
 		}
+		return errs
 	}
-	return nil
+	return []error{err}
```
burgerdev, line 295: For completeness you could cast into `Unwrap() error` and recurse into that one, too.

burgerdev (approved):

internal/manifest/manifest.go
burgerdev: As discussed, not needed.

burgerdev (approved):

PR #1273, opened 2025-03-06
generate: allow envFromRef to reference other files

3u13r:
Upstream work: https://github.com/kata-containers/kata-containers/pull/10986

PR #1288, opened 2025-03-12
generate: allow generating with a one ConfigMap as envFromRef dependency

3u13r:
This is the maximum implementation that we can implement with the intersection of the features of the microsoft genpolicy and the upstream genpolicy.

The full envFromRef feature for secrets and multiple dependency files will be implemented in https://github.com/edgelesssys/contrast/pull/1273.

This PR will be tested with the following regression test: https://github.com/edgelesssys/contrast/pull/1287

### DIFF
--- a/cli/genpolicy/genpolicy.go
+++ b/cli/genpolicy/genpolicy.go
@@ -51,7 +51,7 @@ func New(rulesPath, settingsPath, cachePath string, bin []byte) (*Runner, error)
 // Run runs the tool on the given yaml.
 //
 // Run can be called more than once.
-func (r *Runner) Run(ctx context.Context, yamlPath string, logger *slog.Logger) error {
+func (r *Runner) Run(ctx context.Context, yamlPath string, cmPath []string, logger *slog.Logger) error {
 	args := []string{
 		"--runtime-class-names=contrast-cc",
 		"--rego-rules-path=" + r.rulesPath,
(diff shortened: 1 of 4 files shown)


katexochen:
Please add a leading paragraph to the PR description for users/changelog.

katexochen:
> Opinion needed: is this a feature or a bugfix in cases like this where it is a reasonable assumption that this should work out of the box?

category feature or changelog are ok here, I wouldn't label it as bug fix.

katexochen:
PR title could be more meaningful for users.

cli/genpolicy/genpolicy.go
```diff
@@ -51,14 +51,22 @@ func New(rulesPath, settingsPath, cachePath string, bin []byte) (*Runner, error)
...
 		"--layers-cache-file-path=" + r.cachePath,
 		"--yaml-file=" + yamlPath,
 	}
+
+	if len(cmPath) > 0 {
+		// "-c" is shared between the kata and microsoft genpolicy for different long flags
+		args = append(args, "-c="+cmPath[0])
+	} else if len(cmPath) > 1 {
+		return fmt.Errorf("multiple files with ConfigMaps found, only one is allowed")
+	}
```
katexochen, line 68: This is a limitation that needs to be documented somewhere.

burgerdev: This is actually a regression that I'd like to avoid. Maybe we could join the yamls into a temp file?

cli/genpolicy/genpolicy.go
```diff
@@ -51,14 +51,22 @@ func New(rulesPath, settingsPath, cachePath string, bin []byte) (*Runner, error)
...
 		"--rego-rules-path=" + r.rulesPath,
 		"--json-settings-path=" + r.settingsPath,
 		"--layers-cache-file-path=" + r.cachePath,
 		"--yaml-file=" + yamlPath,
 	}
+
+	if len(cmPath) > 0 {
+		// "-c" is shared between the kata and microsoft genpolicy for different long flags
+		args = append(args, "-c="+cmPath[0])
+	} else if len(cmPath) > 1 {
```
burgerdev, line 66: This branch can't be hit, can it?

cli/genpolicy/genpolicy.go
3u13r: Actually, It's even worse since the current PR already breaks of the file in cmPath contains more than one YAML document (or that one document isn't a ConfigMap) since the upstream genpolicy implementation requires this: https://github.com/kata-containers/kata-containers/blob/fdc0d81198ad556d019e8a9dbb6b2aeb5ecf8fef/src/tools/genpolicy/src/policy.rs#L458

3u13r:
It doesn't make sense to implement this without our upstream changes due to https://github.com/edgelesssys/contrast/pull/1288#discussion_r1991238691. Will be fully implemented in https://github.com/edgelesssys/contrast/pull/1287. Closing this PR.

PR #1291, opened 2025-03-12
coordinator: move meshapi into its own package

burgerdev:
This is an alternative approach to #1272. Both in its current form as well as with the additions from RFC 010, the meshapi only depends on the state of the coordinator as per creation of the channel. We can thus decouple the authority struct and the meshapi server, the remaining interface being the state passed via context.

Before moving the service implementation into its own package, I consolidated the gRPC server creation, which was mostly duplicated code and used to expose the meshapi server to structs it should not concern itself with.

~I will port the test from #1272 in a separate PR.~

### DIFF
--- a/coordinator/internal/meshapi/meshapi_test.go
+++ b/coordinator/internal/meshapi/meshapi_test.go
@@ -1,0 +1,122 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package meshapi
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"crypto/sha256"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"log/slog"
+	"testing"
+
+	"github.com/edgelesssys/contrast/coordinator/internal/authority"
+	"github.com/edgelesssys/contrast/coordinator/internal/seedengine"
+	"github.com/edgelesssys/contrast/internal/ca"
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/testkeys"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/peer"
+)
+
+func TestNewMeshCert(t *testing.T) {
+	require := require.New(t)
+	assert := assert.New(t)
+
+	m := &manifest.Manifest{}
+	policyHash := sha256.Sum256(nil)
+	policyHashHex := manifest.NewHexString(policyHash[:])
+	m.Policies = map[manifest.HexString]manifest.PolicyEntry{
+		policyHashHex: {
+			SANs:             []string{"test"},
+			WorkloadSecretID: "test",
+		},
+	}
+	key := testkeys.New[ecdsa.PrivateKey](t, testkeys.ECDSAP384Keys[0])
+	meshKey := testkeys.New[ecdsa.PrivateKey](t, testkeys.ECDSAP384Keys[1])
+	rootKey := testkeys.New[ecdsa.PrivateKey](t, testkeys.ECDSAP384Keys[2])
+
+	seed := [32]byte{}
+	salt := [32]byte{}
+	se, err := seedengine.New(seed[:], salt[:])
+	require.NoError(err)
+	ca, err := ca.New(rootKey, meshKey)
+	require.NoError(err)
+
+	info := authority.AuthInfo{
+		TLSInfo: credentials.TLSInfo{
+			State: tls.ConnectionState{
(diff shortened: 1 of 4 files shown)


coordinator/internal/meshapi/meshapi_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	require.NotEmpty(t, resp.WorkloadSecret)
+
+	certChain := certFromPEM(t, resp.CertChain)
+	cert, intermediateCert := certChain[0], certChain[1]
+	assert.False(t, cert.IsCA)
+	assert.True(t, intermediateCert.IsCA)
+	assert.Equal(t, cert.AuthorityKeyId, intermediateCert.SubjectKeyId)
+
+	meshCert := certFromPEM(t, resp.MeshCACert)[0]
```
3u13r, line 82: We may also want to require the length here.

coordinator/internal/meshapi/meshapi_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (r *fakeReport) HostData() []byte {
+	return r.hostData
+}
+
+func manifestWithWorkloadOwnerKey(t *testing.T, key *ecdsa.PrivateKey) *manifest.Manifest {
+	t.Helper()
+	require := require.New(t)
+
+	m, err := manifest.Default(platforms.AKSCloudHypervisorSNP)
```
3u13r, line 110: Do we even need a default manifest here, especially since we use a constant platform here? Do we even need the workload owner key in the manifest? As far as I can see `NewMeshCert()` only uses `state.Manifest.Policies[hostData]`

coordinator/internal/meshapi/meshapi_test.go
```diff
@@ -0,0 +1,134 @@
...
+		},
+	}
+	key := testkeys.New[ecdsa.PrivateKey](t, testkeys.ECDSAP384Keys[0])
+	meshKey := testkeys.New[ecdsa.PrivateKey](t, testkeys.ECDSAP384Keys[1])
+	rootKey := testkeys.New[ecdsa.PrivateKey](t, testkeys.ECDSAP384Keys[2])
+
+	seed := [32]byte{}
+	salt := [32]byte{}
+	se, err := seedengine.New(seed[:], salt[:])
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
burgerdev: Done

coordinator/internal/meshapi/meshapi_test.go
```diff
@@ -0,0 +1,134 @@
...
+			Manifest:   m,
+			SeedEngine: se,
+			CA:         ca,
+		},
+	}
+	ctx := peer.NewContext(context.Background(), &peer.Peer{
+		AuthInfo: info,
+	})
+
+	meshapi := New(slog.Default())
```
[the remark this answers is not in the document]
burgerdev: This is what we're doing everywhere in the tests, and while I agree that it could be done better, discarding means that we're losing valuable debug information. How about we fix this globally in a separate PR?

coordinator/internal/meshapi/meshapi_test.go
burgerdev: I copied this over from `userapi_test.go` without thinking too much. You're right that the workload owner key is not used so I'm inlining the `manifest.Default` as suggested.

coordinator/internal/meshapi/meshapi_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+func manifestWithWorkloadOwnerKey(t *testing.T, key *ecdsa.PrivateKey) *manifest.Manifest {
+	t.Helper()
+	require := require.New(t)
+
+	m, err := manifest.Default(platforms.AKSCloudHypervisorSNP)
+	require.NoError(err)
+	if key == nil {
+		return m
+	}
```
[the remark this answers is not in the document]
burgerdev: Done

coordinator/internal/meshapi/meshapi_test.go
burgerdev: Done

burgerdev (commented):
Thanks for the review!

coordinator/internal/meshapi/meshapi_test.go
3u13r: My idea was that we might even not call into `Default`, also because this means that the tests are not executable with plain Go:
```
euler@work:~/projects/coco/coordinator/internal/meshapi$ go test ./...
--- FAIL: TestNewMeshCert (0.00s)
panic: failed to unmarshal embedded reference values: json: cannot unmarshal string into Go value of type manifest.EmbeddedReferenceValues [recovered]
        panic: failed to unmarshal embedded reference values: json: cannot unmarshal string into Go value of type manifest.EmbeddedReferenceValues
```

burgerdev: Oh, that's even nicer - thanks; done!

PR #1301, opened 2025-03-13
atls: encode nonce in ALPN protos

burgerdev:
Encoding aTLS nonces in SNI is problematic when the traffic is routed or inspected by (reverse) proxies. We can add the nonce to the list of ALPN next protocols instead, which suffices at least for the use case of routing by SNI. 

Using ALPN has some more advantages:

1. The specification allows for about 2^16 byte of data in the next protocol list (256 entries with 256B each).  SNI allows for one hostname of 256B per type (and there is only one type, afaik).
2. ALPN protocols are opaque strings (could even be byte slices, actually), while SNI is expected to be a valid domain name.
3. We could transfer even more information to the server using ALPN, for example a range of handshake versions and variants (e.g. TPM), or additional metadata. 

The proposed format for the nonce protocol string is

```abnf
protocol = "atls" ; identifies our protocol
version = "v1"
component = "nonce" ; additional input for the handshake could come as a separate component
encoded-nonce = 64hex-char
next-protocol-string = protocol ":" version ":" component ":" encoded-nonce
; Example:
;   atls:v1:nonce:02f2f9a189459c46c3eb8a40683ca4a07bbe05fc82a18cf023025481de178ab5
```

We should keep a fallback to reading the nonce from SNI, in case an older client connects to the server.

### DIFF
--- a/internal/atls/atls.go
+++ b/internal/atls/atls.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"strings"
 	"time"
 
 	"github.com/edgelesssys/contrast/internal/attestation"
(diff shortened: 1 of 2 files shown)


thomasten (approved):
Thanks!

burgerdev:
I tested manually that a 1.6.0 client receives the expected REPORTDATA when connecting to a Coordinator from this branch.

PR #1302, opened 2025-03-13
transitengineapi: tls authentication/authorization + example Vault deployment

jmxnzo:
This PR introduces mTLS to the transit engine API to secure connections based on Contrast mesh certs. Further it adds an example OpenBao Vault deployment, which utilizes our built-in transit engine API of the coordinator to unseal a user-managed Vault secured by the mTLS connection.

- The client has to authenticate with a valid mesh cert of the current state.
- MeshAPI: A certificate extension holding the workloadSecretID of the requesting client, which is looked up from the manifest while issuing, is added to every meshCert now. Thus we have no need for an additional manifest look-up.
- The client is only authorized to the transit key set of the workloadSecretID in the received meshCert extension. As prior we handle the `<name>` parameter of existing Vault implementations as our workloadSecretID. 
- On initialization the transit engine API derives a private ecdsa key. For each incoming TLS connection of the transit engine API, the coordinator issues itself a fresh mesh cert for this private key of the transit engine API, which is therefore valid for the current state.

### DIFF
--- a/coordinator/internal/transitengineapi/transitengineapi.go
+++ b/coordinator/internal/transitengineapi/transitengineapi.go
@@ -7,10 +7,14 @@
 package transitengine
 
 import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/asn1"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"log/slog"
 	"net/http"
@@ -56,7 +60,11 @@ type stateAuthority interface {
 }
 
 // NewTransitEngineAPI sets up the transit engine API with a provided seedEngineAuthority.
-func NewTransitEngineAPI(authority stateAuthority, port int, logger *slog.Logger) *http.Server {
+func NewTransitEngineAPI(authority stateAuthority, port int, logger *slog.Logger) (*http.Server, error) {
+	privKeyAPI, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return nil, fmt.Errorf("failed creating transit engine API private key")
+	}
 	return &http.Server{
 		Addr: fmt.Sprintf(":%d", port),
 		TLSConfig: &tls.Config{
@@ -80,11 +88,14 @@ func NewTransitEngineAPI(authority stateAuthority, port int, logger *slog.Logger
 					ClientCAs:  meshCAPool,
 					ClientAuth: tls.RequireAndVerifyClientCert,
 					MinVersion: tls.VersionTLS12,
+					GetCertificate: func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+						return getCertificate(privKeyAPI, authority)
+					},
 				}, nil
 			},
 		},
 		Handler: newTransitEngineMux(authority, logger),
-	}
+	}, nil
 }
 
 func newTransitEngineMux(authority stateAuthority, logger *slog.Logger) *http.ServeMux {
@@ -294,4 +305,30 @@ func authorizationMiddleware(next http.HandlerFunc) http.HandlerFunc {
 		}
 		next.ServeHTTP(w, r)
 	})
+}
+
+func getCertificate(privKeyAPI *ecdsa.PrivateKey, authority stateAuthority) (*tls.Certificate, error) {
+	state, err := authority.GetState()
+	if err != nil {
+		return nil, fmt.Errorf("getting state: %w", err)
+	}
+	dnsNames := []string{"coordinator"}
+
+	meshCertPEM, err := state.CA.NewAttestedMeshCert(dnsNames, nil, &privKeyAPI.PublicKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create mesh cert: %w", err)
+	}
+	meshCertDER, _ := pem.Decode(meshCertPEM)
+	if meshCertDER == nil {
+		return nil, fmt.Errorf("failed to decode mesh cert: %w", err)
+	}
+	intermCertDER, _ := pem.Decode(state.CA.GetIntermCACert())
+	if intermCertDER == nil {
+		return nil, fmt.Errorf("failed to decode intermediate cert: %w", err)
+	}
+	certChain := tls.Certificate{
+		Certificate: [][]byte{meshCertDER.Bytes, intermCertDER.Bytes},
+		PrivateKey:  privKeyAPI,
+	}
+	return &certChain, nil
 }
--- a/coordinator/internal/transitengineapi/transitengineapi.go
+++ b/coordinator/internal/transitengineapi/transitengineapi.go
@@ -79,6 +79,9 @@ func NewTransitEngineAPI(authority stateAuthority, port int, logger *slog.Logger
 					ClientCAs:  meshCAPool,
 					ClientAuth: tls.RequireAndVerifyClientCert,
 					MinVersion: tls.VersionTLS12,
+					GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+						return nil, nil
+					},
 				}, nil
 			},
 		},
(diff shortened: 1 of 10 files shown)


coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -192,3 +191,41 @@ func extractVersion(versionStr string) (uint32, error) {
...
+func (rl *responseLogger) Write(b []byte) (int, error) {
+	// Capture the response body only if status code is an error (≥400)
+	if rl.statusCode >= 400 && !rl.bodyCaptured {
+		rl.body = append([]byte{}, b...)
+		rl.bodyCaptured = true
+	}
+	return rl.ResponseWriter.Write(b)
+}
+
+func loggingMiddleware(next http.HandlerFunc, logger *slog.Logger) http.HandlerFunc {
```
burgerdev, line 293: 08a38247cfd1954054ac1d6e1ec698ba60e5d063 should be in a separate PR.

coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -51,6 +53,39 @@ type stateAuthority interface {
 	GetState() (*authority.State, error)
 }
 
+// NewTransitEngineAPI sets up the transit engine API with a provided seedEngineAuthority.
+func NewTransitEngineAPI(authority stateAuthority, port int, logger *slog.Logger) *http.Server {
+	return &http.Server{
+		Addr: fmt.Sprintf(":%d", port),
```
burgerdev, line 69: Can we set this field in `main.go` instead?

coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -51,6 +53,39 @@ type stateAuthority interface {
...
+					return nil, fmt.Errorf("getting state: %w", err)
+				}
+				if len(state.CA.GetMeshCACert()) == 0 {
+					return nil, fmt.Errorf("mesh ca cert not initialized")
+				}
+				meshCAPool := x509.NewCertPool()
+				if !meshCAPool.AppendCertsFromPEM(state.CA.GetMeshCACert()) {
+					logger.Debug("failed parsing mesh CA")
+					return nil, fmt.Errorf("failed to parse mesh CA cert")
+				}
```
burgerdev: We could add a function `GetMeshCAPool` to the `CA` and prepare the pool in the CA constructor, in order to not need to parse certs on every connection.

coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -51,6 +53,39 @@ type stateAuthority interface {
...
+			ClientAuth: tls.RequireAndVerifyClientCert,
+			GetConfigForClient: func(_ *tls.ClientHelloInfo) (*tls.Config, error) {
+				logger.Debug("call getConfigForClient")
+				state, err := authority.GetState()
+				if err != nil {
+					return nil, fmt.Errorf("getting state: %w", err)
+				}
+				if len(state.CA.GetMeshCACert()) == 0 {
+					return nil, fmt.Errorf("mesh ca cert not initialized")
+				}
```
burgerdev, line 81: Can this even happen? I'd say that an initialized CA should always have a mesh cert.

coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -79,6 +79,9 @@ func NewTransitEngineAPI(authority stateAuthority, port int, logger *slog.Logger
 					ClientCAs:  meshCAPool,
 					ClientAuth: tls.RequireAndVerifyClientCert,
 					MinVersion: tls.VersionTLS12,
+					GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+						return nil, nil
+					},
```
burgerdev: The `GetCertificate` addition in 93a91529f65dd2854470f4cf8fa8df215bed92bb does not make sense.

coordinator/main.go
```diff
@@ -135,6 +136,17 @@ func run() (retErr error) {
 		return nil
 	})
 
+	eg.Go(func() error {
+		transitAPI := transitengine.NewTransitEngineAPI(meshAuth, 8200, logger)
```
burgerdev: Needs to be added to graceful shutdown routine below.

internal/kuberesource/parts.go
```diff
@@ -380,6 +380,9 @@ func Coordinator(namespace string) *CoordinatorConfig {
 								ContainerPort().
 									WithName("meshapi").
 									WithContainerPort(7777),
+								ContainerPort().
+									WithName("transitapi").
+									WithContainerPort(8200),
```
burgerdev, line 385: Needs to be added to the service, too.

internal/kuberesource/sets.go
```diff
@@ -698,3 +698,101 @@ func GPU() []any {
...
+# Wait until Vault is ready
+echo "Waiting for Vault to become available..."
+until wget --server-response --no-check-certificate --spider -q -O- http://127.0.0.1:8200 2>&1 | grep -q "HTTP/1.1 200"; do
+    echo "Waiting for Vault to become available..."
+    sleep 2
+done
+
+echo "Vault is now ready. Initializing..."
+bao operator init
+sleep inf`
```
burgerdev, line 713: This should be `wait`, so that we correctly terminate when bao terminates.

internal/kuberesource/sets.go
```diff
@@ -698,3 +698,101 @@ func GPU() []any {
...
+storage "file" {
+        path = "/vault/data"
+}
+
+listener "tcp" {
+  address            = "0.0.0.0:8200"
+  tls_cert_file      = "/contrast/tls-config/certChain.pem"
+  tls_key_file       = "/contrast/tls-config/key.pem"
+  tls_client_ca_file = "/contrast/tls-config/coordinator-root-ca.pem"
+  tls_disable        = "true"
```
burgerdev: That's odd. If you meant to disable, why are you configuring certs?

internal/kuberesource/sets.go
```diff
@@ -698,3 +698,101 @@ func GPU() []any {
...
+seal "transit" {
+  address         = "https://coordinator:8200"
+  disable_renewal = "true"
+  key_name        = "transit_key_name"
+  mount_path      = "transit/"
+  tls_client_cert    = "/contrast/tls-config/certChain.pem"
+  tls_client_key     = "/contrast/tls-config/key.pem"
+
+  tls_ca_cert        = "/contrast/tls-config/coordinator-root-ca.pem"
+  tls_skip_verify    = "true"
```
burgerdev: inconsistent alignment

internal/kuberesource/sets.go
```diff
@@ -698,3 +698,101 @@ func GPU() []any {
...
+
+storage "file" {
+        path = "/vault/data"
+}
+
+listener "tcp" {
+  address            = "0.0.0.0:8200"
+  tls_cert_file      = "/contrast/tls-config/certChain.pem"
+  tls_key_file       = "/contrast/tls-config/key.pem"
+  tls_client_ca_file = "/contrast/tls-config/coordinator-root-ca.pem"
```
burgerdev: Is there a reason for using the root CA here? I'd find the mesh CA more natural.

internal/oid/oid.go
```diff
@@ -12,3 +12,8 @@ var RawSNPReport = asn1.ObjectIdentifier{1, 3, 9901, 2, 1}
 // RawTDXReport is the root OID for the raw TDX report extensions
 // used by the aTLS issuer and validator.
 var RawTDXReport = asn1.ObjectIdentifier{1, 3, 9901, 2, 2}
+
+// WorkloadSecretOID is the root OID for the workloadSecretID report
+// extension, added to the mesh certificates to allow verification
+// and authorization based on the workloadSecretID.
+var WorkloadSecretOID = asn1.ObjectIdentifier{1, 3, 9901, 2, 3}
```
burgerdev, line 19: `1.3.9901` is the "as you like it" prefix, right? Should we take a different next number to distinguish report extensions from claim extensions? Don't we already have an extension number for claims somewhere?

burgerdev (commented):
Sorry, need to interrupt review at afacb4c55dc37a3ecc2d1f6dea4c3362fbbbb5fe. Please split the PR into (at least)

* logging middleware
* workload secret cert extension
* authorization
* http server start in main
* demo

This will make review easier, both considering volume and iterations.

PR #1309, opened 2025-03-18
history: allow watching for LatestTransition changes

burgerdev:
Propagate watcher events from the low-level store to the history, where they can be used by `authority.Authority` to enter recovery mode.

### DIFF
--- a/coordinator/history/history.go
+++ b/coordinator/history/history.go
@@ -24,6 +24,7 @@ const (
 type History struct {
 	store   Store
 	hashFun func() hash.Hash
+	log     *slog.Logger
 }
 
 // New creates a new History backed by the configured store.
@@ -32,14 +33,15 @@ func New(log *slog.Logger) (*History, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating history store: %w", err)
 	}
-	return NewWithStore(store), nil
+	return NewWithStore(log, store), nil
 }
 
 // NewWithStore creates a new History with the given storage backend.
-func NewWithStore(store Store) *History {
+func NewWithStore(log *slog.Logger, store Store) *History {
 	h := &History{
 		store:   store,
 		hashFun: sha256.New,
+		log:     log,
 	}
 	if HashSize != h.hashFun().Size() {
 		panic("mismatch between hashSize and hash function size")
--- a/coordinator/history/history.go
+++ b/coordinator/history/history.go
@@ -5,6 +5,7 @@ package history
 
 import (
 	"bytes"
+	"context"
 	"crypto/ecdsa"
 	"crypto/rand"
 	"crypto/sha256"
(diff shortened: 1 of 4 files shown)


coordinator/history/history.go
```diff
@@ -118,6 +119,43 @@ func (h *History) SetLatest(oldT, newT *LatestTransition, signingKey *ecdsa.Priv
...
+	ctx, cancel := context.WithCancel(context.Background())
+	transitionCh := make(chan LatestTransition)
+	go func() {
+		defer cancelWatch()
+		for {
+			select {
+			case buf, ok := <-ch:
+				if !ok {
+					// TODO(burgerdev): should be logged
+					close(transitionCh)
```
davidweisse, line 141: This can be deferred at the beginning of the goroutine.

coordinator/history/history.go
```diff
@@ -118,6 +119,43 @@ func (h *History) SetLatest(oldT, newT *LatestTransition, signingKey *ecdsa.Priv
 	return nil
 }
 
+// WatchLatestTransitions starts a goroutine that sends LatestTransition structs to the returned
+// channel whenever the latest transition changes in the underlying store.
+//
+// The goroutine should be stopped with the returned cancel function.
+func (h *History) WatchLatestTransitions() (<-chan LatestTransition, func(), error) {
```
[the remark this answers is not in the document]
burgerdev: Funny that you mention it: that's what I implemented first, but then I adopted the signature of `Store.Watcher` for consistency. I think we should pick one flavour and stick to it. But the more I think about it the more I'm leaning towards context.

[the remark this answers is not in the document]
burgerdev: Reworked this function to take a context, leaving aside the store watcher for later.

coordinator/history/history.go
burgerdev: Done

coordinator/history/history.go
```diff
@@ -118,6 +121,46 @@ func (h *History) SetLatest(oldT, newT *LatestTransition, signingKey *ecdsa.Priv
...
+					h.log.Warn("store watcher closed unexpectedly")
+					return
+				}
+				var t LatestTransition
+				if err := t.unmarshalBinary(buf); err != nil {
+					h.log.Error("store watcher sent something that's not a LatestTransition", "error", err)
+					continue
+				}
+				select {
+				case transitionCh <- t:
```
msanft, line 152: Is this a fallthrough in this case? If so, why not use the comma syntax? Or does it not support the redirection here?

msanft (commented):
Generally LGTM

coordinator/history/history.go
burgerdev: There's neither fallthrough nor expression list in select statements:

* https://go.dev/ref/spec#Select_statements
* https://go.dev/ref/spec#Expression_switches

davidweisse (approved):

PR #1310, opened 2025-03-19
meshapi: add recover handler

davidweisse:
The recover handler verifies the peer Coordinator's policy and role and sends a `RecoverResponse` struct so the calling Coordinator can reconstruct its state.

The initial proto definition of the `RecoverResponse` defined in [RFC 10](https://github.com/edgelesssys/contrast/blob/main/rfc/010-distributed-coordinator.md#peer-recovery) is reduced to:
```proto
message RecoverResponse {
  bytes Seed = 1;
  bytes Salt = 2;
  bytes MeshCAKey = 3;
  bytes LatestManifest = 4;
}
```
The root CA key can already be derived by the seed and salt. The certificates can be generated afterward (this happens automatically when calling the constructor of the `CA` with given root CA and mesh CA keys).

### DIFF
--- a/coordinator/internal/meshapi/meshapi.go
+++ b/coordinator/internal/meshapi/meshapi.go
@@ -5,7 +5,9 @@ package meshapi
 
 import (
 	"context"
+	"crypto/ecdsa"
 	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"log/slog"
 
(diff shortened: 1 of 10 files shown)


coordinator/internal/meshapi/meshapi.go
```diff
@@ -100,3 +102,67 @@ func (i *Server) NewMeshCert(ctx context.Context, _ *meshapi.NewMeshCertRequest)
 
 	return resp, nil
 }
+
+// Recover verifies the peers policy and role and recovers the calling Coordinator.
```
burgerdev, line 106: It's a bit confusing that there are two subjects here ("peer", "calling coordinator"). Maybe rewrite along the lines of 

> Recover provides key material to authenticated workloads with the Coordinator role.

coordinator/internal/meshapi/meshapi.go
```diff
@@ -100,3 +102,67 @@ func (i *Server) NewMeshCert(ctx context.Context, _ *meshapi.NewMeshCertRequest)
...
+	state := authInfo.State
+	report := authInfo.Report
+
+	hostData := manifest.NewHexString(report.HostData())
+	entry, ok := state.Manifest().Policies[hostData]
+	if !ok {
+		return nil, status.Errorf(codes.PermissionDenied, "policy hash %s not found in manifest", hostData)
+	}
+	if entry.Role != manifest.RoleCoordinator {
+		return nil, status.Errorf(codes.PermissionDenied, "role %s not allowed to recover", entry.Role)
```
burgerdev, line 128:
```suggestion
		return nil, status.Errorf(codes.PermissionDenied, "role %q not allowed to recover", entry.Role)
```
Role may be empty, so quoting it may help readability.

coordinator/internal/meshapi/meshapi.go
```diff
@@ -100,3 +102,67 @@ func (i *Server) NewMeshCert(ctx context.Context, _ *meshapi.NewMeshCertRequest)
...
+	se := state.SeedEngine()
+
+	rootCAPrivKeyDER, err := x509.MarshalECPrivateKey(se.RootCAKey())
+	if err != nil {
+		return nil, fmt.Errorf("marshaling root CA private key: %w", err)
+	}
+	rootCAPrivKeyPEM := pem.EncodeToMemory(&pem.Block{
+		Type:  "EC PRIVATE KEY",
+		Bytes: rootCAPrivKeyDER,
+	})
```
burgerdev, line 141: How about a helper function for the encoding?

coordinator/internal/meshapi/meshapi_test.go
```diff
@@ -91,6 +92,63 @@ func TestNewMeshCert(t *testing.T) {
 	assert.Equal(intermediateCert.AuthorityKeyId, rootCerts[0].SubjectKeyId)
 }
 
+func TestRecover(t *testing.T) {
```
burgerdev, line 95: We need tests for

* unknown policy hash
* policy hash not a coordinator

burgerdev (approved):

internal/meshapi/meshapi.proto
```diff
@@ -23,3 +24,15 @@ message NewMeshCertResponse {
...
+
+message RecoverRequest {}
+
+message RecoverResponse {
+  bytes Seed = 1;
+  bytes Salt = 2;
+  bytes RootCAKey = 3;
+  bytes RootCACert = 4;
+  bytes MeshCAKey = 5;
+  bytes MeshCACert = 6;
```
davidweisse, line 36: Do we really need all of these? The root CA key can be derived from the salt and seed and given the keys we can generate new certificates. So we would only need the mesh CA key, which is not derived from the seed.

burgerdev (approved):

PR #1311, opened 2025-03-19
retry: add a few helpers

burgerdev:
I'm adding these to allow for easier use of the retrier. First user of this API is #1313.

### DIFF
--- a/internal/retry/retry.go
+++ b/internal/retry/retry.go
@@ -90,3 +90,8 @@ func (f DoerFunc) Do(ctx context.Context) error {
 
 // Ensure DoerFunc implements Doer.
 var _ = Doer(DoerFunc(func(context.Context) error { return nil }))
+
+// Always can be used as an argument to NewIntervalRetrier to always retry.
+func Always(error) bool {
+	return true
+}
--- a/internal/retry/retry.go
+++ b/internal/retry/retry.go
@@ -79,3 +79,14 @@ type Doer interface {
 	// It should return an error that can be checked for retriability.
 	Do(ctx context.Context) error
 }
+
+// DoerFunc is a convenience implementation of Doer for a closure.
+type DoerFunc func(context.Context) error
+
+// Do implements Doer.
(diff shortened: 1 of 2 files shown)


internal/retry/retry.go
```diff
@@ -79,3 +79,19 @@ type Doer interface {
 	// It should return an error that can be checked for retriability.
 	Do(ctx context.Context) error
 }
+
+// DoerFunc is a convenience implementation of Doer for a closure.
+type DoerFunc func(context.Context) error
+
+// Do implements Doer.
```
msanft, line 86: The average comment if your linters require public APIs to be commented.
Makes you question their usefulness. :D

internal/retry/retry.go
```diff
@@ -79,3 +79,19 @@ type Doer interface {
...
+	return f(ctx)
+}
+
+// Ensure DoerFunc implements Doer.
+var _ = Doer(DoerFunc(func(context.Context) error { return nil }))
+
+// Always can be used as an argument to NewIntervalRetrier to always retry.
+func Always(_ error) bool {
+	return true
+}
```
msanft, line 97:
```suggestion
func Always(error) bool { return true }
```

nit: Align this with the inlining of the function above.

msanft (approved):

internal/retry/retry.go
```diff
@@ -79,3 +79,19 @@ type Doer interface {
...
+// Do implements Doer.
+func (f DoerFunc) Do(ctx context.Context) error {
+	return f(ctx)
+}
+
+// Ensure DoerFunc implements Doer.
+var _ = Doer(DoerFunc(func(context.Context) error { return nil }))
+
+// Always can be used as an argument to NewIntervalRetrier to always retry.
+func Always(_ error) bool {
```
msanft, line 95:
```suggestion
func Always(error) bool {
```
nit

msanft (approved):

PR #1312, opened 2025-03-19
e2e: increase timeout for rogue pod test

burgerdev:
We're already spending 20s idle [waiting for readiness checks](https://github.com/edgelesssys/contrast/blob/6d9e46eac8b6eddf4e14f1547bb40f915d757cf3/e2e/internal/kubeclient/deploy.go#L206), so the timeout is probably too aggressive.


miampf (approved):
LGTM, thanks!

PR #1316, opened 2025-03-20
docs: load balancer and registry authentication

burgerdev:
Add some missing details with respect to load balancers and registry authentication.

I have some more content for the `How-To` section in mind:

* Exposing the Coordinator with Ingress/Gateway-API
* Working with genpolicy settings
* Determine correct reference values
* Manual recovery of persistence

### DIFF
--- a/docs/docs/howto/registry-authentication.md
+++ b/docs/docs/howto/registry-authentication.md
@@ -1,0 +1,48 @@
+# Registry authentication
+
+This guide shows how to set up registry credentials for Contrast.
+
+## Contrast CLI
+
+The Contrast CLI, specifically the `contrast generate` subcommand, needs access to the registry to derive policies for the referenced container images.
+The CLI authenticates to the registry using the [`docker_credential`](https://crates.io/crates/docker_credential) crate.
+This crate searches some default locations for a registry authentication file, so it should find credentials created by `docker login` or `podman login`.
+
+The only authentication method that's currently supported is `Basic` HTTP authentication with user name and password (or personal access token).
+Identity token flows, such as the default mechanism of plain `docker login`, don't work.
+Basic authentication can be forced with `docker login -u $REGISTRYUSER`.
+
+You can override the credentials file used by Contrast by setting the environment variable `DOCKER_CONFIG`. This is useful for creating a credential file from scratch, as shown in the following script:
+
+```sh
+#!/bin/sh
+
+registry="<put registry here>"
+user="<put user id here>"
+password="<put client secret here>"
+
+export DOCKER_CONFIG=$(mktemp)
+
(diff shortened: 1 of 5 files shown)


burgerdev:
I need to fix the publish job and I also noticed that there's some more content I need to add. Converting to draft for now, will let you know when I'm ready.

docs/sidebars.js
```diff
@@ -146,6 +146,17 @@ const sidebars = {
         },
       ]
     },
+    {
+      type: 'category',
+      label: 'How-To',
```
m1ghtym0, line 151: We now have "Workflows" in Constellation (which also covers day 1/2 actions of Constellation itself), "Guides" in Privatemode, and "How-To" in Contrast.
I like the explicit naming of "How-To" though:-)
Question is what happens to the "Workload deployment" page.
Should we refactor that into its own section and where we have all the "Contrast operation" references or would it become part of "How-to"?

m1ghtym0 (approved):

docs/sidebars.js
burgerdev: > We now have "Workflows" in Constellation (which also covers day 1/2 actions of Constellation itself), "Guides" in Privatemode, and "How-To" in Contrast.
> I like the explicit naming of "How-To" though:-)

I'm open to renaming to "Guides", but not a big fan of "Workflows". 

> Question is what happens to the "Workload deployment" page.

I'd keep the workload deployment guide where it is right now, but move (and expand) some of its content to How-Tos.

m1ghtym0: Sounds good, let's keep How-To, I like it:-)

docs/docs/howto/registry-authentication.md
```diff
@@ -0,0 +1,48 @@
...
+## AKS
+
+On AKS, images are pulled on the worker nodes using credentials available to Kubernetes and containerd.
+Follow the [official instructions](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/) to set up registry authentication with image pull secrets.
+
+## Bare metal
+
+On bare metal, images are pulled within the confidential guest, which doesn't receive credentials from the host yet.
+You can work around this by mirroring the required images to a private registry that's only exposed to the cluster.
+Such a registry needs to have a valid TLS certificate that's trusted in the web PKI (issued by Let's Encrypt, for example).
```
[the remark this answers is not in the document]
burgerdev: Not sure which examples section you're referring to. I lean towards adding some yaml I've written for a demo right here, but I don't think this should be a maintained feature - we should rather invest that energy into making registry authentication work!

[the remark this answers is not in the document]
burgerdev: As discussed privately, won't add the example because it's too lengthy.

PR #1317, opened 2025-03-21
coordinator: add notion of staleness

burgerdev:
The Coordinator has an internal in-memory state and an external state managed by a `Store` implementation. There's currently a two-way synchronization between the two:

* `SetManifest` requests result in store updates.
* Out-of-band store updates result in state refreshes (implemented in `syncState`).

The latter mechanism is not suitable for distributed Coordinators as specified in RFC010, because it derives new mesh certificates on persistency changes, although the Coordinator that wrote the update already derived its own mesh cert. Thus, we need to replace the store-to-state update happening in `syncState` with a peer recovery attempt. It's important to note that `syncState` was misguided to begin with, even for a single coordinator, because an out-of-band update could not really happen to a running Coordinator.

As a first step towards peer recovery, I'm replacing the `syncState` function with the concept of _state staleness_, which implements the _recovery mode_ outlined in the RFC. A Coordinator starts out uninitialized, becomes ready through a first manifest set, a user recovery or a peer recovery, and eventually becomes stale because other coordinators receive manifest updates.

```mermaid
stateDiagram-v2
    state "out-of-band manifest update" as oob
    [*] --> uninit
    uninit --> userapi.Recover
    userapi.Recover --> ready
    uninit --> meshapi.Recover
    meshapi.Recover -->  ready
    ready --> oob
    oob --> stale
    stale --> userapi.Recover
    stale --> meshapi.Recover
```

The coordinator would be in recovery mode (i.e., should attempt to do peer recovery and should accept user recovery) when it is either uninitialized or stale. Staleness is a property of the state itself, and when a state becomes stale it never becomes fresh again (a new state object could be fresh, though). This is why it's safe to track staleness as a boolean field in State that is only ever flipped in one direction.  

All API methods verify that their state is fresh before responding. However, state can become stale after the Coordinator started responding to a request. This is perfectly fine for the meshapi, where the client is just unlucky to initialize during a manifest update, but still gets certificates for the existing deployment. In the userapi, this means that there are concurrent requests. `GetManifests` returns _a_ state, but not necessarily the latest committed to storage, which is acceptable. Concurrent calls to `SetManifest` will fail eventually due to the `CompareAndSwap` logic in the store.

Staleness is not checked for meshapi functions: they should work on the state they are invoked with, regardless of that state becoming stale.

### DIFF
--- a/coordinator/internal/authority/authority.go
+++ b/coordinator/internal/authority/authority.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"slices"
 	"sync/atomic"
 
 	"github.com/edgelesssys/contrast/coordinator/history"
@@ -57,48 +56,6 @@ func New(hist *history.History, reg *prometheus.Registry, log *slog.Logger) *Aut
 			manifestGeneration: manifestGeneration,
 		},
 	}
-}
-
-// syncState ensures that a.state is up-to-date.
-//
-// This function guarantees to include all state updates committed before it was called.
-func (m *Authority) syncState() error {
-	hasLatest, err := m.hist.HasLatest()
-	if err != nil {
-		return fmt.Errorf("probing latest transition: %w", err)
-	}
-	if !hasLatest {
-		// No history yet -> nothing to sync.
-		return nil
-	}
-
-	oldState := m.state.Load()
-	if oldState == nil {
-		// There are transitions in history, but we don't have a local state -> recovery mode.
-		return ErrNeedsRecovery
-	}
-	nextState, err := m.fetchState(oldState.seedEngine)
-	if err != nil {
-		return fmt.Errorf("fetching latest state: %w", err)
-	}
-
-	// Don't update the state if it did not change.
-	if nextState.latest.TransitionHash == oldState.latest.TransitionHash {
-		return nil
-	}
-
-	// Manifest changed, verify that the seedshare owners did not change.
-	if slices.Compare(oldState.manifest.SeedshareOwnerPubKeys, nextState.manifest.SeedshareOwnerPubKeys) != 0 {
-		return fmt.Errorf("can't update from the current manifest to the latest persisted manifest because the seedshare owners changed")
-	}
-
-	// We consider the sync successful even if the state was updated by someone else.
-	if m.state.CompareAndSwap(oldState, nextState) {
-		// Only set the gauge if our state modification was actually successful - otherwise, it
-		// won't match the active state.
-		m.metrics.manifestGeneration.Set(float64(nextState.generation))
-	}
-	return nil
 }
 
 // fetchState creates a fresh state from the history that's verified by the given SeedEngine.
@@ -168,12 +125,12 @@ func (m *Authority) walkTransitions(transitionRef [history.HashSize]byte, consum
 
 // GetState syncs the current state and returns the loaded current state.
 func (m *Authority) GetState() (*State, error) {
-	if err := m.syncState(); err != nil {
-		return nil, fmt.Errorf("syncing state: %w", err)
-	}
 	state := m.state.Load()
 	if state == nil {
 		return nil, errors.New("coordinator is not initialized")
+	} else if state.stale.Load() {
+		// TODO(burgerdev): we could attempt peer recovery here.
+		return nil, ErrNeedsRecovery
 	}
 	return state, nil
 }
@@ -187,6 +144,10 @@ type State struct {
 
 	latest     *history.LatestTransition
 	generation int
+
+	// stale is set to true when we discover that this State is not current anymore.
+	// This field is only ever flipped from false to true!
+	stale atomic.Bool
 }
 
 // NewState constructs a new State object.
--- a/coordinator/internal/authority/authority.go
+++ b/coordinator/internal/authority/authority.go
@@ -4,11 +4,13 @@
 package authority
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
 	"sync/atomic"
+	"time"
 
 	"github.com/edgelesssys/contrast/coordinator/history"
 	"github.com/edgelesssys/contrast/internal/ca"
@@ -17,6 +19,7 @@ import (
 	"github.com/edgelesssys/contrast/internal/userapi"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"k8s.io/utils/clock"
 )
 
 // ErrNoManifest is returned when a manifest is needed but not present.
@@ -32,6 +35,8 @@ type Authority struct {
 	hist    *history.History
 	logger  *slog.Logger
 	metrics metrics
+
+	clock clock.Clock
 
 	userapi.UnimplementedUserAPIServer
 }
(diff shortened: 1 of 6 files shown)


davidweisse (approved):
Overall LGTM

coordinator/internal/authority/authority.go
```diff
@@ -42,63 +43,59 @@ type metrics struct {
...
-	}
-
-	// Manifest changed, verify that the seedshare owners did not change.
-	if slices.Compare(oldState.manifest.SeedshareOwnerPubKeys, nextState.manifest.SeedshareOwnerPubKeys) != 0 {
-		return fmt.Errorf("can't update from the current manifest to the latest persisted manifest because the seedshare owners changed")
-	}
+	// Set up a watcher that marks the current State stale if it does not correspond to the latest
+	// transition in store.
+
+	go func() {
```
katexochen, line 65: This should be its own method, which is then invoked in the constructor.

katexochen, line 65: time should be fakeable for tests, use kubernetes clock package.

katexochen, line 65: This should be tested through a unit test, ensuring the concurrency is done correct.

burgerdev: You mean a test where, say, `Set`, `Get` and updates from the store happen in parallel, similar to this one?

https://github.com/edgelesssys/contrast/blob/3a14afc060703fa7475fc46f2b9eb3a3b9720f10/coordinator/internal/authority/userapi_test.go#L445-L458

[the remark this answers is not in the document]
burgerdev: The goroutine is cancelled through the context - do you mean direct cancellation through a `Close` method?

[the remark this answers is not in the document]
katexochen: Something has to wait for the routine to finish. Usually done by a close function.

burgerdev: Done

burgerdev: Done and test added

burgerdev: Added a concurrent test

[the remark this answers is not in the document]
burgerdev: Changed the constructor back to not taking a context, added a `Stop` function to `Authority` and added goleak to the test.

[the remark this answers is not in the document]
3u13r: While not being that deep into how to do Go async best, I think that the current approach is also weird with the authority constructor saving it's own cancel func. I'm not sure if we have to decouple the authority context from the main context, but at least I'd pass inline in the `main.go`. This should allow us be be more straightforward and just call e.g., `cancelAuthority()`.

3u13r: I find the solution with an variable number of clock arguments which should only be used in tests counter-intuitive.
I'd align this with the usage of the `clock.Ticker` in `NewCachedHTTPSGetter()` in `cached_client.go`, where the caller explicitly passes the wanted clock. 

Another solution might be to just remove the parameter and set the fake clock directly in the struct for the tests.

coordinator/internal/authority/authority_test.go
```diff
@@ -50,13 +53,52 @@ func TestSNPValidateOpts(t *testing.T) {
...
+	require.Equal(int32(1), store.n.Load())
+	clock.Step(time.Minute)
+
+	require.Eventually(func() bool { return store.n.Load() == 2 }, time.Second, 10*time.Millisecond, "Authority should call store.Watch after watch stopped")
+}
+
+type badStore struct {
+	history.Store
+	ch atomic.Pointer[chan []byte]
+	n  atomic.Int32
```
3u13r, line 85:
```suggestion
	watchCalled  atomic.Int32
```
or `watchCounter`?

coordinator/internal/authority/authority_test.go
```diff
@@ -50,13 +53,52 @@ func TestSNPValidateOpts(t *testing.T) {
...
+	}
+	ch := make(chan []byte)
+	store.ch.Store(&ch)
+	hist := history.NewWithStore(slog.Default(), store)
+	reg := prometheus.NewRegistry()
+	clock := testingclock.NewFakeClock(time.Now())
+	a := New(hist, reg, slog.Default(), clock)
+	t.Cleanup(a.Stop)
+
+	require.Eventually(func() bool { return store.n.Load() == 1 }, time.Second, 10*time.Millisecond, "Authority should call store.Watch after init")
```
3u13r, line 70: There's apparently a debate/issues with the async testify functions: https://github.com/stretchr/testify/pull/1657. Just as an fyi. Stumbled over this when searching for the implementation of https://github.com/stretchr/testify/issues/1087.

coordinator/internal/authority/userapi_test.go
```diff
@@ -458,15 +472,227 @@ func TestUserAPIConcurrent(t *testing.T) {
...
 }
 
-func newCoordinatorWithRegistry(reg *prometheus.Registry) *Authority {
+func TestStoreRaces(t *testing.T) {
+	require := require.New(t)
+	ctx := context.Background()
+	log := slog.Default()
+
+	store := newWatchableStore()
+	hist := history.NewWithStore(slog.Default(), store)
```
3u13r, line 544:
```suggestion
	hist := history.NewWithStore(log, store)
```
Or do they have to be different loggers?

coordinator/internal/authority/userapi_test.go
```diff
@@ -458,15 +472,227 @@ func TestUserAPIConcurrent(t *testing.T) {
...
+
+	passiveCoordinator := New(hist, prometheus.NewRegistry(), log)
+	t.Cleanup(passiveCoordinator.Stop)
+	_, err := passiveCoordinator.GetManifests(ctx, nil)
+	assert.ErrorContains(t, err, ErrNoManifest.Error())
+
+	seedshareOwnerKey := testkeys.RSA(t)
+	workloadOwnerKey := testkeys.ECDSA(t)
+
+	b, err := json.Marshal(&manifest.Manifest{
```
3u13r, line 565:
```suggestion
	manifestBytes, err := json.Marshal(&manifest.Manifest{
```

3u13r (changes requested):

coordinator/internal/authority/authority.go
burgerdev: This is now a standalone method.

burgerdev: Clock is now an explicit argument to `WatchHistory`.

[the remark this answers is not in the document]
burgerdev: I restructured the approach so that no goroutines are started in the library. We now have a blocking function that updates the state, which is put into a goroutine in main. That way, we don't need to deal with cancelling resources in library code at all.

coordinator/internal/authority/authority_test.go
burgerdev: Thanks, that's useful to keep in mind. For the usage at hand, I believe we are fine because the trigger for all of these issues seems to be a func argument that runs longer than the timeout, and with timeouts >10ms we should be safe when the func does only trivial local computation.

coordinator/internal/authority/userapi_test.go
burgerdev: no, that was just an oversight

coordinator/internal/authority/authority_test.go
```diff
@@ -50,6 +53,54 @@ func TestSNPValidateOpts(t *testing.T) {
...
+		cancel()
+		<-doneCh
+	})
+
+	require.Eventually(func() bool { return store.watchCallCounter.Load() == 1 }, time.Second, 10*time.Millisecond, "Authority should call store.Watch after init")
+
+	newCh := make(chan []byte)
+	store.ch.Store(&newCh)
+	close(ch)
+	time.Sleep(10 * time.Millisecond)
```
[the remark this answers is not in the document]
burgerdev: Since the logic under test runs in a goroutine (T1) outside the test body goroutine (T0), there's no guarantee that closing the channel in T0 has an immediate effect in T1. I might get unlucky and the instructions in line 85 and 87 are executed in direct succession without yielding to T1. The sleep is there to give the scheduler a chance to advance T1 and prove that even if T1 was scheduled it does not call watch again immediately, but waits for the clock to step. This could be better expressed with a `require.Never`, though.

katexochen (approved):

PR #1318, opened 2025-03-21
grpc: use default dialer implementation that supports HTTP CONNECT

thomasten:
The grpc client wrapper `internal/grpc/dialer` takes a dialer that should be used to establish connections. This allows to inject a fake dialer for testing.
In the application code, we pass a `net.Dialer` instance. This overrides grpc's default dialer implementation, which supports HTTP CONNECT. To enable this feature, pass `nil` in the application code and then don't set a custom dialer for grpc.

I need some hints on how to test this. I thought about cloning one of the e2e tests and run some HTTP proxy in a goroutine.

### DIFF
--- a/e2e/internal/contrasttest/contrasttest.go
+++ b/e2e/internal/contrasttest/contrasttest.go
@@ -305,14 +305,16 @@ func (ct *ContrastTest) ApplyFromYAML(t *testing.T, yaml []byte) {
 	require.NoError(ct.Kubeclient.Apply(ctx, objects...))
 }
 
-// Set runs the contrast set subcommand.
-func (ct *ContrastTest) Set(t *testing.T) {
-	require := require.New(t)
-
+// RunSet runs the contrast set subcommand.
+func (ct *ContrastTest) RunSet() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
 	defer cancel()
+	return ct.runAgainstCoordinator(ctx, cmd.NewSetCmd(), path.Join(ct.WorkDir, "resources.yml"))
+}
 
-	require.NoError(ct.runAgainstCoordinator(ctx, cmd.NewSetCmd(), path.Join(ct.WorkDir, "resources.yml")))
+// Set runs the contrast set subcommand and fails the test if it is not successful.
+func (ct *ContrastTest) Set(t *testing.T) {
+	require.NoError(t, ct.RunSet())
 }
 
 // RunVerify runs the contrast verify subcommand.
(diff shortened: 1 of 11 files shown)


burgerdev (commented):
A dedicated e2e test for this feature may be a bit too much overhead. Maybe we can add a test to the regression e2e? A CONNECT proxy that tracks connection attempts could probably be implemented directly in our codebase, and then we test combinations of http proxy env vars with `t.Setenv`?

burgerdev (approved):
Thanks a lot, lgtm!

burgerdev:
I wonder how this change interferes with the `get-logs` script, though.

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -411,6 +413,10 @@ func (ct *ContrastTest) runAgainstCoordinator(ctx context.Context, cmd *cobra.Co
 	cmd.Flags().String("log-level", "debug", "")
 
 	return ct.Kubeclient.WithForwardedPort(ctx, ct.Namespace, "port-forwarder-coordinator", "1313", func(addr string) error {
+		// Go never uses a proxy for connections to localhost. To enable proxy tests, we
+		// replace localhost with 0.0.0.0, which can be used as localhost on Linux and BSD.
+		addr = strings.Replace(addr, "localhost", "0.0.0.0", 1)
```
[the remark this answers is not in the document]
thomasten: I found the replacement a bit hacky, so I kept the scope small. I don't have a strong opinion on it though.

e2e/regression/regression_test.go
```diff
@@ -89,11 +97,128 @@ func TestRegression(t *testing.T) {
 			require.NoError(c.WaitFor(ctx, kubeclient.Ready, kubeclient.Deployment{}, ct.Namespace, deploymentName))
 		})
 	}
+
+	t.Run("http-proxy", func(t *testing.T) { testHTTPProxy(t, ct) })
+}
+
+func testHTTPProxy(t *testing.T, ct *contrasttest.ContrastTest) {
```
thomasten, line 104: > I wonder how this change interferes with the `get-logs` script, though.

The e2e test setup expects exactly one namespace per run. So there should only be one `func Test...` per e2e test.

This happened:
* first `func Test...` runs
* `get-logs start` reads the first namespace name
* second `func Test...` runs and overwrites the namespace file
* `get-logs download` reads the second namespace name
* `grep log-collector` causes pipefail

(namespace cleanup is affected similarly)

burgerdev: Thanks for debugging this, I forgot about the generated namespace.

burgerdev (approved):
lgtm, but needs rebase

katexochen:
This PR needs a description suitable for users/changelog.

PR #1320, opened 2025-03-26
docs: correct sysctl setting recommendation

burgerdev:
The instructions pointed to a non-existent sysctl. While replacing it with the actual setting, I used the opportunity to also

* not strictly prescribe the file the setting needs to go into
* suggest that users should check whether their setting is too low, rather than telling them to blindly change the setting
* use systemd commands for applying changes
* tell users how to verify


m1ghtym0 (approved):
As someone who usually just follow such instructions without being that deep into the systemd/sysctl config stuff, the change looks good to me:-)

I would probably stumble a second over what `creating a config override in /etc/sysctl.d` means exactly, but that should become clear immediately after.

PR #1321, opened 2025-03-26
grpc: fix backoff

thomasten:
Setting `ConnectParams` without setting `Backoff` results in a backoff of 0. This causes `contrast set` to flood the server until connection succeeds. With `-c localhost` you can see 100% cpu usage.


burgerdev (approved):
lgtm, thanks!

PR #1324, opened 2025-03-31
Add peer recovery functionality to enterprise packages

davidweisse:
This adds the `RecoverFromPeer` method to the `Authority` which tries to recover its state from a given Coordinator host as specified in the RFC. For this, the authority first needs an issuer to build a TLS client for the connection.


coordinator/internal/authority/recovery.go
```diff
@@ -0,0 +1,158 @@
...
+	ticker := clock.RealClock{}.NewTicker(9 * month)
+	kdsGetter := certcache.NewCachedHTTPSGetter(memstore.New[string, []byte](), ticker, logger.NewNamed(m.logger, "kds-getter-validator"))
+	defer ticker.Stop()
+
+	validators, err := m.validatorsFromManifest(mnfst, kdsGetter)
+	if err != nil {
+		return fmt.Errorf("generating validators: %w", err)
+	}
+	dial := dialer.New(m.issuer, validators, atls.NoMetrics, &net.Dialer{}, m.logger)
+	conn, err := dial.Dial(ctx, net.JoinHostPort(host, meshapi.Port))
```
burgerdev, line 61: I'd prefer an `addr` argument and leaving the port choice to the caller.

coordinator/internal/authority/recovery.go
```diff
@@ -0,0 +1,158 @@
...
+	month := 30 * 24 * time.Hour
+	ticker := clock.RealClock{}.NewTicker(9 * month)
+	kdsGetter := certcache.NewCachedHTTPSGetter(memstore.New[string, []byte](), ticker, logger.NewNamed(m.logger, "kds-getter-validator"))
+	defer ticker.Stop()
+
+	validators, err := m.validatorsFromManifest(mnfst, kdsGetter)
+	if err != nil {
+		return fmt.Errorf("generating validators: %w", err)
+	}
+	dial := dialer.New(m.issuer, validators, atls.NoMetrics, &net.Dialer{}, m.logger)
```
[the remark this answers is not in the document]
davidweisse: I'll need to rebase once #1318 is merged.

burgerdev:
I'm currently piecing together a working HA branch from the various enabling commits, and taking over this branch for now to address feedback on #1317 and for putting this functionality into the enterprise tree.

enterprise/coordinator/recovery/recovery.go
```diff
@@ -0,0 +1,180 @@
...
+func dial(ctx context.Context, issuer atls.Issuer, validators []atls.Validator, logger *slog.Logger, addr string) (meshapi.MeshAPIClient, func() error, error) {
+	dial := dialer.New(issuer, validators, atls.NoMetrics, nil, logger)
+	conn, err := dial.Dial(ctx, addr)
+	if err != nil {
+		return nil, nil, fmt.Errorf("dialing coordinator: %w", err)
+	}
+
+	client := meshapi.NewMeshAPIClient(conn)
+	return client, conn.Close, nil
+}
```
katexochen, line 180: I don't get why these have to be funcs, can't we just use a struct/method here to fulfill the interfaces?

enterprise/coordinator/recovery/recovery.go
```diff
@@ -0,0 +1,180 @@
...
+		errs = append(errs, err)
+	}
+
+	return errors.Join(errs...)
+}
+
+// recoverFromPeer sends a recovery request to the peer coordinator and recovers the state.
+func (r *Recoverer) recoverFromPeer(ctx context.Context, oldState *stateguard.State, peer string) error {
+	r.logger.Info("attempting recovery", "peer", peer)
+	_, err := r.guard.ResetState(oldState, authorizerFunc(func(mnfst *manifest.Manifest) (*seedengine.SeedEngine, *ecdsa.PrivateKey, error) {
```
katexochen, line 112: nit: define func upfront to ease readability.

enterprise/coordinator/recovery/recovery.go
```diff
@@ -0,0 +1,180 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package recovery
```
katexochen, line 4: should we call this autorecovery to distinguish it from the cli-based recovery?

enterprise/coordinator/recovery/recovery_test.go
```diff
@@ -0,0 +1,272 @@
...
+			return newFakeClient(t), func() error { return nil }, nil
+		},
+		logger: logger,
+	}
+
+	require.NoError(r.recoverFromPeer(ctx, nil, expectedAddr))
+	require.True(dialCalled)
+}
+
+type fakeGuard struct {
```
katexochen, line 156: This seems to be a lot of code to implement the fake logic. Why aren't we using stubs instead?

enterprise/coordinator/recovery/recovery_test.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+type fakePeerGetter struct {
+	peers []string
+	err   error
+}
+
+func (pg *fakePeerGetter) GetPeers(context.Context) ([]string, error) {
+	return pg.peers, pg.err
+}
```
katexochen, line 234: this is a stub, not a fake

enterprise/coordinator/recovery/recovery_test.go
```diff
@@ -0,0 +1,272 @@
...
+		},
+		"one bad peer": {
+			peerGetter: &fakePeerGetter{[]string{"a", "b"}, nil},
+			guard:      newStaleGuard(t),
+			dial: func(_ context.Context, _ atls.Issuer, _ []atls.Validator, _ *slog.Logger, addr string) (meshapi.MeshAPIClient, func() error, error) {
+				if addr == "b:7777" {
+					return newFakeClient(t), func() error { return nil }, nil
+				}
+				return nil, nil, assert.AnError
+			},
```
katexochen, line 113: hard to read, a stubDialer could have a map from addr to result.

enterprise/coordinator/recovery/recovery.go
burgerdev: Renamed it to `peerrecovery` to avoid confusion with a potential recovery from sealed secrets.

enterprise/coordinator/recovery/recovery.go
```diff
@@ -0,0 +1,180 @@
...
+
+	clock clock.WithTicker
+	dial  func(context.Context, atls.Issuer, []atls.Validator, *slog.Logger, string) (meshapi.MeshAPIClient, func() error, error)
+}
+
+// guard is the public API of stateguard.Guard used by Recoverer.
+type guard interface {
+	// GetState returns the current state. If the error is nil, the state must be set.
+	GetState() (*stateguard.State, error)
+	// ResetState recovers to the latest persisted state, authorizing the recovery seed with the passed func.
```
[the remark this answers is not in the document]
burgerdev: done

enterprise/coordinator/recovery/recovery.go
```diff
@@ -0,0 +1,180 @@
...
+		if err := r.recoverOnce(ctx); err != nil {
+			r.logger.Warn("Could not recover from any peer.", "err", err)
+		}
+	})
+}
+
+var errNoPeers = errors.New("no peers found")
+
+// recoverOnce performs one round of recovery attempts over all discovered peers.
+func (r *Recoverer) recoverOnce(ctx context.Context) error {
```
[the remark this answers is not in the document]
burgerdev: Renamed it, but not too happy with the result. Let me know if you have an alternative preference.

enterprise/coordinator/recovery/recovery.go
```diff
@@ -0,0 +1,180 @@
...
+		return fmt.Errorf("resetting state: %w", err)
+	}
+	return nil
+}
+
+func periodically(ctx context.Context, clock clock.WithTicker, interval time.Duration, f func()) error {
+	t := clock.NewTicker(interval)
+	defer t.Stop()
+	for {
+		f()
```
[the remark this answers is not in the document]
burgerdev: Done

enterprise/coordinator/recovery/recovery_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	cancel()
+	err := receiveEventually(t, time.Second, errs)
+	require.ErrorIs(err, context.Canceled)
+}
+
+func receiveEventually[A any](t *testing.T, d time.Duration, ch chan A) A {
+	// t.FailNow, require.Fail etc. are not considered function returns, so the compiler complains
+	// about a missing return after the select. Wrapping the select in a for loop makes the error
+	// disappear.
```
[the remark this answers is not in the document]
burgerdev: Done

enterprise/coordinator/recovery/recovery.go
burgerdev: 1. The dialer is a struct now.
2. The authorizer used to be a func so that I can capture all of the relevant arguments that only exist in that call. However, it turns out that I can just make the Recoverer itself an authorizer, I only need to pass a context and add the peer.

enterprise/coordinator/recovery/recovery.go
burgerdev: Obsolete

enterprise/coordinator/recovery/recovery_test.go
burgerdev: Used stubs where possible. The guard, however, needs to call the authorizer, so we need at least some fake implementation there.

enterprise/coordinator/recovery/recovery_test.go
burgerdev: Renamed according to our convention.

enterprise/coordinator/recovery/recovery_test.go
burgerdev: Done

coordinator/internal/userapi/userapi.go
```diff
@@ -46,7 +46,7 @@ type guard interface {
 	// UpdateState advances the state to the given manifest and policies.
 	UpdateState(oldState *stateguard.State, se *seedengine.SeedEngine, manifest []byte, policies [][]byte) (newState *stateguard.State, err error)
 	// ResetState recovers to the latest persisted state, authorizing the recovery seed with the passed func.
-	ResetState(oldState *stateguard.State, a stateguard.SecretSourceAuthorizer) (newState *stateguard.State, err error)
+	ResetState(ctx context.Context, oldState *stateguard.State, a stateguard.SecretSourceAuthorizer) (newState *stateguard.State, err error)
```
katexochen, line 49: What about the other interface methods, shouldn't they have a context, too?

katexochen: Ah, comment says this is only used for the authorizer.

burgerdev: I think the other methods need a context, too, but wanted to tackle that separately including history. I can add them now, though.

katexochen (approved):
thanks!

enterprise/coordinator/recovery/recovery.go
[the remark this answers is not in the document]
burgerdev: I decided against `recoverFromPeers` because it's too similar to `recoverFromPeer`, but adding an `Available` sounds good.

PR #1327, opened 2025-04-02
manifest: re-organize code

katexochen:


katexochen:
TODO: add fist commit to git-blame-ignore-revs in a follow up PR.

PR #1335, opened 2025-04-04
history: allow insecure access to the latest transition

burgerdev:
This is required for RFC010, so that a recovering client can compare the latest manifest with what the recovering server sent. I factored this out from #1324.

### DIFF
--- a/coordinator/history/history_test.go
+++ b/coordinator/history/history_test.go
@@ -30,6 +30,8 @@ func TestHistory_GetLatestAndHasLatest(t *testing.T) {
 		wantErr bool
 		// wants for HasLatest
 		wantHasLatest bool
+		// wants for GetLatestInsecure
+		wantInsecureErr bool
 	}{
 		"success": {
 			fsContent: map[string]string{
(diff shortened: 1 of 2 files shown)


davidweisse (approved):

coordinator/history/history_test.go
```diff
@@ -59,18 +61,20 @@ func TestHistory_GetLatestAndHasLatest(t *testing.T) {
...
 			fsContent: map[string]string{
 				"transitions/latest": fromHex(rq, "2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824"),
 			},
-			signingKey:    testkeys.New[ecdsa.PrivateKey](t, testkeys.ECDSAP256Keys[0]),
-			wantErr:       true,
-			wantHasLatest: true,
+			signingKey:      testkeys.New[ecdsa.PrivateKey](t, testkeys.ECDSAP256Keys[0]),
+			wantErr:         true,
+			wantHasLatest:   true,
+			wantInsecureErr: true,
```
katexochen, line 77: Shouldn't we have new case where `GetLatestInsecure` succeeds but `GetLatest` donesn't?

burgerdev: That should be the case in `signature modified` above.

katexochen: makes sense.

katexochen (approved):

PR #1340, opened 2025-04-07
kata: disable genpolicy integration tests

burgerdev:
With this, we should not need to manually disable all tests that are added upstream in the future.


katexochen (approved):
thanks, lgtm beside formatter failure.

PR #1343, opened 2025-04-07
nodeinstaller: backup existing containerd config

davidweisse:
With this PR, the containerd config will be backed up before the node installer overwrites it to configure the new runtime. For an atomic update, the new config will be written to a temporary file first and then replace the old config.

### DIFF
--- a/nodeinstaller/node-installer.go
+++ b/nodeinstaller/node-installer.go
@@ -249,8 +249,27 @@ func patchContainerdConfig(runtimeHandler, basePath, configPath string, platform
 		return nil
 	}
 
+	// Backup the existing config.
+	if len(existingRaw) != 0 {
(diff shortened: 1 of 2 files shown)


nodeinstaller/node-installer.go
```diff
@@ -249,8 +249,27 @@ func patchContainerdConfig(runtimeHandler, basePath, configPath string, platform
 		return nil
 	}
 
+	// Backup the existing config.
+	if len(existingRaw) != 0 {
+		if err := os.WriteFile(fmt.Sprintf("%s.%d.bak", configPath, time.Now().Unix()), existingRaw, 0o666); err != nil {
```
burgerdev, line 254: Could you please add a sentence to  https://docs.edgeless.systems/contrast/components/runtime#node-level-components that describes the process and the backup, and add a link to that page to https://docs.edgeless.systems/contrast/deployment#deploy-the-contrast-runtime?

burgerdev (commented):
Thanks!

burgerdev (approved):

PR #1344, opened 2025-04-07
cli/generate: validate existing manifest

davidweisse:
When running `contrast generate` with an invalid manifest, the returned error would always say `Please fill in the reference values` even if the manifest validation error was unrelated. This PR fixes this by validating any existing manifest on `contrast generate` before proceeding. The `Please fill in the reference values` message will still be printed for bare metal platforms if a new manifest is being generated by the CLI and no embedded reference values can be used.


3u13r (approved):
LGTM

PR #1346, opened 2025-04-07
transitengineapi: tls client cert auth

jmxnzo:
Adds the required auth logic to the transit engine API to accomplish the client authentication of the mTLS handshake. A client can be authenticated based on a valid mesh cert issued by the mesh CA of the current state.

- [x] Add GetMeshCAPool to CA in state


burgerdev (commented):
Please squash the commits - at least the first one is partially reverted by the last one.

burgerdev (approved):

PR #1348, opened 2025-04-07
transitengineapi: mTLS with client authorization by workloadSecret 

jmxnzo:
This PR introduces mTLS to the transit engine API to secure connections based on Contrast mesh certs.

- The client is only authorized to the transit key set of the workloadSecretID in the received meshCert extension. As prior we handle the <name> parameter of existing Vault implementations as our workloadSecretID.

- On initialization the transit engine API derives a private ecdsa key. For each incoming TLS connection of the transit engine API, the coordinator issues itself a fresh mesh cert for this private key of the transit engine API, which is therefore valid for the current state.

 The logging mechanism introduced in #1345 and the new authorizationMiddleware is bypassed during testing using a mock transitengine API.

### DIFF
--- a/coordinator/internal/transitengineapi/transitengineapi.go
+++ b/coordinator/internal/transitengineapi/transitengineapi.go
@@ -7,10 +7,14 @@
 package transitengine
 
 import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/asn1"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"log/slog"
 	"net/http"
@@ -19,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/edgelesssys/contrast/coordinator/internal/authority"
+	"github.com/edgelesssys/contrast/internal/manifest"
 	"github.com/edgelesssys/contrast/internal/oid"
 )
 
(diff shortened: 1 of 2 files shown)


coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -216,6 +237,20 @@ func extractVersion(versionStr string) (uint32, error) {
 	return uint32(version), nil
 }
 
+func extractCertExtension(cert *x509.Certificate, oid asn1.ObjectIdentifier) (string, error) {
```
burgerdev, line 240: Do we need to pass the constant OID as an argument?

coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -253,3 +288,40 @@ func loggingMiddleware(next http.HandlerFunc, logger *slog.Logger) http.HandlerF
 		}
 	})
 }
+
+func authorizationMiddleware(next http.HandlerFunc) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		workloadSecretID := r.PathValue("name")
+		if err := authorizeWorkloadSecret(workloadSecretID, r); err != nil {
+			http.Error(w, fmt.Sprintf("Unauthorized: %v", err), http.StatusUnauthorized)
```
burgerdev, line 296: `http.StatusUnauthorized` is only correct if the client is not _authenticated_ (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401). Since we authenticate at the TLS layer, we can always return 403 here.

coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -253,3 +288,40 @@ func loggingMiddleware(next http.HandlerFunc, logger *slog.Logger) http.HandlerF
...
+		next.ServeHTTP(w, r)
+	})
+}
+
+func getCertificate(privKeyAPI *ecdsa.PrivateKey, authority stateAuthority) (*tls.Certificate, error) {
+	state, err := authority.GetState()
+	if err != nil {
+		return nil, fmt.Errorf("getting state: %w", err)
+	}
+	dnsNames := []string{"coordinator"}
```
burgerdev, line 308: The SANs should come from the manifest (`Role == RoleCoordinator`).

PR #1349, opened 2025-04-07
meshapi: add workload secret as extension

jmxnzo:
This PR adds an extension to the issued mesh certificates holding the workloadSecretID, if set in the manifest entry. Therefore allowing for faster authorization without any manifest look-ups.

### DIFF
--- a/coordinator/internal/meshapi/meshapi.go
+++ b/coordinator/internal/meshapi/meshapi.go
@@ -17,6 +17,9 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
+
+	"github.com/edgelesssys/contrast/internal/attestation/extension"
+	"github.com/edgelesssys/contrast/internal/oid"
 )
 
 // Server implements the meshapi service.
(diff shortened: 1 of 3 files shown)


burgerdev:
> > 1.3.9901 is the "as you like it" prefix, right? Should we take a different next number to distinguish report extensions from claim extensions? Don't we already have an extension number for claims somewhere? ( [#1302 (comment)](https://github.com/edgelesssys/contrast/pull/1302#discussion_r1995808661) )
> 
> @burgerdev can you elaborate on the number for claims some more? I couldn't find it myself, i would be happy about a suggestion here.

The extensions we have right now are for TEE reports:

```go
var RawSNPReport = asn1.ObjectIdentifier{1, 3, 9901, 2, 1}
var RawTDXReport = asn1.ObjectIdentifier{1, 3, 9901, 2, 2}
```

The `1.3.9901` prefix is for application-defined extensions, and I was suggesting to _not_ use `1.3.9901.2.*` for adding a new extension type, but rather start a new hierarchy at e.g. `1.3.9901.3.*`. This would allow using `1.3.9901.2.3` for an ARM CCA report, for example.

The OIDs I had in mind for individual claims are actually sub-OIDs of the reports, so using a new OID under `1.3.9901` is fine: https://github.com/edgelesssys/contrast/blob/9536dad955ea1ac697ccd867026932e284c5295c/internal/attestation/tdx/extensions.go#L17

burgerdev (approved):
lgtm

burgerdev:
The TDX runner is broken, and I think the code changes are sufficiently tested on the other runners - force-merging.

PR #1350, opened 2025-04-07
coordinator: move WalkTransitions function to history package

burgerdev:
I'd like to use this function from enterprise code to implement RFC010, so I moved it to the history package (where it arguably makes more sense, too) and added a test.

### DIFF
--- a/coordinator/history/history.go
+++ b/coordinator/history/history.go
@@ -170,6 +170,23 @@ func (h *History) WatchLatestTransitions(ctx context.Context) (<-chan LatestTran
 	return transitionCh, nil
 }
 
+// WalkTransitions executes a function for the referenced transition and all its ancestors.
+//
+// The all-zero transition is the root node of all transition trees and is not passed to the closure.
+func (h *History) WalkTransitions(transitionHash [HashSize]byte, consume func([HashSize]byte, *Transition) error) error {
+	for transitionHash != [HashSize]byte{} {
+		transition, err := h.GetTransition(transitionHash)
+		if err != nil {
+			return fmt.Errorf("getting transition %x: %w", transitionHash, err)
+		}
+		if err := consume(transitionHash, transition); err != nil {
+			return fmt.Errorf("running consume function for transition %x: %w", transitionHash, err)
+		}
+		transitionHash = transition.PreviousTransitionHash
+	}
+	return nil
+}
+
 func (h *History) getContentaddressed(pathFmt string, hash [HashSize]byte) ([]byte, error) {
 	hashStr := hex.EncodeToString(hash[:])
 	data, err := h.store.Get(fmt.Sprintf(pathFmt, hashStr))
--- a/coordinator/history/history_test.go
+++ b/coordinator/history/history_test.go
@@ -644,6 +644,83 @@ func TestHistory_WatchLatestTransitions(t *testing.T) {
 	}, 10*time.Millisecond, time.Millisecond)
 }
 
+func TestHistory_WalkTransitions(t *testing.T) {
+	t.Run("empty history", func(t *testing.T) {
+		require := require.New(t)
+		h := &History{
+			store:   &AferoStore{fs: &afero.Afero{Fs: afero.NewMemMapFs()}},
+			hashFun: sha256.New,
+		}
+		doNotCall := func(_ [32]byte, _ *Transition) error {
+			require.Fail("closure should not be called without any transitions")
+			return nil
+		}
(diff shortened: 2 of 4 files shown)


burgerdev:
Found a new use case while decoupling authority, userapi and enterprise, so let's get this in.

coordinator/history/history.go
```diff
@@ -170,6 +170,23 @@ func (h *History) WatchLatestTransitions(ctx context.Context) (<-chan LatestTran
 	return transitionCh, nil
 }
 
+// WalkTransitions executes a function for the referenced transition and all its ancestors.
+//
+// The all-zero transition is the root node of all transition trees and is not passed to the closure.
+func (h *History) WalkTransitions(transitionRef [HashSize]byte, consume func([HashSize]byte, *Transition) error) error {
```
katexochen, line 176: The rest of the package refers to `transitionRef` as `hash`, so it should be called `hash`/`transitionHash` here, too.

coordinator/history/history_test.go
```diff
@@ -644,6 +644,61 @@ func TestHistory_WatchLatestTransitions(t *testing.T) {
...
+
+	t.Run("walk transitions", func(t *testing.T) {
+		require := require.New(t)
+		closureCallCount := 0
+		require.NoError(h.WalkTransitions(latestTransition, func(_ [32]byte, _ *Transition) error {
+			closureCallCount++
+			return nil
+		}))
+		require.Equal(expectedTransitionChainSize, closureCallCount)
+	})
```
katexochen, line 699: As I understand it, "populate history" isn't actually a test case here, but preparation for the "walk transitions" test case, right? I should be in the same test case then.

coordinator/history/history_test.go
```diff
@@ -644,6 +644,61 @@ func TestHistory_WatchLatestTransitions(t *testing.T) {
 	}, 10*time.Millisecond, time.Millisecond)
 }
 
+func TestHistory_WalkTransitions(t *testing.T) {
+	h := &History{
```
katexochen, line 648: There shouldn't be shared state between test cases. Move into test case or make it a func.

coordinator/history/history.go
```diff
@@ -170,6 +170,23 @@ func (h *History) WatchLatestTransitions(ctx context.Context) (<-chan LatestTran
...
+	for transitionRef != [HashSize]byte{} {
+		transition, err := h.GetTransition(transitionRef)
+		if err != nil {
+			return fmt.Errorf("getting transition: %w", err)
+		}
+		if err := consume(transitionRef, transition); err != nil {
+			return err
```
katexochen, line 183: Maybe we should wrap the error here?

katexochen, line 183: There isn't a test case covering this line.

katexochen (changes requested):

coordinator/history/history.go
```diff
@@ -170,6 +170,23 @@ func (h *History) WatchLatestTransitions(ctx context.Context) (<-chan LatestTran
...
+//
+// The all-zero transition is the root node of all transition trees and is not passed to the closure.
+func (h *History) WalkTransitions(transitionHash [HashSize]byte, consume func([HashSize]byte, *Transition) error) error {
+	for transitionHash != [HashSize]byte{} {
+		transition, err := h.GetTransition(transitionHash)
+		if err != nil {
+			return fmt.Errorf("getting transition %x: %w", transitionHash, err)
+		}
+		if err := consume(transitionHash, transition); err != nil {
+			return fmt.Errorf("running closure for transition %x: %w", transitionHash, err)
```
katexochen, line 183:
```suggestion
			return fmt.Errorf("running consume function for transition %x: %w", transitionHash, err)
```

coordinator/history/history_test.go
```diff
@@ -644,6 +644,83 @@ func TestHistory_WatchLatestTransitions(t *testing.T) {
...
+
+		closureCallCount := 0
+		require.NoError(h.WalkTransitions(latestTransition, func(_ [32]byte, _ *Transition) error {
+			closureCallCount++
+			return nil
+		}))
+		require.Equal(expectedTransitionChainSize, closureCallCount)
+	})
+
+	t.Run("failing closure", func(t *testing.T) {
```
katexochen, line 702:
```suggestion
	t.Run("failing consume func", func(t *testing.T) {
```

katexochen (approved):

burgerdev:
The changed code is not platform dependent, so I'm going to ignore the failing TDX tests (that are due to a broken runner).

PR #1355, opened 2025-04-08
docs: improve language of SNP attestation

thomasten:
I reviewed the docs page. Content LGTM, so I didn't make changes in that respect.

Apart from language improvements, notable changes are:
* Added one or two introductory sentences to some of the sections.
* Removed some fields from the attestation report table. Please let me know if you think these should stay.

### DIFF
--- a/docs/docs/architecture/snp.md
+++ b/docs/docs/architecture/snp.md
@@ -1,13 +1,17 @@
-# SNP Attestation
+# AMD SEV-SNP attestation
 
-The key component for attesting AMD SEV-SNP is the Security Processor (SP),
-which measures the CVM and metadata and returns an attestation report reflecting those
+This page explains details of the AMD SEV-SNP attestation that are especially relevant for the implementation in Contrast.
+You should be familiar with [attestation basics](attestation.md).
+For more details on SEV-SNP, see the [whitepapers and specifications from AMD](https://www.amd.com/en/developer/sev.html).
+
+The key component for attesting AMD SEV-SNP is the Secure Processor (SP).
+It measures the confidential VM (CVM) and metadata and returns an attestation report reflecting those
 measurements.
-
 
 ## Startup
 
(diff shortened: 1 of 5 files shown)


docs/docs/architecture/snp.md
```diff
@@ -1,44 +1,52 @@
...
-The `HOSTDATA` is opaque for the SP. Kata writes the hash of the kata policy in
-this field to bind the policy to the CVM. `HOSTDATA` is later reflected in the
+The `HOST_DATA` is opaque to the SP. Kata writes the hash of the kata policy in
+this field to bind the policy to the CVM. `HOST_DATA` is later reflected in the
 attestation report.
 
-### ID Block Structure
+### ID block structure
+
+The ID block defines the identity of the CVM.
```
3u13r, line 28: I think it makes sense to move the paragraph below the ID block table to here so that the reader knows _how_ the ID block defines the identity (because it's enforced by the SP) and what the "identity" entails (at least the launch digest and the policy). Maybe we can spell that out that this is what we mean by identity. Though, this would be our own concept. Another option is to spell out that the ID block defines the expected contents (via LD) and setting (via policy).

3u13r (changes requested):
Thanks for the review/PR. Only one thing I tripped over.

docs/docs/architecture/snp.md
thomasten: I actually just wanted to give a minimal description of the ID block because the section shouldn't start with "The complete structure ...".
I could replace this sentence with "The ID block contains fields that identify the VM." (derived from a similar sentence in the AMD whitepaper)

We can also go with yours, but I'm not sure if I understood everything. If you think we should go with that, can you phrase the full "ID block structure" section?

3u13r: > The ID block contains fields that identify the VM.

I like that

thomasten:
```suggestion
The ID block contains fields that identify the VM.
```

3u13r (approved):
LGTM

PR #1357, opened 2025-04-09
coordinator: expose methods for external recovery

burgerdev:
We're going to add an external recovery mechanism to the enterprise package, which does not have accesss to the internal structs of the authority. Thus, we add two functions that can be called from outside:

* `RecoverWith` allows the enterprise package to perform peer recovery with all the checks from RFC010, while the authority only updates its internal state without checking.
* `NeedsRecovery` allows for early abort of recovery attempts, to be used in the peer recovery loop and the readiness probe.


burgerdev:
This is not the API we want.

PR #1358, opened 2025-04-09
coordinator: make KDS cache shareable

burgerdev:
The enterprise code for peer recovery is going to validate the recovery server, so we need a certcache for it. It seems like a good idea to share the cert cache with the one used for meshapi server validation, because the Coordinators will be recovering each other after each `Setmanifest`. Thus, create the cache in main and inject it into the meshapi server (for now) and into the peer recovery routine (in the future).


3u13r (approved):
LGTM

PR #1361, opened 2025-04-09
enterprise: make peerdiscovery a struct

burgerdev:
This makes it easier to use `GetPeers`, because I don't need to hold on to the clientset. It also allows for future improvements, like an informer cache.

### DIFF
--- a/enterprise/coordinator/peerdiscovery/peerdiscovery.go
+++ b/enterprise/coordinator/peerdiscovery/peerdiscovery.go
@@ -16,9 +16,22 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+type Discovery struct {
+	client    kubernetes.Interface
+	namespace string
+}
+
+func New(client kubernetes.Interface, namespace string) *Discovery {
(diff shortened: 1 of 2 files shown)


katexochen:
please rebase on main to see golangci-lint fail. ;)

davidweisse (approved):

enterprise/coordinator/peerdiscovery/peerdiscovery.go
```diff
@@ -14,9 +14,22 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+type Discovery struct {
+	client    kubernetes.Interface
+	namespace string
+}
+
+func New(client kubernetes.Interface, namespace string) *Discovery {
```
katexochen, line 22: This is exported and should have comments, right? (like I said, expected golangci-lint to fail)

burgerdev: Looks like that check is still disabled, then.

PR #1363, opened 2025-04-09
contrast-enterprise: init

katexochen:
Adding the build targets, container and justfile parts. You can already configure it via `enterprise="true"` in justfile.env and the enterprise coordinator will be just in the just workflows.


burgerdev (approved):
lgtm, thanks!

katexochen:
only log collection failing, merging anyway

PR #1364, opened 2025-04-10
node-installer: fix timestamp precision in unit restart detection

katexochen:
Previously, we only parsed the containerd/k3s restart time with second precision, which lead to unnecessary restarts as we couldn't if the restart happened after or before the config mtime. With this change, we will use micro second precision.

However, there is a bug in the systemd version/build used by azurelinux on AKS, which won't include the micro seconds for unknown reason. So instead of invoking systemctl, we talk to dbus directly to get the raw start time.

However, while we already mount the host path under `/host` into the node-installer, and the dbus socket is under `/var/run/dbus` on the host, `/var/run` is usually a symlink to `/run`, and depending on the way distros create this symlink, it might point out of `/host` to the `/run` inside the container. This is the case on Ubuntu, as the symlink is `/var/run -> /run`. This is a problem, as there is noting under `/run` of the container, especially no dbus socket. On azurelinux, this isn't a problem, as the symlink is relative `/var/run -> ../run`, so it is still pointing to correctly to `/host/run`. To solve this, we mount the dbus socket as separate volume.

### DIFF
--- a/nodeinstaller/node-installer.go
+++ b/nodeinstaller/node-installer.go
@@ -293,24 +293,28 @@ func restartHostContainerd(containerdConfigPath, service string) error {
 	configMtime := info.ModTime()
 
 	// get containerd start time
-	// Note that "--timestamp=unix" is not supported in the installed version of systemd (v250) at the time of writing.
 	serviceStartTime, err := exec.Command(
 		"nsenter", "--target", "1", "--mount", "--",
-		"systemctl", "show", "--timestamp=utc", "--property=ActiveEnterTimestamp", service,
+		"systemctl", "show", "--timestamp=us+utc", "--property=ActiveEnterTimestamp", service,
 	).CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("getting service (%s) start time: %w %q", service, err, serviceStartTime)
 	}
 
-	// format: ActiveEnterTimestamp=Day YYYY-MM-DD HH:MM:SS UTC
+	// format: ActiveEnterTimestamp=Day YYYY-MM-DD HH:MM:SS.000000 UTC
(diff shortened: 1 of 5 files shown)


burgerdev (approved):

nodeinstaller/node-installer.go
```diff
@@ -44,6 +45,7 @@ func main() {
 	fetcher := asset.NewDefaultFetcher()
 	if err := run(context.Background(), fetcher, platform, *shouldRestartContainerd); err != nil {
 		fmt.Println(err)
+		time.Sleep(10 * time.Minute)
```
burgerdev, line 48: Do you want to keep this?

nodeinstaller/node-installer.go
```diff
@@ -284,33 +286,21 @@ func parseExistingContainerdConfig(path string) ([]byte, config.ContainerdConfig
...
-		return fmt.Errorf("getting service (%s) start time: %w %q", service, err, serviceStartTime)
-	}
-
-	// format: ActiveEnterTimestamp=Day YYYY-MM-DD HH:MM:SS UTC
-	dayUTC := strings.TrimPrefix(strings.TrimSpace(string(serviceStartTime)), "ActiveEnterTimestamp=")
-	startTime, err := time.Parse("Mon 2006-01-02 15:04:05 MST", dayUTC)
+	startTime, err := getSystemdServiceRestartTime(ctx, service)
 	if err != nil {
-		return fmt.Errorf("parsing service (%s) start time: %w", service, err)
+		return fmt.Errorf("retreiving service (%s) start time: %w", service, err)
```
burgerdev, line 299:
```suggestion
		return fmt.Errorf("retrieving service (%s) start time: %w", service, err)
```

PR #1367, opened 2025-04-11
packages: replace rec with finalAttrs on buildGoModule

katexochen:
This was made possible through recent upstream changes, see https://github.com/NixOS/nixpkgs/commit/06a152f9a174b9131c0c4e6ff08e662f493e2e44


katexochen:
@burgerdev this should fix your problem with the tags in checkPhase not being correct on the contrast-enterprise package.

burgerdev (approved):
Thanks!

PR #1374, opened 2025-04-11
kata-image: add default-deny systemd unit

3u13r:
The goal of this PR is to alleviate the user from needing to manually check that the service-mesh sidecar has implemented its iptables rules. This should allow for safe default usage of the service-mesh feature.
See commits messages for more information.

Point of discussion: The Envoy binary is 89MB. Since we need to remove the default deny rule from all podVMs, we currently use the "disabled" service-mesh image as initContainer to remove them from e.g. the coordinator's podVM. This necessitates the memoryRequest increases in this PR. We could also use a smaller image that doesn't contain Envoy to try to safe some memory. Though I think this can be done as a follow up if needed. Let me know what you think.


packages/by-name/microsoft/kata-image/package.nix
```diff
@@ -57,6 +57,31 @@ let
...
+      EOF
+
+      # create a "RequiredBy" dependency where the kata-agent.service requires
+      # the deny-incoming-traffic.service to be started and successfully exited
+      # before the kata-agent.service can be started
+      # Usually this is done via the "RequiredBy" directive under the [Install]
+      # section of the systemd unit file, but the filesystem is read only
+      # and systemd wants the create the following symlink.
+      mkdir -p /build/rootfs/etc/systemd/system/kata-agent.service.requires/
+      ln -s /build/rootfs/usr/lib/systemd/system/deny-incoming-traffic.service /build/rootfs/etc/systemd/system/kata-agent.service.requires/deny-incoming-traffic.service
```
burgerdev, line 82: Is the symlink target actually correct in the final image, after the `/build/rootfs` was stripped?

packages/by-name/service-mesh/package.nix
```diff
@@ -12,21 +12,25 @@ buildGoModule (finalAttrs: {
...
       inherit root;
       fileset = fileset.unions [
         (path.append root "go.mod")
         (path.append root "go.sum")
-        (path.append root "golden/defaultEnvoy.json")
-        (fileset.fileFilter (file: hasSuffix ".go" file.name) root)
+        (path.append root "service-mesh/go.mod")
+        (path.append root "service-mesh/go.sum")
+        (path.append root "service-mesh/golden/defaultEnvoy.json")
+        (fileset.fileFilter (file: hasSuffix ".go" file.name) (path.append root "internal/defaultdeny"))
```
burgerdev, line 25: Why don't we add the `defaultdeny` package to the service-mesh module?

service-mesh/iptables.go
```diff
@@ -19,18 +19,11 @@ const EnvoyIngressPortNoClientCert = 15007
...
+	if err := defaultdeny.EnsureXTablesFile(); err != nil {
+		return fmt.Errorf("ensuring xtables file: %w", err)
 	}
-	file, err := os.Create("/run/xtables.lock")
-	if err != nil {
-		return fmt.Errorf("failed to create /run/xtables.lock: %w", err)
-	}
-	_ = file.Close()
 
+	// Create missing `/run/xtables.lock` file.
```
burgerdev, line 26: The comment looks misplaced now.

docs/docs/components/service-mesh.md
```diff
@@ -14,6 +14,20 @@ all workloads which have a specified configuration. The service mesh container f
 sets up `iptables` rules based on its configuration and then starts
 [Envoy](https://www.envoyproxy.io/) for TLS origination and termination.
 
+## Service mesh startup enforcement
+Since Contrast doesn't yet enforce the order in which the containers are started
+(see [Limitations](../features-limitations.md)), we deny all incoming connections
+until the service mesh resolves this deny rule.
```
burgerdev, line 20: I stumbled over "this deny rule", because the implicit definition a few words earlier was not apparent. Maybe

> we deny all incoming connections until the service mesh is fully configured.

docs/docs/components/service-mesh.md
```diff
@@ -14,6 +14,20 @@ all workloads which have a specified configuration. The service mesh container f
...
+A systemd unit inside the podVM creates this deny rule.
+The kata-agent systemd unit requires that this unit successfully runs and exists.
+Therefore, the deny rule is in place, before any containers can be started.
```
burgerdev, line 23:
```suggestion
Therefore, the deny rule is in place before any containers can be started.
```

docs/docs/components/service-mesh.md
```diff
@@ -14,6 +14,20 @@ all workloads which have a specified configuration. The service mesh container f
...
+
+Contrast ensures that this deny rule doesn't drop traffic in scenarios that don't
+require a service mesh by injecting the service mesh container not as a sidecar
+but as an init container. During generation, the CLI also sets `CONTRAST_SERVICE_MESH_DISABLE`
+as an environment on the container to disable any further service mesh setup.
+The container just removes the deny rule and exits.
```
burgerdev, line 29: Did you consider lifting the deny rule in the initializer instead? That would avoid the need for an additional image pull.

3u13r: Yes, but then we'd also need NET_ADMIN for the initializer. Also the coordinator pod, who doesn't have an initializer (yet) needs either NET_ADMIN to remove the rule itself or needs an "disabled" initializer. In summary, using only/one additional initializer might be optimal. Though I have no string favorite. Do you have a preference?

burgerdev: The Coordinator does have `SYS_ADMIN`, and I would not mind the `NET_ADMIN` too much in a real (terminating) init container. But I see the point of not duplicating the removal rule between the coordinator and the initializer. Do I understand correctly that the Coordinator now gets a (disabled) mesh sidecar, too? I missed that in review :sweat_smile:

3u13r: > understand correctly that the Coordinator now gets a (disabled) mesh sidecar, too?
yes. The Coordinator currently doesn't have NET_ADMIN and is a long-running (albeit a highly trusted) component. Therefore, I didn't want to add it, if it's not needed in the long term. But I think it's a discussion of a new container pull for _every_ confidential pod or _one_ additional image pull for the coordinator. So we should go with option 2, right? This is also the reason why the defaultdeny package is already in the shared /internal folder, because I first implemented it in the initializer.

packages/by-name/microsoft/kata-image/package.nix
katexochen: This should be a relative symlink (remember recent issues with /var/run -> /run)

service-mesh/go.mod
```diff
@@ -2,9 +2,12 @@ module github.com/edgelesssys/contrast/service-mesh
 
 go 1.23.0
 
+replace github.com/edgelesssys/contrast => ..
```
katexochen, line 5: needs update of renovate config!

service-mesh/iptables.go
```diff
@@ -19,18 +19,11 @@ const EnvoyIngressPortNoClientCert = 15007
 
 // IngressIPTableRules sets up the iptables rules for the ingress proxy.
 func IngressIPTableRules(ingressEntries []ingressConfigEntry) error {
-	// Create missing `/run/xtables.lock` file.
-	if err := os.Mkdir("/run", 0o755); err != nil {
-		if !os.IsExist(err) {
-			return fmt.Errorf("failed to create /run directory: %w", err)
-		}
+	if err := defaultdeny.EnsureXTablesFile(); err != nil {
+		return fmt.Errorf("ensuring xtables file: %w", err)
```
katexochen, line 23: I think this is currently in the wrong commit.

service-mesh/main.go
```diff
@@ -38,8 +40,14 @@ func run() error {
 
 	adminPort := os.Getenv(adminPortEnvVar)
 	log.Println("Port for Envoy admin interface:", adminPort)
+
```
katexochen, line 43: Comment and commented line are in different commits.

docs/docs/components/service-mesh.md
```diff
@@ -14,6 +14,20 @@ all workloads which have a specified configuration. The service mesh container f
 sets up `iptables` rules based on its configuration and then starts
 [Envoy](https://www.envoyproxy.io/) for TLS origination and termination.
 
+## Service mesh startup enforcement
```
katexochen, line 17:
```suggestion
## Service mesh startup enforcement

```

docs/docs/components/service-mesh.md
katexochen, line 23: Maybe this is to technical for the docs? We should rather describe the guarantees and consequences as how this is implemented.

docs/docs/components/service-mesh.md
```diff
@@ -14,6 +14,20 @@ all workloads which have a specified configuration. The service mesh container f
...
+Since Contrast doesn't yet enforce the order in which the containers are started
+(see [Limitations](../features-limitations.md)), we deny all incoming connections
+until the service mesh resolves this deny rule.
+A systemd unit inside the podVM creates this deny rule.
+The kata-agent systemd unit requires that this unit successfully runs and exists.
+Therefore, the deny rule is in place, before any containers can be started.
+
+Contrast ensures that this deny rule doesn't drop traffic in scenarios that don't
+require a service mesh by injecting the service mesh container not as a sidecar
```
katexochen, line 26: This should be more connected to action of a user: "by setting flag -xy during generate, the service mesh can be disabled,... in this case, the service mesh will be injected as init container to lift the deny rule..."

docs/docs/components/service-mesh.md
burgerdev: The SYS_ADMIN cap should include NET_ADMIN, I think? 

I don't understand the options you're mentioning - afaiu we have either

1. Add defaultdeny lifting to the initializer (conditional on service mesh absence) and the coordinator (unconditional).
2. Add service mesh container to the coordinator and all non-mesh pods.

The more I think about it, the more I prefer (1).

packages/by-name/microsoft/kata-image/package.nix
3u13r: This should work now with a relative symlink: 
```
root [ /etc/systemd/system/kata-agent.service.requires ]# ls
ls
deny-incoming-traffic.service
root [ /etc/systemd/system/kata-agent.service.requires ]# ls -la
ls -la
total 12
drwxr-xr-x  2 1000 users  68 Jan  1  1970 .
drwxr-xr-x 10 1000 users 483 Jan  1  1970 ..
lrwxrwxrwx  1 1000 users  67 Jan  1  1970 deny-incoming-traffic.service -> ../../../../../usr/lib/systemd/system/deny-incoming-traffic.service
root [ /etc/systemd/system/kata-agent.service.requires ]# ls -la ../../../../../usr/lib/systemd/system/deny-incoming-traffic.service
<sr/lib/systemd/system/deny-incoming-traffic.service
-rw-r--r-- 1 1000 users 252 Jan  1  1970 ../../../../../usr/lib/systemd/system/deny-incoming-traffic.service
```

The currently best way to test this is to replace the iptables command in the deny unit with something that fails (e.g. cat this-file-does-not-exist).
Can you think of a good way to test that the deny unit is required for the kata-agent? I think this leads to the more general question of how to test "CDH"-like Contrast components.
I could write a test that at lest checks that the symlink here is "correct" and that the deny system unit was started and ask systemd about the ordering.
What do you think? In any case though, I think this should be a follow up, since the good path is already tested.

service-mesh/go.mod
3u13r: Does it? We already exclude dependencies on the contrast module, what am I missing?

docs/docs/components/service-mesh.md
3u13r: Should I just delete the marked sentences? Do you think the docs are sufficient without it?

3u13r:
re-worked the way the default deny rule is removed as described in https://github.com/edgelesssys/contrast/pull/1374#discussion_r2043936926.

coordinator/main.go
```diff
@@ -70,6 +71,13 @@ func run() (retErr error) {
 
 	logger.Info("Coordinator started")
 
+	// If we don't have a config then we are running as an init container and not
+	// as a sidecar container remove the default deny rule and return early.
+	logger.Info("removing default deny rule")
+	if err := defaultdeny.RemoveDefaultDenyRule(logger); err != nil {
+		return fmt.Errorf("failed to remove default deny rule: %w", err)
```
burgerdev, line 78:
```suggestion
		return fmt.Errorf("removing default deny rule: %w", err)
```

docs/docs/components/service-mesh.md
```diff
@@ -14,6 +14,17 @@ all workloads which have a specified configuration. The service mesh container f
 sets up `iptables` rules based on its configuration and then starts
 [Envoy](https://www.envoyproxy.io/) for TLS origination and termination.
 
+## Service mesh startup enforcement
+Since Contrast doesn't yet enforce the order in which the containers are started
+(see [Limitations](../features-limitations.md)), we deny all incoming connections
+until the service mesh is fully configured.
+A systemd unit inside the podVM creates this deny rule.
+The kata-agent systemd unit requires that this unit successfully runs and exists.
```
burgerdev, line 22:
```suggestion
The kata-agent systemd unit requires that this unit successfully runs and exits.
```
?

docs/docs/components/service-mesh.md
```diff
@@ -14,6 +14,17 @@ all workloads which have a specified configuration. The service mesh container f
...
+Therefore, the deny rule is in place before any containers can be started.
+
+If the user specifies no service mesh annotations, the Initializer takes care
+of removing the deny rule.
```
burgerdev, line 26: I just realized that this is a mildly breaking change, since you can't run containers without an initializer anymore (typical debug workflow for me). This is acceptable, but should be pointed out in release notes, I suppose.

initializer/main.go
```diff
@@ -75,6 +76,15 @@ func run(cmd *cobra.Command, _ []string) (retErr error) {
 
 	log.Info("Initializer started")
 
+	// If we don't have a config then we are running as an init container and not
+	// as a sidecar container remove the default deny rule and return early.
```
burgerdev, line 80: I believe this comment dates to a prior version of this PR.

coordinator/main.go
```diff
@@ -70,6 +71,13 @@ func run() (retErr error) {
 
 	logger.Info("Coordinator started")
 
+	// If we don't have a config then we are running as an init container and not
+	// as a sidecar container remove the default deny rule and return early.
```
burgerdev, line 75: Does not apply to Coordinator.

packages/by-name/microsoft/kata-image/package.nix
```diff
@@ -57,6 +57,32 @@ let
...
+
+      # create a "RequiredBy" dependency where the kata-agent.service requires
+      # the deny-incoming-traffic.service to be started and successfully exited
+      # before the kata-agent.service can be started
+      # Usually this is done via the "RequiredBy" directive under the [Install]
+      # section of the systemd unit file, but the filesystem is read only
+      # and systemd wants the create the following symlink.
+      mkdir -p /build/rootfs/etc/systemd/system/kata-agent.service.requires/
+      (cd /build/rootfs/etc/systemd/system/kata-agent.service.requires/ && \
+        ln -s ../../../../../usr/lib/systemd/system/deny-incoming-traffic.service deny-incoming-traffic.service)
```
burgerdev, line 83: You should not need to `cd` for this - 

```suggestion
      ln -s ../../../../../usr/lib/systemd/system/deny-incoming-traffic.service /build/rootfs/etc/systemd/system/kata-agent.service.requires/deny-incoming-traffic.service
```

docs/docs/components/service-mesh.md
3u13r: Note that the containers will still start but any readiness probe/incoming network will fail. But for your debug workflow you should be able to manually remove them via execing into the container. Though changing you debug container to just delete the rule should be easiest.

burgerdev: Yeah, it's niche enough and workarounds exist.

burgerdev (approved):

PR #1377, opened 2025-04-14
attestation: get product from report if available

thomasten:
v3 Reports don't have the product set in the attestation. Prefer product from report and fall back to attestation if needed.

### DIFF
--- a/internal/attestation/snp/validator.go
+++ b/internal/attestation/snp/validator.go
@@ -156,7 +156,14 @@ func addCRLtoVerifyOptions(attestationData *sevsnp.Attestation, verifyOpts *veri
 		return fmt.Errorf("could not parse CRL from attestation data: %w", err)
 	}
 
-	productLine := kds.ProductLine(attestationData.GetProduct())
+	// Attestation with v3 report: FMS is set in the report; Product is nil
+	// Attestation with v2 report: FMS is 0; Product is set
+	var productLine string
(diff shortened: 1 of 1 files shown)


internal/attestation/snp/validator.go
```diff
@@ -156,7 +156,14 @@ func addCRLtoVerifyOptions(attestationData *sevsnp.Attestation, verifyOpts *veri
 		return fmt.Errorf("could not parse CRL from attestation data: %w", err)
 	}
 
-	productLine := kds.ProductLine(attestationData.GetProduct())
+	// Attestation with v3 report: FMS is set in the report; Product is nil
+	// Attestation with v2 report: FMS is 0; Product is set
```
3u13r, line 160: See: https://github.com/edgelesssys/contrast/pull/1377#issuecomment-2801358817

3u13r (changes requested):

internal/attestation/snp/validator.go
thomasten: I observed this. `getAttestationWithoutCertChain` is not called, but `getAttestationFromKdsOrExtendedReport`, which leads to https://github.com/edgelesssys/go-sev-guest/blob/deaf0dfccf3c9643088729e6d51af10000b34ba3/verify/verify.go#L901-L903

3u13r (approved):
LGTM

edgelessci:
Successfully created backport PR for `release/v1.7`:
- #1381

PR #1380, opened 2025-04-14
tools/bm-maintenance: add nix garbage collection cronjob

davidweisse:
This periodically runs `nix store gc` on the CI nodes with a Kubernetes cronjob. This was previously configured in the crontab on the node itself.


tools/bm-maintenance/deployment_tdx_snp.yml
```diff
@@ -82,3 +82,25 @@ spec:
...
+          hostPID: true
+          containers:
+          - name: nix-gc
+            image: alpine@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
+            securityContext:
+              privileged: true
+            args:
+            - '/bin/sh'
+            - '-c'
+            - 'nsenter -t 1 -m -- /root/.nix-profile/bin/nix store gc'
```
3u13r, line 105: nit: I think we can can use the long flags here, to that it is a bit easier to read
```suggestion
            - 'nsenter -target 1 -mount -- /root/.nix-profile/bin/nix store gc'
```

tools/bm-maintenance/deployment_tdx_snp.yml
```diff
@@ -82,3 +82,25 @@ spec:
...
+spec:
+  schedule: "0 0 */7 * *" # 0:00 every 7 days
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          hostPID: true
+          containers:
+          - name: nix-gc
+            image: alpine@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
```
3u13r, line 99: I'd be more inclined to extend the existing `ghcr.io/edgelesssys/contrast/cleanup-bare-metal` image. It's likely not much overhead since the image can be shared. I think it's better so we have better control over updates of the packages inside the images. Though then we'd also need a version tag on them, which I currently find superfluous

3u13r (changes requested):

tools/bm-maintenance/deployment_tdx_snp.yml
davidweisse: Is it not enough if we push the cleanup image in the step before, and use the latest reference in the deployment, as we are doing with the cleanup job?

3u13r: I think as long as you don't specify `ImagePullPolicy: Always` you will get the `latest` image that is present on the bare-metal host: https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy and always pulling strains the GitHub container registry quite a lot. Though after some research I can't tell you the exact limit: https://github.com/github/docs/issues/24504#issuecomment-1498367223

3u13r: Ah then I'd suggest to do that (use the cleanup-bare-metal container with imagePullPolicy: Always). Though I'm fine with any solution as to not block this PR any further.

3u13r (approved):
LGTM

PR #1382, opened 2025-04-14
treefmt: restrict enterprise tag usage

katexochen:


packages/scripts.nix
```diff
@@ -572,4 +572,23 @@
...
+  lint-buildtags = writeShellApplication {
+    name = "lint-enterprise-buildtags";
+    runtimeInputs = with pkgs; [ gnugrep ];
+    text = ''
+      IFS=',' read -r -a tags <<< "$1"
+      shift
+      exitcode=0
+      for tag in "''${tags[@]}"; do
+        for file in "$@"; do
+          if grep -q "//go:build $tag" "$file"; then
```
burgerdev, line 585: This does not catch, e.g.,

```go
//go:build e2e && enterprise
```

Maybe a bit drastic, but we could just match `-E ://go:build.*$tag"` (and assume that build tags are regex-safe).

burgerdev (approved):

PR #1387, opened 2025-04-15
ci: fix enabled linters

burgerdev:
Adding a rules section to the revive plugin disables all rules. Since we don't care about the default package-comment rule right now, I'm explicitly enabling all other recommended linters in this PR. Also includes a linter fix for #1361.


katexochen (approved):
thanks!

PR #1389, opened 2025-04-15
sdk: take cache as input for ValidatorsFromManifest

burgerdev:
This allows reusing an existing cache, which will be used from the enterprise recovery code that verifies the peer coordinator. 

I'm leaving the `GetCoordinatorState` signature for now because we have a rework of the `CachedHTTPSGetter` scheduled, which would require changing the signature once more.

### DIFF
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -6,12 +6,16 @@ package cmd
 import (
 	"context"
 	_ "embed"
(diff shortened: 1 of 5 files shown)


derpsteb (approved):
As Markus also noted I don't see this breaking anything in Continuum.

PR #1390, opened 2025-04-15
ci: use the right GPU runner for manual tests

burgerdev:
* [ ] [manual test](https://github.com/edgelesssys/contrast/actions/runs/14473919357)


3u13r (approved):
I assume the test failure is what you are trying to debug right now. LGTM, since the test ran on discovery.

PR #1393, opened 2025-04-16
transiten

jmxnzo:
kuberesource: activate mTLS for entire Vault


internal/kuberesource/sets.go
```diff
@@ -698,3 +698,99 @@ func GPU() []any {
...
+				),
+			),
+		)
+	vaultService := ServiceForStatefulSet(vaultSfSets).
+		WithAnnotations(map[string]string{exposeServiceAnnotation: "true"})
+
+	configMap := applycorev1.ConfigMap("vault-config", namespace).WithData(
+		map[string]string{
+			"config.hcl": `ui = true
+api_addr     = "https://vault:8200"
```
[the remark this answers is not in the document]
jmxnzo: As outlined in the [OpenBao docs](https://openbao.org/docs/configuration/#high-availability-parameters), env variable `VAULT_API_ADDR` maps to the `api_addr` field in config. For the [listener config](https://openbao.org/docs/configuration/listener/tcp/#tcp-listener-parameters) they do not map to any of the environmental variables. As all of the used environmental variables are on the client side of our OpenBao container for the `bao operator init` (see [env variable CLI](https://openbao.org/docs/commands/#environment-variables)), i suggest keeping this separation and setting `api_addr` in the config.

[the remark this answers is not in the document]
burgerdev: Maybe setting them in the script itself, after the server starts, could prevent this type of confusion.

internal/kuberesource/sets.go
```diff
@@ -698,3 +698,100 @@ func GPU() []any {
...
+			WithServiceName("vault").
+			WithTemplate(
+				PodTemplateSpec().
+					WithLabels(map[string]string{"app.kubernetes.io/name": "vault"}).
+					WithSpec(PodSpec().
+						WithContainers(
+							Container().
+								WithName("openbao").
+								WithImage("quay.io/openbao/openbao:2.2.0@sha256:19612d67a4a95d05a7b77c6ebc6c2ac5dac67a8712d8df2e4c31ad28bee7edaa").
+								WithCommand("/bin/sh", "-c", startVaultCmd).
```
burgerdev, line 733:
```suggestion
								WithCommand("/bin/sh", "-ec", startVaultCmd).
```

internal/kuberesource/sets.go
```diff
@@ -698,3 +698,100 @@ func GPU() []any {
...
+func Vault(namespace string) []any {
+	startVaultCmd := `bao server -config=/vault/config/config.hcl &
+# Wait until Vault is ready
+echo "Waiting for Vault to become available..."
+until wget --server-response --no-check-certificate --spider -q -O- https://127.0.0.1:8200 2>&1 | grep -q "HTTP/1.1 200"; do
+    sleep 2
+done
+
+echo "Vault is now ready. Initializing..."
+bao operator init
```
burgerdev, line 712: Sorry, I just realized it now: why do we always init here? Shouldn't the Bao instance recover itself, at least after restarting?

jmxnzo: Thanks for pointing out, that's right.

e2e/vaultstatefulset/vaultstatefulset_test.go
```diff
@@ -0,0 +1,296 @@
...
+	t.Run("transit engine api auto-unseals Vault", func(t *testing.T) {
+		require := require.New(t)
+		ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(2*60*time.Second))
+		defer cancel()
+
+		vaultLogs, stdErr, err := ct.Kubeclient.ExecContainer(ctx, ct.Namespace, pods[0].Name, "openbao-server", []string{"cat", loggingPath})
+		require.NoError(err, "stdout: %s, stderr: %s", vaultLogs, stdErr)
+
+		coordinatorLogs, err := ct.Kubeclient.GetContainerLogs(ctx, ct.Namespace, "StatefulSet", "coordinator", "coordinator")
+		require.NoError(err)
```
burgerdev, line 87: I don't think we should be looking at the Coordinator logs here, but rather focus on the result: does the Vault eventually become unsealed? It should be possible to use https://developer.hashicorp.com/vault/docs/commands/status or an equivalent HTTP request.

e2e/vaultstatefulset/vaultstatefulset_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+		stdOut, stdErr, err := ct.Kubeclient.ExecContainer(
+			ctx,
+			ct.Namespace,
+			pods[0].Name,
+			"openbao-client",
+			[]string{
+				"sh",
+				"-c",
+				fmt.Sprintf("export VAULT_TOKEN=%s && bao secrets enable -path=mykv kv", token),
```
burgerdev, line 115: Nit: you can simplify these scripts like this:

```suggestion
				fmt.Sprintf("VAULT_TOKEN=%s bao secrets enable -path=mykv kv", token),
```

burgerdev: Considering that you're executing such scripts 4 times, defining a function that just takes the `bao` command might make sense.

e2e/vaultstatefulset/vaultstatefulset_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+		require.NoError(err, "stdout: %s, stderr: %s", stdOut, stdErr)
+		t.Logf("%s", stdOut)
+	})
+
+	// Restarts the Vault deployment and verifies it auto-unseals using the transit engine API.
+	// Additionally confirms that the previously written secret still exists, ensuring data persistence.
+	t.Run("Vault auto-unseals after restart and secrets are persistent", func(t *testing.T) {
+		require := require.New(t)
+		ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(60*time.Second))
```
burgerdev, line 162:
```suggestion
		ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(time.Minute))
```

internal/kuberesource/sets.go
```diff
@@ -698,3 +698,138 @@ func GPU() []any {
...
+    sleep 2
+done
+
+
+# Get init status using wget
+echo "Vault is responding. Checking initialization status..."
+init_check=$(wget --no-check-certificate -q -O - https://vault:8200/v1/sys/init)
+
+# Check if Vault is uninitialized
+if echo "$init_check" | grep -q '"initialized"[ ]*:[ ]*false'; then
```
burgerdev, line 720: This might actually be easier to understand if inlined:

```sh
if wget --no-check-certificate -q -O - https://vault:8200/v1/sys/init | grep -q '"initialized"[ ]*:[ ]*false'; then
```

internal/kuberesource/sets.go
```diff
@@ -698,3 +698,138 @@ func GPU() []any {
...
+								WithImage("quay.io/openbao/openbao:2.2.0@sha256:19612d67a4a95d05a7b77c6ebc6c2ac5dac67a8712d8df2e4c31ad28bee7edaa").
+								WithCommand("bao", "server", "-config=/vault/config/config.hcl", "-log-file=/vault/data/openbao.log").
+								/*These environmental variables are required for the Vault server instance:
+								- VAULT_CA_CERT accept the transit engine API to be set-up for auto-unsealing
+								- VAULT_CLIENT_CERT to be accepted by the transit engine API as a valid communication peer
+								- VAULT_CLIENT_KEY is the corresponding private key
+								*/
+								WithEnv(EnvVar().WithName("VAULT_CACERT").WithValue("/contrast/tls-config/mesh-ca.pem")).
+								WithEnv(EnvVar().WithName("VAULT_CLIENT_CERT").WithValue("/contrast/tls-config/certChain.pem")).
+								WithEnv(EnvVar().WithName("VAULT_CLIENT_KEY").WithValue("/contrast/tls-config/key.pem")).
```
burgerdev, line 756: I still believe these belong into the `seal transit` config, as shown in https://developer.hashicorp.com/vault/docs/configuration/seal/transit#transit-example.

burgerdev (approved):
lgtm, thanks!

jmxnzo:
I noticed some unintended behavior with the PVC mounting, backing up the Vault storage on our TDX runners. I'm currently examining the occurrence of the issue. So i'll leave this open until all platforms do work as intended.

burgerdev (commented):
Now that the tests pass, please re-squash into

* changes to kuberesource
* coordinator port exposure
* e2e test

burgerdev (approved):

PR #1395, opened 2025-04-16
coordinator: fix TestRecovery overrides

burgerdev:
While refactoring the userapi, I found out that the overrides for TestRecovery requests are broken. The intention was to override the request data with what's in the test case struct, but this did not work as expected for empty seed and empty salt. We did not detect this because the error message was unchecked and we got the expected `InvalidArgument`, but from the latest transition signature check and not from the seedengine constructor.

### DIFF
--- a/coordinator/internal/authority/userapi_test.go
+++ b/coordinator/internal/authority/userapi_test.go
@@ -253,32 +253,35 @@ func TestRecovery(t *testing.T) {
 	var seed [32]byte
 	var salt [32]byte
 	testCases := []struct {
-		name     string
-		seed     []byte
-		salt     []byte
-		wantCode codes.Code
+		name        string
+		seed        *[]byte
+		salt        *[]byte
+		wantCode    codes.Code
+		wantMessage string
 	}{
 		{
-			name:     "empty seed",
-			salt:     salt[:],
-			wantCode: codes.InvalidArgument,
+			name:        "empty seed",
+			seed:        toPtr[[]byte](nil),
+			wantCode:    codes.InvalidArgument,
+			wantMessage: "seed must be",
 		},
 		{
-			name:     "empty salt",
-			seed:     seed[:],
-			wantCode: codes.InvalidArgument,
+			name:        "empty salt",
+			salt:        toPtr[[]byte](nil),
+			wantCode:    codes.InvalidArgument,
+			wantMessage: "salt must be",
 		},
 		{
-			name:     "short seed",
-			seed:     seed[:16],
-			salt:     salt[:],
-			wantCode: codes.InvalidArgument,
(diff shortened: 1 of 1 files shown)


coordinator/internal/authority/userapi_test.go
```diff
@@ -253,32 +253,35 @@ func TestRecovery(t *testing.T) {
...
+		salt        *[]byte
+		wantCode    codes.Code
+		wantMessage string
 	}{
 		{
-			name:     "empty seed",
-			salt:     salt[:],
-			wantCode: codes.InvalidArgument,
+			name:        "empty seed",
+			seed:        toPtr[[]byte](nil),
```
3u13r, line 264: The `toPtr` is superfluous here, since type of the member variable should make nil already to a typed nil.

coordinator/internal/authority/userapi_test.go
```diff
@@ -253,32 +253,35 @@ func TestRecovery(t *testing.T) {
 	var seed [32]byte
 	var salt [32]byte
 	testCases := []struct {
-		name     string
-		seed     []byte
-		salt     []byte
-		wantCode codes.Code
+		name        string
+		seed        *[]byte
+		salt        *[]byte
```
3u13r, line 258: Since `nil` is a valid slice, why not use `[]byte` as the type. The default initialization is also nil:
```go
package main

import "fmt"

func main() {
	var s []int

	fmt.Println(s==nil) // prints true
}
```

3u13r (changes requested):

coordinator/internal/authority/userapi_test.go
burgerdev: The idea is to decide whether to override the value in the test body based on this pointer being set or nil, and I want to be able to override to nil.

coordinator/internal/authority/userapi_test.go
burgerdev: There's a difference: I'm creating a non-nil pointer to a nil slice here, which makes the test body override the seed with a nil slice.

3u13r: Makes sense, but I think there's still another way of testing the emptiness test cases with initialized 0 length slices and then using the nil slice to decide whether to replace or not. If you want to explicitly test nil inputs then I'd create different test cases (one with nil and one with zero length slices), if not then I'd go with just zero length slices.

burgerdev: I explicitly want to test with nil slices, there is a test with a short slice already.

3u13r: Alright, good point

3u13r (approved):

PR #1398, opened 2025-04-17
e2e: introduce snp atls test

3u13r:
What we are trying to test:
* Do the grpc/net dailers correctly error when there is a attestation problem?
* Which mismatches in PlatformInfo return an error? One can argue that this is more of a test of the go-sev-guest library, but I think given the [recent changes](https://github.com/google/go-sev-guest/pull/161) it's also a nice place to test those
* Does the IDKeyDigest derivation/check fails if the policy or the measurement changes?

Alternative approach to also test the validator derivation from the manifest (i.e., call `sdk.ValidatorsFromManifest` in the test). In order to test the failure cases (i.e. modified MEASUREMENT), we need access to the underlying validateOpts in those `atls.Validators`, by:
* Casting the Validators form the Manifest back to `snp.Validator`.
* Exporting the `validateOpts` and `verifyOpts` fields in `snp.Validator`

### DIFF
--- a/e2e/atls/atls_test.go
+++ b/e2e/atls/atls_test.go
@@ -1,0 +1,188 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//go:build e2e
+
+package atls
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha512"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log/slog"
+	"os"
+	"path"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/edgelesssys/contrast/e2e/internal/contrasttest"
+	"github.com/edgelesssys/contrast/internal/atls"
+	"github.com/edgelesssys/contrast/internal/attestation/certcache"
+	"github.com/edgelesssys/contrast/internal/attestation/snp"
+	"github.com/edgelesssys/contrast/internal/grpc/dialer"
+	"github.com/edgelesssys/contrast/internal/idblock"
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/memstore"
+	"github.com/edgelesssys/contrast/internal/platforms"
+	"github.com/edgelesssys/contrast/internal/userapi"
+	"github.com/google/go-sev-guest/abi"
+	"github.com/stretchr/testify/require"
+)
+
+// TestATLSSNP runs e2e tests for the atls layer.
+func SNPValidatorsTest(t *testing.T) {
+	platform, err := platforms.FromString(contrasttest.Flags.PlatformStr)
+	require.NoError(t, err)
+
+	if platform != platforms.AKSCloudHypervisorSNP &&
+		platform != platforms.K3sQEMUSNP &&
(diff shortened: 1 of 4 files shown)


e2e/atls/snp/snp_test.go
```diff
@@ -0,0 +1,191 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//go:build e2e
+
+package snp
```
katexochen, line 6: imo this shouldn't be a SNP specific test package. Instead you could just skip the snp test on platforms that don't support snp and make it a generic atls test.

3u13r: Sure, I also wondered about how we might want to extend this to TDX in the future. I originally thought about 2 packages, but your suggestion makes sense.

katexochen (approved):

e2e/atls/atls_test.go
```diff
@@ -0,0 +1,191 @@
...
+		},
+	}
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			require.NoError(t, ct.Kubeclient.WithForwardedPort(ctx, ct.Namespace, "port-forwarder-coordinator", "1313", func(addr string) error {
+				assert := require.New(t)
+				require := require.New(t)
+				// Go never uses a proxy for connections to localhost. To enable proxy tests, we
+				// replace localhost with 0.0.0.0, which can be used as localhost on Linux and BSD.
+				addr = strings.Replace(addr, "localhost", "0.0.0.0", 1)
```
burgerdev, line 153:
```suggestion
```
This should only be needed for the HTTP_PROXY regression test.

e2e/atls/atls_test.go
```diff
@@ -0,0 +1,191 @@
...
+				for i, opt := range opts {
+					opt.ValidateOpts.HostData = coordPolicyHashBytes
+					assert.NoError(tc.manifestModifyFunc(&opt))
+					name := fmt.Sprintf("snp-%d-%s", i, strings.TrimPrefix(opt.VerifyOpts.Product.Name.String(), "SEV_PRODUCT_"))
+					validators = append(validators, snp.NewValidator(opt.VerifyOpts, opt.ValidateOpts, logger.WithGroup("validator"), name))
+				}
+
+				dialer := dialer.New(atls.NoIssuer, validators, atls.NoMetrics, nil, logger)
+				conn, err := dialer.Dial(ctx, addr)
+				assert.NoError(err, "dialing coordinator")
```
burgerdev, line 169:
```suggestion
				require.NoError(err, "dialing coordinator")
```
It does not make sense to continue if dialing failed, does it?

burgerdev (approved):

e2e/atls/atls_test.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/memstore"
+	"github.com/edgelesssys/contrast/internal/platforms"
+	"github.com/edgelesssys/contrast/internal/userapi"
+	"github.com/google/go-sev-guest/abi"
+	"github.com/stretchr/testify/require"
+)
+
+// TestATLSSNP runs e2e tests for the atls layer.
+func SNPValidatorsTest(t *testing.T) {
```
katexochen, line 38: This should be TestSNPValidators, and the comment above must be updated

PR #1400, opened 2025-04-17
coordinator: move userapi into own package

burgerdev:
This PR refactors the authority package, moving out the userapi and changing the authority API to allow manifest manipulation from other packages (i.e., enterprise). A second commit adds diagrams and documentation to the packages,  which should facilitate high-level orientation on the package layout and interactions.

---

**Notes to reviewers**: 

1. Please review the doc commit first, as it describes the target architecture after the refactoring. Rich-diff-mode should render the mermaid diagrams.
2. At this stage, I don't consider the authority API final, but rather a draft of how it could be restructured. What I found tricky in particular is where to draw the line of responsibilities for managing history. I had several drafts, one tried to avoid managing history in the userapi as far as possible, one where the userapi manages everything except the LatestTransition, and the current one which is a mix. We could also consider hiding the history altogether and adding the necessary history manipulation functions to the authority. I don't have a clear favourite here. 
3. The authority package is very thin on tests now. However, I tried to keep the existing tests for the Coordinator API (`GetManifests|SetManifest|Recover`) intact while moving, so we at least don't reduce test coverage with this PR and don't change the public API. Once we decided on the final authority API, I can add tests for it either here or in a follow-up.

### DIFF
--- a/coordinator/README.md
+++ b/coordinator/README.md
@@ -1,0 +1,40 @@
+# Contrast Coordinator
+
+The Coordinator is the central service in a Contrast deployment.
+It exposes an API to set manifests and to get the history of manifests, and it verifies workloads according to the current manifest.
+The below diagram shows how the Coordinator packages interact in the enterprise version.
+
+```mermaid
+graph TB
+    subgraph Coordinator
+            userapi -->|update state in<br/>reset state in| stateguard
+            meshapi -->|get state from| stateguard
+            stateguard -->|manage latest state in| history
+            history -->|store data by hash<br/>atomically update state| store
+    end
+
+    subgraph Workload
+        initializer -->|NewMeshCert| meshapi
+        main
+    end
+
+    subgraph Coordinator2
+        enterprise2(enterprise/recovery)
+        stateguard2(stateguard)
+        enterprise2 -->|reset state in| stateguard2
+    end
+
+    subgraph Kubernetes
+        Workload
+        Coordinator
+        Coordinator2
+        ConfigMap
+    end
+
+
+    store -->|store data in<br/>receive data from| ConfigMap
+
+    user -->|GetManifests<br/>SetManifest<br>Recover| userapi
+
+    enterprise2 -->|Recover| meshapi
+```
--- a/coordinator/internal/authority/authority.go
+++ b/coordinator/internal/authority/authority.go
@@ -408,11 +408,13 @@ type State struct {
 	stale atomic.Bool
 }
 
-// NewState constructs a new State object.
+// NewStateForTest constructs a new State object.
 //
-// This function is intended for other packages that work on State objects. It does not produce a
-// State that is valid for this package.
-func NewState(seedEngine *seedengine.SeedEngine, manifest *manifest.Manifest, manifestBytes []byte, ca *ca.CA) *State {
+// This function is intended for testing packages that work on State objects. It fills the fields
+// that are observable outside this package, but does not manage the fields only relevant for this
+// package. State objects created with this function can't be used as arguments to this package's
+// public API functions.
+func NewStateForTest(seedEngine *seedengine.SeedEngine, manifest *manifest.Manifest, manifestBytes []byte, ca *ca.CA) *State {
 	return &State{
 		seedEngine:    seedEngine,
 		manifest:      manifest,
--- a/coordinator/internal/authority/authority.go
+++ b/coordinator/internal/authority/authority.go
@@ -1,14 +1,34 @@
 // Copyright 2024 Edgeless Systems GmbH
 // SPDX-License-Identifier: AGPL-3.0-only
 
+// Package authority guards the current state of the Coordinator.
+//
+// The authority.Authority struct is the single source of truth for the currently enforced manifest
+// and related data (secrets, metrics). It manages an authority.State object that can be handed
+// out to other packages that need to operate on a current state. A State object does not change
+// and can be safely used as long as necessary, but callers need to ensure that they consistently
+// operate on a single state. For example, gRPC calls authenticated with the Authority.Credentials
+// must only work with the state object added to the request context.
+// The Authority exposes methods to manipulate the state by either updating to a new manifest or
+// resetting to a manifest that was once active. To make these manipulations consistent, the
+// method signatures resemble a compare-and-swap operation: callers need to first get the current
+// state, then decide based on that state whether/how to change the state, and finally call the
+// state manipulation method with the old state and the desired next state.
+// For scenarios with multiple Coordinator instances, the Authority exposes a watcher routine that
+// keeps track of changes to the persistency and marks the current state as stale if something else
+// persisted changes. The state is also marked as stale if an inconsistency is discovered during
+// state manipulation. A stale state can still be operated upon, since it was valid at some point,
+// but recovery needs to happen before this Coordinator instance can update the state again.
 package authority
 
 import (
 	"context"
+	"crypto/ecdsa"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
+	"slices"
 	"sync/atomic"
 	"time"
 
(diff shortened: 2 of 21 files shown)


coordinator/README.md
```diff
@@ -0,0 +1,33 @@
...
+    subgraph Kubernetes
+        Workload
+        Coordinator
+        ConfigMap
+    end
+
+
+    store -->|store data in<br/>receive data from| ConfigMap
+
+    user -->|GetManifests<br/>SetManifest<br>Recover| userapi
```
katexochen, line 32: maybe we could add a connection from another coordinator to enterprise/recovery, just so it is easier to understand what is happening there.

coordinator/internal/authority/authority.go
```diff
@@ -1,29 +1,57 @@
 // Copyright 2024 Edgeless Systems GmbH
 // SPDX-License-Identifier: AGPL-3.0-only
 
+// Package authority guards the current state of the Coordinator.
```
katexochen, line 4: nit: Maybe we could rename to `stateguard`? It was called authority in the beginning because it issued the certs, not sure that's still its main function.

coordinator/internal/authority/authority.go
```diff
@@ -130,16 +156,23 @@ func (m *Authority) WatchHistory(ctx context.Context) error {
...
+//
+// This function is intended to be called in recovery scenarios, where a latest state exists but
+// can't be verified without the necessary secrets. The state is recovered by first loading the
+// state insecurely, then passing the corresponding manifest to the caller-provided authorization
+// function, and finally using the key material to verify the loaded state.
+//
+// The authorizeSeedSource function must check that the source of the seed (a user or a peer
+// Coordinator) are authorized to hold the secret seed, according to the manifest that's being
+// recovered to. See RFC 010 for more details on the security considerations for handling seeds.
+func (m *Authority) ResetState(oldState *State, authorizeSeedSource func(*manifest.Manifest) (*seedengine.SeedEngine, *ecdsa.PrivateKey, error)) (*State, error) {
```
katexochen, line 169: maybe use an interface instead of a function for `authorizedSeedSource`.

coordinator/internal/userapi/userapi.go
```diff
@@ -0,0 +1,279 @@
...
+type stateUpdateAuthorizer = func(*manifest.Manifest) (*seedengine.SeedEngine, *ecdsa.PrivateKey, error)
+
+// auth is the public API of authority.Authority.
+type auth interface {
+	// GetState returns the current state. If the error is nil, the state must be set.
+	GetState() (*authority.State, error)
+	// GetHistory returns a slice of manifests and a map of policies referenced in the manifests.
+	GetHistory() ([][]byte, map[manifest.HexString][]byte, error)
+	// UpdateState advances the state to the given manifest and policies.
+	UpdateState(*authority.State, *seedengine.SeedEngine, []byte, [][]byte) (*authority.State, error)
```
katexochen, line 51: While parameter names can be left out, I think it is really helpful to add them in the interface definition, too.

coordinator/README.md
burgerdev: Added a second coordinator.

coordinator/internal/authority/authority.go
burgerdev: Renamed

coordinator/internal/authority/authority.go
burgerdev: Converted to interface

coordinator/internal/userapi/userapi.go
burgerdev: Named parameters

coordinator/internal/authority/authority.go
```diff
@@ -177,21 +220,180 @@ func (m *Authority) fetchState(se *seedengine.SeedEngine) (*State, error) {
...
+		return nextState, nil
+	}
+	m.metrics.manifestGeneration.Set(float64(nextState.generation))
+	return nextState, nil
+}
+
+// GetHistory returns a list of manifests, the current manifest being last, and the policies
+// referenced in at least one of the manifests.
+// TODO should return a policy map.
+func (m *Authority) GetHistory() ([][]byte, map[manifest.HexString][]byte, error) {
```
burgerdev: TODO is outdated

coordinator/internal/authority/userapi.go
```diff
@@ -1,6 +1,17 @@
 // Copyright 2024 Edgeless Systems GmbH
 // SPDX-License-Identifier: AGPL-3.0-only
 
+// Package userapi implements the userapi.UserAPI gRPC service.
```
burgerdev: Wrong commit

coordinator/internal/authority/userapi.go
```diff
@@ -28,14 +38,34 @@ import (
...
 
+// Server serves the userapi.UserAPI. Servers need to be constructed with New.
+type Server struct {
+	logger *slog.Logger
+	auth   *Authority
+
+	userapi.UnimplementedUserAPIServer
+}
+
+// New constructs a new Server instance.
```
burgerdev: Comments refer to names after move in subsequent commit

coordinator/internal/userapi/userapi.go
```diff
@@ -6,13 +6,11 @@
...
 //   - authorizing users
-//   - managing content-addressed history elements (but not the LatestTransition)
 //   - creating secrets (seed and mesh CA key)
 //
-// The package has access to an authority.Authority, which is the source of truth for the current
-// state of the system, and a history.History for storing supporting content (manifests and
-// policies).
-package stateguard
+// The package has access to an stateguard.Guard, which is the source of truth for the current
+// state of the system.
```
burgerdev, line 12: Should be in the stateguard rename commit

coordinator/internal/stateguard/stateguard_test.go
```diff
@@ -33,26 +35,48 @@ contrast_coordinator_manifest_generation %d
 `
 )
 
-var keyDigest = manifest.HexString("0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef")
-
-func TestSNPValidateOpts(t *testing.T) {
+func TestMetrics(t *testing.T) {
```
katexochen, line 38: Should the test also cover case where state.CompareAndSwap in UpdateState fails?

coordinator/internal/stateguard/stateguard.go
```diff
@@ -0,0 +1,444 @@
...
+
+	// stale is set to true when we discover that this State is not current anymore.
+	// This field is only ever flipped from false to true!
+	stale atomic.Bool
+}
+
+// NewState constructs a new State object.
+//
+// This function is intended for other packages that work on State objects. It does not produce a
+// State that is valid for this package.
```
katexochen, line 415: I'm not sure I understand what this means.

coordinator/internal/userapi/userapi_test.go
```diff
@@ -354,16 +340,15 @@ func TestRecoveryFlow(t *testing.T) {
 
 	// 1. A Coordinator is created from empty state.
 
-	a := newCoordinator()
+	logger := slog.Default()
+	fs := afero.NewMemMapFs()
+	store := history.NewAferoStore(&afero.Afero{Fs: fs})
+	hist := history.NewWithStore(slog.Default(), store)
+	auth := stateguard.New(hist, prometheus.NewRegistry(), logger)
```
katexochen, line 347: I guess we could stub the guard interface here instead of using a real stateguard?

coordinator/internal/stateguard/stateguard.go
```diff
@@ -0,0 +1,445 @@
...
+	// This field is only ever flipped from false to true!
+	stale atomic.Bool
+}
+
+// NewStateForTest constructs a new State object.
+//
+// This function is intended for testing packages that work on State objects. It fills the fields
+// that are observable outside this package, but does not manage the fields only relevant for this
+// package. State objects created with this function can't be used as arguments to this package's
+// public API functions.
```
katexochen, line 416: imo it is bad practice exposing such test related functions as part of the normal package api. It hints at the fact that there should be an interface and stubbing used at the other side.

katexochen (approved):
only the points left regarding stubbing/testing, but I'm fine with merging this and addressing those points later.

PR #1401, opened 2025-04-18
runtime: remove nydus on bare metal

burgerdev:
This is a PoC of guest pull without Nydus. The Kata patch forces to enter guest pull mode, even if the snapshotter did not provide the Kata virtual volume option. The rest of this PR just removes Nydus from the runtime, to show that it actually works.


burgerdev:
Topic is taken over by @katexochen

PR #1402, opened 2025-04-22
tools/bm-maintenance: support containerd config v3 in cleanup

davidweisse:
The cleanup script would error due to an incorrect path in the containerd config. It now tries both paths and continues in case one of them errors.

### DIFF
--- a/packages/cleanup-bare-metal.sh
+++ b/packages/cleanup-bare-metal.sh
@@ -53,8 +53,12 @@ for runtimeClass in "${unusedRuntimeClasses[@]}"; do
 
   # Remove references from containerd config
   echo "Removing ${runtimeClass} from ${CONFIG} ..."
-  dasel delete --file "${CONFIG}" --indent 0 --read toml --write toml "plugins.io\.containerd\.grpc\.v1\.cri.containerd.runtimes.${runtimeClass}" 2>/dev/null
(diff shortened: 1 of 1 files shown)


packages/cleanup-bare-metal.sh
```diff
@@ -53,8 +53,11 @@ for runtimeClass in "${unusedRuntimeClasses[@]}"; do
 
   # Remove references from containerd config
   echo "Removing ${runtimeClass} from ${CONFIG} ..."
-  dasel delete --file "${CONFIG}" --indent 0 --read toml --write toml "plugins.io\.containerd\.grpc\.v1\.cri.containerd.runtimes.${runtimeClass}" 2>/dev/null
-  dasel delete --file "${CONFIG}" --indent 0 --read toml --write toml "proxy_plugins.${SNAPSHOTTER}-${runtimeClass}" 2>/dev/null
+  dasel delete --file "${CONFIG}" --indent 0 --read toml --write toml "plugins.io\.containerd\.grpc\.v1\.cri.containerd.runtimes.${runtimeClass}" 2>/dev/null || true
+  # Containerd config v3
+  dasel delete --file "${CONFIG}" --indent 0 --read toml --write toml "plugins.io\.containerd\.cri\.v1\.runtime.containerd.runtimes.${runtimeClass}" 2>/dev/null || true
```
burgerdev, line 58: I wonder how we could detect a cleanup breakage here. The `|| true` is probably right, since we want this to be idempotent, but shouldn't we at least log the errors?

burgerdev (approved):

PR #1405, opened 2025-04-23
treewide: print unified preface string on startup

katexochen:
This will help to differentiate enterprise from non-enterprise components, identify the version of running containers as well as guide users to the right place to report issues.


burgerdev (approved):
lgtm

3u13r (approved):
LGTM

PR #1408, opened 2025-04-24
e2e: fix snp atls test name

3u13r:
on TDX (which is skipped): https://github.com/edgelesssys/contrast/actions/runs/14638555551
on TDX (with 2nd commit): https://github.com/edgelesssys/contrast/actions/runs/14639561249
on SNP: https://github.com/edgelesssys/contrast/actions/runs/14638577116

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -105,4 +105,10 @@ jobs:
       - name: Cleanup
         if: always() && !inputs.skip-undeploy
         run: |
+          # Check if namespace file exists and is not empty
+          if [[ ! -f workspace/e2e.namespace ]]; then
+            echo "No namespace file found, skipping cleanup." >&2
+            exit 0
+          fi
+
           kubectl delete ns "$(cat workspace/e2e.namespace)" --timeout 10m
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -105,4 +105,10 @@ jobs:
       - name: Cleanup
         if: always() && !inputs.skip-undeploy
         run: |
+          # Check if namespace file exists and is not empty
+          if [[ ! -f workspace/e2e.namespace ]]; then
+            echo "No namespace file found, skipping cleanup."
```
burgerdev, line 110: This might obscure other issues with the run. Why do we need to run the test on runners that do not support it in the first place?

3u13r: I think there are two points here:
1. In general tests can be skipped. Should the cleanup error if no test have been run? In my opinion not.
2. The atls test is a general atls test but currently only has a snp implementation. So _in general_ the atls test can be run on TDX. It only contains one test function that skips the test on non-TDX.

katexochen, line 110:
```suggestion
            echo "No namespace file found, skipping cleanup." >&2
```

burgerdev: Ok, thanks

katexochen (approved):

PR #1409, opened 2025-04-24
docs: configure GPU operator for CC mode

burgerdev:
Our instructions were lacking the Helm value that makes cc-manager set the GPU to confidential mode. These instructions were taken from [upstream documentation](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-operator-confidential-containers.html#procedure), which is either just wrong or assumes some other component activates CC mode.

---

The other commit includes some minor fixes I found useful while reconfiguring discovery:

* Add instructions for k3s config override. We had these files on some runners in the past, but it appears that they've been lost between reinstalls. The config I'm including here
  * allow the `github` user (and others) to access `k3s.yaml` via `sudo` membership
  * disable the default storage class, since we're installing longhorn
  * increase the CRI request timeout (while not strictly needed right now, it may be for larger VMs; also, it does not hurt and documents how to do this).
* Pin the k3s version.
* Customize longhorn installation to need less _free_ space on the rootfs and to not allocate unneeded replicas.
* Use upstream Nix instead of determinate (determinate comes with a bunch of settings that I'm not sure we want, and I don't see why _not_ to use upstream).
* Configure nix for Cachix.
* Remove `hetzner-*` references where they were hindering my copy-paste routine.

### DIFF
--- a/dev-docs/e2e/bare-metal-runner.md
+++ b/dev-docs/e2e/bare-metal-runner.md
@@ -71,9 +71,22 @@ Install helm
 curl -fsL https://get.helm.sh/helm-v3.17.1-linux-amd64.tar.gz | tar -C /tmp -xz linux-amd64/helm && mv /tmp/linux-amd64/helm /usr/local/bin
 ```
 
+Add K3s configuration override
+```bash
+mkdir -p /etc/rancher/k3s
+cat > /etc/rancher/k3s/config.yaml <<EOF
+write-kubeconfig-mode: "0640"
+write-kubeconfig-group: sudo
+disable:
+  - local-storage
+kubelet-arg:
+  - "runtime-request-timeout=5m"
+EOF
+```
+
 Install K3s
 ```bash
-curl -sfL https://get.k3s.io | sh -
+curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.31.5+k3s1 sh -
 ```
 The K3s docs state:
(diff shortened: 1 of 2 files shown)


dev-docs/e2e/bare-metal-runner.md
```diff
@@ -71,9 +71,22 @@ Install helm
...
+  - local-storage
+kubelet-arg:
+  - "runtime-request-timeout=5m"
+EOF
+```
+
 Install K3s
 ```bash
-curl -sfL https://get.k3s.io | sh -
+curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.31.5+k3s1 sh -
```
katexochen, line 89: Should we also mention that version in the user docs?

dev-docs/e2e/bare-metal-runner.md
```diff
@@ -129,15 +144,25 @@ useradd -s /bin/bash -m -G sudo,docker github
...
-sudo chown -c github /home/github/.kube/config
 ```
 
-The CI jobs build things with nix, therefore install it with the
-Determinate Nix Installer:
-https://github.com/DeterminateSystems/nix-installer?tab=readme-ov-file#determinate-nix-installer
+The CI jobs build things with nix, therefore install it following the official instructions:
+https://nixos.org/download/#nix-install-linux.
 
+Customize the Nix configuration for flakes, the GitHub runner and Cachix:
```
katexochen, line 153: Not easy to understand how this supports cachix.

katexochen (approved):

katexochen (approved):

PR #1430, opened 2025-05-02
cli: sensible error message if coordinator doesn't meet expectations

miampf:


katexochen (dismissed):
How do we know that's the reason the set failed?

miampf:
> How do we know that's the reason the set failed?

I got this info from @burgerdev, but maybe I understood something wrong?

cli/cmd/set.go
```diff
@@ -140,7 +141,11 @@ func runSet(cmd *cobra.Command, args []string) error {
 				fmt.Fprintln(cmd.OutOrStdout(), msg)
 			}
 		}
-		return fmt.Errorf("setting manifest: %w", err)
+		additionalHelp := ""
+		if strings.Contains(err.Error(), "validating report data") {
```
burgerdev, line 145: This match may be too broad: invalid fields in report data might point to insecure hardware config, wrong VM image, etc.

miampf: Would also matching on `MR_CONFIG_ID` be specific enough?

burgerdev: We should match these cases:

https://github.com/google/go-tdx-guest/blob/9efd53b4a100e467dfd00c79fbb3de19f71b1ba4/validate/validate_test.go#L238

https://github.com/google/go-sev-guest/blob/8023c5f041bd9a64c7fb522288d166c0ea112a95/validate/validate_test.go#L475

(with field `HOST_DATA`)

burgerdev (approved):

PR #1434, opened 2025-05-08
runtime: remove nydus-snapshotter, force guest pull via config

katexochen:
This PR removes the need for nydus-snapshotter on all k3s / bare metal platforms. Instead we configure the kata runtime on the host to always add annotation for guest pull. As nydus-snapshotter has been a source of unreliability and ongoing issues, this is a great stability improvement for users on bare metal.


burgerdev (approved):
Amazing!

PR #1437, opened 2025-05-12
kata: remove unneeded tools patch

katexochen:
We don't use the update build script anymore, so there the patch isn't needed anymore.


burgerdev (approved):
thanks

PR #1439, opened 2025-05-12
attestation: add cached HTTPSGetter to TDX validator

davidweisse:
The HTTPSGetter from `go-tdx-guest` is a superset of the implementation from `go-sev-guest` so we can change the implementation of the `certcache` to use the TDX variant and use an adapter for SNP. We can then pass the `kdsGetter` to the TDX validator (as we are currently doing for the SNP validator) and utilize the cache functionality there.


internal/attestation/certcache/cached_client.go
```diff
@@ -5,16 +5,20 @@ package certcache
...
-	"github.com/google/go-sev-guest/verify/trust"
+	"github.com/google/go-tdx-guest/verify/trust"
 	"k8s.io/utils/clock"
 	testingclock "k8s.io/utils/clock/testing"
 )
 
-var crlURL = regexp.MustCompile(`^https://kdsintf\.amd\.com/vcek/v1/[A-Za-z]*/crl$`)
+var (
+	snpCrlURL = regexp.MustCompile(`^https://kdsintf\.amd\.com/vcek/v1/[A-Za-z]*/crl$`)
+	tdxCrlURL = regexp.MustCompile(`^https://api\.trustedservices\.intel\.com/sgx/certification/v4/pckcrl?ca=(platform|processor)&encoding=der$`)
```
burgerdev, line 20: The question mark needs escaping.

internal/attestation/certcache/cached_client_test.go
```diff
@@ -142,27 +164,30 @@ func TestContextCancellation(t *testing.T) {
...
 	}
 
-	getter := &CachedHTTPSGetter{
+	getter := (&CachedHTTPSGetter{
 		ContextHTTPSGetter: fakeGetter,
 		gcTicker:           ticker,
 		cache:              memstore.New[string, []byte](),
 		logger:             slog.Default(),
-	}
+	})
```
burgerdev, line 172: Why does this need parens?

internal/attestation/certcache/cached_client.go
```diff
@@ -42,47 +46,94 @@ func NewCachedHTTPSGetter(s store, ticker clock.Ticker, log *slog.Logger) *Cache
...
+		header, body, err := c.ContextHTTPSGetter.GetContext(ctx, url)
 		if err == nil {
-			c.cache.Set(url, res)
-			return res, nil
+			data, err := json.Marshal(cacheEntry{header, body})
+			if err != nil {
+				c.logger.Error("Failed to marshal CRL response", "error", err)
+				return nil, nil, err
+			}
+			c.cache.Set(url, data)
```
burgerdev, line 72: Did you consider changing the store to accept a struct instead?

davidweisse: We either use a file-base store or an in-memory store for the cache, so it's hard to change the interface here. At some point we have to write `[]byte`.

burgerdev: Alright, I forgot about the file-based store.

burgerdev (approved):

PR #1440, opened 2025-05-13
microsoft.genpolicy: fix regression in env var checks

burgerdev:
Addresses an issue with images that set HOSTNAME in their configuration, and an unrelated issue with `generateName` resources.

Fixed regression test: https://github.com/edgelesssys/contrast/actions/runs/14983444437


katexochen (approved):
Thanks!

burgerdev:
TDX runner is broken, unfortunately.

PR #1444, opened 2025-05-14
release: upload artifacts into public bucket

katexochen:
:green_circle: release run: https://github.com/edgelesssys/contrast/actions/runs/15000310432

I didn't test the the cleanup on-release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -219,6 +219,7 @@ jobs:
     permissions:
       contents: write
       packages: write
+      id-token: write
     env:
       RELEASE_BRANCH: ${{ needs.process-inputs.outputs.RELEASE_BRANCH }}
       WORKING_BRANCH: ${{ needs.process-inputs.outputs.WORKING_BRANCH }}
@@ -372,7 +373,7 @@ jobs:
           cp image-replacements-enterprise.txt cli/cmd/assets/image-replacements.txt
           git add cli/cmd/assets/image-replacements.txt
           git diff --staged --quiet || git commit -m "release: update cli assets (enterprise)"
-      - name: Build CLI
+      - name: Build CLI (enterprise)
         run: |
           nix build -L .#contrast-enterprise-cli-release --out-link workspace/contrast-enterprise
       - name: Upload release artifacts (for main branch PR)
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -389,6 +390,26 @@ jobs:
...
+            workspace/contrast-enterprise-cli/bin/contrast-enterprise
+            workspace/coordinator*.yml
+            workspace/runtime-*.yml
+            workspace/emojivoto-demo.yml
+            workspace/mysql-demo.yml
+          )
+          for file in "${files[@]}"; do
+            aws s3 cp "$file" "s3://contrast-public/pre-releases/${{ inputs.version }}/$unix/"
+          done
+          echo "Artifacts available at https://contrast-public.s3.eu-central-1.amazonaws.com/releases/${{ inputs.version }}/$unix/" | tee -a "$GITHUB_STEP_SUMMARY"
```
burgerdev, line 412: I think the artifacts might be easier to consume programmatically if they were not timestamped. Having two release candidates simultaneously sounds invalid anyway, and maybe we can live with the race condition between a re-run of the release pipeline and artifacts being consumed?

burgerdev, line 412: `amazonaws.com/releases/` vs. `s3://contrast-public/pre-releases` is a typo, right?

.github/workflows/release.yml
```diff
@@ -389,6 +390,26 @@ jobs:
...
+        run: |
+          unix=$(date +%s)
+          files=(
+            workspace/contrast-cli/bin/contrast
+            workspace/contrast-enterprise-cli/bin/contrast-enterprise
+            workspace/coordinator*.yml
+            workspace/runtime-*.yml
+            workspace/emojivoto-demo.yml
+            workspace/mysql-demo.yml
+          )
```
burgerdev, line 408: Could we download the artifact produced earlier as zip, and then loop over the contained files? I fear that it's easy to forget updating paths consistently.

.github/workflows/release.yml
burgerdev, line 412: Is this expected to be publicly accessible? I tried the URL from the test run, but got a permission denied for the directory as well as for individual files.

katexochen: Not sure how that makes it easier. I think the team consuming this can easily pick up the URL in the summary and use it. I added this because it allows to re-release with a minor change while the previous pre-release artifacts are still under test, in situations where we are confident that this won't affect the test result.

Alternatively, I would need to ensure we delete all the existing files before uploading the new ones.

.github/workflows/release.yml
katexochen: Sadly, we already have this block twice, so I think adding it a third time doesn't make things work. Would like to have this in a step as single source of truth and then consume it, but $time.

.github/workflows/release.yml
katexochen: Sorry, I tested the delete command, so now the files are gone. But I verified that they can be curled.

burgerdev: sgtm if programmatic access is not a priority.

burgerdev (approved):

PR #1456, opened 2025-05-19
ci: cleanup data from already deleted runtime classes

davidweisse:
Previously, the script determined unused runtime classes by listing existing runtime classes and excluding the ones which are currently in use. If a runtime class gets deleted on its own (during development, etc.) the data is still there. We can still get the runtime handler by directly looking into the `/opt/edgeless` directory and deleting those files, as well as cleaning up the config.

Without the nydus snapshotter, we also cannot get the runtime handler from the node installer arguments. Instead, we can now directly get it from the resource name.

### DIFF
--- a/packages/cleanup-bare-metal.sh
+++ b/packages/cleanup-bare-metal.sh
@@ -16,30 +16,38 @@ resourcesToCheck=(
   "cronjobs"
 )
 
+# Extract used runtime classes from running pods
 touch usedRuntimeClasses
 for resource in "${resourcesToCheck[@]}"; do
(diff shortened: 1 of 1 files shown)


packages/cleanup-bare-metal.sh
```diff
@@ -22,24 +22,27 @@ for resource in "${resourcesToCheck[@]}"; do
...
+      ls -1 "${OPTEDGELESS}"
+    } | sort -u
   )
 )
 
 for runtimeClass in "${unusedRuntimeClasses[@]}"; do
   # Delete unused runtime classes
   echo "Deleting runtimeclass ${runtimeClass} ..."
-  kubectl delete runtimeclass "${runtimeClass}"
+  kubectl delete runtimeclass "${runtimeClass}" || true
```
katexochen, line 45: Maybe log a warning if this doesn't succeed? We would want to know if it's always failing..

packages/cleanup-bare-metal.sh
katexochen, line 1: Script could use some documenting comments..

katexochen (approved):
only some nits

PR #1455, opened 2025-05-19
ci: regularly clean up old namespaces

davidweisse:
This adds a CronJob to the CI clusters that runs every 15 mins and deletes all development/testing namespaces that are older than one hour. The CronJob is deployed using the `bm_maintenace` GitHub workflow.

PR #1462, opened 2025-05-20
ci: update tcb specs on pull request

davidweisse:
This triggers the bare metal maintenance job on pull requests where the TCB specs change and does not trigger other maintenance jobs in that case.


katexochen:
how does this fit together with https://github.com/edgelesssys/contrast/pull/1455#discussion_r2097419246? Are you planning to split this into two jobs, or will we just re-apply the cron anyway?

.github/workflows/bm_maintenance.yml
```diff
@@ -2,6 +2,9 @@ name: bare metal maintenance
 
 on:
   workflow_dispatch:
+  pull_request:
```
katexochen, line 5: better: push on main instead of pr trigger.

katexochen:
do we need similar triggers for the deployment files etc?

davidweisse:
> how does this fit together with [#1455 (comment)](https://github.com/edgelesssys/contrast/pull/1455#discussion_r2097419246)? Are you planning to split this into two jobs, or will we just re-apply the cron anyway?

If we have two steps, we can simply move the if condition into the step that only updates the tcb specs.

> do we need similar triggers for the deployment files etc?

Should this also be a push on main trigger? In that case, I don't think we can separate changes to the tcb specs or changes to other deployment files.

.github/workflows/bm_maintenance.yml
```diff
@@ -2,8 +2,13 @@ name: bare metal maintenance
...
+    paths:
+      - ".github/workflows/bm_maintenance.yml"
+      - "tools/bm-maintenance/**"
+      - "packages/**"
```
katexochen: Why do we need packages/** ?

davidweisse: Removing this, so the cleanup job does not get triggered too often on pull requests.

PR #1467, opened 2025-05-20
coordinator: consider instances with stale or no manifests unready

burgerdev:
This PR adds a readiness probe to the Coordinator that checks whether the Coordinator is ready to serve `meshapi` requests, i.e. has a manifest configured that is not stale. The `userapi` can still be reached via the `coordinator:1313` service endpoint, even if unready, to allow initial `SetManifest` calls. Users may need to update any logic that waits for a Coordinator instance to be `Ready` before calling `contrast set`: it now only needs to be in phase `Running`.

---

Internally, we add a new service - `coordinator-ready` - backed by `Ready` Coordinators. This service is used by initializers and in peerrecovery, so that we only hit Coordinator endpoints that can actually serve requests. The idea is described in RFC 010:

https://github.com/edgelesssys/contrast/blob/cc41f2021a307f0dd9c00f4aa7a166b2f1eafe79/rfc/010-distributed-coordinator.md?plain=1#L113-L121

### DIFF
--- a/packages/scripts.nix
+++ b/packages/scripts.nix
@@ -207,6 +207,34 @@
     '';
   };
 
+  kubectl-wait-coordinator = writeShellApplication {
+    name = "kubectl-wait-coordinator";
+    runtimeInputs = with pkgs; [ kubectl ];
+    text = ''
+      namespace=$1
+      name="coordinator-0"
+
+      echo "Waiting for $name.$namespace to become ready" >&2
+
+      timeout=180
+
+      interval=4
+      while [ $timeout -gt 0 ]; do
+        if kubectl -n "$namespace" get pods "$name"; then
+          break
+        fi
+        sleep "$interval"
+        timeout=$((timeout - interval))
+      done
+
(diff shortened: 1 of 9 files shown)


packages/scripts.nix
```diff
@@ -207,6 +207,34 @@
...
+    text = ''
+      namespace=$1
+      name="coordinator-0"
+
+      echo "Waiting for $name.$namespace to become ready" >&2
+
+      timeout=180
+
+      interval=4
+      while [ $timeout -gt 0 ]; do
```
katexochen, line 222:
```suggestion
      while [[ $timeout -gt 0 ]]; do
```
nit: prefer [[ for consistency.

katexochen (approved):

packages/scripts.nix
burgerdev: It iis currently consistent with `kubectl-wait-ready` above. Let's fix this in a follow-up.

PR #1486, opened 2025-05-26
coordinator: remove state volume

katexochen:
The state is now managed as kubernetes objects, so a volume isn't needed anymore.


internal/kuberesource/parts.go
```diff
@@ -292,15 +292,11 @@ func Coordinator(namespace string) *CoordinatorConfig {
...
-							WithVolumeDevices(applycorev1.VolumeDevice().
-								WithName("state-device").
-								WithDevicePath("/dev/csi0"),
-							).
 							WithSecurityContext(SecurityContext().
 								WithCapabilities(applycorev1.Capabilities().
 									WithAdd(
 										"SYS_ADMIN",
-										"NET_ADMIN", // Needed for removing the default deny iptales rule.
+										"NET_ADMIN", // Needed for removing the default deny iptables rule.
```
burgerdev, line 299: Can we remove `SYS_ADMIN` now?

PR #1496, opened 2025-05-28
stateguard: add unit tests

burgerdev:


coordinator/internal/stateguard/stateguard_test.go
```diff
@@ -35,13 +40,207 @@ contrast_coordinator_manifest_generation %d
...
+		assert.True(pool.AppendCertsFromPEM(pemBytes), name)
+		block, rest := pem.Decode(pemBytes)
+		assert.NotEmpty(block.Bytes, name)
+		assert.Empty(rest, name)
+	}
+}
+
+// TestUpdateStateRace tests that parallel changes to the internal state pointer don't affect the
+// outcome of UpdateState calls.
+func TestUpdateStateRace(t *testing.T) {
```
katexochen: I was confused by the test name here, I expected to see a race test (something trying to trigger the race detector by hammering with goroutines). I think the test makes sense, but maybe frame it more like a "background update" than parallel.

burgerdev: Renamed

PR #1505, opened 2025-06-03
attestation/certcache: treat malformed cache entries as cache miss

davidweisse:
The previous change to the `certcache` to include the header information in the response in #1439 changed the cache data from plain text to JSON data. In case of a cache hit, this JSON data has to be unmarshalled, which leads to errors when trying to unmarshal old cache entries, which are not in JSON format. This PR changes the behavior of the `certcache` such that any JSON unmarshalling error will simply result in a cache miss. Additionally, any error caused by JSON marshalling a response will just not write to the cache instead of returning an error.


burgerdev (approved):
Thanks!

PR #1507, opened 2025-06-03
cli: first invocation of `generate` should fail if resource does not have a coordinator

charludo:
- running `contrast generate` on a resource without a coordinator will now ~~fail~~ inform the user about the missing coordinator on the first invocation, instead of (previously) giving an unrelated info message and failing on the second invocation
- if reference values described in [the docs](https://docs.edgeless.systems/contrast/next/deployment?platform=k3s-qemu-tdx#generate-policy-annotations-and-manifest) are (expectedly) missing, show the old info message, since this is the error case it applies to
- for all other errors, including unexpectedly missing reference values, return the error and fail manifest generation

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -173,8 +173,14 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	}
 
 	mnf.Policies = policyMap
-	// Existing manifests are already validated above, but newly generated manifests may be missing the reference values.
-	if err := mnf.Validate(); err != nil {
+	// Existing manifests are already validated above, but newly generated manifests may be missing reference values or a coordinator.
+	var ce *manifest.CoordinatorCountError
+	if err := mnf.Validate(); errors.As(err, &ce) {
+		if ce.Count == 0 {
+			fmt.Fprintln(cmd.OutOrStdout(), "  No Coordinator resource found, did you forget to add it to your resources?")
+		}
+		return ce
+	} else if err != nil {
 		fmt.Fprintf(cmd.OutOrStdout(), "  Please fill in the reference values for %s\n", flags.referenceValuesPlatform.String())
 	}
 
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -60,14 +60,14 @@ func (m *Manifest) Validate() error {
 		}
 	}
 
-	var coordinatorCount int
+	var coordinatorCount uint
 	for _, policy := range m.Policies {
 		if policy.Role == RoleCoordinator {
 			coordinatorCount++
 		}
 	}
 	if coordinatorCount != 1 {
-		return fmt.Errorf("expected exactly 1 policy with role 'coordinator', got %d", coordinatorCount)
+		return &CoordinatorCountError{Count: coordinatorCount}
 	}
 
 	if err := m.ReferenceValues.Validate(); err != nil {
@@ -431,4 +431,13 @@ func flattenValidationError(err error) (errs []error) {
 		return errs
 	}
 	return []error{err}
+}
+
+
+	}
+// CoordinatorCountError occurs during manifest validation when either zero, or more than one coordinators have been defined.
+type CoordinatorCountError struct{ Count uint }
+
+func (e *CoordinatorCountError) Error() string {
+	return fmt.Sprintf("expected exactly 1 policy with role 'coordinator', got %d", e.Count)
 }
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -371,15 +371,15 @@ func validateHexString(value HexString, expectedNumBytes int) error {
 	return err
 }
 
-// validationError contains a JSON path and a list of errors.
+// ValidationError contains a JSON path and a list of errors.
 // Nested validation errors are printed on newlines with the full path.
-type validationError struct {
+type ValidationError struct {
 	path string
 	errs []error
 }
 
 func newValidationError(path string, errs ...error) error {
-	e := &validationError{
+	e := &ValidationError{
 		path: path,
 		errs: make([]error, 0, len(errs)),
 	}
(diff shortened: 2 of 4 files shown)


cli/cmd/generate.go
```diff
@@ -174,7 +174,9 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 
 	mnf.Policies = policyMap
 	// Existing manifests are already validated above, but newly generated manifests may be missing the reference values.
```
katexochen: Comment needs to be slightly updated.

internal/manifest/manifest.go
```diff
@@ -432,3 +434,7 @@ func flattenValidationError(err error) (errs []error) {
 	}
 	return []error{err}
 }
+
+// ErrCoordinatorMissing is a possible error type returned by the manifest's Validate method,
+// for which an error is thrown instead of only informaing the user.
```
katexochen: This should rather describe what the error case is that this is returned on, and not make assumptions what a caller reaction to such error is.

internal/manifest/manifest.go
```diff
@@ -432,3 +434,7 @@ func flattenValidationError(err error) (errs []error) {
...
+var ErrCoordinatorMissing = errors.New("coordinator definition is missing")
```
katexochen: I feel like the error message that was thrown before was more precise. Maybe we can create an error type like `CoordinatorCountError`, check for that type with `errors.As` in generate.go and then check for count stored in the error and print the suggestion (like "No Coordinator resource found, did you forget to add it to your resources?") as log statement.

internal/manifest/manifest.go
```diff
@@ -432,3 +432,15 @@ func flattenValidationError(err error) (errs []error) {
...
+// CoordinatorCountError occurs during manifest validation when either zero, or more than coordinators have been defined.
+type CoordinatorCountError struct{ count uint }
+
+func (e *CoordinatorCountError) Error() string {
+	switch e.count {
+	case 0:
+		return "No Coordinator resource found, did you forget to add it to your resources?"
+	default:
+		return fmt.Sprintf("expected exactly 1 policy with role 'coordinator', got %d", e.count)
+	}
```
katexochen: I'd prefer to always return the default case here, and rather check on the caller side (where errors.As is used for the count and only print the suggestion instead of using it as error string.

charludo: The main reason I went with this solution is that this also improves the error message shown when validation due to a missing coordinator fails for an existing manifest, with no extra code changes.

Would keeping the `Error()` fn from above work for you, with the change of only logging the error in the case of a new manifest?

internal/manifest/manifest.go
```diff
@@ -432,3 +433,47 @@ func flattenValidationError(err error) (errs []error) {
...
+		}
+	}
+
+	if isLeaf {
+		return slices.Contains(expectedMissingReferenceValues, e.path)
+	}
+	return true
+}
+
+// CoordinatorCountError occurs during manifest validation when either zero, or more than coordinators have been defined.
```
burgerdev:
```suggestion
// CoordinatorCountError occurs during manifest validation when either zero, or more than one coordinators have been defined.
```

cli/cmd/generate.go
```diff
@@ -173,8 +173,11 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	}
 
 	mnf.Policies = policyMap
-	// Existing manifests are already validated above, but newly generated manifests may be missing the reference values.
-	if err := mnf.Validate(); err != nil {
+	// Existing manifests are already validated above, but newly generated manifests may be missing reference values or a coordinator.
+	var ce *manifest.CoordinatorCountError
+	if err := mnf.Validate(); errors.As(err, &ce) {
+		fmt.Println("  ", err)
```
burgerdev: Should this be using `fmt.Fprintf(cmd.OutOrStdout(), ...)`?

cli/cmd/generate.go
```diff
@@ -177,7 +177,10 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	var ce *manifest.CoordinatorCountError
 	var ve *manifest.ValidationError
 	if err := mnf.Validate(); errors.As(err, &ce) {
-		fmt.Println("  ", err)
+		if ce.Count == 0 {
+			fmt.Fprintf(cmd.OutOrStdout(), "  No Coordinator resource found, did you forget to add it to your resources?\n")
```
katexochen:
```suggestion
			fmt.Fprintln(cmd.OutOrStdout(), "  No Coordinator resource found, did you forget to add it to your resources?")
```

internal/manifest/manifest_test.go
```diff
@@ -437,3 +440,55 @@ func TestHexStrings(t *testing.T) {
...
+func TestExpectedMissingReferenceValues(t *testing.T) {
+	manifestTDXExpectedOnly := newTestManifestTDX()
+	manifestTDXExpectedOnly.ReferenceValues.TDX[0].MrSeam = ""
+	manifestTDXWithUnexpected := newTestManifestTDX()
+	manifestTDXWithUnexpected.ReferenceValues.TDX[0].MrTd = ""
+
+	manifestSNPExpectedOnly := newTestManifestSNP()
+	manifestSNPExpectedOnly.ReferenceValues.SNP[0].MinimumTCB.TEEVersion = nil
+	manifestSNPWithUnexpected := newTestManifestSNP()
+	manifestSNPWithUnexpected.ReferenceValues.SNP[0].TrustedMeasurement = ""
```
katexochen: prefer to not use global state between test cases.

internal/manifest/manifest_test.go
```diff
@@ -437,3 +440,55 @@ func TestHexStrings(t *testing.T) {
...
+			m:    newTestManifestSNP(),
+			want: true,
+		},
+	}
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			var ve *ValidationError
+			err := tc.m.Validate()
```
katexochen, line 486:
```suggestion
			err := tc.m.Validate()
			assert.Error(err)
			require.ErrorAs(err, &ve)
			assert.Equal(tc.want, ve.Only...)
```

katexochen (approved):
Good job!

PR #1508, opened 2025-06-03
ci: add weekly containerd store cleanup

davidweisse:
This separates and revises the step to clean up the containerd store on the bare metal servers. This is no longer executed daily, but rather weekly or by `workflow_dispatch`. It uses the `--sync` option from `ctr image rm` to perform garbage collection on all associated resources on all images. If any images were otherwise deleted, leftover content can be cleaned up using `ctr content prune references`.

Successful run with containerd cleanup: https://github.com/edgelesssys/contrast/actions/runs/15418145640

### DIFF
--- a/packages/scripts.nix
+++ b/packages/scripts.nix
@@ -555,4 +555,19 @@
     ];
     text = builtins.readFile ./cleanup-namespaces.sh;
   };
+
+  cleanup-containerd = writeShellApplication {
+    name = "cleanup-containerd";
+    runtimeInputs = with pkgs; [ containerd ];
+    text = ''
+      while read -r image; do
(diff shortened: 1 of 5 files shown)


packages/scripts.nix
```diff
@@ -555,4 +555,20 @@
     ];
     text = builtins.readFile ./cleanup-namespaces.sh;
   };
+
+  cleanup-containerd = writeShellApplication {
+    name = "cleanup-containerd";
+    runtimeInputs = with pkgs; [ containerd ];
+    text = ''
+      set -euo pipefail
```
katexochen:
```suggestion
```

already injected by writeShellApplication

PR #1513, opened 2025-06-04
msft.genpolicy: correctly represent binaryData in ConfigMaps

miampf:
This PR adds a patch that fixes `contrast generate` on ConfigMaps that contain `binaryData` fields. The patch is taken from this commit on `kata-containers/kata-containers`: https://github.com/kata-containers/kata-containers/commit/c06bf2e3bb696016be0357c373b0c68e10602b57

### DIFF
--- a/packages/by-name/microsoft/genpolicy/0015-genpolicy-correctly-represent-binaryData-in-ConfigMa.patch
+++ b/packages/by-name/microsoft/genpolicy/0015-genpolicy-correctly-represent-binaryData-in-ConfigMa.patch
@@ -1,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: miampf <miampf@proton.me>
+Date: Wed, 4 Jun 2025 12:21:28 +0200
+Subject: [PATCH] genpolicy: correctly represent binaryData in ConfigMaps
+
+While Kubernetes defines `binaryData` as `[]byte`,
+when defined in a YAML file the raw bytes are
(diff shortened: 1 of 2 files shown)


miampf:
I tested this using the regression test cherry picked from here: https://github.com/edgelesssys/contrast/pull/1394

packages/by-name/microsoft/genpolicy/0014-genpolicy-correctly-represent-binaryData-in-ConfigMa.patch
```diff
@@ -0,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: miampf <miampf@proton.me>
```
burgerdev: When cherry-picking commits without significant changes, I'd consider retaining the original author. I somewhat expected `git cherry-pick` to do that automatically...

miampf: I believe git cherry-pick does that, I just didn't use it but added the changes myself since I did not think that changes to the testdata of genpolicy would be relevant. Can cherry pick the original commit instead if you want me to

burgerdev: Ah, I see, and the testdata change would probably not even apply. Ok then.

burgerdev (approved):

PR #1516, opened 2025-06-06
fix: correct policy generation for `ReplicationController`

miampf:
Previously, you couldn't add a ReplicationController to a contrast cluster since you would get a "policy hash not found in manifest" error. This PR fixes that.


miampf:
I don't know if this is super relevant since ReplicationController is legacy so I added the `no changelog` label. Feel free to remove it if this should appear in the changelog :)

katexochen:
@miampf is this ready for review?

katexochen (approved):

PR #1517, opened 2025-06-10
Fix e2e tests `genpolicy` and `getdents` failing after #1507

charludo:


katexochen (changes requested):
Please change commit titles to state what changed. The reference to the CI failure and the previous change can go in to commit message body.

charludo:
> "bug fix" label is for user facing bugs only. This wasn't released yet, so there is no need to mention it as bug fix in the release notes.

Got it. "no changelog" then for anything not user-visible?

katexochen:
> > "bug fix" label is for user facing bugs only. This wasn't released yet, so there is no need to mention it as bug fix in the release notes.
> 
> Got it. "no changelog" then for anything not user-visible?

Exactly.

PR #1519, opened 2025-06-10
kata: add patch preventing corruption of genpolicy's layer cache file

charludo:
`fs2` file locks where used incorrectly: the lock was only being applied to write procedures, not to read (or delete, though this is not directly relevant for our issue).

For large enough strings, parsing takes long enough for a race condition to occur and the file is corrupted. The warning from the logs is the result, since the json file can no longer be parsed.

The reproducer from the ticket actually does not exhibit this behavior since approximately kata-containers 3.17.0, since improvements were made to the string parsing; however, for large enough json strings in the initial state of the later cache file, it can still be triggered.

Upstream PR: kata-containers/kata-containers#11426

### DIFF
--- a/packages/by-name/microsoft/genpolicy/package.nix
+++ b/packages/by-name/microsoft/genpolicy/package.nix
@@ -108,6 +108,11 @@ rustPlatform.buildRustPackage rec {
       # base64 instead of a Vec<u8>.
       # Taken from https://github.com/kata-containers/kata-containers/commit/c06bf2e3bb696016be0357c373b0c68e10602b57.
       ./0015-genpolicy-correctly-represent-binaryData-in-ConfigMa.patch
+
+      # This patch fixes an issue where genpolicy can corrupt the layer cache file due to simultaneous
+      # read/write operations on the file. Instead of the upstream implementation, the cache file is opened
+      # read-only, changes are written to a tempfile, and the original file replaced by the tempfile atomically.
+      ./0016-genpolicy-prevent-corruption-of-the-layer-cache-file.patch
     ];
   };
 
--- a/packages/by-name/kata/kata-runtime/0026-genpolicy-keep-layers-cache-in-memory.patch
+++ b/packages/by-name/kata/kata-runtime/0026-genpolicy-keep-layers-cache-in-memory.patch
@@ -1,0 +1,640 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: charludo <git@charlotteharludo.com>
+Date: Tue, 17 Jun 2025 15:39:32 +0200
+Subject: [PATCH] genpolicy: keep layers cache in-memory to prevent corruption
+
+The locking mechanism around the layers cache file was insufficient to
+prevent corruption of the file. This commit moves the layers cache's
+management in-memory, only reading the cache file once at the beginning
+of `genpolicy`, and only writing to it once, at the end of `genpolicy`.
+
+In the case that obtaining a lock on the cache file fails,
+reading/writing to it is skipped, and the cache is not used/persisted.
+
+Signed-off-by: Charlotte Hartmann Paludo <git@charlotteharludo.com>
+---
+ src/tools/genpolicy/src/layers_cache.rs       |  97 +++++++++
+ src/tools/genpolicy/src/lib.rs                |   1 +
+ src/tools/genpolicy/src/main.rs               |   2 +
+ src/tools/genpolicy/src/registry.rs           | 193 ++++--------------
+ .../genpolicy/src/registry_containerd.rs      | 111 ++++------
+ src/tools/genpolicy/src/utils.rs              |   5 +-
+ src/tools/genpolicy/tests/policy/main.rs      |   2 +-
+ 7 files changed, 186 insertions(+), 225 deletions(-)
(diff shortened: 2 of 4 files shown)


packages/by-name/kata/kata-runtime/0026-genpolicy-prevent-corruption-of-the-layer-cache-file.patch
```diff
@@ -0,0 +1,96 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: charludo <git@charlotteharludo.com>
+Date: Thu, 12 Jun 2025 10:33:33 +0200
+Subject: [PATCH] genpolicy prevent corruption of the layer cache file
```
katexochen:
```suggestion
Subject: [PATCH] genpolicy: prevent corruption of the layer cache file
```

katexochen: Commit needs a commit message body.

packages/by-name/kata/kata-runtime/0026-genpolicy-prevent-corruption-of-the-layer-cache-file.patch
```diff
@@ -0,0 +1,97 @@
...
+-        vec
+-    } else {
+-        // Delete the malformed file here if it's present
+-        Vec::new()
++    // Open the json file in read-only mode
++    let read_file = match OpenOptions::new().read(true).open(cache_file) {
++        Ok(file) => file,
++        // If the file did not exist, try creating it.
++        // This is necessary on the first invocation of this function.
++        // create_new() fails if the file was created elsewhere in the meantime
```
katexochen: Why not use `OpenOptions.create(true)` for this instead of doing a second call?

Thinking more about it: If we know the files doesn't exist, why create it? We know its empty, we don't want to write to it.

packages/by-name/kata/kata-runtime/0026-genpolicy-prevent-corruption-of-the-layer-cache-file.patch
```diff
@@ -0,0 +1,97 @@
...
+-    // Serialize in pretty format
+-    let serialized = serde_json::to_string_pretty(&data)?;
++    // Write the serialized data to a temporary file
++    let temp_file = tempfile::NamedTempFile::new()?;
++    serde_json::to_writer_pretty(&temp_file, &data)?;
+ 
+-    // Open the JSON file to write
+-    let file = OpenOptions::new().write(true).open(cache_file)?;
++    // Atomically replace the original cache file
++    temp_file.persist(cache_file)?;
```
thomasten: `persist` doesn't work across filesystems. And many distros use tmpfs for `/tmp`.
You could do something like
`let temp_file = tempfile::NamedTempFile::new_in(Path::new(cache_file).parent().unwrap())?;`

thomasten (changes requested):

packages/by-name/kata/kata-runtime/0026-genpolicy-prevent-corruption-of-the-layer-cache-file.patch
charludo: Oh, good catch! Thanks, that seems to work great.

packages/by-name/kata/kata-runtime/0026-genpolicy-prevent-corruption-of-the-layer-cache-file.patch
charludo: Done

burgerdev (approved):
LGTM, thanks!

packages/by-name/microsoft/genpolicy/package.nix
```diff
@@ -108,6 +108,11 @@ rustPlatform.buildRustPackage rec {
       # base64 instead of a Vec<u8>.
       # Taken from https://github.com/kata-containers/kata-containers/commit/c06bf2e3bb696016be0357c373b0c68e10602b57.
       ./0015-genpolicy-correctly-represent-binaryData-in-ConfigMa.patch
+
+      # This patch fixes an issue where genpolicy can corrupt the layer cache file due to simultaneous
+      # read/write operations on the file. Instead of the upstream implementation, the cache file is opened
+      # read-only, changes are written to a tempfile, and the original file replaced by the tempfile atomically.
+      ./0016-genpolicy-prevent-corruption-of-the-layer-cache-file.patch
```
thomasten, line 115: Is it intended that we use the initial fix for microsoft genpolicy?

burgerdev: Yes: there are a couple of refactorings upstream that make porting hard, and I think investing here is not worth the effort right now. Rather, we should see to replacing the MS fork with upstream asap.

charludo: Yes, because https://github.com/kata-containers/kata-containers/commit/349ce8c33940cf95337857171fde900e7e6863b4 is not backported to the microsoft fork, adapting the upstream fix isn't exactly straightforward. I discussed this with @burgerdev who advised to use the initial fix, since usage of the fork is supposed to be phased out anyways.

katexochen (approved):

PR #1520, opened 2025-06-10
kata-debug-shell: rework, make nix pkgs

katexochen:
The previous script was broken as it used the now removed snapshotter
to find the runtimeClass name. It was also inconvenient as it
took the container id, where now namespace/pod-name can be passed.


katexochen:
Instructions for testing:

1. Enable debug access to pod vm image:
    ```sh
    sed -i 's/debugRuntime ? false/debugRuntime ? true/' packages/by-name/kata/contrast-node-installer-image/package.nix
     ```
3. Get credentials for a bare metal CI runner (not tdx, that's where I developed this):
    ```sh
    sed -i 's/AKS-CLH-SNP/K3s-QEMU-SNP/' justfile.env`
    just get-credentials
    ```
3. ```
    just
    ```
4. Get access to the node (nodeshell, nsenter, tailscale ssh)
5. Copy the script over
6. Run with a pod you created

You should then get a shell in the pod vm.

packages/by-name/kata-debug-shell/kata-debug-shell.sh
```diff
@@ -0,0 +1,75 @@
...
+else
+  die "Neither 'k3s ctr' nor 'ctr' command found. Please install one of them."
+fi
+
+pod_namespace="${1%/*}"
+pod_name="${1##*/}"
+
+if [[ -z $pod_namespace ]] || [[ -z $pod_name ]]; then
+  die "Could not parse namespace or podname from input '$1'."
+fi
```
charludo:
```suggestion
pod_namespace="${1%/*}"
pod_name="${1##*/}"

if [[ "$1" != */*  ]] || [[ -z $pod_namespace ]] || [[ -z $pod_name ]]; then
  die "Could not parse namespace or podname from input '$1'."
fi
```

Assuming that `$1` not containing any `/` should also result in the same error; right now, something like `./kata-debug-shell no-slashes` results in
```
Error from server (NotFound): namespaces "no-slashes" not found
Error: Failed to get runtime class for pod 'no-slashes' in namespace 'no-slashes'.
```
since both `pod_namespace` and `pod_name` become copies of `$1`.

charludo (changes requested):
Seems to work as intended! 🎉 
image

packages/by-name/kata-debug-shell/kata-debug-shell.sh
katexochen: good catch :)

charludo (approved):
🎉

PR #1522, opened 2025-06-12
e2e/regression: binaryData in ConfigMap

miampf:
This tests consumption of `binaryData` in the regression test

### DIFF
--- a/e2e/regression/testdata/config-map.yml
+++ b/e2e/regression/testdata/config-map.yml
@@ -5,4 +5,25 @@ metadata:
   namespace: "@@REPLACE_NAMESPACE@@"
 data:
   foo: bar
-# TODO(miampf): Add binaryData field
+binaryData:
+  bar: Zm9v
+---
+apiVersion: v1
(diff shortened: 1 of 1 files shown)


e2e/regression/testdata/config-map.yml
```diff
@@ -5,4 +5,24 @@ metadata:
   namespace: "@@REPLACE_NAMESPACE@@"
 data:
   foo: bar
-# TODO(miampf): Add binaryData field
+binaryData:
+  bar: Zm9v
+---
+apiVersion: v1
+kind: Pod
```
katexochen, line 12: Why are we adding the pod here?

miampf: The ticket specifies that a pod should consume the config map to make sure it can be used.

burgerdev: It would need to be a `contrast-cc` pod, though...

burgerdev (approved):

PR #1524, opened 2025-06-13
node-installer: split constants package

katexochen:
Ordering some code in the node-installer. We should also move the containerd tests/testdata into containerdconf by time and maybe move the snp stuff out of kata.


charludo (approved):
LGTM - structure makes a lot more sense now 👍🏼 

Is there a specific reason this:

> We should also move the containerd tests/testdata into containerdconf

should not be done in this PR?

katexochen:
> Is there a specific reason this:
> 
> > We should also move the containerd tests/testdata into containerdconf
> 
> should not be done in this PR?

Time. :D That stuff is a bit more entangled and I need to think about the right interface.

PR #1526, opened 2025-06-16
dev-shells: add a rust dev shell for upstream work

charludo:
Depending on the project, additional packages may need to be added to the list.

### DIFF
--- a/dev-shells/by-name/kata.nix
+++ b/dev-shells/by-name/kata.nix
@@ -1,0 +1,62 @@
+# Copyright 2025 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+{
+  # keep-sorted start
+  clang,
+  cmake,
+  fenix,
+  lib,
+  lld,
+  llvm,
+  llvmPackages,
+  pkg-config,
+  protobuf,
+  # keep-sorted end
+  mkShell,
+  pkgsCross,
+}:
+
+let
+  pinnedCommit = "26c7f941aa3b378ded647709418f6ff4637bfb1c";
+  toolchainFile = builtins.fetchurl {
+    url = "https://raw.githubusercontent.com/kata-containers/kata-containers/${pinnedCommit}/rust-toolchain.toml";
+    sha256 = "sha256:079kqn2kpsp7q5xjgxdhgxdqddvhacq75akwylfa1adjsjlynp1f";
+  };
+
+  toolchainSpec = {
+    name = (lib.importTOML toolchainFile).toolchain.channel;
+    sha256 = "sha256-Hn2uaQzRLidAWpfmRwSRdImifGUCAb9HeAqTYFXWeQk=";
+  };
+  muslToolchain = fenix.targets."x86_64-unknown-linux-musl".fromToolchainName toolchainSpec;
+in
+mkShell {
+  buildInputs = [
+    # keep-sorted start
+    clang
+    cmake
+    lld
+    llvm
+    llvmPackages.libclang
+    pkg-config
+    protobuf
+    # keep-sorted end
+    pkgsCross.musl64.buildPackages.gcc
+    (fenix.combine [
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -33,7 +36,7 @@
       let
         pkgs = import nixpkgs {
           inherit system;
-          overlays = [ (import ./overlays/nixpkgs.nix) ];
+          overlays = [ (import ./overlays/nixpkgs.nix { inherit (self) inputs; }) ];
```
katexochen: nit: file is called overlays/nixpkgs.nix, maybe add fenix as a second overlay inline:

```suggestion
          overlays = [
            (final: _prev: { fenix = self.inputs.fenix.packages.${final.system}; })
            (import ./overlays/nixpkgs.nix)
          ];
```

dev-shells/by-name/rust.nix
```diff
@@ -0,0 +1,36 @@
...
+mkShell {
+  packages = [
+    clang
+    cmake
+    lld
+    llvm
+    llvmPackages.libclang
+    pkg-config
+    protobuf
+    (fenix.latest.withComponents [
```
katexochen: Should we use the toolchain version actually used by kata instead?

dev-shells/by-name/rust.nix
katexochen, line 1: maybe rename to kata.nix, I could image that we will add some kata-specific stuff here.

katexochen (changes requested):

flake.nix
charludo: Yeah, makes sense.

dev-shells/by-name/rust.nix
charludo: Oh! Yes, good idea.

katexochen: https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L416

charludo: > https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L416

Since kata uses a `rust-toolchain.toml` file and fenix supports reading those, I would rather go that route

katexochen: Not sure I understand, how would you pass that file from kata to fenix?

charludo: I admit, it's not a pretty solution due to the two hashes... But at least it's explicit where the version comes from? 😅

flake.lock
```diff
@@ -1,5 +1,24 @@
 {
   "nodes": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
```
katexochen: We should use inputs.nixpkgs.follows = "nixpkgs"; for fenix.

dev-shells/by-name/rust.nix
katexochen: Fine for me. :)

dev-shells/by-name/kata.nix
```diff
@@ -0,0 +1,59 @@
...
+  toolchainFile = builtins.fetchurl {
+    url = "https://raw.githubusercontent.com/kata-containers/kata-containers/${pinnedCommit}/rust-toolchain.toml";
+    sha256 = "sha256:079kqn2kpsp7q5xjgxdhgxdqddvhacq75akwylfa1adjsjlynp1f";
+  };
+
+  toolchainSpec = {
+    name = (lib.importTOML toolchainFile).toolchain.channel;
+    sha256 = "sha256-Hn2uaQzRLidAWpfmRwSRdImifGUCAb9HeAqTYFXWeQk=";
+  };
+  crossToolchain = fenix.targets."x86_64-unknown-linux-musl".fromToolchainName toolchainSpec;
```
katexochen:
```suggestion
  muslToolchain = fenix.targets."x86_64-unknown-linux-musl".fromToolchainName toolchainSpec;
```

charludo:
@katexochen `make` should just work now.

PR #1531, opened 2025-06-18
just: update k3s-qemu-tdx runner; update tcb-specs

katexochen:


justfile
```diff
@@ -299,7 +299,7 @@ get-credentials platform=default_platform:
                 --name "$azure_resource_group"
         ;;
         "K3s-QEMU-TDX")
-            nix run -L .#scripts.get-credentials "projects/796962942582/secrets/m50-ganondorf-kubeconf/versions/latest"
+            nix run -L .#scripts.get-credentials "projects/796962942582/secrets/olimar-kubeconfig/versions/latest"
```
charludo, line 302: Out of curiosity, where do these come from?

charludo (approved):

justfile
katexochen: https://github.com/edgelesssys/contrast/blob/main/dev-docs/e2e/bare-metal-runner.md?plain=1#L226-L246

PR #1532, opened 2025-06-19
cli: warn volume mount sharing

jmxnzo:
This PR adds a mechanism to detect and warn about volumes mounted by more than one container—whether within the same Pod or across different Pods (e.g. PVCs, ConfigMaps). We introduce a `volumeInfoMap` in `policiesFromKubeResources()`—where we already have direct access to each resource’s `PodTemplateSpec`—to accumulate mount data as we process every workload.

- **Accumulation Phase:**  
  As we iterate each resource, we increment mount counts and record which resources reference each volume in `volumeInfoMap`.

- **Reporting Phase:**  
  Once all resources are processed, `reportVolumeSharing()` walks the map and emits one warning per shared volume.

- **Unique Map Keys:**  
  Each entry’s key combines the volume’s name and its type, aiming to create UID, as it is missing before the resources are applied.

- **`emptyDir` Isolation:**  
  Because `emptyDir` volumes are confined to a single Pod, we prefix their keys with the owning resource name. This prevents false positives when multiple Pods declare emptyDirs with identical names.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -158,7 +158,7 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 
 	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Generated workload policy annotations")
 
-	policies, err := policiesFromKubeResources(paths)
+	policies, err := policiesFromKubeResources(paths, log)
 	if err != nil {
 		return fmt.Errorf("find kube resources with policy: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


jmxnzo:
Volume mount sharing was not referring to shared volumes between pods, but to shared volumes between host and guest. Also we decided to add support for the shared filesystem mounts, so the warning mechanism likely won't be needed anymore.

PR #1533, opened 2025-06-20
node-installer: don't require snpIDBlock on TDX

katexochen:
This broke in the last refactoring.


charludo (commented):
As far as I can tell, this should currently work without problem. I am a bit concerned about passing the uninitialized `snpIDBlock` to `KataRuntimeConfig` though, from what I've gathered that's undefined behavior in Go?

PR #1535, opened 2025-06-20
e2e/regression: move testdata to own directories

miampf:
To make the regression test more flexible, this PR splits resources to be tested into own directories.


e2e/regression/regression_test.go
```diff
@@ -60,84 +61,115 @@ func TestRegression(t *testing.T) {
...
 
+				if resourceType == "CronJob" {
+					// Resource waiting is overly complicated so CronJob is only tested for generate.
+					require.True(t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
+					return
+				}
+
+				wg.Add(1)
+				goroutineCounter++
+				go func() {
```
katexochen: no need to do this concurrently here, we should gain nearly nothing from waiting in parallel.

e2e/regression/regression_test.go
```diff
@@ -60,84 +61,115 @@ func TestRegression(t *testing.T) {
...
-				if err != nil {
-					t.Log("error unmarshaling yaml to unstructured resources: ", err)
-				}
-				bgDeletion := metav1.DeletePropagationBackground
-				for _, r := range unstructuredResources {
-					client, err := ct.Kubeclient.ResourceInterfaceFor(r)
-					// Using WithoutCancel here since t.Context would get canceled in cleanup and context.Background triggers the linter
-					ctx, cancel := context.WithTimeoutCause(context.WithoutCancel(t.Context()), ct.FactorPlatformTimeout(1*time.Minute), errors.New("deletion took to long"))
-					defer cancel()
+				resourceName, _ := strings.CutSuffix(file.Name(), ".yml")
```
katexochen: Instead of relying on the file name and assuming there is only a single resource in that file, we should look for resources we want to wait for in intermediateResources.

e2e/regression/regression_test.go
```diff
@@ -60,84 +61,115 @@ func TestRegression(t *testing.T) {
...
+
```
katexochen, line 92: We should do the rest of the testCase outside of the file loop.

katexochen (dismissed):

e2e/regression/testdata/prometheus/prometheus-service.yml
katexochen, line 1: we should be able to undo all splits like these

e2e/regression/regression_test.go
```diff
@@ -60,83 +62,103 @@ func TestRegression(t *testing.T) {
...
+				newResources, err := kuberesource.UnmarshalApplyConfigurations(resourceYAML)
+				require.NoError(err)
+
+				newResources = kuberesource.PatchRuntimeHandlers(newResources, runtimeHandler)
+				newResources = kuberesource.AddPortForwarders(newResources)
+
+				intermediateResources = append(intermediateResources, newResources...)
+
+				// Check if we are testing a cron job
+				unmarshalledYAML := make(map[string]interface{})
```
katexochen:
```suggestion
				unmarshalledYAML := make(map[string]any)
```

katexochen:
@miampf what about the test failures?

e2e/regression/regression_test.go
```diff
@@ -60,83 +62,108 @@ func TestRegression(t *testing.T) {
 	require.True(t, t.Run("set", ct.Set), "contrast set needs to succeed for subsequent tests")
 	require.True(t, t.Run("verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
 
-	for _, file := range files {
-		t.Run(file.Name(), func(t *testing.T) {
+	for _, dir := range dirs {
+		t.Run(dir.Name(), func(t *testing.T) {
```
burgerdev: Please skip the entry if it's not a directory.

e2e/regression/regression_test.go
```diff
@@ -60,83 +62,108 @@ func TestRegression(t *testing.T) {
...
+				require.NoError(yaml.Unmarshal(resourceYAML, &unmarshalledYAML))
+				resourceType, ok := unmarshalledYAML["kind"].(string)
+				require.True(ok)
+				if resourceType == "CronJob" {
+					resourceBytes, err := kuberesource.EncodeResources(intermediateResources...)
+					require.NoError(err)
+					require.NoError(os.WriteFile(path.Join(ct.WorkDir, "resources.yml"), resourceBytes, 0o644))
+					// Resource waiting is overly complicated so CronJob is only tested for generate.
+					require.True(t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
+					return
```
burgerdev: What if there are two files, of which one contains a CronJob? Looks like we're skipping the other file in that case.

e2e/regression/regression_test.go
```diff
@@ -60,83 +62,108 @@ func TestRegression(t *testing.T) {
...
+				}
+			}
 			// write the new resources.yml
-			resourceBytes, err := kuberesource.EncodeResources(append(resources, newResources...)...)
+			resourceBytes, err := kuberesource.EncodeResources(intermediateResources...)
```
burgerdev, line 103: The point of moving the test data into separate files was to test whether the resources can be in separate files - like the separate CM and Secret files in #1549. If we're writing them all to the same file again here, we've effectively reverted to the prior structure of a single file per test.

e2e/regression/regression_test.go
```diff
@@ -267,3 +294,29 @@ func runCommandImpl(ctx context.Context, cmd, namespace, workDir string) error {
 
 	return errors.New("unknown command: " + cmd)
 }
+
+func cleanupResource(ctx context.Context, resource any, name string, ct *contrasttest.ContrastTest) error {
+	if strings.Contains(name, "coordinator") {
+		return nil
+	}
```
burgerdev: You could push this down to the loop below - and then you can remove the name parameter (because it's already present on the unstructured type). 

But it could also make sense to track the Coordinator and the testdata content independently (instead of combining them into `intermediateResources`), and only clean up the latter. That would make the code less hacky.

e2e/regression/regression_test.go
miampf: Depends on the file order since it loops alphabetically, but in principle yes. I can `continue` instead if you want.

e2e/regression/regression_test.go
miampf: That is true. Initially I had everything split up in separate files, however, IIRC @katexochen requested everything to remain in one file for now. https://github.com/edgelesssys/contrast/pull/1549 will explicitly use multiple files. I can also revert and use multiple files in this PR already, or make a follow up PR if this is wanted, but I'd like @katexochen's input about that :)

katexochen: We can test multiple files are working by rebasing the other PR. Not changing existing tests keeps the diff small. Also, we want to test we can handle multiple resources in one file.

burgerdev: sgtm

e2e/regression/regression_test.go
burgerdev: Since we're not going for the multi-file support in this PR, let's shelve this for now.

e2e/regression/regression_test.go
```diff
@@ -60,83 +62,82 @@ func TestRegression(t *testing.T) {
...
-						t.Log("error creating client for resource deletion: ", err)
-					}
-					if err := client.Delete(ctx, resourceName, metav1.DeleteOptions{
-						PropagationPolicy: &bgDeletion,
-					}); err != nil {
-						t.Log("error deleting resource: ", err)
-					}
+				// Check if we are testing a cron job
+				unmarshalledYAML := make(map[string]any)
+				require.NoError(yaml.Unmarshal(resourceYAML, &unmarshalledYAML))
```
burgerdev: This seems to assume a single YAML doc per file, which is not necessarily the case. Also, we already deserialized the file above. Can't we just loop over `newResources` instead?

e2e/regression/regression_test.go
```diff
@@ -267,3 +268,29 @@ func runCommandImpl(ctx context.Context, cmd, namespace, workDir string) error {
 
 	return errors.New("unknown command: " + cmd)
 }
+
+func cleanupResource(ctx context.Context, resource any, ct *contrasttest.ContrastTest) error {
+	unstructuredResources, err := kuberesource.ResourcesToUnstructured([]any{resource})
+	if err != nil {
+		return err
+	}
+	bgDeletion := metav1.DeletePropagationBackground
```
burgerdev: Why are we doing background deletion? Doesn't this introduce potential for interference between test dirs? At least add a comment here explaining the reason.

miampf: Was a remainder of the original regression test. I don't remember my reasoning back then and it also seems to work fine with `DeletePropagationForeground`, so I just switched to that :)

burgerdev (approved):

PR #1539, opened 2025-06-23
tools/bm-maintenance: only use one host mount

davidweisse:
In cases where no `/var/lib/nydus-snapshotter` directory is on the host (since we don't use it anymore), the volume mount would fail. This simplifies the deployment with only one `/host` mount, so if `/var/lib/nydus-snapshotter` does not exist, nothing happens.

### DIFF
--- a/packages/cleanup-bare-metal.sh
+++ b/packages/cleanup-bare-metal.sh
@@ -39,7 +39,7 @@ mapfile -t unusedRuntimeClasses < <(
         tr ' ' '\n' |
         grep '^contrast-cc' || true
       # Get all (maybe already deleted) runtime classes with a reference in /opt/edgeless
-      ls -1 "${OPTEDGELESS}"
+      ls -1 "${HOST_MOUNT:?}/opt/edgeless"
     } | sort -u
   )
 )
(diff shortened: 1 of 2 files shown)


packages/cleanup-bare-metal.sh
```diff
@@ -39,7 +39,7 @@ mapfile -t unusedRuntimeClasses < <(
         tr ' ' '\n' |
         grep '^contrast-cc' || true
       # Get all (maybe already deleted) runtime classes with a reference in /opt/edgeless
-      ls -1 "${OPTEDGELESS}"
+      ls -1 "${HOST_MOUNT:?}/opt/edgeless"
```
katexochen, line 42: What are you trying to archive with `:?`? As far as I understand, this if for throwing a message if undefined, but you aren't giving any message.

> ${foo:?message}	Show error message and exit if $foo is unset (or null)
https://devhints.io/bash#parameter-expansions

davidweisse: If the error message is not set, `${foo:?}` will just print `parameter null or not set` if it is null or not set.

PR #1542, opened 2025-06-24
coordinator: don't fail liveness probe if Kubernetes API server is unavailable

burgerdev:
If the Kubernetes API becomes unavailable, the Coordinator will fail the liveness probe and be restarted eventually. This defeats the point of automatic peer recovery, because it is very likely that all Coordinators are experiencing the API downtime at the same time, so will all be restarted and not have any peers to recover from.

This PR replaces the liveness probe with the startup probe, because there are no liveness conditions within the Coordinator that do not explicitly lead to process termination. Checking the startup probe again should suffice for detecting network or VM problems. This is aligned with [k8s best practices](https://codefresh.io/learn/kubernetes-management/kubernetes-liveness-probes/), h/t @thomasten. 

Bundled with this change, the startup probe now includes one round of peer recovery, as was [originally planned](https://github.com/edgelesssys/contrast/blob/177a84f03ed9c00cb4df8d881c62ceab3b4af345/rfc/010-distributed-coordinator.md?plain=1#L113).

### DIFF
--- a/coordinator/internal/peerrecovery/recovery.go
+++ b/coordinator/internal/peerrecovery/recovery.go
@@ -70,7 +70,7 @@ func New(guard guard, peerGetter peerGetter, issuer atls.Issuer, httpsGetter *ce
 // The function returns only when the context expires, with the error returned from the context.
 func (r *Recoverer) RunRecovery(ctx context.Context) error {
 	return periodically(ctx, r.clock, peerRecoveryInterval, func(ctx context.Context) {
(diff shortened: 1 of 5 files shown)


burgerdev:
Following @thomasten's suggestion, I added the liveness probe again, but it's now backed by the startup probe handler and mostly verifies network connectivity to the Coordinator process.

katexochen (approved):

PR #1544, opened 2025-06-25
kata.genpolicy: fix EphemeralVolumeSource

katexochen:
We accidentally used  `PersistentVolumeClaim` instead of `PersistentVolumeClaimTemplate` as payload of `EphemeralVolumeSource`. This lead to a bug where `PersistentVolumeClaimTemplate` without a `ObjectMeta` couldn't be unmarshaled by genpolicy. While the requirement of `ObjectMeta` on `PersistentVolumeClaim` makes sense, it is fully optional for `PersistentVolumeClaimTemplate`.

<details>
<summary>Tested with the following resource, which is based on https://github.com/edgelesssys/contrast/pull/1254/files.</summary>

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ephemeral-volume
  namespace: custom-p
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: ephemeral-volume
  template:
    metadata:
      labels:
        app.kubernetes.io/name: ephemeral-volume
    spec:
      containers:
        - name: ephemeral-volume
          image: quay.io/quay/busybox@sha256:92f3298bf80a1ba949140d77987f5de081f010337880cd771f7e7fc928f8c74d
          command:
            - /bin/sh
            - -ec
            - |
              mkfs.ext2 /dev/foo
              mount /dev/foo /tmp
              tail -f /dev/null
          securityContext:
            privileged: true
          volumeDevices:
            - name: ephemeral
              devicePath: /dev/foo
          resources:
            requests:
              memory: 400Mi
            limits:
              memory: 400Mi
      volumes:
        - name: ephemeral
          ephemeral:
            volumeClaimTemplate:
              # metadata:
              #   labels:
              #     foo: bar
              spec:
                accessModes:
                  - ReadWriteOnce
                resources:
                  requests:
                    storage: 1Mi
                volumeMode: Block
      runtimeClassName: contrast-cc
```

</details>


packages/by-name/kata/kata-runtime/0016-genpolicy-support-ephemeral-volume-source.patch
```diff
@@ -53,11 +74,11 @@ index 0bb908a81c2fd7ffcd7dd59587db316f609d0c13..8592b1f9e69ac054b3e06c3778ad0879
...
  }
  
-+/// See Reference / Kubernetes API / Config and Storage Resources / Volume.
++/// See Reference / Kubernetes API / Config and Storage Resources / EphemeralVolumeSource.
 +#[derive(Clone, Debug, Serialize, Deserialize)]
 +pub struct EphemeralVolumeSource {
-+    #[serde(skip_serializing_if = "Option::is_none")]
-+    pub volumeClaimTemplate: Option<persistent_volume_claim::PersistentVolumeClaim>,
++    // Required.
++    pub volumeClaimTemplate: Option<persistent_volume_claim::PersistentVolumeClaimTemplate>,
```
burgerdev: Should we just remove the `Option`, then?

packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -170,7 +170,7 @@ buildGoModule (finalAttrs: {
       # This patch fixes an issue where genpolicy can corrupt the layer cache file due to simultaneous
       # read/write operations on the file, instead implementing an in-memory caching solution.
       # Upstream PR: https://github.com/kata-containers/kata-containers/pull/11426
-      ./0026-genpolicy-keep-layers-cache-in-memory.patch
+      ./0026-genpolicy-keep-layers-cache-in-memory-to-prevent-cor.patch
```
burgerdev, line 173: Interesting - do you know why this happened?

burgerdev (approved):

packages/by-name/kata/kata-runtime/package.nix
katexochen: @charludo named it by hand when it was added.

PR #1552, opened 2025-06-26
imagepuller: init

charludo:
This is still an early draft.

- Open questions:
  - [ ] why are calls to `ghcr.io` failing after the first 2-3? (Rate limiting?)
  - [ ] what is the cause of the `Failed to apply layer sha256xxxxx: that ID is already in use` errors?
  - [ ] can the kata-runtime patch be handled in a nicer way?

- ToDo:
  - [ ] solve questions above
  - [ ] refactor `imagepuller/main.go`
  - [ ] tests


imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,123 @@
...
+func (s *ImagePullerService) PullImage(ctx context.Context, r *api.ImagePullRequest) (*api.ImagePullResponse, error) {
+	log := s.Logger.With(
+		slog.String("image_url", r.ImageUrl),
+		slog.String("mount_path", r.BundlePath),
+	)
+	log.Info("Handling image pull request")
+
+	ref, err := name.ParseReference(r.ImageUrl)
+	if err != nil {
+		return nil, s.Error("Failed to parse the image URL as a references", err)
```
katexochen:
```suggestion
		return nil, s.Error("parsing the image URL as a references", err)
```
Error messages should be lower case, start with a verb, not contain fail etc

imagepuller/main.go
```diff
@@ -0,0 +1,121 @@
...
+		return
+	}
+	flag.Parse()
+
+	log = slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
+	}))
+
+	if err := run(); err != nil {
+		os.Exit(1)
```
katexochen, line 43:
```suggestion
    log.Fatal(err)
```

imagepuller/main.go
```diff
@@ -0,0 +1,121 @@
...
+
+var (
+	version = "0.0.0-dev"
+	log     *slog.Logger
+)
+
+func main() {
+	if reexec.Init() {
+		return
+	}
```
katexochen, line 35: What is happening here? This should have a comment.

imagepuller/main.go
```diff
@@ -0,0 +1,121 @@
...
+	}))
+
+	if err := run(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	if err := os.MkdirAll(filepath.Dir(api.Socket), os.ModePerm); err != nil {
+		return logErrAndReturn("Failed to create directory for socket", err)
```
katexochen: just return the error instead.

imagepuller/main.go
```diff
@@ -0,0 +1,121 @@
...
+	defer os.RemoveAll(api.Socket)
+
+	s, err := ttrpc.NewServer()
+	if err != nil {
+		return logErrAndReturn("Failed to create ttRPC server", err)
+	}
+	defer s.Close()
+
+	sigintCh := make(chan os.Signal, 1)
+	signal.Notify(sigintCh, syscall.SIGINT, syscall.SIGTERM)
```
katexochen: Prefere `signal.NotifyContext`

imagepuller/main.go
```diff
@@ -0,0 +1,121 @@
...
+
+	errCh := make(chan error, 1)
+	api.RegisterImagePullServiceService(s, &service.ImagePullerService{
+		Error: func(msg string, err error) error {
+			errCh <- fmt.Errorf("%s: %w", msg, err)
```
katexochen: Why do we need a chan here? We can just put the logger into service.

packages/by-name/imagepuller/package.nix
```diff
@@ -0,0 +1,43 @@
...
+  src =
+    let
+      inherit (lib) fileset path hasSuffix;
+      root = ../../../.;
+    in
+    fileset.toSource {
+      inherit root;
+      fileset = fileset.unions [
+        (path.append root "go.mod")
+        (path.append root "go.sum")
```
katexochen: why do we need these?

packages/nixos/imagepuller.nix
```diff
@@ -0,0 +1,26 @@
...
+
+{
+  pkgs,
+  lib,
+  ...
+}:
+
+{
+  config = {
+    systemd.services.imagepuller = {
```
katexochen: maybe put it into kata.nix, add a config option imagePull and add this unit if value is "guest" .

imagepuller/main.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/edgelesssys/contrast/imagepuller/internal/api"
+	"github.com/edgelesssys/contrast/imagepuller/internal/service"
+)
+
+var (
+	version = "0.0.0-dev"
+	log     *slog.Logger
+)
+
+func main() {
```
katexochen, line 29:
```suggestion
func main() {
	fmt.Fprintf(os.Stderr, "Contrast image-puller %s\n", version)
	fmt.Fprintln(os.Stderr, "Report issues at https://github.com/edgelesssys/contrast/issues")
```

imagepuller/client/main.go
katexochen, line 1: Should the client be in-tree, or is this only for testing?

charludo: Just for testing.

imagepuller/main.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/containerd/ttrpc"
+	"github.com/containers/storage/pkg/reexec"
+	"github.com/edgelesssys/contrast/imagepuller/internal/api"
+	"github.com/edgelesssys/contrast/imagepuller/internal/service"
+)
+
+var (
+	version = "0.0.0-dev"
+	logger  *slog.Logger
```
katexochen: We shouldn't use a global var for this, rather pass it via function arg or/and put it as field into the service struct.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		log.Error("parsing the image URL as a references", "err", err)
+		return nil, err
+	}
+	remoteImg, err := remote.Image(ref, remote.WithContext(ctx))
+	if err != nil {
+		if errors.Is(err, context.Canceled) || errors.Is(ctx.Err(), context.Canceled) {
+			log.Error("pull aborted (context cancelled)", "err", ctx.Err())
+			return nil, err
+		}
```
katexochen: this can be un-nested (if .. else if ..)

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,148 @@
...
+		slog.String("image_url", r.ImageUrl),
+		slog.String("mount_path", r.BundlePath),
+	)
+	log.Info("handling image pull request")
+
+	ref, err := name.ParseReference(r.ImageUrl)
+	if err != nil {
+		log.Error("parsing the image URL as a references", "err", err)
+		return nil, err
+	}
```
katexochen:
```suggestion
// PullImage is a ttRPC service which pulls and mounts docker images.
func (s *ImagePullerService) PullImage(ctx context.Context, r *api.ImagePullRequest) (resp *api.ImagePullResponse, retErr error) {
	log := s.Logger.With(
		slog.String("image_url", r.ImageUrl),
		slog.String("mount_path", r.BundlePath),
	)
	log.Info("handling image pull request")

	defer func() {
		if retErr != nil {
			log.Error("request failed", "err", retErr)
		}
	}()

	ref, err := name.ParseReference(r.ImageUrl)
	if err != nil {
		return nil, fmt.Errorf("parsing the image URL as a references: %w", err)
	}
```

imagepuller/client/main.go
charludo, line 1: Will be removed, included for testing

internal/kuberesource/parts.go
```diff
@@ -331,7 +331,7 @@ func Coordinator(namespace string) *CoordinatorConfig {
 									WithPath("/probe/readiness")),
 							).
 							WithResources(ResourceRequirements().
-								WithMemoryLimitAndRequest(200),
+								WithMemoryLimitAndRequest(4000),
```
charludo: Probably way too much, but it seems it will need *some* amount of increase to account for pulling/mounting images

burgerdev: This value was sufficient when pulling with `image-rs` - why do you expect it to be insufficient for the imagepuller?

imagepuller/internal/api/imagepuller.proto
```diff
@@ -0,0 +1,25 @@
...
+
+package api;
+
+option go_package = "github.com/edgelesssys/contrast/imagepuller/internal/api";
+
+service ImagePullService {
+  rpc PullImage (ImagePullRequest) returns (ImagePullResponse);
+}
+
+  // source: https://github.com/kata-containers/kata-containers/blob/997a1f35ab68ad8cc2b24e8e20ea63f7de82acf9/src/libs/protocols/protos/confidential_data_hub.proto
```
burgerdev: Might make more sense to move the source ref to the top of the file.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+	}()
+
+	ref, err := name.ParseReference(r.ImageUrl)
+	if err != nil {
+		return nil, fmt.Errorf("parsing the image URL as a references: %w", err)
+	}
+
+	tr := transport.NewRetry(remote.DefaultTransport)
+	remoteImg, err := remote.Image(ref, remote.WithContext(ctx), remote.WithTransport(tr))
+	if errors.Is(err, context.Canceled) || errors.Is(ctx.Err(), context.Canceled) {
```
burgerdev: I think that the second condition is not needed:

1. If the context expired during the call, the returned error is likely to wrap that context error.
2. If it only expired afterwards, we could just as well continue until the next remote call.

You could also consider `context.Timeout` as a special error case here.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+		RunRoot:        filepath.Join(os.TempDir(), "run"),
+		GraphRoot:      filepath.Join(os.TempDir(), "graph"),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("opening storage: %w", err)
+	}
+
+	var parentIDs []string
+	for idx, layer := range layers {
+		rc, err := layer.Compressed() // Uncompressed?
```
burgerdev: Interestingly, the `stream` package does not even implement `Uncompressed`: https://github.com/google/go-containerregistry/blob/59a4b85930392a30c39462519adc8a2026d47181/pkg/v1/stream/layer.go#L128

Now I do wonder what happens if a layer is not compressed to begin with, which is supported by the spec but uncommon, afaik.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return nil, fmt.Errorf(fmt.Sprintf("reading layer %d", idx), err)
+		}
+
+		// Wrap in func to close each rc immediately
+		err = func() error {
+			defer rc.Close()
+			putLayer, n, err := store.PutLayer(
+				"",        // empty ID -> let store decide
+				"",        // parent == top of parentIDs (empty for first)
```
burgerdev: I don't understand the "empty for first" - it looks like it's always empty?

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+				parentIDs, // full parent chain
+				"",        // mount label
+				false,     // readonly
+				nil,       // mount options
+				rc,        // tar stream
+			)
+			if err != nil {
+				return err
+			}
+			log.Info("Applied layer", "id", putLayer.ID, "bytes", n)
```
burgerdev:
```suggestion
			log.Info("Applied layer", "id", putLayer.ID, "size", n)
```
I half expected `bytes` to be the actual content.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+			parentIDs = append(parentIDs, putLayer.ID)
+			return nil
+		}()
+		if errors.Is(err, context.Canceled) || errors.Is(ctx.Err(), context.Canceled) {
```
burgerdev: dito

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+			return nil, fmt.Errorf("pull aborted while applying layer (context cancelled): %w, context error: %w", err, ctx.Err())
+		} else if err != nil {
+			return nil, fmt.Errorf("applying layer: %w", err)
+		}
+	}
+
+	if len(parentIDs) == 0 {
+		return nil, fmt.Errorf("no layers found (empty image?)")
```
burgerdev: That is technically not an error condition (`FROM scratch`), although I don't see how this could happen in practice.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	log.Info("Created container", "id", container.ID)
+
+	mountPoint, err := store.Mount(container.ID, "")
+	if err != nil {
+		return nil, fmt.Errorf("mounting container: %w", err)
+	}
+	log.Debug("Mounted in tmpdir", "mountPoint", mountPoint)
+
+	mode := os.FileMode(int(755))
```
burgerdev: This should be octal, right?

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+	imagePath := filepath.Join(r.BundlePath, "rootfs")
+	if err := os.MkdirAll(imagePath, mode); err != nil {
+		return nil, fmt.Errorf("creating bundle path: %w", err)
+	}
+	if err := unix.Unmount(imagePath, 0); err == nil {
```
burgerdev: Why is that necessary? Not sure whether we should care about what the agent did so far to the rootfs.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,142 @@
...
+		log.Warn("Unmounted existing mount")
+	} else if !errors.Is(err, unix.EINVAL) {
+		return nil, fmt.Errorf("unmounting existing mount: %w", err)
+	}
+
+	if err := unix.Mount(mountPoint, imagePath, "", unix.MS_BIND, ""); err != nil {
+		return nil, fmt.Errorf(fmt.Sprintf("moving mount %s to %s", mountPoint, imagePath), err)
```
burgerdev: Nit:
```suggestion
		return nil, fmt.Errorf(fmt.Sprintf("binding mount %s to %s", mountPoint, imagePath), err)
```

packages/nixos/kata.nix
```diff
@@ -99,5 +100,20 @@ in
       "dummy file, to be bind-mounted by the Kata agent when writing network configuration";
     environment.etc."kata-opa/default-policy.rego".source =
       "${pkgs.kata-runtime.src}/src/kata-opa/allow-set-policy.rego";
+
+    systemd.services.imagepuller = lib.mkIf cfg.guestImagePull {
+      description = "Image Puller";
+      documentation = [ "https://github.com/edgelesssys/contrast" ];
+      wants = [ "network.target" ];
```
burgerdev:
```suggestion
      after = [ "network.target" ];
```
See https://systemd.io/NETWORK_ONLINE/

internal/kuberesource/parts.go
burgerdev: Old pending comment, sry.

imagepuller/internal/service/service.go
charludo: Whoops, yes, remnant from before code changes

imagepuller/internal/service/service.go
charludo: During testing, when the pull would be restarted, sometimes an image was already mounted where it was tried again, and then multiple mounts where pointing to the same dir. I have not seen this happen in a while, and I do believe it could be removed. Up to you.

imagepuller/internal/service/service.go
charludo: > `context.Timeout`

`context.DeadlineExceeded`?

imagepuller/internal/service/service.go
charludo: What would you recommend doing in this case?

burgerdev: You might be able to just continue with the empty string: https://github.com/containers/storage/blob/19ffd5bbbcceaba60eb9f60a7b361c61dcf50d13/store.go#L1569.

imagepuller/internal/service/service.go
burgerdev: As long as we don't know for sure that _we_ mounted something there, I don't think we should be unmounting.

charludo: Alright. Should we just mount over whatever might be mounted, or return an error?

burgerdev: Just mount over it - we were asked by the agent to mount the image there, and that's what we should do.

charludo:
@burgerdev https://github.com/edgelesssys/contrast/pull/1552/commits/91befc46d5a9b33da7f1567a186c31077eabb358

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,161 @@
...
+			log.Error("request failed", "err", retErr)
+			response = nil
+		}
+	}()
+
+	ref, err := name.ParseReference(r.ImageUrl)
+	if err != nil {
+		return nil, fmt.Errorf("parsing the image URL as a references: %w", err)
+	}
+	digestExpected := ref.Identifier()
```
katexochen: Looking at the implementation of `ParseReference`, `Identifier` will be tag and not digest if the image has a tag. Maybe we want to call NewDigest directly?

packages/nixos/system.nix
```diff
@@ -73,6 +73,7 @@
     );
 
   networking.firewall.enable = false;
+  networking.dhcpcd.enable = false;
```
katexochen, line 79: Please document why in a comment.

packages/by-name/imagepuller/package.nix
```diff
@@ -0,0 +1,41 @@
...
+# SPDX-License-Identifier: BUSL-1.1
+
+{ lib, buildGoModule }:
+
+buildGoModule (finalAttrs: {
+  pname = "imagepuller";
+  version = builtins.readFile ../../../version.txt;
+
+  # The source of the main module of this repo. We filter for Go files so that
+  # changes in the other parts of this repo don't trigger a rebuild.
```
katexochen: Comment is wrong, this isn't the main module.

packages/by-name/imagepuller/package.nix
```diff
@@ -0,0 +1,41 @@
...
+  src =
+    let
+      inherit (lib) fileset path hasSuffix;
+      root = ../../../.;
```
katexochen: maybe set root directly to imagepuller dir if we don't need anything outside?

imagepuller/main.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return fmt.Errorf("creating ttRPC server: %w", err)
+	}
+	defer s.Close()
+
+	api.RegisterImagePullServiceService(s, &service.ImagePullerService{Logger: log})
+	eg, ctx := errgroup.WithContext(ctxSignal)
+
+	eg.Go(func() error {
+		log.Info("Started image-puller", "version", version, "socket", api.Socket)
```
katexochen: nit: we already printed the version before

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,161 @@
...
+	log := s.Logger.With(
+		slog.String("image_url", r.ImageUrl),
+		slog.String("bundle_path", r.BundlePath),
+	)
+	log.Info("Handling image pull request")
+
+	defer func() {
+		if retErr != nil {
+			log.Error("request failed", "err", retErr)
+			response = nil
```
katexochen:
```suggestion
```

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,161 @@
...
+func (s *ImagePullerService) PullImage(ctx context.Context, r *api.ImagePullRequest) (response *api.ImagePullResponse, retErr error) {
+	log := s.Logger.With(
+		slog.String("image_url", r.ImageUrl),
+		slog.String("bundle_path", r.BundlePath),
+	)
+	log.Info("Handling image pull request")
+
+	defer func() {
+		if retErr != nil {
+			log.Error("request failed", "err", retErr)
```
katexochen:
```suggestion
			log.Error("Request failed", "err", retErr)
```

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,161 @@
...
+	manifest, err := remoteImg.Manifest()
+	if err != nil {
+		return nil, fmt.Errorf("obtaining image manifest: %w", err)
+	}
+
+	var parentIDs []string
+	for idx, layer := range layers {
+		rc, err := layer.Compressed()
+		if err != nil {
+			return nil, fmt.Errorf(fmt.Sprintf("reading layer %d", idx), err)
```
katexochen:
```suggestion
			return nil, fmt.Errorf("reading layer %d: %w", idx, err)
```

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,161 @@
...
+				parentIDs, // full parent chain
+				"",        // mount label
+				false,     // readonly
+				nil,       // mount options
+				rc,        // tar stream
+			)
+			if err != nil {
+				return errors.Join(err, rc.Close())
+			}
+			if err := rc.Close(); err != nil {
```
katexochen, line 76: nit: errors could use some wrapping info

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,161 @@
...
+
+	mountPoint, err := store.Mount(container.ID, "")
+	if err != nil {
+		return nil, fmt.Errorf("mounting container: %w", err)
+	}
+	log.Debug("Mounted in tmpdir", "mountPoint", mountPoint)
+
+	mode := os.FileMode(int(0o755))
+	imagePath := filepath.Join(r.BundlePath, "rootfs")
+	if err := os.MkdirAll(imagePath, mode); err != nil {
```
katexochen:
```suggestion
	imagePath := filepath.Join(r.BundlePath, "rootfs")
	if err := os.MkdirAll(imagePath, 0o755); err != nil {
```

imagepuller/internal/service/service.go
charludo: I retract my previous comment. Is using a tag instead of a digest even supported in our usecase?

If we only ge a tag, then the only thing we can get a digest from is the remote image, whose digest we need to validate... Or am I missing something?

imagepuller/internal/service/service.go
charludo: Ah yep, the other two uses were just used in the re-mount logic

imagepuller/internal/service/service.go
katexochen: Exactly. So we should rather fail than pulling a tag we can't verify.

charludo: Ah, got it. 

(Though pulling would already be prevented here, just with an unhelpful error)

internal/kuberesource/sets.go
```diff
@@ -720,7 +720,7 @@ func GPU() []any {
 					WithContainers(
 						Container().
 							WithName("gpu-tester").
-							WithImage("ghcr.io/edgelesssys/contrast/ubuntu:24.04").
+							WithImage("ghcr.io/edgelesssys/contrast/ubuntu@sha256:b0c08a4b639b5fca9aa4943ecec614fe241a0cebd1a7b460093ccaeae70df698").
```
katexochen: we should keep the tag

charludo: There's a second usage of this exact image further up, should I add the tag there as well?

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+		}
+
+		if digestFound == "" {
+			return nil, fmt.Errorf("obtaining image digest for linux/amd64: platform missing from image index")
+		}
+
+		log.Info("Obtained actual image digest, re-executing image-puller with it", "actual_image_digest", digestFound)
+		r.ImageUrl = strings.Replace(r.ImageUrl, digestExpected, digestFound, 1)
+		return s.PullImage(ctx, r)
```
burgerdev: The recursion here is a bit confusing. How about a helper function 

```go
func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, ref string) (v1.Image, error) 
```

that 

1. extracts the expected digest from the reference to `expectedDigest`
1. tries index first, if successful
   1.  verifies index digest == `expectedDigest`
   2. looks for `linux/amd64` image and assigns its digest to `expectedDigest`
   3. sets `img := index.Image(expectedDigest)`
1. if not successful, `img := remote.Image(ref)`
2. verify that `img.Digest() == expectedDigest`.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,218 @@
...
+
+	layers, err := remoteImg.Layers()
+	if err != nil {
+		return nil, fmt.Errorf("obtaining the image layers: %w", err)
+	}
+
+	store, err := storage.GetStore(types.StoreOptions{
+		TransientStore: true,
+		RunRoot:        filepath.Join(os.TempDir(), "run"),
+		GraphRoot:      filepath.Join(os.TempDir(), "graph"),
```
burgerdev: I believe we should use real temporary directories from `os.MkdirTemp` here, but reuse them throughout the process lifetime.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	log.Info("Created container", "id", container.ID)
+
+	mountPoint, err := store.Mount(container.ID, "")
+	if err != nil {
+		return nil, fmt.Errorf("mounting container: %w", err)
+	}
+	log.Debug("Mounted in tmpdir", "mountPoint", mountPoint)
+
+	imagePath := filepath.Join(r.BundlePath, "rootfs")
```
burgerdev: Nit: consider renaming the variable `rootfs`, imho `imagePath` sounds more like the source (or a lowerdir of it) in this case.

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := os.MkdirAll(imagePath, 0o755); err != nil {
+		return nil, fmt.Errorf("creating bundle path: %w", err)
+	}
+
+	if err := unix.Mount(mountPoint, imagePath, "", unix.MS_BIND, ""); err != nil {
+		return nil, fmt.Errorf(fmt.Sprintf("binding mount %s to %s", mountPoint, imagePath), err)
```
burgerdev:
```suggestion
		return nil, fmt.Errorf("binding mount %s to %s: %w", mountPoint, imagePath, err)
```

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,218 @@
...
+func newDigest(name string) (string, error) {
+	parts := strings.Split(name, "@")
+	if len(parts) != 2 {
+		return "", fmt.Errorf("a digest must contain exactly one '@' separator (e.g. registry/repository@digest) saw: %s", name)
+	}
+	dig := parts[1]
+	prefix := digest.Canonical.String() + ":"
+	if !strings.HasPrefix(dig, prefix) {
+		return "", fmt.Errorf("unsupported digest algorithm: %s", dig)
+	}
```
burgerdev: This is too strict, imho. What if someone were to use `sha512` (not sure whether it's implemented right now in go-containerregistry, but it might be eventually)? Can we just use the digest and trust the libraries to do the right thing?

imagepuller/internal/service/service.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := unix.Mount(mountPoint, imagePath, "", unix.MS_BIND, ""); err != nil {
+		return nil, fmt.Errorf(fmt.Sprintf("binding mount %s to %s", mountPoint, imagePath), err)
+	}
+	log.Info("Pulled and mounted image", "mount_path", imagePath)
+
+	return &api.ImagePullResponse{}, nil
+}
+
+func newDigest(name string) (string, error) {
+	parts := strings.Split(name, "@")
```
burgerdev: I'd feel better if we used a library for reference parsing. How about https://pkg.go.dev/github.com/distribution/reference#Parse?

imagepuller/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+	eg.Go(func() error {
+		<-ctx.Done()
+		if ctxSignal.Err() != nil {
+			log.Info("Received signal, shutting down.")
+		} else {
+			log.Info("Context done, shutting down", "err", ctx.Err())
+		}
+
+		if err := s.Shutdown(ctxSignal); err != nil {
```
burgerdev: If we were signaled, the context is already expired. Maybe create a fresh one with a low timeout for cleanup, as in https://github.com/edgelesssys/contrast/blob/7acc59fda34d8b6620565694c73d3e0077d0f0bc/coordinator/main.go#L251-L252

packages/by-name/kata/kata-runtime/package.nix
```diff
@@ -155,6 +155,11 @@ buildGoModule (finalAttrs: {
       # under these volumes, by failing the policy generation if volumes without mounts are found.
       # TODO(burgerdev): open upstream issue after disclosure.
       ./0020-genpolicy-don-t-allow-mount-storage-for-declared-VOL.patch
+
+      # We are starting our own imagepuller instead of the upstream one. These changes avoid invoking the CDH,
+      # instead using a slightly modified version of upstream's PullImage ttRPC client implementation to
+      # communicate with our imagepuller ttRPC server.
+      ./0021-agent-use-custom-implementation-for-image-pulling.patch
```
burgerdev: The docstring misses the long-term plan for this patch.

https://github.com/edgelesssys/contrast/blob/7acc59fda34d8b6620565694c73d3e0077d0f0bc/dev-docs/patches.md?plain=1#L112-L118

I think referring to https://github.com/kata-containers/kata-containers/issues/11532 should be sufficient.

imagepuller/internal/service/service.go
charludo: Just realized that the `NewDigest` func the implementation above weas imitating is indeed being exported from go-containerreigstry, so using that now. However, that one also hardcoded the prefix, but since it is used internally by e.g. our layer validation anyways, I don't really see a way around this.

imagepuller/internal/service/service.go
charludo: I extracted the function as you suggested, but with some deviations:

- it's still recursive... mostly in order to not have to re-do the `ParseReferece` logic. If that's preferred though, let me know.
- adding to that, `getAndVerifyImage` receives the image url, not the parsed ref

It's certainly an improvement over the previous iteration, but I'm not really happy with the logger situation yet: I'd like to add the actual image ref as a structured attribute, but can not think of a "nice" way to do so without having to pass the logger around and return it from `getAndVerifyImage`. Is there an idiomatic way in go to handle this?

(Adding an attribute to the `PullImageRequest` struct would probably be nice, except that then the logger setup is shared between ttrpc requests...)

imagepuller/internal/service/verify.go
```diff
@@ -0,0 +1,84 @@
...
+	ref, err := name.ParseReference(imageURL)
+	if err != nil {
+		return nil, fmt.Errorf("parsing the image URL as a reference: %w", err)
+	}
+
+	digestExpectedRaw, err := name.NewDigest(imageURL)
+	if err != nil {
+		return nil, fmt.Errorf("parsing image digest: %w", err)
+	}
+	digestExpected := digestExpectedRaw.DigestStr()
```
burgerdev: `name.Digest` implements `name.Reference` - I think we could skip the `name.ParseReference` here.

```suggestion
	ref, err := name.NewDigest(imageURL)
	if err != nil {
		return nil, fmt.Errorf("parsing image digest: %w", err)
	}
	digestExpected := ref.DigestStr()
```

imagepuller/internal/service/verify.go
```diff
@@ -0,0 +1,84 @@
...
+		log.Info("Obtained actual image digest", "image_url_linux", ImageURLLinux)
+		return s.getAndVerifyImage(ctx, log, ImageURLLinux)
+	}
+
+	remoteImg, err := remote.Image(ref, remote.WithContext(ctx), remote.WithTransport(tr))
+	if errors.Is(err, context.DeadlineExceeded) {
+		return nil, fmt.Errorf("pull aborted (deadline exceeded): %w", err)
+	} else if err != nil {
+		return nil, fmt.Errorf("accessing the remote image URL: %w", err)
+	}
```
burgerdev: You're doing the recursion again. Instead, you could simply do

```suggestion
		digestExpected = digestFound
		
		remoteImg, err = remoteImgIndex.Image(digestFound) // actually, arg needs to be `m.Digest`
	} else {
		remoteImg, err = remote.Image(ref, remote.WithContext(ctx), remote.WithTransport(tr))
		if err != nil {
			return nil, fmt.Errorf("accessing the remote image URL: %w", err)
		}
	}
```

imagepuller/internal/service/verify.go
charludo: Oh, great catch!

imagepuller/internal/service/verify.go
charludo: Ah! Wasn't aware about remoteImgIndex.Index, thank you!

katexochen (approved):

charludo:
🎉

PR #1555, opened 2025-06-26
node-installer: disable config overrides via annotations

katexochen:
Disable all annotations, as we don't support these. Some will mess up measurements, others bypass things you can archive via correct resource declaration anyway.


burgerdev (approved):
Thanks. Can you please update https://docs.edgeless.systems/contrast/features-limitations#overriding-kata-configuration, too?

burgerdev (approved):

PR #1563, opened 2025-07-03
coordinator: don't allow user recovery when ready peers are present

burgerdev:
Recovery via userapi resets the mesh CA key, whereas recovery via peers does not. If user recovery is triggered while there are ready Coordinators, the cluster enters an inconsistent state.

In order to avoid such situations, we reject user recovery when peers are available, because we still expect the Coordinator to recover automatically. For situations where this does not happen, we include a `--force` flag for the CLI and a corresponding argument for the Recover method that overrides this check.


coordinator/internal/userapi/userapi.go
```diff
@@ -187,6 +194,17 @@ func (s *Server) Recover(ctx context.Context, req *userapi.RecoverRequest) (*use
...
+	if !req.Force {
+		// Unless forced, check whether recovery is a good idea.
+		peers, err := s.discovery.GetPeers(ctx)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "listing peers: %v", err)
+		}
+		if len(peers) > 0 {
+			return nil, status.Errorf(codes.FailedPrecondition, "rejecting user recovery because %d recovered peers are available", len(peers))
+		}
+	}
```
katexochen, line 208: Maybe log that force was used?

burgerdev: Added a log line

PR #1566, opened 2025-07-03
microsoft.kata-runtime: 3.2.0.azl5 -> 3.15.0.aks0

katexochen:


packages/by-name/microsoft/kata-runtime/package.nix
```diff
@@ -6,17 +6,82 @@
...
+      # We could think about moving the specific path into the settings and set it
+      # to the expected value.
+      #
+      # This patch is not upstreamable.
+      ./0001-genpolicy-regex-check-contrast-specific-layer-src-pr.patch
+      # This patch builds on top of the Azure CSI patches specific to the msft
+      # version of genpolicy. Therefore, we don't attempt to upstream those changes.
+      # We can revisit this if microsoft upstreamed
+      # https://github.com/microsoft/kata-containers/pull/174
+      # TODO
```
burgerdev: Leftover?

burgerdev (approved):

packages/by-name/microsoft/kata-runtime/package.nix
katexochen: Ah yes, there was some confusion with that patch.. But I think it's okay now.

PR #1571, opened 2025-07-08
e2e/gpu: check for mounted libs/binaries

katexochen:
The libraries we expose into the container are relied upon by the user, so we need to ensure these are stable, or at least detect changes to warn about them. Note this PR doesn't include library ABI version stability guarantees, which is something we could think about checking, too.

### DIFF
--- a/e2e/gpu/gpu_test.go
+++ b/e2e/gpu/gpu_test.go
@@ -8,7 +8,11 @@ package gpu
 import (
 	"context"
 	"flag"
+	"fmt"
 	"os"
+	"path"
+	"regexp"
+	"strings"
 	"testing"
 	"time"
 
@@ -16,7 +20,9 @@ import (
 	"github.com/edgelesssys/contrast/internal/kuberesource"
 	"github.com/edgelesssys/contrast/internal/manifest"
 	"github.com/edgelesssys/contrast/internal/platforms"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


e2e/gpu/gpu_test.go
```diff
@@ -62,10 +69,121 @@ func TestGPU(t *testing.T) {
...
+
+			// Run `ls` to check what libraries with that name exist.
+			getLibsCmd := fmt.Sprintf("ls %s*", pathThisLib)
+			stdout, stderr, err := ct.Kubeclient.Exec(ctx, ct.Namespace, pod.Name, []string{"/usr/bin/env", "bash", "-c", getLibsCmd})
+			assert.NoError(err, "running %q:\nstdout:\n%s\nstderr:\n%s", getLibsCmd, stdout, stderr)
+			lsLibs := strings.Split(strings.TrimSpace(stdout), "\n")
+			assert.GreaterOrEqual(len(lsLibs), 1, "expected at least one library for %q", lib)
+			if len(lsLibs) == 0 {
+				continue
+			}
```
burgerdev: You could do `if !assert.Positive(len(lsLibs), ...) { continue }`

e2e/gpu/gpu_test.go
```diff
@@ -62,10 +69,121 @@ func TestGPU(t *testing.T) {
...
+			"libnvidia-allocator.so":       {abiLink: true},
+			"libnvidia-pkcs11-openssl3.so": {abiLink: false},
+			"libnvidia-nvvm.so":            {abiLink: true},
+		}
+		for lib, libChecks := range expectLibs {
+			pathThisLib := path.Join(libPath, lib)
+
+			// Run `ls` to check what libraries with that name exist.
+			getLibsCmd := fmt.Sprintf("ls %s*", pathThisLib)
+			stdout, stderr, err := ct.Kubeclient.Exec(ctx, ct.Namespace, pod.Name, []string{"/usr/bin/env", "bash", "-c", getLibsCmd})
```
burgerdev, line 122: Did you consider running one `find` on the libPath, outside the loop, and then processing the results locally? I'm a bit worried that the large number of `Kubeclient.Exec` calls might increase flakiness.

katexochen: Thought about it, but it looked like it worked quite well (and fast) during development, and it helps a bit with debugability and readability of the code.
I'm fine with switching the ls in loop to a single find, but we still call exec 1-3 times for each lib, so we will still do many execs.

We could also do all this in a bash script that we copy over for the test, but I thought the feedback we get in ci is better this way.

burgerdev: Ok, then let's go ahead and see.

PR #1574, opened 2025-07-08
[release/v1.9] kata: don't add storage for implicit VOLUME mounts

burgerdev:
* [x] [e2e regression TDX](https://github.com/edgelesssys/contrast/actions/runs/16145958079)
* [x] e2e regression SNP executed manually

### DIFF
--- a/packages/by-name/kata/kata-runtime/0026-genpolicy-don-t-allow-mount-storage-for-declared-VOL.patch
+++ b/packages/by-name/kata/kata-runtime/0026-genpolicy-don-t-allow-mount-storage-for-declared-VOL.patch
@@ -1,0 +1,162 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Markus Rudy <mr@edgeless.systems>
+Date: Mon, 30 Jun 2025 11:18:47 +0200
+Subject: [PATCH] genpolicy: don't allow mount/storage for declared VOLUMEs
+
+Signed-off-by: Markus Rudy <mr@edgeless.systems>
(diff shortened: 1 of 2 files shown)


burgerdev:
Imho we can ignore the govulncheck findings:

> Vulnerability #1: GO-2025-3751
    Sensitive headers not cleared on cross-origin redirect in net/http
  More info: https://pkg.go.dev/vuln/GO-2025-3751
  Standard library
    Found in: net/http@go1.24.3
    Fixed in: net/http@go1.24.4
    Example traces found:
      #1: nodeinstaller/internal/asset/http.go:100:30: asset.HTTPFetcher.FetchUnchecked calls http.Client.Do

Nodeinstaller is not setting sensitive headers.

> Vulnerability #2: GO-2025-3750
    Inconsistent handling of O_CREATE|O_EXCL on Unix and Windows in os in
    syscall
  More info: https://pkg.go.dev/vuln/GO-2025-3750

Windows is not a supported target, behaviour on UNIX is ok before and after.

> Vulnerability #3: GO-2025-3749
    Usage of ExtKeyUsageAny disables policy validation in crypto/x509
  More info: https://pkg.go.dev/vuln/GO-2025-3749
  Standard library
    Found in: crypto/x509@go1.24.3
    Fixed in: crypto/x509@go1.24.4
    Example traces found:
      #1: internal/attestation/snp/validator.go:100:40: snp.Validator.Validate calls verify.SnpAttestationContext, which eventually calls x509.Certificate.Verify

As far as I know, we don't have policy graphs in certificates relevant to attestation.

PR #1580, opened 2025-07-10
e2e/gpu: fix version regex match

katexochen:


e2e/gpu/gpu_test.go
```diff
@@ -127,7 +127,7 @@ func TestGPU(t *testing.T) {
 			}
 
 			// Determine what library paths we got.
-			fullVerRegex := regexp.MustCompile(fmt.Sprintf(`%s\.\d+.\d+$`, strings.ReplaceAll(pathThisLib, ".", "\\.")))
+			fullVerRegex := regexp.MustCompile(fmt.Sprintf(`%s\.\d+\.\d+(\.\d+)?`, strings.ReplaceAll(pathThisLib, ".", "\\.")))
```
burgerdev: Did you drop the `$` on purpose?

burgerdev: Would it make sense to do `%s\.\d+(\.\d+)+`?

katexochen: no

PR #1582, opened 2025-07-10
dev-docs: document NVIDIA RIM IDs & update automatically

katexochen:
We pin the gpu driver based on this, so we want to be notified if there are any changes and we can bump the driver to a newer version.


.github/workflows/rim_updates.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 18 * * 0"
+
+jobs:
+  update-rim-ids:
+    name: update NVIDIA RIM IDs
+    runs-on: ubuntu-22.04
```
burgerdev, line 11: Do we want to go with 24.04 right away?

.github/workflows/rim_updates.yml
```diff
@@ -0,0 +1,44 @@
...
+          token: ${{ !github.event.pull_request.head.repo.fork && secrets.NUNKI_CI_COMMIT_PUSH_PR || github.token }}
+      - uses: ./.github/actions/setup_nix
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - name: Update RIM IDs
+        run: |
+          nix run .#scripts.get-nvidia-rim-ids > dev-docs/nvidia/rim-ids.json
+      - name: Create PR
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
```
burgerdev, line 26: Does this action fail if there are no changes? In that case, we'd need to adjust the notification logic somehow.

burgerdev: I guess https://github.com/edgelesssys/contrast/pull/1582#issuecomment-3056533109 means it does not.

.github/workflows/rim_updates.yml
katexochen: I'd rather migrate all the jobs at once and keep the versions in sync. Will create a ticket for it.

PR #1583, opened 2025-07-10
packages: build rust projects with crane

charludo:
## Draft: currently uses crane only for kata-agent.

Reduces re-build times significantly, esp. when working on upstream kata and frequently updating patches. In the case of kata-agent,  ~2min are saved on my local machine when rebuilding (and not needing to re-build dependencies, of course).

Note that tests are missing, crane handles them a bit differently (see link to gist below).
## How to see in action:

> CURRENTLY BROKEN, works in this gist: https://gist.github.com/charludo/a75388d33e8e244bbbb0d4e4d30c4ad8 but full recompiles are triggered in the "integrated" version, investigating

1. run `nix build .#kata.kata-agent --print-build-logs`, observe that all of `kata-agent`'s dependencies are being built
2. run `nix build .#kata.kata-agent --print-build-logs` again, no build should take place at all, since no inputs have changed
3. make a change to the kata source, either through a new patch or (for testing) by e.g. changing a log message string in the latest patch
4. run `nix build .#kata.kata-agent --print-build-logs` once again, observe that only kata-agent and related crates from the `kata-containers` project are being recompiled, *not* any of the dependencies


packages/by-name/kata/kata-agent/package.nix
```diff
@@ -16,88 +18,113 @@
...
+      let
+        additionalFilter = path: _type: builtins.match ".*[in|proto]$" path != null;
+        additionalOrCargo =
+          path: type: (additionalFilter path type) || (craneLib.filterCargoSources path type);
+      in
+      lib.cleanSourceWith {
+        src = kata.kata-runtime.src;
+        filter = additionalOrCargo;
+        name = "source";
+      };
```
charludo: This entire block is definitely doable in a nicer fashion. 

`cleanSourceWith` should probably move to `kata-runtime`.

burgerdev:
> Hmm, looks to me like the folllowing is happening:
> 
> * in the gist flake, Cargo.toml/Cargo.lock are referenced by a nix path, which points to the actual files
> * these don't change between rebuilds, unless they are actually being edited
> * in contrast, that's not possible, so instead, the `${kata.kata-runtim.src}/src.agent/Cargo.[lock|toml]` files are used
> * (here I am not not 100% sure) now if e.g. a patch is edited, the store path to the source changes, including the path to the toml/lock files; Crane assumes the files have changed, and initiates a complete rebuild of all deps
> 
> Edit: not making progress here, asked in the crane repo: [ipetkov/crane#863](https://github.com/ipetkov/crane/discussions/863)

Shouldn't it be possible to use `${kata.kata-runtim.src.src}`, that is, refer to the source before `applyPatches` took place?

PR #1588, opened 2025-07-11
docs: move supported kinds to policy page

burgerdev:


docs/docs/architecture/components/policies.md
```diff
@@ -80,3 +80,34 @@ After the last step, we know that the policy hasn't been tampered with and, thus
 Contrast uses different rules and data sections for different platforms.
 This results in different policy hashes for different platforms.
 The `generate` command automatically derives the correct set of rules and data sections from the `reference-values` flag.
+
+## Supported resource kinds
+
+Contrast policies can be generated for all Kubernetes resource kinds that spawn pods, including:
+
+<!-- keep-sorted start by_regex=`(\w+)` -->
```
katexochen, line 88:
```suggestion
<!-- keep-sorted start  -->
```
Why the by_regex? Seems to work fine without it.

katexochen (approved):
otherwise lgtm

docs/docs/architecture/components/policies.md
burgerdev: It's the right way™ of sorting these, but for the pod generating kinds it does not make a difference (for the others below, it does). Since we're not expecting any additions to the first list, should I remove it?

katexochen: Ah, got it, its about the ending quote. No, keept it.

PR #1599, opened 2025-07-14
attestation/certcache: always fetch for TDX requests

davidweisse:
The TDX collaterals (TCBInfo and QeIdentity) have an expiration date, which is verified by the validator. If the validator uses a cached entry, the validation can fail if the entry is expired. To fix this, always try to fetch the collateral from the PCS first, not just the CRL. If the PCS is not available, fallback to cache.

### DIFF
--- a/internal/attestation/certcache/cached_client.go
+++ b/internal/attestation/certcache/cached_client.go
@@ -16,8 +16,9 @@ import (
 )
 
 var (
-	snpCrlURL = regexp.MustCompile(`^https://kdsintf\.amd\.com/vcek/v1/[A-Za-z]*/crl$`)
-	tdxCrlURL = regexp.MustCompile(`^https://api\.trustedservices\.intel\.com/sgx/certification/v4/pckcrl\?ca=(platform|processor)&encoding=der$`)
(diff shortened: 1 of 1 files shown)


internal/attestation/certcache/cached_client.go
```diff
@@ -59,27 +60,27 @@ func (c *CachedHTTPSGetter) GetContext(ctx context.Context, url string) (map[str
...
 			if data, err := json.Marshal(cacheEntry{header, body}); err == nil {
 				c.cache.Set(url, data)
 			} else {
-				c.logger.Warn("Failed to marshal CRL response, not writing to cache", "error", err)
+				c.logger.Warn("Failed to marshal response, not writing to cache", "error", err)
 			}
 			return header, body, nil
 		}
-		c.logger.Warn("Failed requesting CRL from KDS/PCS", "error", err)
+		c.logger.Warn("Failed requesting URL from KDS/PCS", "error", err)
```
burgerdev: At this point, it would probably make sense to log the failing URL, too. Preferably as `log := c.logger.With("url", url)` at the beginning of the function, then you could remove a bunch of log arguments.

burgerdev (approved):

PR #1601, opened 2025-07-14
e2e: fix and improve vault test

burgerdev:
* repairs the Vault test to be compatible with changes for https://github.com/edgelesssys/contrast/security/advisories/GHSA-phhq-63jg-fp7r
* enables the vault test in nightly e2e runs
* removes the sidecar `openbao-client` and replaces it with a separate deployment
  - the sidecar served no purpose in the demo, and could just be replaced by the main container in tests
  - the new deployment can be integrated into Vault policies, as demonstrated in the last test
  - this setup should be closer to what a user would actually want to demo.
* removes the sealing status checks:
  - accessing the keys after restart sufficiently proves unsealed status 

- [x] [Vault e2e on TDX](https://github.com/edgelesssys/contrast/actions/runs/16312646321)

### DIFF
--- a/.github/workflows/e2e_nightly.yml
+++ b/.github/workflows/e2e_nightly.yml
@@ -30,6 +30,7 @@ jobs:
           - peerrecovery
           - policy
           - servicemesh
+          - vault
           - volumestatefulset
           - workloadsecret
           # keep-sorted end


internal/kuberesource/sets.go
```diff
@@ -739,6 +739,18 @@ func GPU() []any {
 	return []any{tester}
 }
 
+const vaultImage = "quay.io/openbao/openbao:2.2.0@sha256:19612d67a4a95d05a7b77c6ebc6c2ac5dac67a8712d8df2e4c31ad28bee7edaa"
```
katexochen: nit: we could move these into the Vault func.

internal/kuberesource/sets.go
```diff
@@ -854,5 +854,47 @@ seal "transit" {
...
+					).
+					WithContainers(
+						Container().
+							WithName("vault-client").
+							WithImage(vaultImage).
+							WithCommand("/bin/sh", "-c", vaultClientEntrypoint).
+							WithEnv(EnvVar().WithName("VAULT_ADDR").WithValue("https://vault:8200")).
+							WithEnv(EnvVar().WithName("VAULT_CACERT").WithValue("/contrast/tls-config/mesh-ca.pem")).
+							WithEnv(EnvVar().WithName("VAULT_CLIENT_CERT").WithValue("/contrast/tls-config/certChain.pem")).
+							WithEnv(EnvVar().WithName("VAULT_CLIENT_KEY").WithValue("/contrast/tls-config/key.pem")).
```
katexochen: nit: WithEnv can take multiple EnvVars.

internal/kuberesource/sets.go
burgerdev: moved image and client script into `Vault`, and moved the config file to a const, too.

internal/kuberesource/sets.go
burgerdev: done

PR #1603, opened 2025-07-15
ci: don't run docs-publish on PRs from forks

msanft:
To harden the workflow, don't run it on any PRs originating from forks.

### DIFF
--- a/.github/workflows/docs_publish.yml
+++ b/.github/workflows/docs_publish.yml
@@ -30,6 +30,7 @@ jobs:
   deploy:
     name: Publish docs to GitHub Pages
     runs-on: ubuntu-22.04
+    if: github.repository_owner == github.event.pull_request.head.repo.owner.login
     env:
       PREVIEW: ${{ !(github.event_name == 'push' && github.ref_name == 'main') }}
     steps:
@@ -57,7 +58,7 @@ jobs:
         run: |
           nix build -L --impure --expr "(builtins.getFlake \"git+file://$(pwd)?shallow=1\").outputs.legacyPackages.x86_64-linux.contrast-docs.override { docusaurusBaseUrl = \"contrast/pr-preview/pr-${{ github.event.number }}\"; }"
       - name: Deploy preview
-        if: env.PREVIEW == 'true' && !github.event.pull_request.head.repo.fork
+        if: env.PREVIEW == 'true'
         uses: rossjrw/pr-preview-action@9f77b1d057b494e662c50b8ca40ecc63f21e0887 # v1.6.2
         with:
           source-dir: ./result


.github/workflows/docs_publish.yml
```diff
@@ -40,24 +37,35 @@ jobs:
...
           folder: ./result
           branch: gh-pages
           clean-exclude: pr-preview
           force: false
 
+  deploy-preview:
+    name: Preview docs
+    runs-on: ubuntu-22.04
+    permissions:
+      pull-requests: write
```
burgerdev: We still need to push the preview to the `gh-pages` branch, don't we?

msanft: Seems so. I didn't know this action implicitly uses that branch, while the other one explicitly specifies it.

.github/workflows/docs_publish.yml
```diff
@@ -40,24 +37,36 @@ jobs:
           cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
       - name: Build production website
-        if: env.PREVIEW == 'false'
         run: |
           nix build -L .#contrast-docs
+
```
katexochen:
```suggestion
```

.github/workflows/docs_publish.yml
```diff
@@ -40,24 +37,36 @@ jobs:
...
+      - uses: ./.github/actions/setup_nix
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
       - name: Build preview website
-        if: env.PREVIEW == 'true'
         run: |
           nix build -L --impure --expr "(builtins.getFlake \"git+file://$(pwd)?shallow=1\").outputs.legacyPackages.x86_64-linux.contrast-docs.override { docusaurusBaseUrl = \"contrast/pr-preview/pr-${{ github.event.number }}\"; }"
+
```
katexochen:
```suggestion
```

.github/workflows/docs_publish.yml
```diff
@@ -40,24 +37,36 @@ jobs:
...
         uses: JamesIves/github-pages-deploy-action@6c2d9db40f9296374acc17b90404b6e8864128c8 # v4.7.3
         with:
           folder: ./result
           branch: gh-pages
           clean-exclude: pr-preview
           force: false
 
+  deploy-preview:
+    name: Preview docs
+    runs-on: ubuntu-22.04
```
katexochen: This needs an if, too, or we will deploy twice on main.

katexochen (changes requested):

.github/workflows/docs_publish.yml
```diff
@@ -40,24 +37,35 @@ jobs:
...
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
       - name: Build preview website
-        if: env.PREVIEW == 'true'
         run: |
           nix build -L --impure --expr "(builtins.getFlake \"git+file://$(pwd)?shallow=1\").outputs.legacyPackages.x86_64-linux.contrast-docs.override { docusaurusBaseUrl = \"contrast/pr-preview/pr-${{ github.event.number }}\"; }"
       - name: Deploy preview
-        if: env.PREVIEW == 'true' && !github.event.pull_request.head.repo.fork
+        if: github.repository_owner == github.event.pull_request.head.repo.owner.login
```
burgerdev: Shouldn't we rather lift this logic to the job level - that is, don't do previews on forks?

msanft: Can you elaborate? I don't really understand that. This conditional should ensure that the preview isn't deployed on forks.

burgerdev: If execution of this workflow is problematic when it comes from a fork, we could just condition the whole job with `!github.event.pull_request.head.repo.fork`, couldn't we? Otherwise, I don't understand why we should constrain the `pr-preview-action`, but not the build above.

msanft: This conditional doesn't work unfortunately. But the one I've added (i.e. `github.repository_owner == github.event.pull_request.head.repo.owner.login`) should do. 

If you're fine with only having the preview on non-fork PRs, that's the easiest fix possible. However I thought you might actually want to see if the docs build on fork PRs too. If this is not a hard requirement, I'm open to do that.

burgerdev: > This conditional doesn't work unfortunately. 

Oh, because of https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request: "The pull_request webhook event payload is empty for [...] pull requests that come from forked repositories."?

> If you're fine with only having the preview on non-fork PRs, that's the easiest fix possible. However I thought you might actually want to see if the docs build on fork PRs too. If this is not a hard requirement, I'm open to do that.

I think we could work around that when needed - so far, the list of external contributors to docs is pretty small :)

.github/workflows/docs_publish.yml
```diff
@@ -40,24 +37,34 @@ jobs:
...
+    permissions:
+      pull-requests: write
+      contents: write
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: ./.github/actions/setup_nix
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
```
katexochen:
```suggestion
```

.github/workflows/docs_publish.yml
katexochen: this wasn't fixed

.github/workflows/docs_publish.yml
msanft: FYI: I kept this whitespace to align with what was there previously.

.github/workflows/docs_publish.yml
```diff
@@ -40,24 +37,34 @@ jobs:
...
         with:
           folder: ./result
           branch: gh-pages
           clean-exclude: pr-preview
           force: false
 
+  deploy-preview:
+    name: Preview docs
+    runs-on: ubuntu-22.04
+    if: github.repository_owner == github.event.pull_request.head.repo.owner.login && github.ref_name != 'main'
```
burgerdev: Hey @msanft, sorry for the back-and-forth. I think this line here is an actual improvement to defense in depth, but at the current state of this PR I don't think we need to split the workflows in two - the reduced permission (`pr:write`) for the job running on main is imho not worth the duplication of the job. Could you please go back to the original job definition, but with the "same repo" condition?

katexochen (approved):

burgerdev (approved):
Thanks!

PR #1610, opened 2025-07-16
docs: CPU limit usage

miampf:


docs/docs/howto/workload-deployment/deployment-file-preparation.md
```diff
@@ -132,4 +132,21 @@ Altogether, setting the limit to 10x the compressed image size should be suffici
 
 :::
 
+Getting a predictable number of CPUs in your containers can be a bit more complicated than just setting a resource request since contrast runs on kata containers.
```
burgerdev: Please capitalize proper nouns: Contrast; Kata Containers

docs/docs/howto/workload-deployment/deployment-file-preparation.md
```diff
@@ -132,4 +132,21 @@ Altogether, setting the limit to 10x the compressed image size should be suffici
...
+In your YAML, set a resource limit instead of a request:
```
burgerdev: This could be more enlightening - we should mention the need to allocate resources at VM start, which prevents us from implementing requests and limits as documented by k8s. 

This can (and should) reference the recommendation for memory above.

docs/docs/howto/workload-deployment/deployment-file-preparation.md
```diff
@@ -132,4 +132,21 @@ Altogether, setting the limit to 10x the compressed image size should be suffici
...
+    - name: my-container
+      image: "my-image@sha256..."
+      resources:
+        limits:
+          cpu: "1"
+  runtimeClassName: contrast-cc
+```
+
+This limit is rounded up to the next whole integer. The final number of CPUs you can observe in a container is this ceiling + 1 (`numCPUs = ceil(limit) + 1`), so
+in this case, the final number of CPUs would be 2. A limit of 1.25 would result in 3 CPUs.
```
burgerdev: Please mention the default (0 limit --> 1 VM CPU, I guess).

docs/docs/howto/workload-deployment/deployment-file-preparation.md
miampf: How exactly do you imagine referencing the recommendation above? Do you mean also recommending 10x the normal CPU limit? Or just something along the lines of "Due to the limitations mentioned above, Contrast needs to allocate resources at VM start"?

burgerdev: I was refering to this:

> Contrast workloads are deployed as one confidential virtual machine (CVM) per pod.
The resources provided to the CVM are static and can't be adjusted at runtime.

The point being: normal runc containers can just define a request, and then saturate whatever CPUs are available on the host.

PR #1613, opened 2025-07-16
e2e: increase timeout limits for tests that restart pods

miampf:
Some pods may fail to restart fast enough to satisfy context timeouts. This PR increases the timeouts for all cases where a pod is restarted to 2 minutes.

### DIFF
--- a/e2e/policy/policy_test.go
+++ b/e2e/policy/policy_test.go
@@ -82,12 +82,11 @@ func TestPolicy(t *testing.T) {
 	})
 
 	// initial deployment with pod allowed
-
-	ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(1*time.Minute))
-	defer cancel()
-
 	require.True(t, t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
 	require.True(t, t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
+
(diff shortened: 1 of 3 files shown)


e2e/policy/policy_test.go
```diff
@@ -83,7 +83,7 @@ func TestPolicy(t *testing.T) {
 
 	// initial deployment with pod allowed
 
-	ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(1*time.Minute))
+	ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(2*time.Minute))
 	defer cancel()
```
burgerdev: Could you please move the context creation below `generate` and `apply`? Those have their own context and should not count towards the 2min.

e2e/workloadsecret/workloadsecret_test.go
```diff
@@ -183,7 +183,7 @@ func TestWorkloadSecrets(t *testing.T) {
 	t.Run("workload secrets are not created if not configured in the manifest", func(t *testing.T) {
 		assert := assert.New(t)
 		require := require.New(t)
-		ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(60*time.Second))
+		ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(2*time.Minute))
```
burgerdev: Same here, please move the context creation to after `set`.

burgerdev (approved):
Nice, thanks!

PR #1615, opened 2025-07-17
imagepuller: do not re-pull known images

charludo:
Memory is sufficient again 🎉

### DIFF
--- a/imagepuller/internal/service/service.go
+++ b/imagepuller/internal/service/service.go
@@ -8,12 +8,9 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"os"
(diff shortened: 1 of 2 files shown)


imagepuller/internal/service/service.go
```diff
@@ -36,6 +33,16 @@ func (s *ImagePullerService) PullImage(ctx context.Context, r *api.ImagePullRequ
 		}
 	}()
 
+	cachedID, err := s.Store.Lookup(r.ImageUrl)
```
burgerdev, line 36: We should _not_ be reusing the container. Consider naming the image, and then creating a fresh container with every mount.

PR #1626, opened 2025-07-18
tools: add a script to run release test matrix locally

katexochen:
This is brittle AI code, but it should do for now.

### DIFF
--- a/tools/run-release-test-matrix/run-tests.sh
+++ b/tools/run-release-test-matrix/run-tests.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+# Copyright 2025 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+set -euo pipefail
+
+# Create an associative array to hold platform.name -> test list
+declare -A platform_tests
+
+# This should point to the workflows called in release.yml, excluding the release test.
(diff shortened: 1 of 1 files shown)


tools/run-release-test-matrix/run-tests.sh
```diff
@@ -0,0 +1,78 @@
...
+
+  for ((i = 0; i < platform_count; i++)); do
+    name=$(yq ".$MATRIX_PATH.platform[$i].name" "$MATRIX_FILE")
+    self_hosted=$(yq ".$MATRIX_PATH.platform[$i].self-hosted" "$MATRIX_FILE")
+
+    valid_tests=()
+    for test in "${tests[@]}"; do
+      excluded=$(yq -o=json ".$MATRIX_PATH.exclude[]?" "$MATRIX_FILE" |
+        jq -e --argjson sh "$self_hosted" --arg t "$test" 'select(."test-name" == $t and .platform."self-hosted" == $sh)' >/dev/null && echo true || echo false)
+      if [[ $excluded == "false" ]]; then
```
burgerdev: This might be clearer as 

```sh
if 
 yq ... | \
  jq ...
then
  echo test excluded
else
  valid_tests+=("$test")
fi
```

PR #1628, opened 2025-07-21
imagepuller: test against evil registry

charludo:
Things found during testing:
-  `ImageIndex` and `Image` already perform digest validation.
- layer validation was always trivially true 😬

One further issue: the tests (just as the linter and building the imagepuller binary itself) requires `CGO_ENABLED=0`, not sure if there's a good "central" place to put it?

### DIFF
--- a/imagepuller/internal/service/service.go
+++ b/imagepuller/internal/service/service.go
@@ -5,7 +5,6 @@ package service
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log/slog"
 
@@ -49,62 +48,13 @@ func (s *ImagePullerService) PullImage(ctx context.Context, r *api.ImagePullRequ
 	}
 	log.Info("Validated image")
 
-	layers, err := remoteImg.Layers()
+	finalLayer, err := s.storeAndVerifyLayers(log, remoteImg)
 	if err != nil {
-		return nil, fmt.Errorf("obtaining the image layers: %w", err)
+		return nil, fmt.Errorf("verifying and putting layers in store: %w", err)
 	}
+	log.Info("Verified and put in store layers")
 
-	manifest, err := remoteImg.Manifest()
-	if err != nil {
-		return nil, fmt.Errorf("obtaining image manifest: %w", err)
-	}
-
-	previousLayer := ""
-	for idx, layer := range layers {
-		rc, err := layer.Compressed()
-		if err != nil {
-			return nil, fmt.Errorf("reading layer %d: %w", idx, err)
-		}
-
-		// Wrap in func to close each rc immediately
-		err = func() error {
-			putLayer, n, err := s.Store.PutLayer(
-				"",            // empty ID -> let store decide
-				previousLayer, // parent is previous layer
-				nil,           // empty parent chain -> let store decide
-				"",            // mount label
-				false,         // readonly
-				nil,           // mount options
-				rc,            // tar stream
-			)
-			if err != nil {
-				return errors.Join(err, rc.Close())
-			}
-			if err := rc.Close(); err != nil {
-				return fmt.Errorf("closing the layer reader: %w", err)
-			}
-
-			ldManifest := manifest.Layers[idx].Digest
-			ld, err := layer.Digest()
-			if err != nil {
-				return fmt.Errorf("obtaining digest: %w", err)
-			}
-			if ldManifest != ld {
-				return fmt.Errorf("validation failed, expected digest '%s' but got digest '%s'", ldManifest, ld)
-			}
-
-			log.Info("Applied and validated layer", "id", putLayer.ID, "size", n, "digest", ld)
-			previousLayer = putLayer.ID
-			return nil
-		}()
-		if errors.Is(err, context.DeadlineExceeded) {
-			return nil, fmt.Errorf("pull aborted while applying layer (deadline exceeded): %w", err)
-		} else if err != nil {
-			return nil, fmt.Errorf("applying layer: %w", err)
-		}
-	}
-
-	newImg, err := s.Store.CreateImage("", nil, previousLayer, "", nil)
+	newImg, err := s.Store.CreateImage("", nil, finalLayer, "", nil)
 	if err != nil {
 		return nil, fmt.Errorf("creating image: %w", err)
 	}
--- a/imagepuller/internal/service/verify.go
+++ b/imagepuller/internal/service/verify.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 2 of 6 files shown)


imagepuller/internal/service/verify_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	arbitraryDigest := "sha256:6ad6bbb5735b84b10af42d2441e8d686b1d9a6cbf096b53842711ef5ddabd28d"
+
+	// TODO(charludo): import from evil-registry
+	wrongIndexDigest := "sha256:1111111111111111111111111111111111111111111111111111111111111111"
+	indexForEvilManifestDigest := "sha256:91ba34e46d6c6d2a484360d07521a7b790dd3e3427afd71c1fb21702cbe26521"
+
+	tests := map[string]struct {
+		digest    string
+		wantedErr string
```
katexochen:
```suggestion
		wantErr string
```

imagepuller/internal/service/verify_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"log/slog"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestImagePullerService_getAndVerifyImage(t *testing.T) {
+	log := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
```
katexochen: iirc there is a test logger

imagepuller/internal/service/verify_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			require := require.New(t)
+
+			// TODO(charludo): start evil-registry from here
+			var s ImagePullerService
+			_, err := s.getAndVerifyImage(context.Background(), log, fmt.Sprintf("127.0.0.1:8080/busybox:v0.0.1@%s", tc.digest))
+
+			require.ErrorContains(err, tc.wantedErr)
```
katexochen: prefer using assert. require should be used when the test should be aborted early (for example, because following steps cannot be executed)

imagepuller/internal/service/verify_test.go
```diff
@@ -0,0 +1,97 @@
...
+			digest:    indexForManifestForEvilBlobDigest,
+			wantedErr: "validating layer:",
+		},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			require := require.New(t)
+
+			// TODO(charludo): start evil-registry from here
+			var s ImagePullerService
```
katexochen: shouldn't we be configuring the store here (potentially to a stub-store)?

charludo: I don't think that's necessary - we're essentially just using the `ImagePullerService` struct as a means of accessing the `getAndVerifyImage` and `storeAndVerifyLayers` functions, we're not actually putting anything in the store or accessing `s.Store`.

charludo: ...I take it back 😄

imagepuller/internal/service/verify_test.go
charludo: Not quite sure I understand - there's `t.Log()`, but `t` does not implement `Info()` etc - or do you mean the [slogt](https://github.com/neilotoole/slogt) package?

imagepuller/internal/service/service.go
```diff
@@ -5,18 +5,29 @@ package service
...
-	"errors"
 	"fmt"
+	"io"
 	"log/slog"
 
 	"github.com/containers/storage"
 	"github.com/edgelesssys/contrast/imagepuller/internal/api"
 )
 
+// ContainersStorageStore defines the functions used from the storage.Store interface to allow for easier testing.
```
burgerdev: Did you know that you can embed an interface into a struct to make it implement the interface, but fail any calls that you did not override explicitly?

```go
type stubStore struct {
	storage.Store
}

func (*stubStore) Lookup(name string) (string, error) {
	return "foo", nil
}
```

imagepuller/internal/service/verify.go
```diff
@@ -55,28 +50,80 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
...
+			return "", fmt.Errorf("applying layer: %w", err)
+		}
 	}
 
-	return remoteImg, nil
+	return previousLayer, nil
+}
+
+func isDigestMismatch(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "does not match requested digest")
```
burgerdev, line 122: Could you please document where this string comes from?

imagepuller/internal/service/verify.go
```diff
@@ -55,28 +50,80 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
...
+
+	previousLayer := ""
+	for idx, layer := range layers {
+		rc, err := layer.Compressed()
+		if err != nil {
+			return "", fmt.Errorf("reading layer %d: %w", idx, err)
+		}
+
+		// Wrap in func to close each rc immediately
+		err = func() error {
```
burgerdev: I believe this is now obsolete, since we don't `defer rc.Close()` anymore.

imagepuller/internal/service/verify_test.go
burgerdev: We don't have a test logger right now, but I guess we should have one eventually? Other unit tests tend to use `slog.Default()`, maybe consider doing that here, too, so we'll find it easier when we add a test logger.

imagepuller/internal/service/verify_test.go
```diff
@@ -0,0 +1,139 @@
...
+	tests := map[string]struct {
+		digest  string
+		wantErr string
+	}{
+		"missing digest is rejected": {
+			digest:  "",
+			wantErr: "parsing image digest",
+		},
+		"wrong manifest digest is caught": {
+			digest:  "sha256:6ad6bbb5735b84b10af42d2441e8d686b1d9a6cbf096b53842711ef5ddabd28d",
```
burgerdev, line 42: Consider adding a comment along the lines of "the evil registry responds to unknown digests with a default manifest".

imagepuller/internal/service/verify_test.go
```diff
@@ -0,0 +1,139 @@
...
+			assert := assert.New(t)
+
+			mux := http.NewServeMux()
+			server := httptest.NewUnstartedServer(mux)
+			go registry.Run(server.Listener, mux)
+			server.Start()
+			defer server.Close()
+
+			var s ImagePullerService
+			_, err := s.getAndVerifyImage(context.Background(), log, fmt.Sprintf("%s/busybox:v0.0.1@%s", server.URL[7:], tc.digest))
```
burgerdev: Instead of indexing the URL, you could use `server.Listener.Addr().String()`.

PR #1635, opened 2025-07-23
e2e/aks-runtime: install openssl-1.1

miampf:
This fixes the E2E test since `confcom` depends on openssl-1.1, but it wasn't installed.


miampf:
Also opened an issue on the `azure-cli-extensions` repo (https://github.com/Azure/azure-cli-extensions/issues/8995), if the dependency for openssl-1.1 can be removed for confcom we should revert this change

PR #1636, opened 2025-07-23
just: update namespaces for custom target

miampf:
Previously, namespaces for custom deployments needed to be correct in the `.custom` directory. This PR lets just automatically update the namespace to the correct value for custom deployments.


justfile
```diff
@@ -90,13 +90,15 @@ populate target=default_deploy_target platform=default_platform:
     if [[ "${target}" == "custom" ]]; then
         target=""
         cp -r ./.custom/* ./{{ workspace_dir }}/deployment/
+        ns="{{ target }}${namespace_suffix-}"
+        env ns="$ns" yq -i '.metadata.namespace = strenv(ns)' ./{{ workspace_dir }}/deployment/*.yml
```
burgerdev: The `-i` does not quite do what you'd expect.

```console
> yq --version
yq (https://github.com/mikefarah/yq/) version v4.46.1
> cat 1.yaml 
kind: Foo
source: 1.yaml
---
kind: Bar
source: 1.yaml
> cat 2.yaml 
kind: Baz
source: 2.yaml
> yq -i '.metadata.namespace = "mynamespace"' *.yaml
> cat 1.yaml
kind: Foo
source: 1.yaml
metadata:
  namespace: mynamespace
---
kind: Bar
source: 1.yaml
metadata:
  namespace: mynamespace
---
kind: Baz
source: 2.yaml
metadata:
  namespace: mynamespace
> cat 2.yaml 
kind: Baz
source: 2.yaml
```

justfile
```diff
@@ -90,13 +90,15 @@ populate target=default_deploy_target platform=default_platform:
...
     fi
     if [[ -f ./{{ workspace_dir }}/deployment/deployment.yml ]]; then
         echo "---" >> ./{{ workspace_dir }}/deployment/deployment.yml
     fi
     nix shell .#contrast --command resourcegen \
         --image-replacements ./{{ workspace_dir }}/just.containerlookup \
-        --namespace {{ target }}${namespace_suffix-} \
+        --namespace ${ns-} \
```
burgerdev: Don't you need to set `ns` for non-custom targets, too?

justfile
burgerdev: Please also support `.yaml` as extension.

justfile
miampf: Yeah sorry this was a leftover from a previous iteration that slipped through (which it shouldn't have), I'll revert that part.

justfile
miampf: That is weird behavior... I'll research an alternative flag

miampf: Didn't find an alternative flag but using a loop fixes the issue

justfile
```diff
@@ -112,7 +116,7 @@ generate cli=default_cli platform=default_platform:
         --workspace-dir ./{{ workspace_dir }} \
         --image-replacements ./{{ workspace_dir }}/just.containerlookup \
         --reference-values {{ platform }}\
-        ./{{ workspace_dir }}/deployment/*.yml
+        ./{{ workspace_dir }}/deployment/*.{yml,yaml}
```
burgerdev:
```suggestion
        ./{{ workspace_dir }}/deployment/
```

justfile
```diff
@@ -234,7 +238,7 @@ set cli=default_cli:
     nix run -L .#{{ cli }} -- set \
         --workspace-dir ./{{ workspace_dir }} \
         -c localhost:1313 \
-        ./{{ workspace_dir }}/deployment/*.yml
+        ./{{ workspace_dir }}/deployment/*.{yml,yaml}
```
burgerdev:
```suggestion
        ./{{ workspace_dir }}/deployment/
```

burgerdev (approved):
Should be good with just the directory: https://docs.edgeless.systems/contrast/next/howto/workload-deployment/generate-annotations#how-to

burgerdev (approved):

PR #1637, opened 2025-07-24
e2e/gpu: fix handling of empty ls stdout

katexochen:


e2e/gpu/gpu_test.go
```diff
@@ -121,7 +121,11 @@ func TestGPU(t *testing.T) {
 			getLibsCmd := fmt.Sprintf("ls %s*", pathThisLib)
 			stdout, stderr, err := ct.Kubeclient.Exec(ctx, ct.Namespace, pod.Name, []string{"/usr/bin/env", "bash", "-c", getLibsCmd})
 			assert.NoError(err, "running %q:\nstdout:\n%s\nstderr:\n%s", getLibsCmd, stdout, stderr)
-			lsLibs := strings.Split(strings.TrimSpace(stdout), "\n")
+			stdout = strings.TrimSpace(stdout)
+			var lsLibs []string
+			if stdout == "" {
```
burgerdev: Shouldn't that be `!=`? We want to populate `lsLibs` with the split only if there actually is something to split.

PR #1638, opened 2025-07-24
ci: allow unrestricted namespaces in setup_nix

katexochen:
As we upgraded hosted CI runners to Ubuntu 24.04, we run into this issues in various situations. The fix was previously present in the e2e_runtime_reproducibility, where we used 24.04 before.


charludo (approved):
🎉

PR #1640, opened 2025-07-24
ci: add pr release artifact workflow

katexochen:


katexochen:
successful cleanup: https://github.com/edgelesssys/contrast/actions/runs/16517221191/job/46710570868

.github/actions/release_artifacts/action.yml
```diff
@@ -0,0 +1,178 @@
...
+      with:
+        name: image-replacements
+        path: |
+          image-replacements.txt
+    - name: Update cli assets
+      shell: bash
+      run: |
+        cp image-replacements.txt cli/cmd/assets/image-replacements.txt
+        git add cli/cmd/assets/image-replacements.txt
+        git diff --staged --quiet || git commit -m "release: update cli assets"
```
burgerdev, line 95: Should this happen on every PR?

.github/actions/release_artifacts/action.yml
```diff
@@ -0,0 +1,178 @@
...
+          --image-replacements ./image-replacements.txt \
+          --add-load-balancers mysql > workspace/mysql-demo.yml
+        nix shell .#contrast --command resourcegen \
+          --image-replacements ./image-replacements.txt \
+          --add-load-balancers vault > workspace/vault-demo.yml
+    - name: Build CLI
+      shell: bash
+      run: |
+        nix build -L .#contrast-cli-release --out-link workspace/contrast
+    - name: AWS login (IAM role)
```
burgerdev, line 154: Not sure what the `IAM role` is supposed to tell me - I was half expecting the role to be an IAM admin...

.github/workflows/pr_release_artifacts.yml
```diff
@@ -0,0 +1,137 @@
+name: pr release artifacts
+
+on:
+  schedule:
+    - cron: '0 15 * * 4'
```
burgerdev, line 5: I was expecting a `pull_request` trigger here, to match the workflow name.

.github/workflows/pr_release_artifacts.yml
```diff
@@ -0,0 +1,137 @@
...
+        if: ${{ steps.get-pr-number.outputs.pr_number != '' }}
+        uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043 # v4.0.0
+        with:
+          issue-number: ${{ steps.get-pr-number.outputs.pr_number }}
+          body: |
+            ### Pre-release artifacts on ${{ github.sha }}
+
+            The pre-release artifacts for this commit are available at the following link:
+
+            ${{ steps.create-artifacts.outputs.s3-bucket-url }}
```
burgerdev, line 78: Is this URL supposed to be publicly accessible? It's not for me.

katexochen: It is, but you can't list. You need to append the name of an existing artifact to the path.

.github/workflows/pr_release_artifacts.yml
katexochen: That would mean it would trigger on every PR. I think the vast majority of PRs won't need to build the artifacts, so I decided to go with a `workflow_dispatch` trigger instead.

.github/actions/release_artifacts/action.yml
katexochen: Yes, this has to happen on release, and it doesn't hurt to do it on PR, too, as long as we don't push the change.

.github/actions/release_artifacts/action.yml
katexochen: just says we are logging in via IAM role as opposed to some bearer token login flow. that's the only way we log in to aws in our ci, I can remove it if you prefer.

.github/workflows/pr_release_artifacts.yml
burgerdev: Got it. But do we need the schedule trigger at all? And should we rename the workflow so it does not sound like we're running it on PR?

katexochen: Well, it is intended to build the artifacts on PR (preferably) or some commit (but it has specific support for PR with the comment posting). If I remove the PR from the name, it sounds like it is part of the release process. Open for suggestions.

The schedule trigger is for cleanup. As the workflow doesn't have a PR trigger, we cannot cleanup on PR merge, so we cleanup on schedule instead.

.github/actions/release_artifacts/action.yml
burgerdev: Now I'm even more confused - aren't we using the OIDC integration here? 

https://github.com/aws-actions/configure-aws-credentials/blob/68858db075f07f5ade3752254dfda34afd669f9b/README.md?plain=1#L84

.github/workflows/pr_release_artifacts.yml
burgerdev: Can we fix that? It's a bit inconvenient without the dir listing.

.github/actions/release_artifacts/action.yml
katexochen: Yes, this is the "Assume [IAM] Role directly using GitHub OIDC provider" flow. Just as always. I just copied that from another workflow without any change. Not sure where the confusion is.

.github/workflows/pr_release_artifacts.yml
katexochen: Why should we grant more permissions than needed? The permissions are sufficient for the intended way to consume the artifacts (which is: inset that url into your nix code, fetchurl the artifacts). If you want to check you can just open the s3 bucket in the web UI.

.github/actions/release_artifacts/action.yml
burgerdev: The added confusion came from "as opposed to some bearer token login flow", which I'd argue OIDC is. 

Now I see that the difference is between "login as IAM user" and "login somehow and assume IAM role", and understand better why it's in the step name.

.github/workflows/pr_release_artifacts.yml
burgerdev: Guess I'm just not the target audience for the links.

PR #1643, opened 2025-07-25
cli: add `verifier` package

miampf:
This prepares us to later run verification pipelines against kubernetes YAML.


cli/verifier/pipeline.go
```diff
@@ -0,0 +1,22 @@
...
+// VerificationPipeline represents a pipeline for resource verification.
+type VerificationPipeline struct {
+	Pipeline []Verifier
+}
+
+// Verify runs a given kubernetes object through the pipeline. If a verifier fails it returns an error.
+func (v *VerificationPipeline) Verify(toVerify *unstructured.Unstructured) error {
+	for _, toRun := range v.Pipeline {
+		if err := toRun(toVerify); err != nil {
+			return err
```
katexochen: As with manifest validation, we should collect and join all errors instead returning on the first one that fails.

cli/verifier/pipeline.go
```diff
@@ -0,0 +1,30 @@
...
+	for _, toRun := range v.Pipeline {
+		if err := toRun(toVerify); err != nil {
+			allErrors = errors.Join(allErrors, err)
+		}
+	}
+	if allErrors != nil {
+		return allErrors
+	}
+
+	return nil
```
burgerdev:
```suggestion
	return allErrors
```

cli/verifier/pipeline_test.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			require := require.New(t)
+			pipelineToTest := verifier.VerificationPipeline{
+				Pipeline: tc.pipeline,
+			}
+			// using nil since individual pipelines are tested on their own
```
burgerdev: Can you please double check this comment? Afaiu "pipeline" is a collection of verifiers + an evaluation function, and with this understanding I can't parse the comment.

cli/verifier/pipeline.go
```diff
@@ -0,0 +1,30 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// VerificationPipeline represents a pipeline for resource verification.
+type VerificationPipeline struct {
+	Pipeline []Verifier
+}
+
+// Verify runs a given kubernetes object through the pipeline. If a verifier fails it returns an error.
+func (v *VerificationPipeline) Verify(toVerify *unstructured.Unstructured) error {
```
burgerdev: Returning a single error makes it harder to use the pipeline for issuing warnings or a structured report. Did you consider returning a slice of errors, or maybe even a `Finding` struct of sorts?

cli/verifier/verifiers.go
```diff
@@ -0,0 +1,11 @@
...
+// SPDX-License-Identifier: BUSL-1.1
+
+package verifier
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// Verifier is a function that verifies a given unstructured object and returns an error if verification fails.
+type Verifier func(toVerify *unstructured.Unstructured) error
```
burgerdev: It's usually a good idea to make this an interface, and add a `VerifierFunc` type for when the user only has a closure. Compare with `http.Handler` and [`http.HandlerFunc`](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/net/http/server.go;l=2286-2295).

cli/verifier/pipeline.go
```diff
@@ -0,0 +1,30 @@
...
+	var allErrors error
+	for _, toRun := range v.Pipeline {
+		if err := toRun(toVerify); err != nil {
+			allErrors = errors.Join(allErrors, err)
+		}
+	}
+	if allErrors != nil {
+		return allErrors
+	}
```
burgerdev: This implementation is almost trivial - do we need a type for it? I think I'd rather work like this at the call site:

```go
var verifiers []Verifier // from somewhere
var findings []error
for _, v := range verifiers {
	findings = append(findings, v.Verify(obj))
}
// Handle depending on flags or something
if !warnOnly {
	return errors.Join(findings)
}
for _, err := range findings {
	logger.Warn(err)
}
```

cli/verifier/pipeline_test.go
miampf: Sorry, I meant the individual verification functions that will eventually be added in `verifiers.go`. The new comment should be clearer

cli/verifier/pipeline.go
miampf: Then I'd remove the type (`pipeline.go` and `pipeline_test.go`) if that's okay with you?

cli/verifier/pipeline.go
miampf: I believe this becomes obsolete if `VerificationPipeline` is removed? Or do you want me to add a `Finding` struct that is used as a return value for each verifier? If yes, what data do you think it should contain (e.g. name, kind and error of the failing object)?

burgerdev: Let's maybe start with errors and see whether we'll eventually need findings.

cli/verifier/pipeline.go
burgerdev: Yes, but could you please add code using the verifiers to the `generate` command? Will probably need to export a `AllVerifiers() []Verifier` here.

miampf: Sure, but I'd write the verifiers in a separate PR

cli/cmd/generate.go
```diff
@@ -109,6 +110,30 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
+	var findings []error
+	for _, path := range paths {
+		fileContent, err := os.ReadFile(path)
+		if err != nil {
+			return fmt.Errorf("reading file to run generate on: %w", err)
+		}
+		resources, err := kubeapi.UnmarshalUnstructuredK8SResource(fileContent)
+		if err != nil {
+			return fmt.Errorf("parsing file to run generate on: %w", err)
+		}
```
burgerdev: You might want to add the `path` to the error, in case it's something nondescript like `syscall.Errno`.

cli/cmd/generate.go
```diff
@@ -109,6 +110,30 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
+		for _, v := range verifiers {
+			for _, r := range resources {
+				findings = append(findings, v.Verify(r))
+			}
+		}
+	}
+
+	for _, f := range findings {
+		if f != nil {
+			log.Warn("failed to verify resource", "error", f)
```
burgerdev: This loses information about the file/resource where the finding came from. How about logging above, instead of `findings = append(findings, v.Verify(r))`, including the file and `(kind, metadata.name)` as slog values?

cli/verifier/verifiers.go
```diff
@@ -0,0 +1,26 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// AllVerifiers returns all verifiers for k8s objects.
+func AllVerifiers() []Verifier {
+	return []Verifier{}
+}
+
+// VerificationFunc is a function that verifies a given unstructured object and returns an error if verification fails.
+type VerificationFunc func(toVerify *unstructured.Unstructured) error
```
burgerdev, line 16: I think it'd be more consistent if this was called `VerifierFunc`.

miampf: Then golangci-lint complains that this sounds bad and should be called `Func` instead which I think is a very unclear name, so I decided on this instead.

burgerdev: Ah, I see - thanks.

burgerdev (approved):

PR #1650, opened 2025-07-28
imagepuller: vendor evil registry

burgerdev:


imagepuller/internal/service/verify_test.go
```diff
@@ -39,7 +41,7 @@ func TestGetAndVerifyImage_EvilRegistry(t *testing.T) {
 		},
 		"wrong manifest digest is caught": {
 			// the evil registry responds to unknown digests with a default manifest
-			digest:  "sha256:6ad6bbb5735b84b10af42d2441e8d686b1d9a6cbf096b53842711ef5ddabd28d",
+			digest:  registry.WrongManifestDigest,
```
charludo: The comment above can probably be removed now

charludo (approved):
Alright, the evil registry refactor seems a lot nicer than the previous version 😄

Does this mean there's no interest in maintaining the separate repo? Seemed like something potentially useful for other projects as well.

imagepuller/internal/service/verify_test.go
```diff
@@ -29,6 +27,10 @@ func (s *StubStore) PutLayer(_, _ string, _ []string, _ string, _ bool, _ *stora
 }
 
 func TestGetAndVerifyImage_EvilRegistry(t *testing.T) {
+	registry := registry.New(t)
+	server := httptest.NewServer(registry)
+	t.Cleanup(server.Close)
```
katexochen: why move this out of the test case? that's global state between test cases!

burgerdev: It's not mutable state, though, and I need the instantiated registry in order to know the digests for the test matrix. I think it's better than hard-coding digests, and also better than the `init` method currently used by `burgerdev/evil-registry`.

katexochen: > It's not mutable state

That's hard to verify when looking at this test.

> init method currently used 

Those aren't allowed in contrast anyway. 

> wrongManifestDigest := digest([]byte("BAD MANIFEST"))

isn't the only value we use from the instantiated registry `wrongManifestDigest`, which is defined as the result of a pure function and a constant?

burgerdev: > > It's not mutable state
> 
> That's hard to verify when looking at this test.

Is it? The registry only serves GET requests, and its fields are write-only in the registry package and read-only in the test. 

> > wrongManifestDigest := digest([]byte("BAD MANIFEST"))
> 
> isn't the only value we use from the instantiated registry `wrongManifestDigest`, which is defined as the result of a pure function and a constant?

All digests exposed as public fields are used by the test, except `WrongBlobDigest`.

I guess I could expose functions for what are currently registry fields, and use them internally in the registry. Would you prefer that?

burgerdev: Implemented, ptal.

katexochen: Thanks, sorry for the slow response.

PR #1651, opened 2025-07-29
kata: gzip policy annotation

burgerdev:
This PR modifies policy generation (i.e., `contrast generate`) such that the resulting policy annotation is gzipped. This should allow staying within the limit for annotation values, even for large pods.

Comparison:

* Old Coordinator policy: 100kiB
* New Coordinator policy: 18kiB

#### TODO

- [x] add patch comments


packages/by-name/kata/kata-runtime/0023-misc-use-compressed-policy.patch
```diff
@@ -0,0 +1,223 @@
...
+             std::io::stdout().write_all(policy.as_bytes()).unwrap();
+         }
+-        general_purpose::STANDARD.encode(policy.as_bytes())
++
++        let mut compressed_policy = Vec::new();
++        let mut gz = flate2::GzBuilder::new()
++            .write(&mut compressed_policy, flate2::Compression::best());
++        gz.write_all(policy.as_bytes()).unwrap();
++        gz.finish().unwrap();
++
```
charludo:
```suggestion
+
```

`check_patches` does not seem to care, but `git am` does 🤷🏼‍♀️

charludo (approved):
LGTM - seems to work as intended, and patches look good (except for the comment below).
Also, 100->18kb is great! 🎉

packages/by-name/kata/kata-runtime/0023-misc-use-compressed-policy.patch
burgerdev: This is a valid patch, isn't it? I mean, I have spurious whitespace in the Rust source, and I'll remove it, but why should git care about that?

burgerdev:
Thanks for the review! @katexochen spotted a more minimal way, unzipping in the runtime instead of the agent, so I'll pivot to that. Sorry 😅

packages/by-name/kata/kata-runtime/0023-misc-use-compressed-policy.patch
charludo: It does, try it out with `git am`, git complains 🤷🏼‍♀️

burgerdev:
```
[...]patch:213: trailing whitespace.
        
warning: 1 line adds whitespace errors.
```
This one? You can actually configure this: https://stackoverflow.com/a/12396793. I probably did not notice because I'm using `-3` (maybe I shouldn't be using it).

burgerdev:
If the agent never sees zipped data, HOSTDATA needs to be the hash of the unzipped data, and now I find myself manipulating in `manifest.go`. That means that I'll need to implement the change for AKS, too. 

I'm currently leaning towards reverting to https://github.com/edgelesssys/contrast/commit/9d56c3eb411cc79c38dc509ad69d398afbaa5b18, but we could also delay this change until AKS is removed. wdyt @katexochen?

katexochen:
> I'm currently leaning towards reverting to [9d56c3e](https://github.com/edgelesssys/contrast/commit/9d56c3eb411cc79c38dc509ad69d398afbaa5b18), but we could also delay this change until AKS is removed. wdyt @katexochen?

No strong opinion, as you prefer.

PR #1652, opened 2025-07-30
ci: imagepuller-benchmark

charludo:
- on non-main branches making changes to imagepuller files: evaluate resource usage, fail if memory or storage consumption exceeds baseline by more than 20%
- on main branch: run evaluation, store result as new baseline in s3


imagepuller/client/client.go
charludo, line 1: Putting this under the `imagepuller/` path seemed like the simplest way of accessing the `internal/api` definitions.

tools/imagepuller-eval/go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/edgelesssys/contrast/imagepuller-eval
+
+go 1.24.0
+
+replace github.com/edgelesssys/contrast/imagepuller => ../../imagepuller
```
charludo: This probably shouldn't be required in the workspace...? Could not get go to *not* try and use the (non-existent) remote version from Github otherwise though

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,354 @@
...
+	"ghcr.io/edgelesssys/contrast/dmesg:v0.0.1@sha256:6ad6bbb5735b84b10af42d2441e8d686b1d9a6cbf096b53842711ef5ddabd28d",
+	// "ghcr.io/charludo/contrast/coordinator@sha256:6f966a922cc9a39d7047ed41ffafc7eb7a3c6a4fd8966cbf30fa902b455789f7",
+	// "quay.io/quay/busybox@sha256:92f3298bf80a1ba949140d77987f5de081f010337880cd771f7e7fc928f8c74d",
+	// "ghcr.io/edgelesssys/nginx-unprivileged@sha256:1d5be2aa3c296bd589ddd3c9bf2f560919e31ac32bae799a15dd182b6fdb042b",
+	// "quay.io/prometheus/prometheus@sha256:f20d3127bf2876f4a1df76246fca576b41ddf1125ed1c546fbd8b16ea55117e6",
+	// "ghcr.io/charludo/contrast/initializer@sha256:25b5ff1bd5259b6bd8c112b2321b8dc1857a9e63e0f2698c7ed4929c71ae514d",
+
+	// Large number of layers, and large file size, respectively. Make sure to adjust the client timeout!
+	// "tensorflow/tensorflow:latest-gpu@sha256:73fe35b67dad5fa5ab0824ed7efeb586820317566a705dff76142f8949ffcaff",
+	// "floriangeigl/datascience@sha256:7bd3d9827056abfd87ef089a18ac3815e2e1e0ea360cf429cc8c6060788c8050",
```
charludo: Not sure which/how many to choose.

.github/workflows/imagepuller-upload.yml
katexochen, line 1: I'd prefer to have this in one workflow, you can skip the steps based on the active trigger.

.github/workflows/imagepuller-eval.yml
```diff
@@ -0,0 +1,48 @@
+name: imagepuller evaluate resource usage
```
katexochen: nit: maybe s/eval(uate)?/benchmark/ everywhere, imagepuller-eval sounds weird for nix ears.. :rabbit:

.github/workflows/imagepuller-eval.yml
```diff
@@ -0,0 +1,48 @@
...
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - imagepuller/**
+      - tools/imagepuller/**
```
katexochen:
```suggestion
      - tools/imagepuller-eval/**
```

.github/workflows/imagepuller-eval.yml
```diff
@@ -0,0 +1,48 @@
...
+          S3_BUCKET_PATH: "imagepuller-eval"
+        run: |
+          aws s3 cp "s3://contrast-public/${S3_BUCKET_PATH}/baseline.json" "baseline.json"
+      - name: Run imagepuller-eval script
+        run: |
+          imagepullereval="$(nix build .#imagepuller-eval --print-out-paths)/bin/imagepuller-eval"
+          imagepuller="$(nix build .#imagepuller --print-out-paths)/bin/imagepuller"
+          sudo "$imagepullereval" "$imagepuller" -c baseline.json -o baseline-new.json
+      - name: Upload new baseline.json to S3
+        if: github.event_name != 'pull_request'
```
katexochen:
```suggestion
        if: github.event_name == 'push' && github.ref_name == 'main'
```
better to be specific here. Current condition won't exclude workflow_dispatch.

.github/workflows/imagepuller-eval.yml
```diff
@@ -0,0 +1,48 @@
...
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@b47578312673ae6fa5b5096b330d9fbac3d116df # v4.2.1
+        with:
+          role-to-assume: arn:aws:iam::795746500882:role/ContrastPublicBucketRW
+          aws-region: eu-central-1
+      - name: Download baseline JSON from S3
+        env:
+          S3_BUCKET_PATH: "imagepuller-eval"
```
katexochen: you don't need to use an env here, just put it inline. same below.

imagepuller/main
katexochen, line 1: oops

packages/by-name/imagepuller-eval/package.nix
```diff
@@ -0,0 +1,42 @@
+# Copyright 2025 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+{ lib, buildGoModule }:
+
+buildGoModule (finalAttrs: {
+  pname = "imagepuller-eval";
+  version = "0.1.0";
```
katexochen:
```suggestion
  name = "imagepuller-eval";
```

packages/by-name/imagepuller-eval/package.nix
```diff
@@ -0,0 +1,42 @@
...
+
+  sourceRoot = "${finalAttrs.src.name}/tools/imagepuller-eval";
+
+  env.CGO_ENABLED = 0;
+
+  ldflags = [ "-s" ];
+
+  preCheck = ''
+    export CGO_ENABLED=1
+  '';
```
katexochen: rather `doCheck = false`? imagepuller-eval has no tests, and imagepuller tests are already executed elsewhere.

tools/imagepuller-eval/go.mod
katexochen: Please add imagepuller to https://github.com/edgelesssys/contrast/blob/main/renovate.json5#L96-L101

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+	"golang.org/x/sys/unix"
+
+	"github.com/edgelesssys/contrast/imagepuller/client"
+	"github.com/shirou/gopsutil/v3/disk"
+	"github.com/spf13/cobra"
+)
+
+var imageList = []string{
+	"ghcr.io/edgelesssys/contrast/dmesg:v0.0.1@sha256:6ad6bbb5735b84b10af42d2441e8d686b1d9a6cbf096b53842711ef5ddabd28d",
+	"ghcr.io/charludo/contrast/coordinator@sha256:6f966a922cc9a39d7047ed41ffafc7eb7a3c6a4fd8966cbf30fa902b455789f7",
```
katexochen: Please don't use images from your own registry.

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+	slash := strings.LastIndex(name[:at], "/")
+	return name[slash+1 : at]
+}
+
+func cleanup(storagePath string) {
+	cmd := exec.Command("findmnt", "-rn", "-o", "TARGET")
+	output, err := cmd.Output()
+	if err != nil {
+		log.Fatalf("failed to run findmnt: %v", err)
+	}
```
katexochen: 1) use exec.CommandContext instead.

2) there is a specific pattern checking cmd.Output errors:

```go
    stdout, err := cmd.Output()
	var exitErr *exec.ExitError
	if errors.As(err, &exitErr) {
		return "", fmt.Errorf("program foo returned non-zero exit code, stderr is: %s", string(exitErr.Stderr))
	} else if err != nil {
		return "", fmt.Errorf("failed to execute foo: %w", err)
	}
```

Please apply everywhere.

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+	var mountpoints []string
+	scanner := bufio.NewScanner(strings.NewReader(string(output)))
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.Contains(line, storagePath) {
+			mountpoints = append(mountpoints, line)
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		log.Fatalf("error reading findmnt output: %v", err)
```
katexochen: Fatalf should only be used in a top-level main/run function. Return an error instead.

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+		return pid, nil
+	}
+	return 0, fmt.Errorf("no child found for PID %d", ppid)
+}
+
+func startServerWithMemoryTracking(serverPath, args string) (func() (int, error), int, error) {
+	var stderr bytes.Buffer
+	var cmd *exec.Cmd
+
+	timeCmd := exec.Command("which", "time")
```
katexochen: nit: always use command -v instead of which ;) which is not posix.

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+	out, err := timeCmd.Output()
+	if err != nil {
+		return nil, 0, fmt.Errorf("finding time bin: %w", err)
+	}
+	timePath := strings.TrimSpace(string(out))
+
+	if len(args) == 0 {
+		cmd = exec.Command(timePath, "-v", serverPath)
+	} else {
+		cmd = exec.Command(timePath, "-v", serverPath, args)
```
katexochen: args should be varadic:

func startServerWithMemoryTracking(serverPath, args ...string) (func() (int, error), int, error) {


```suggestion
		cmd = exec.Command(timePath, "-v", serverPath, args...)
```

then you won't need the check for len(args)

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	cmd.Stdout = nil
+	cmd.Stderr = &stderr
+
+	err = cmd.Start()
+	if err != nil {
```
katexochen:
```suggestion
    if err := err = cmd.Start(); err != nil {
```

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+
+	childPid, err := findChildPid(cmd.Process.Pid)
+	if err != nil {
+		return nil, 0, fmt.Errorf("failed to find server child PID: %w", err)
+	}
+
+	// Closure that will wait and extract MaxRSS after process exit
+	waitAndGetMaxRSS := func() (int, error) {
+		_ = cmd.Wait()
+		if err != nil {
```
katexochen: why aren't you checking the error of `cmd.Wait` here?

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+			if strings.Contains(line, "Maximum resident set size (kbytes)") {
+				parts := strings.Fields(line)
+				if len(parts) >= 6 {
+					kb, err := strconv.Atoi(parts[len(parts)-1])
+					if err != nil {
+						return 0, fmt.Errorf("parsing MaxRSS failed: %w", err)
+					}
+					return kb, nil
+				}
+			}
```
katexochen: reduce nesting by inverting the conditions and continuing in that case.

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+		if err != nil {
+			log.Fatalf("Failed to start server %s: %v", label, err)
+		}
+		time.Sleep(500 * time.Millisecond)
+
+		diskBefore, _ := getDiskUsage(storagePath)
+
+		start := time.Now()
+		err = client.Request(image, mountPoint)
+		if err != nil {
```
katexochen:
```suggestion
		if err = client.Request(image, mountPoint); err != nil {
```

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+		cleanup(mountPoint)
+		fmt.Printf("[%s]\n", extractName(image))
+
+		waitForRSS, childPid, err := startServerWithMemoryTracking(serverPath, args)
+		if err != nil {
+			log.Fatalf("Failed to start server %s: %v", label, err)
+		}
+		time.Sleep(500 * time.Millisecond)
+
+		diskBefore, _ := getDiskUsage(storagePath)
```
katexochen: Check the error?

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+		err = client.Request(image, mountPoint)
+		if err != nil {
+			log.Printf("Error pulling image %s: %v", image, err)
+		}
+
+		duration := time.Since(start)
+		if err := syscall.Kill(childPid, syscall.SIGKILL); err != nil {
+			log.Fatalf("Failed to kill server: %v", err)
+		}
+		diskAfter, _ := getDiskUsage(storagePath)
```
katexochen: Check the error?

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+
+	return cmd
+}
+
+func run(cmd *cobra.Command, args []string) error {
+	binPath := args[0]
+	output, _ := cmd.Flags().GetString("output")
+	compare, _ := cmd.Flags().GetString("compare")
+	threshold, _ := cmd.Flags().GetFloat64("threshold")
+	delta, _ := cmd.Flags().GetInt("delta")
```
katexochen: check the errors

tools/imagepuller-eval/main.go
```diff
@@ -0,0 +1,357 @@
...
+
+	// Dirs to clean up after each pull
+	imagepullerDir := "tmp_imagepuller"
```
katexochen: const?

katexochen (changes requested):

tools/imagepuller-eval/main.go
charludo: Isn't `command` a shell built-in? Using it here gives me `"command": executable file not found in PATH`.

tools/imagepuller-eval/main.go
charludo: Good question. Checking `err` immediately after always leads to a 137 exit code.

Moving the error check to after the `for` loop avoids this, but is also not really useful I think...? Added that version for now.

tools/imagepuller-eval/main.go
katexochen: Yes, you need to invoke it in shell. `which` isn't guaranteed to be there at all. Even better, use https://pkg.go.dev/os/exec#LookPath instead of shelling out.

tools/imagepuller-benchmark/main.go
```diff
@@ -0,0 +1,425 @@
...
+	return name[slash+1 : at]
+}
+
+func cleanup(storagePath string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, "findmnt", "-rn", "-o", "TARGET")
+	output, err := cmd.Output()
+
```
katexochen:
```suggestion
```

tools/imagepuller-benchmark/main.go
```diff
@@ -0,0 +1,425 @@
...
+	if err := os.RemoveAll(storagePath); err != nil {
+		return fmt.Errorf("removing directory %s: %w", storagePath, err)
+	}
+	return nil
+}
+
+func findChildPid(ctx context.Context, ppid int) (int, error) {
+	out, err := exec.CommandContext(ctx, "ps", "-o", "pid=", "--ppid", fmt.Sprint(ppid)).Output()
+
+	var exitErr *exec.ExitError
```
katexochen:
```suggestion
	out, err := exec.CommandContext(ctx, "ps", "-o", "pid=", "--ppid", fmt.Sprint(ppid)).Output()
	var exitErr *exec.ExitError
```

tools/imagepuller-benchmark/main.go
```diff
@@ -0,0 +1,425 @@
...
+}
+
+func startServerWithMemoryTracking(ctx context.Context, serverPath string, args ...string) (func() (int, error), int, error) {
+	var stderr bytes.Buffer
+	var cmd *exec.Cmd
+
+	timeCmd := exec.CommandContext(ctx, "which", "time")
+	out, err := timeCmd.Output()
+
+	var exitErr *exec.ExitError
```
katexochen:
```suggestion
	out, err := timeCmd.Output()
	var exitErr *exec.ExitError
```

tools/imagepuller-benchmark/main.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+
+	timePath := strings.TrimSpace(string(out))
+
+	argsFull := append([]string{"-v", serverPath}, args...)
+	cmd = exec.CommandContext(ctx, timePath, argsFull...)
+	cmd.Stdout = nil
+	cmd.Stderr = &stderr
+
+	if err := cmd.Start(); err != nil {
```
katexochen:
```suggestion
	cmd.Stderr = &stderr
	if err := cmd.Start(); err != nil {
```
never empty line between a command invocation and its error handling

baseline.json
katexochen, line 1: we don't need that in the repo, right?

tools/imagepuller-eval/main.go
katexochen: That doesn't look right, please investigate further.

baseline.json
charludo: No 🤦🏼‍♀️

tools/imagepuller-eval/main.go
charludo: Ah, thanks, changed it to `LookPath`. Still not sure how `command` could have been used though tbh, invoking `ExecCommand(ctx, "bash command -v time")`?

tools/imagepuller-eval/main.go
charludo: Looks like it was an issue with the error pipe handling. Should be resolved now.

.github/workflows/imagepuller-benchmark.yml
```diff
@@ -0,0 +1,44 @@
+name: imagepuller evaluate resource usage
```
katexochen:
```suggestion
name: imagepuller benchmark
```

katexochen (approved):

PR #1660, opened 2025-07-30
e2e/genpolicy-unsupported: add test

miampf:
Add an e2e test to check to make sure assumed unsupported configurations for genpolicy are actually unsupported.

### DIFF
--- a/.github/workflows/e2e_manual.yml
+++ b/.github/workflows/e2e_manual.yml
@@ -11,6 +11,7 @@ on:
           # keep-sorted start
           - atls
           - genpolicy
+          - genpolicy-unsupported
           - getdents
           - gpu
           - multiple-cpus
--- a/.github/workflows/e2e_regression.yml
+++ b/.github/workflows/e2e_regression.yml
@@ -38,7 +38,7 @@ jobs:
           - name: K3s-QEMU-TDX
             runner: TDX
             self-hosted: true
-        test-name: [getdents, genpolicy, regression]
+        test-name: [getdents, genpolicy, genpolicy-unsupported, regression]
         exclude:
           # getdents is a regression test for tardev-snapshotter
           - platform:


e2e/genpolicy-unsupported/genpolicy-unsupported_test.go
```diff
@@ -0,0 +1,75 @@
...
+			filePath := path.Join(testdata, file.Name())
+			yamlContent, err := os.ReadFile(filePath)
+			require.NoError(err)
+			require.NoError(os.WriteFile(path.Join(ct.WorkDir, file.Name()), yamlContent, 0o644))
+			t.Cleanup(func() {
+				require.NoError(os.Remove(path.Join(ct.WorkDir, file.Name())))
+			})
+
+			cfg := genpolicy.NewConfig(platform)
+			cli, err := genpolicy.New(path.Join(ct.WorkDir, "rules.rego"), path.Join(ct.WorkDir, "settings.json"), path.Join(ct.WorkDir, "layers-cache.json"), cfg.Bin)
```
burgerdev: Is there a reason why you can't use contrasttest for this? I fear that this invocation here might get out of sync (when we update CLI flags, for example), and then fail but for the wrong reason.

miampf: I thought invoking genpolicy would be more specific and thus less prone to false alarms but I can see your point. I'll run `contrast generate` instead here :)

e2e/genpolicy-unsupported/genpolicy-unsupported_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	resources = kuberesource.AddPortForwarders(resources)
+
+	// required for making everything available for genpolicy
+	ct := contrasttest.New(t)
+	ct.Init(t, resources)
+	require.True(t, t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
+	require.True(t, t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
+	require.True(t, t.Run("set", ct.Set), "contrast set needs to succeed for subsequent tests")
+	require.True(t, t.Run("contrast verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
```
burgerdev: These are not needed for this test's scope, are they?

e2e/genpolicy-unsupported/genpolicy-unsupported_test.go
```diff
@@ -0,0 +1,69 @@
...
+			require := require.New(t)
+			filePath := path.Join(testdata, file.Name())
+			yamlContent, err := os.ReadFile(filePath)
+			require.NoError(err)
+			require.NoError(os.WriteFile(path.Join(ct.WorkDir, file.Name()), yamlContent, 0o644))
+			t.Cleanup(func() {
+				require.NoError(os.Remove(path.Join(ct.WorkDir, file.Name())))
+			})
+
+			require.Error(ct.RunGenerate())
```
burgerdev: This is still a bit too generic for me. Could we add an expected substring per test, or something similar that ensures the test is failing for the right reason and we're not hitting any random errors?

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -183,26 +188,49 @@ func (ct *ContrastTest) Generate(t *testing.T) {
...
+		return err
+	}
+	return nil
 }
 
 // PatchManifestFunc defines a function type allowing the given manifest to be modified.
 type PatchManifestFunc func(manifest.Manifest) manifest.Manifest
 
-// PatchManifest modifies the current manifest by executing a provided PatchManifestFunc on it.
+// PatchManifest modifies the current manifest by executing a provided PatchManifestFunc on it. This function fail the test if something errors.
```
burgerdev:
```suggestion
// PatchManifest modifies the current manifest by executing a provided PatchManifestFunc on it. This function fails the test if it encounters an error.
```

e2e/genpolicy-unsupported/genpolicy-unsupported_test.go
```diff
@@ -0,0 +1,76 @@
...
+			require.NoError(os.WriteFile(path.Join(ct.WorkDir, file.Name()), yamlContent, 0o644))
+			t.Cleanup(func() {
+				require.NoError(os.Remove(path.Join(ct.WorkDir, file.Name())))
+			})
+
+			err = ct.RunGenerate()
+			require.Error(err)
+			expectedError, ok := expectedErrorMessages[file.Name()]
+			if !ok {
+				t.Fatal("test case does not have expected error message", "file", file.Name())
```
burgerdev: Consider `t.Fatalf`, otherwise these will be joined by space and the sentence does not make sense.

e2e/genpolicy-unsupported/genpolicy-unsupported_test.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		t.Run(file.Name(), func(t *testing.T) {
+			require := require.New(t)
+			filePath := path.Join(testdata, file.Name())
+			yamlContent, err := os.ReadFile(filePath)
+			require.NoError(err)
+			require.NoError(os.WriteFile(path.Join(ct.WorkDir, file.Name()), yamlContent, 0o644))
+			t.Cleanup(func() {
+				require.NoError(os.Remove(path.Join(ct.WorkDir, file.Name())))
```
burgerdev: I believe you can't `require` in cleanups because it's too late to fail the test.

e2e/genpolicy-unsupported/genpolicy-unsupported_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	platform, err := platforms.FromString(contrasttest.Flags.PlatformStr)
+	require.NoError(t, err)
+
+	runtimeHandler, err := manifest.RuntimeHandler(platform)
+	require.NoError(t, err)
+
+	resources := kuberesource.PatchRuntimeHandlers(kuberesource.CoordinatorBundle(), runtimeHandler)
+
+	resources = kuberesource.AddPortForwarders(resources)
```
burgerdev:
```suggestion
```

e2e/genpolicy-unsupported/genpolicy-unsupported_test.go
miampf: Oh okay, I thought this would work like `slog`. Will change that

PR #1661, opened 2025-07-31
manifest: remove TDX SVNs

katexochen:
This removes the following fields from the Contrast manifest:

- `.ReferenceValues.tdx.[].MinimumQeSvn`
- `.ReferenceValues.tdx.[].MinimumPceSvn`
- `.ReferenceValues.tdx.[].MinimumTeeTcbSvn`

These values are verified via collateral from the Intel PCS already, so there is no need for users to configure specific values.

### DIFF
--- a/docs/docs/howto/workload-deployment/generate-annotations.md
+++ b/docs/docs/howto/workload-deployment/generate-annotations.md
@@ -114,19 +114,13 @@ This must be done on a trusted machine, with a secure and trusted connection to 
 contrast generate --reference-values k3s-qemu-tdx resources/
 ```
 
-On bare-metal TDX, `contrast generate` is unable to fill in the `MinimumTeeTcbSvn` and `MrSeam` TCB values as they can vary between platforms.
-They will have to be filled in manually.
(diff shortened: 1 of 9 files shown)


docs/docs/howto/workload-deployment/generate-annotations.md
```diff
@@ -114,21 +114,15 @@ This must be done on a trusted machine, with a secure and trusted connection to
 contrast generate --reference-values k3s-qemu-tdx resources/
 ```
 
-On bare-metal TDX, `contrast generate` is unable to fill in the `MinimumTeeTcbSvn` and `MrSeam` TCB values as they can vary between platforms.
-They will have to be filled in manually.
+On bare-metal TDX, `contrast generate` is unable to fill in the `MrSeam` value as it depends on your platform configuration.
```
thomasten, line 117: MinimumTeeTcbSvn is also mentioned on `getting-started/deployment.md` and `howto/encrypted-storage.md`

katexochen: I know, there is some duplication we have to remove. The previous PR also didn't touch those. Will fix in a follow-up.

thomasten (approved):

PR #1662, opened 2025-08-01
docs: clarify significance of mesh annotations

burgerdev:
... and some minor fixes on the service mesh page.

### DIFF
--- a/docs/docs/architecture/components/service-mesh.md
+++ b/docs/docs/architecture/components/service-mesh.md
@@ -40,31 +40,28 @@ After policy generation, the annotations themselves aren't interpreted by the ru
 
 ### Ingress
 
-All TCP ingress traffic is routed over Envoy by default. Since we use
-[TPROXY](https://docs.kernel.org/networking/tproxy.html), the destination address
-remains the same throughout the packet handling.
+The service mesh ingress rule redirects all incoming TCP traffic transparently to the Envoy proxy.
+The rule is activated if either the ingress or the egress annotation is present on the pod.
+If the ingress annotation value is empty, incoming connections must present a client certificate signed by the [mesh CA certificate](#summary-of-certificate-roles).
+Envoy presents a certificate chain of the mesh certificate of the workload and the intermediate CA certificate.
 
-Any incoming connection is required to present a client certificate signed by the
-[mesh CA certificate](#summary-of-certificate-roles).
-Envoy presents a certificate chain of the mesh
-certificate of the workload and the intermediate CA certificate as the server certificate.
-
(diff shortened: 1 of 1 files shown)


docs/docs/architecture/components/service-mesh.md
```diff
@@ -166,6 +160,8 @@ spec:
           image: ghcr.io/edgelesssys/conversion:v1.2.3@...
 ```
 
+Setting the `contrast.edgeless.systems/servicemesh-egress` annotation without a value adds the service mesh sidecar, but has no effects beyond.
```
katexochen: Is that something we should actually support, or rather add a check to prevent this?
I don't like that we have `contrast.edgeless.systems/servicemesh-ingress: ""`  as most secure configuration and `contrast.edgeless.systems/servicemesh-egress: ""` as insecure one. I could imagine people mistaken one for the other.

burgerdev: Good question. I don't think there's a use case for adding the envoy without any config, so we should probably not support it. We could check this in the generic verifier (#1643 ff) so that it's reported for all resources at once, rather than failing an individual `AddServiceMesh`.

docs/docs/architecture/components/service-mesh.md
```diff
@@ -36,37 +34,31 @@ The service mesh container can be configured using the following object annotati
...
-
-Any incoming connection is required to present a client certificate signed by the
-[mesh CA certificate](#summary-of-certificate-roles).
-Envoy presents a certificate chain of the mesh
-certificate of the workload and the intermediate CA certificate as the server certificate.
+When the `contrast.edgeless.systems/servicemesh-ingress` annotation is set on the workload, all TCP ingress traffic is transparently proxied by Envoy.
+Any incoming connection is required to present a client certificate signed by the [mesh CA certificate](#summary-of-certificate-roles).
+Envoy presents a certificate chain of the mesh certificate of the workload and the intermediate CA certificate.
 
+The annotation value consists of triples of the form `<name>#<port>#<disable-TLS>`, defining exemptions from the mTLS requirement.
```
katexochen: We should document how to configure ingress in case there shouldn't be any exceptions, I think that's not totally clear here.

katexochen (approved):

docs/docs/architecture/components/service-mesh.md
burgerdev: Reworded

docs/docs/architecture/components/service-mesh.md
burgerdev: Changed in anticipation of #1717

burgerdev (commented):
Addressed outstanding comments and updated for #1717 and #1725

PR #1663, opened 2025-08-01
imagepuller: add unit tests

charludo:
This PR adds more in-depth tests of `getAndVerifyImage` and `storeAndVerifyLayers`, not focusing on the validation side like the evil registry tests did.

However, I've still used the (sometimes) evil registry extensively in this; AFAICT, it's by far the easiest way to test the functions' behaviors depending on what the registry answers, as well as getting an actual oci image for use in the layer-specific tests.

As a result, these tests look very similar to those against the evil registry; I'm ambivalent on keeping them separate or combining them. They *do* serve slightly different purposes, and it might be nice to tell just from the names.

### DIFF
--- a/imagepuller/internal/remote/remote.go
+++ b/imagepuller/internal/remote/remote.go
@@ -1,0 +1,28 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package remote
+
+import (
+	"github.com/google/go-containerregistry/pkg/name"
+	gcr "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+// DefaultRemote implements Remote, passing all function calls to the remote module.
+type DefaultRemote struct{}
+
+// Head returns a gcr.Descriptor for the given reference by issuing a HEAD request.
+func (DefaultRemote) Head(ref name.Reference, opts ...remote.Option) (*gcr.Descriptor, error) {
+	return remote.Head(ref, opts...)
+}
+
+// Image provides access to a remote image reference.
+func (DefaultRemote) Image(ref name.Reference, opts ...remote.Option) (gcr.Image, error) {
+	return remote.Image(ref, opts...)
+}
+
+// Index provides access to a remote index reference.
+func (DefaultRemote) Index(ref name.Reference, opts ...remote.Option) (gcr.ImageIndex, error) {
+	return remote.Index(ref, opts...)
+}
--- a/imagepuller/internal/service/verify.go
+++ b/imagepuller/internal/service/verify.go
@@ -17,21 +17,21 @@ import (
 
 // Error types which should be differentiable in tests.
 var (
-	ErrParseDigest   = errors.New("parsing image digest")
-	ErrRemoteImage   = errors.New("obtaining remote image")
-	ErrRemoteIndex   = errors.New("obtaining remote image index")
-	ErrValidateLayer = errors.New("validating layer")
+	errParseDigest         = errors.New("parsing image digest")
+	errUnexpectedMediaType = errors.New("unexpected media type")
+	errMissingPlatform     = errors.New("obtaining image digest for linux/amd64: platform missing from image index")
+	errValidateLayer       = errors.New("validating layer")
 )
 
 func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Logger, imageURL string) (gcr.Image, error) {
 	ref, err := name.NewDigest(imageURL)
 	if err != nil {
-		return nil, fmt.Errorf("%w: %w", ErrParseDigest, err)
+		return nil, fmt.Errorf("%w: %w", errParseDigest, err)
 	}
 
 	tr := transport.NewRetry(remote.DefaultTransport)
 
-	desc, err := remote.Head(ref, remote.WithContext(ctx), remote.WithTransport(tr))
+	desc, err := s.Remote.Head(ref, remote.WithContext(ctx), remote.WithTransport(tr))
 	if err != nil {
 		return nil, fmt.Errorf("obtaining descriptor: %w", err)
 	}
@@ -42,9 +42,9 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
 	case desc.MediaType.IsIndex():
 		log.Info("Received manifest list")
 
-		remoteImgIndex, err := remote.Index(ref, remote.WithContext(ctx), remote.WithTransport(tr))
+		remoteImgIndex, err := s.Remote.Index(ref, remote.WithContext(ctx), remote.WithTransport(tr))
 		if err != nil {
-			return nil, fmt.Errorf("%w: %w", ErrRemoteIndex, err)
+			return nil, fmt.Errorf("obtaining remote image index: %w", err)
 		}
 
 		manifest, err := remoteImgIndex.IndexManifest()
@@ -52,29 +52,28 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
 			return nil, fmt.Errorf("obtaining index manifest: %w", err)
 		}
 
-		var digestFound gcr.Hash
+		var digestFound *gcr.Hash
 		for _, m := range manifest.Manifests {
+			log.Info("MANIFEST", "name", m.Platform.String())
 			if m.Platform.String() == "linux/amd64" {
-				digestFound = m.Digest
+				digestFound = &m.Digest
 				break
 			}
 		}
-		if digestFound.String() == "" {
-			return nil, fmt.Errorf("obtaining image digest for linux/amd64: platform missing from image index")
+		if digestFound == nil {
+			return nil, errMissingPlatform
 		}
 		log.Info("Obtained actual image digest", "image_digest_linux", digestFound.String())
 
-		remoteImg, imgErr = remoteImgIndex.Image(digestFound)
+		remoteImg, imgErr = remoteImgIndex.Image(*digestFound)
 	case desc.MediaType.IsImage():
-		remoteImg, imgErr = remote.Image(ref, remote.WithContext(ctx), remote.WithTransport(tr))
+		remoteImg, imgErr = s.Remote.Image(ref, remote.WithContext(ctx), remote.WithTransport(tr))
 	default:
-		return nil, fmt.Errorf("unexpected media type %q", desc.MediaType)
+		return nil, fmt.Errorf("%w: %q", errUnexpectedMediaType, desc.MediaType)
 	}
 
-	if errors.Is(imgErr, context.DeadlineExceeded) {
-		return nil, fmt.Errorf("pull aborted (deadline exceeded): %w", imgErr)
-	} else if imgErr != nil {
-		return nil, fmt.Errorf("%w: %w", ErrRemoteImage, imgErr)
+	if imgErr != nil {
+		return nil, fmt.Errorf("obtaining remote image: %w", imgErr)
 	}
 
 	return remoteImg, nil
@@ -120,7 +119,7 @@ func (s *ImagePullerService) storeAndVerifyLayers(log *slog.Logger, remoteImg gc
 		ldManifest := manifest.Layers[idx].Digest.String()
 		ld := putLayer.CompressedDigest.String()
 		if ldManifest != ld {
-			return "", fmt.Errorf("%w: expected digest '%s' but got digest '%s'", ErrValidateLayer, ldManifest, ld)
+			return "", fmt.Errorf("%w: expected digest '%s' but got digest '%s'", errValidateLayer, ldManifest, ld)
 		}
 
 		log.Info("Applied and validated layer", "id", putLayer.ID, "size", n, "digest", ld)
--- a/imagepuller/internal/service/verify_test.go
+++ b/imagepuller/internal/service/verify_test.go
@@ -4,62 +4,153 @@
 package service
 
 import (
+	"context"
 	"fmt"
-	"io"
 	"log/slog"
 	"net/http/httptest"
 	"testing"
 
-	"github.com/containers/storage"
+	"github.com/edgelesssys/contrast/imagepuller/internal/remote"
 	"github.com/edgelesssys/contrast/imagepuller/internal/test/registry"
+	gcr "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/opencontainers/go-digest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-type StubStore struct {
-	putLayerLayer digest.Digest
-	storage.Store
-}
+var zeroDigest = "0000000000000000000000000000000000000000000000000000000000000000"
 
-func (s *StubStore) PutLayer(_, _ string, _ []string, _ string, _ bool, _ *storage.LayerOptions, _ io.Reader) (*storage.Layer, int64, error) {
-	return &storage.Layer{CompressedDigest: s.putLayerLayer}, 0, nil
-}
-
-func TestGetAndVerifyImage_EvilRegistry(t *testing.T) {
+// TestGetAndVerifyImage contains unit tests for the getAndVerifyImage function.
+func TestGetAndVerifyImage(t *testing.T) {
 	tests := map[string]struct {
-		digest  string
-		wantErr error
+		digest     string
+		imageRef   string
+		wantErr    error
+		stubRemote stubRemote
 	}{
-		"missing digest is rejected": {
-			digest:  "",
-			wantErr: ErrParseDigest,
+		"digest missing, no tag": {
+			imageRef: "busybox",
+			wantErr:  errParseDigest,
 		},
-		"wrong manifest digest is caught": {
-			digest:  registry.WrongManifestDigest(),
-			wantErr: ErrRemoteImage,
+		"digest malformed, no tag": {
+			digest:   "sha256:000",
+			imageRef: "busybox",
+			wantErr:  errParseDigest,
 		},
-		"wrong index digest is caught": {
-			digest:  registry.WrongIndexDigest(),
-			wantErr: ErrRemoteIndex,
+		"digest missing algorithm, no tag": {
+			digest:   zeroDigest,
+			imageRef: "busybox",
+			wantErr:  errParseDigest,
 		},
-		"correct index digest, wrong manifest digest is caught": {
-			digest:  registry.IndexForWrongManifestDigest(),
-			wantErr: ErrRemoteImage,
+		"digest missing, with tag": {
+			imageRef: "busybox:v0.0.1",
+			wantErr:  errParseDigest,
+		},
+		"digest malformed, with tag": {
+			digest:   "sha256:000",
+			imageRef: "busybox:v0.0.1",
+			wantErr:  errParseDigest,
+		},
+		"digest missing algorithm, with tag": {
+			digest:   zeroDigest,
+			imageRef: "busybox:v0.0.1",
+			wantErr:  errParseDigest,
+		},
+		"head request error": {
+			digest:     fmt.Sprintf("sha256:%s", zeroDigest),
+			imageRef:   "busybox:v0.0.1",
+			stubRemote: stubRemote{errDescriptor: assert.AnError},
+			wantErr:    assert.AnError,
+		},
+		"remote image failure": {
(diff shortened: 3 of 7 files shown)


imagepuller/internal/service/verify.go
```diff
@@ -17,10 +17,18 @@ import (
...
+	ErrUnexpectedMediaType = errors.New("unexpected media type")
+	ErrRemoteImage         = errors.New("obtaining remote image")
+	ErrRemoteIndex         = errors.New("obtaining remote image index")
+	ErrMissingPlatform     = errors.New("obtaining image digest for linux/amd64: platform missing from image index")
+
+	ErrObtainingLayers   = errors.New("obtaining the image layers")
+	ErrObtainingManifest = errors.New("obtaining image manifest")
+	ErrValidateLayer     = errors.New("validating layer")
+	ErrReadLayer         = errors.New("reading layer")
+	ErrPutLayer          = errors.New("putting layer to store")
```
katexochen: I really think this is to much. It's okay if we need to identify a specific error when it is important, but this shouldn't be the default. We don't need to differentiate each specific error.

imagepuller/internal/service/verify.go
```diff
@@ -59,21 +67,20 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
 				break
 			}
 		}
-		if digestFound.String() == "" {
-			return nil, fmt.Errorf("obtaining image digest for linux/amd64: platform missing from image index")
+		var emptyDigest gcr.Hash
+		if digestFound == emptyDigest {
```
katexochen:
```suggestion
		if digestFound == *new(gcr.Hash) {
```

not sure this is nicer..

imagepuller/internal/service/verify_test.go
```diff
@@ -102,3 +110,175 @@ func TestStoreAndVerifyLayers_EvilRegistry(t *testing.T) {
...
+			var s ImagePullerService
+			_, err := s.getAndVerifyImage(context.Background(), log, imageURL)
+
+			assert.ErrorIs(err, tc.wantErr)
+		})
+	}
+}
+
+type StubImage struct {
+	Real gcr.Image
```
katexochen:
```suggestion
```
You can access the embedded type via its name: s.Image.Layers()

imagepuller/internal/service/verify_test.go
```diff
@@ -102,3 +110,175 @@ func TestStoreAndVerifyLayers_EvilRegistry(t *testing.T) {
...
+
+	LayersFunc   func() ([]gcr.Layer, error)
+	ManifestFunc func() (*gcr.Manifest, error)
```
katexochen: Don't fill the stub with functions, but return types.

imagepuller/internal/service/verify.go
charludo: Then the tests lose meaning though, if we only check for *any* failure.

imagepuller/internal/service/verify.go
```diff
@@ -59,21 +92,19 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
 				break
 			}
 		}
-		if digestFound.String() == "" {
-			return nil, fmt.Errorf("obtaining image digest for linux/amd64: platform missing from image index")
+		if digestFound == *new(gcr.Hash) {
```
burgerdev: It would be clearer if digestFound was a pointer and this was checking for nil.

imagepuller/internal/service/verify.go
burgerdev: Now that we have the `Remote` interface, you could inject test-only errors into the stub (make `headErr` and friends `error`s instead of `bool`s) and verify that they are propagated? That would at least remove the need for the "obtaining" errors, afaict.

imagepuller/internal/service/verify_test.go
```diff
@@ -12,17 +14,25 @@ import (
...
 type StubStore struct {
 	putLayerLayer digest.Digest
+	putLayerErr   bool
+
 	storage.Store
 }
 
 func (s *StubStore) PutLayer(_, _ string, _ []string, _ string, _ bool, _ *storage.LayerOptions, _ io.Reader) (*storage.Layer, int64, error) {
+	if s.putLayerErr {
+		return nil, 0, errors.New("")
```
burgerdev: [`assert.AnError`](https://pkg.go.dev/github.com/stretchr/testify/assert#AnError) is useful here.

imagepuller/internal/service/verify_test.go
```diff
@@ -102,3 +118,225 @@ func TestStoreAndVerifyLayers_EvilRegistry(t *testing.T) {
 		})
 	}
 }
+
+type StubRemote struct {
+	headErr  bool
+	imageErr bool
+	indexErr bool
+
+	DefaultRemote
```
burgerdev: I agree with @katexochen that this delegation here is unexpected and complicates the test. On the other hand, faking the interfaces provided by the library would require significant amounts of code, so I would rather take the evil registry.

Did you consider https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/fake? That might be the sweet spot in between, but I did not check how easy it is to use.

imagepuller/internal/service/verify_test.go
```diff
@@ -102,3 +118,225 @@ func TestStoreAndVerifyLayers_EvilRegistry(t *testing.T) {
...
+				imageURL = fmt.Sprintf("busybox:v0.0.1@%s", tc.digest)
+			}
+
+			if tc.startRegistry {
+				server := httptest.NewServer(registry.New())
+				t.Cleanup(server.Close)
+
+				baseURL := server.Listener.Addr().String()
+				imageURL = fmt.Sprintf("%s/%s", baseURL, imageURL)
+			}
```
burgerdev: None of these are technically URLs. The correct term for the full thing is _image reference_, and it's built up like this:

```abnf
image-reference = [registry "/"] repository [":" tag] ["@" digest]
```

In a URL, `registry` would be the host and port and `repository` the path.

imagepuller/internal/service/verify_test.go
```diff
@@ -102,3 +118,225 @@ func TestStoreAndVerifyLayers_EvilRegistry(t *testing.T) {
...
+		},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			log := slog.Default()
+
+			var imageURL string
+			if tc.imageURLNoTag {
+				imageURL = fmt.Sprintf("@%s", tc.digest)
```
burgerdev: In combination with `tc.startRegistry`, the resulting reference looks invalid: `localhost:123/@sha256:...`.

charludo: oops, forgot `busybox`!

Invalid nonetheless 😄

imagepuller/internal/service/verify_test.go
charludo: I did, did not really work out / manual implementation turned out to be less work.

imagepuller/internal/service/verify.go
charludo: That would work. But I am not sure if moving half of those error definitions into the test really improves the overall code quality 😅

imagepuller/internal/service/verify_test.go
```diff
@@ -102,3 +118,280 @@ func TestStoreAndVerifyLayers_EvilRegistry(t *testing.T) {
...
+
+var zeroDigest = "0000000000000000000000000000000000000000000000000000000000000000"
+
+type StubRemote struct {
+	mediaType  types.MediaType
+	imageIndex StubImageIndex
+
+	headErr  bool
+	imageErr bool
+	indexErr bool
```
burgerdev: Here is where I meant to introduce configurable errors, like e.g.

https://github.com/edgelesssys/contrast/blob/0c18004dfc697745199fb1acaba646c7d0074a24/coordinator/internal/userapi/userapi_test.go#L927-L930

imagepuller/internal/test/stubs/stubs.go
katexochen, line 1: This can be just declared in the _test.go file where it is used, no need to make it a package.

charludo: So just drop the (fixup) commit?

I thought it was all getting a bit unreadable and long, so separated the stubs out.

imagepuller/internal/service/verify_test.go
```diff
@@ -55,14 +51,20 @@ func TestGetAndVerifyImage_EvilRegistry(t *testing.T) {
...
+			}
 			_, err := s.getAndVerifyImage(
 				t.Context(),
 				slog.Default(),
 				fmt.Sprintf("%s/busybox:v0.0.1@%s", server.Listener.Addr().String(), tc.digest),
 			)
 
-			assert.ErrorIs(err, tc.wantErr)
+			if tc.stubRemote.Err != nil {
+				assert.ErrorIs(err, tc.stubRemote.Err)
```
katexochen: I'd prefer to have this explicitly in `wantErr`. Looking at the test cases, it looks like the test cases using the stub should succeed.

imagepuller/internal/service/verify_test.go
```diff
@@ -4,48 +4,44 @@
...
-			wantErr: ErrRemoteIndex,
+			digest:     registry.WrongIndexDigest(),
+			stubRemote: stubs.StubRemote{Err: errors.New("remote index")},
 		},
 		"correct index digest, wrong manifest digest is caught": {
-			digest:  registry.IndexForWrongManifestDigest(),
-			wantErr: ErrRemoteImage,
+			digest:     registry.IndexForWrongManifestDigest(),
+			stubRemote: stubs.StubRemote{Err: errors.New("remote image")},
 		},
```
katexochen: None of these test cases is actually using the registry.

imagepuller/internal/service/verify_test.go
```diff
@@ -102,3 +108,159 @@ func TestStoreAndVerifyLayers_EvilRegistry(t *testing.T) {
...
+			} else {
+				digest = tc.digest
+			}
+
+			var imageRef string
+			if tc.imageRefNoTag {
+				imageRef = fmt.Sprintf("busybox@%s", digest)
+			} else {
+				imageRef = fmt.Sprintf("busybox:v0.0.1@%s", digest)
+			}
```
katexochen: move this directly into the test cases as digest/imageRef and remove emptyDigest/imageRefNoTag.

imagepuller/internal/service/verify_test.go
```diff
@@ -102,3 +108,159 @@ func TestStoreAndVerifyLayers_EvilRegistry(t *testing.T) {
...
+
+			s := ImagePullerService{Remote: tc.stubRemote}
+			_, err := s.getAndVerifyImage(context.Background(), log, imageRef)
+
+			if tc.stubRemote.Err != nil {
+				assert.ErrorIs(err, tc.stubRemote.Err)
```
katexochen: use explicit error in wantErr instead.

imagepuller/internal/test/stubs/stubs.go
katexochen: If you need, put it in a stubs_test.go in the package where it is needed.

imagepuller/internal/remote/remote.go
```diff
@@ -0,0 +1,35 @@
...
+	gcr "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+// Remote allows stubbing remote calls.
+type Remote interface {
+	Head(ref name.Reference, options ...remote.Option) (*gcr.Descriptor, error)
+	Image(ref name.Reference, opts ...remote.Option) (gcr.Image, error)
+	Index(ref name.Reference, opts ...remote.Option) (gcr.ImageIndex, error)
+}
```
katexochen: interface should be defined where it is used.

imagepuller/internal/service/verify_test.go
```diff
@@ -305,71 +202,37 @@ func TestGetAndVerifyImage(t *testing.T) {
...
-
-type StubLayer struct {
-	gcr.Layer
-}
-
-func (s StubLayer) Compressed() (io.ReadCloser, error) {
-	return nil, errStubLayerRead
-}
-
 func TestStoreAndVerifyLayers(t *testing.T) {
```
katexochen, line 162: This is using the registry, too, isn't it?

imagepuller/internal/service/verify_test.go
```diff
@@ -250,36 +147,36 @@ func TestGetAndVerifyImage(t *testing.T) {
...
-				err:        errors.New("remote image"),
-				mediaType:  types.DockerManifestList,
-				imageIndex: StubImageIndex{includeLinux: true, imageErr: true},
+			stubRemote: stubs.StubRemote{
+				Err:        errors.New("remote image"),
+				MediaType:  types.DockerManifestList,
+				ImageIndex: stubs.StubImageIndex{IncludeLinux: true, ImageErr: true},
 			},
 		},
 		"index success": {
```
katexochen: test names confuse me here. What does "index success" mean? Is the function executed successfully here? Is this a happy path?

imagepuller/internal/test/stubs/stubs.go
```diff
@@ -0,0 +1,169 @@
...
+	gcr "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/opencontainers/go-digest"
+	"github.com/stretchr/testify/assert"
+)
+
+// StubStore stubs the containers/storage Store.
+type StubStore struct {
+	PutLayerLayer digest.Digest
```
katexochen: name is confusing, this isn't the layer but the digest that is embedded in a layer. rename or move the wrapping into the test cases (personally prefer the latter).

imagepuller/internal/test/stubs/stubs.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	return &storage.Layer{CompressedDigest: s.PutLayerLayer}, 0, nil
+}
+
+// StubRemote stubs the DefaultRemote.
+type StubRemote struct {
+	MediaType  types.MediaType
+	ImageIndex StubImageIndex
+
+	Err error
```
katexochen: If you only have a single Err field, how can I stub a successful `Head` call and a then failing `Index` call?
Better stick to the pattern of `<FunName><ReturnName>` fields in stubs.

imagepuller/internal/test/stubs/stubs.go
```diff
@@ -0,0 +1,169 @@
...
+	if s.Err != nil {
+		return nil, s.Err
+	}
+	return s.ImageIndex, nil
+}
+
+// StubImageIndex stubs the go-containerregistry ImageIndex.
+type StubImageIndex struct {
+	IncludeLinux bool
+	ImageErr     bool
```
katexochen: again, better split stub implementation and payload construction. You can provide a func that returns a grc.IndexManifest to fill the stub.

imagepuller/internal/test/stubs/stubs.go
```diff
@@ -0,0 +1,169 @@
...
+var ErrStubLayerRead = errors.New("read layer")
+
+// StubLayer stubs the go-containerregistry Layer.
+type StubLayer struct {
+	gcr.Layer
+}
+
+// Compressed stubs the Layers's Compressed.
+func (s StubLayer) Compressed() (io.ReadCloser, error) {
+	return nil, ErrStubLayerRead
```
katexochen: error should be filled by the test case.

imagepuller/internal/test/stubs/stubs.go
```diff
@@ -0,0 +1,169 @@
...
+
+// RawManifest stubs the ImageIndex's RawManifest.
+func (s StubImageIndex) RawManifest() ([]byte, error) {
+	return nil, nil
+}
+
+// Image stubs the ImageIndex's Image.
+func (s StubImageIndex) Image(_ gcr.Hash) (gcr.Image, error) {
+	if s.ImageErr {
+		return nil, assert.AnError
```
katexochen: error should be filled by the test case.

imagepuller/internal/service/verify_test.go
charludo: Yes, `TestStoreAndVerifyLayers` is using the registry, as mentioned in our call on Monday. I need an image from somewhere. The call to the registry is identical between all test cases, so this can't really be argued as changing the SUT.

imagepuller/internal/service/verify_test.go
charludo: Yes

imagepuller/internal/service/verify_test.go
katexochen: The question is why this is another test as the registry test above.

charludo: No technical reason, just a separation of the integration tests against the evil registry, and the unittests.

katexochen: as the unit tests use the registry, too, how are they not integration tests? How should I decide to which of those test functions I should add my test case?

charludo: The integration tests against the evil registry test the behavior of the imagepuller against the responses from a (mocked, evil) registry. The unittests use a single, constant call to that registry as a convenience wrapper to obtain a remote image.

That should make it simple to sort new test cases into the two functions. As another factor: if a test requires anything other than the constant image used in the unittests, it should not go in the unittests.

As I said, there's no technical requirement to keep these two test functions separate; we can combine them. I just thought it prudent to keep separate concerns separate.

katexochen: I'm not against keeping the separation, just document your thoughts a little.

charludo: OK, sure thing. Will add comments to the test fns.

imagepuller/internal/service/verify_test.go
katexochen: And where is the difference to the other success cases? Should also check there is no error for the happy path.

imagepuller/internal/service/verify_test.go
charludo: Hm, true, with the unittest and the refactoring, this is covered by the unittests. I'll remove this one.

imagepuller/internal/service/verify_test.go
charludo: Renamed them. Both test a different happy path (simple manifest MediaType vs index, requiring an additional intermediate step).

> Should also check there is no error for the happy path.

Isn't that what is happening here? 😅 `wantErr` is nil => `assert.ErrorIs` fails if `err` is not `nil`

imagepuller/internal/service/verify_test.go
```diff
@@ -4,81 +4,232 @@
...
-}
+var (
+	errDescriptor        = errors.New("descriptor")
+	errRemoteImage       = errors.New("remote image")
+	errRemoteIndex       = errors.New("remote index")
+	errCompressed        = errors.New("compressed layer")
+	errObtainingLayers   = errors.New("remote index")
+	errObtainingManifest = errors.New("compressed layer")
+	errPutLayer          = errors.New("put layer")
+)
```
katexochen: nit: I don't get why you would need these in the test. Just use errors.New("my unique error") for each test case, then you know you got the error you injected, right?

imagepuller/internal/service/verify_test.go
```diff
@@ -4,81 +4,232 @@
...
-				fmt.Sprintf("%s/busybox:v0.0.1@%s", server.Listener.Addr().String(), tc.digest),
-			)
+			t.Log(registry.BlobDigest())
+
+			store := &stubStore{
+				putLayerDigest: digest.NewDigestFromEncoded(digest.SHA256, registry.BlobDigest()[7:]),
+				putLayerErr:    tc.wantErr,
+			}
+			s := ImagePullerService{Logger: log, Store: store, Remote: tc.stubRemote.DefaultRemote}
+			realImg, err := s.getAndVerifyImage(t.Context(), log, fmt.Sprintf("%s/busybox:v0.0.1@%s", server.Listener.Addr().String(), registry.ManifestDigest()))
```
katexochen: nit: wrap long line

katexochen (approved):

imagepuller/internal/service/verify_test.go
charludo: Yes, but I need them both in the `tc.WantErr` and the stubs.

katexochen: > them

no, you just need a single one. You could use assert.AnError

charludo: Hm, that'd work, yes. Will change.

PR #1673, opened 2025-08-05
service-mesh: allow egress without ingress

burgerdev:
The service mesh used to always set up a secure ingress configuration even if the annotation was not set, preventing the exclusive use of egress. This PR allows setting up egress without ingress by setting the special ingress annotation value `DISABLED`.

For testing, we're configuring the vote-bot with a disabled ingress and confirm that there is no `TPROXY` redirection.

### DIFF
--- a/internal/kuberesource/sets.go
+++ b/internal/kuberesource/sets.go
@@ -556,7 +556,8 @@ func Emojivoto(smMode serviceMeshMode) []any {
 		smEgressConfigAnnotationKey:  "emoji#127.137.0.1:8081#emoji-svc:8080##voting#127.137.0.2:8081#voting-svc:8080",
 	})
 	voteBot.WithAnnotations(map[string]string{
-		smEgressConfigAnnotationKey: "web#127.137.0.3:8081#web-svc:443",
+		smIngressConfigAnnotationKey: "DISABLED",
+		smEgressConfigAnnotationKey:  "web#127.137.0.3:8081#web-svc:443",
 	})
 
 	return resources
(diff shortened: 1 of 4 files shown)


internal/kuberesource/sets.go
```diff
@@ -546,6 +547,9 @@ func Emojivoto(smMode serviceMeshMode) []any {
 		smIngressConfigAnnotationKey: "web#8080#false",
 		smEgressConfigAnnotationKey:  "emoji#127.137.0.1:8081#emoji-svc:8080##voting#127.137.0.2:8081#voting-svc:8080",
 	})
+	voteBot.WithAnnotations(map[string]string{
+		smEgressConfigAnnotationKey: "web#127.137.0.3:8081#web-svc:443",
+	})
```
katexochen: not sure if this makes sense. The vote bot isn't part of the confidential deployment, it simulates a user accessing the service. Why are we setting a service mesh annotation on it? It shouldn't have a contrast runtime class.

katexochen (commented):
Update of the relevant docs sections needed.

internal/kuberesource/sets.go
burgerdev: Oh, that's interesting - this is not what's currently implemented. None of our resources have a `runtimeClassName` set, e.g. 

https://github.com/edgelesssys/contrast/blob/e115304b5d273e1cbff981e142b7ed71e0293580/internal/kuberesource/sets.go#L286-L312

Instead, the runtime class is added globally with

https://github.com/edgelesssys/contrast/blob/e115304b5d273e1cbff981e142b7ed71e0293580/e2e/servicemesh/servicemesh_test.go#L44

which is why the egress annotation here works.

burgerdev: As discussed in private, we're keeping the bot in the mesh in order to have more test targets.

burgerdev:
I'll document this in #1662.

PR #1685, opened 2025-08-11
secure image store: init

charludo:
An ephemeral volume will now be provided to each pod, which is automatically mounted after setting up a LUKS-encryption and integrity-protection on it. Instead of storing container image layers in memory, they are stored on this encrypted device, significantly reducing memory requirements for many deployments.

Previously, the sum of uncompressed container image sizes needed to be added to the actual memory requirements of the containers. This is no longer necessary.

This feature is enabled by default. Please reduce memory requirements of your deployments accordingly. In order to retain the old behavior, please see the docs on how to disable securemounting globally or on a per-pod-basis.


katexochen:
I'd assume we need to

- Update the documentation regarding image pulling, memory requirements etc
- Reduce the memory requirements for your e2e tests
- Add a tests we are actually pulling to disk and not mem
- Write a PR description that explains provides a short explanation of the change to users

charludo:
> * Update the documentation regarding image pulling, memory requirements etc

Ah right, I meant to ask: is this something that should just always be enabled? Or enabled/disabled by default, and somehow toggleable? That also somewhat affects all of the other subtasks.

docs/docs/_media/runtime.drawio.svg
katexochen, line 1: I think the previous proportions looked much better. Feels like there is also enough space in the old graphic to fit the additional box.

docs/docs/howto/workload-deployment/deployment-file-preparation.md
```diff
@@ -147,12 +147,16 @@ spec: # v1.PodSpec
 
 :::note
 
-On bare metal platforms, container images are pulled from within the guest CVM and stored in encrypted memory.
-The images are pulled and uncompressed into encrypted memory, so the uncompressed image size needs to be taken into account when setting the container limits.
+Container images are pulled from within the pod-VMs using the [Contrast image puller](../../architecture/components/runtime#pod-vm-image).
+The images are pulled and uncompressed either into encrypted memory, or onto an encrypted storage device, depending on whether the [Contrast secure mount](../../architecture/components/runtime#pod-vm-image) feature is being used.
```
katexochen: This should state what the default is.

docs/docs/architecture/components/runtime.md
```diff
@@ -70,6 +70,49 @@ For each container scheduled to run in the pod, the kata agent will request the
...
+    contrast.edgeless.systems/trusted-store-size: 0
+```
+
+To disable Contrast secure mount globally, specify the `--skip-trusted-store` flag in the `generate` command.
+
+:::note
+
+If the Contrast secure mount feature is disabled, container images are pulled and uncompressed into encrypted memory.
+Memory limits must be adjusted accordingly.
+See [Pod resources](../../howto/workload-deployment/deployment-file-preparation#pod-resources) for details.
```
katexochen: This should be a how-to.

docs/docs/architecture/components/runtime.md
```diff
@@ -70,6 +70,49 @@ For each container scheduled to run in the pod, the kata agent will request the
 The image puller verifies the checksums of both the image manifest, as well as of all subsequently pulled image layers against the provided digest.
 For this reason, unpinned images aren't supported.
 
+### Contrast secure mount
```
katexochen: IMO the wording here is confusing and should be changed (not only here, but in most places). It should be clear that this is about **storing pulled images on secured storage**. Maybe call it "# Contrast secured image pull storage" or something similar. The way it is currently worded makes to differentiate from the encrypted volume feature.

docs/docs/architecture/components/runtime.md
```diff
@@ -70,6 +70,49 @@ For each container scheduled to run in the pod, the kata agent will request the
...
+If the Contrast secure mount feature is disabled, container images are pulled and uncompressed into encrypted memory.
+Memory limits must be adjusted accordingly.
+See [Pod resources](../../howto/workload-deployment/deployment-file-preparation#pod-resources) for details.
+
+:::
+
+:::warning
+
+Currently, it's not possible to detect a specific replay attack by the host system wherein the host replaces the contents of the encrypted block device by a previously captured snapshot of the disk.
+Since the snapshot consists purely of ciphertext generated by the pod-VM, it reads as correctly integrity-protected and the pod-VM can't distinguish it from fresh data.
```
katexochen: Needs a source.

docs/docs/architecture/components/runtime.md
```diff
@@ -70,6 +70,49 @@ For each container scheduled to run in the pod, the kata agent will request the
...
+
+Importantly, this **doesn't** allow a compromised host to decrypt the ciphertext stored on the device or to inject arbitrary data.
```
katexochen:
```suggestion
Importantly, this **doesn't** allow a malicious host to decrypt the ciphertext stored on the device or to inject arbitrary data.
```

docs/docs/architecture/components/runtime.md
```diff
@@ -70,6 +70,49 @@ For each container scheduled to run in the pod, the kata agent will request the
...
+Plausible attacks are thus limited to reverting the block device's data to an earlier state, possibly causing a Denial of Service, a risk already inherent to a compromised host.
```
katexochen: I disagree with this statement. replay/rollback attacks offer a more attack surface than just DoS. Using dm-integrity protected storage instead of encrypted memory weakens the security posture. This should be clearly stated here and presented as a trade-off, where the decision is left to the user.

Please also mention/link that in the hardening section, as it is pretty relevant in that context.

docs/docs/_media/runtime.drawio.svg
charludo: Changed it

docs/docs/architecture/components/runtime.md
charludo: Ah, good idea. I didn't know where exactly to put it. "Preparing the deployment files" was my hunch, but didn't feel quite right. Will move them.

docs/docs/architecture/components/runtime.md
charludo: "Contrast trusted image store"? to keep roughly with the naming scheme from the rest of the PR?

katexochen: I don't find _trusted_ to be a good fit. Trusting is something a relying party might do (or not, that's why it can be disabled).

charludo: Hm, I see your point. "secure image storage"? (Just to keep it a bit short?) Then I'd also swap that out everywhere else.

imagestore/internal/service/mount_linux.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	var mountArgs []string
+	if len(r.Flags) > 0 {
+		mountArgs = append([]string{"-o", strings.Join(r.Flags, ",")}, p.MapperDevice, r.MountPoint)
+	} else {
+		mountArgs = []string{p.MapperDevice, r.MountPoint}
+	}
+	if out, err := runCmd(ctx, "mount", mountArgs...); err != nil {
+		_, closeErr := runCmd(ctx, "cryptsetup", "luksClose", p.MapperName)
```
katexochen: there is a close func on cryptsetup.Device

imagestore/internal/service/mount_linux.go
```diff
@@ -0,0 +1,75 @@
...
+	return nil
+}
+
+func runCmd(ctx context.Context, name string, args ...string) (string, error) {
+	c, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+	cmd := exec.CommandContext(c, name, args...)
+	out, err := cmd.CombinedOutput()
+	return string(out), err
+}
```
katexochen: check for exit err, include out in the error if it is an exist err. Don't return out, it's not used anywhere.

imagestore/internal/service/service.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/edgelesssys/contrast/imagestore/internal/api"
+	"golang.org/x/sync/singleflight"
+)
+
+var mountGroup singleflight.Group
```
katexochen: no global state, this should be part of the service.

imagestore/internal/service/service.go
```diff
@@ -0,0 +1,57 @@
...
+	p, err := getAndVerifyParams(r)
+	if err != nil {
+		return nil, fmt.Errorf("verifying request parameters: %w", err)
+	}
+	log = log.With(
+		slog.String("device_path", p.DevicePath),
+		slog.String("mapper", p.MapperDevice),
+	)
+
+	_, err, repeat := mountGroup.Do(r.MountPoint, func() (any, error) {
```
katexochen: What's the intention of using a singleflight here? What are we trying to prevent? We are expecting this function to only be called once right?

imagestore/internal/service/service.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, setupLuksAndMount(ctx, log, r, p)
+	})
+	if err != nil {
+		return nil, fmt.Errorf("creating and mounting LUKS device: %w", err)
+	}
+	if repeat {
+		return nil, fmt.Errorf("repeated mount attempt")
+	}
```
katexochen: That's not what a single flight guarantees. It just prevents _parallel_ calls of the function. See https://goplay.tools/snippet/qVUptJzNY8_p

packages/by-name/kata/kata-runtime/0026-agent-use-separate-unix-socket-for-image-pulling.patch
katexochen, line 1: why is the patch order changed here?

packages/scripts.nix
```diff
@@ -46,6 +46,8 @@
       nix-update --version=skip --flake legacyPackages.x86_64-linux.snp-id-block-generator
       echo "Updating imagepuller package" >&2
       nix-update --version=skip --flake legacyPackages.x86_64-linux.imagepuller
+      echo "Updating imagestore package" >&2
+      nix-update --version=skip --flake legacyPackages.x86_64-linux.imagestore
```
katexochen: I think that's the wrong spot, isn't it? imagestore depends on Contrast, so it must be ordered after. See the comment.

imagestore/internal/service/service.go
```diff
@@ -0,0 +1,57 @@
...
+
+var mountGroup singleflight.Group
+
+// SecureImageStoreService is the struct for which the SecureImageStore ttRPC service is implemented.
+type SecureImageStoreService struct {
+	Logger *slog.Logger
+}
+
+// SecureMount is a ttRPC service which pulls and mounts docker images.
+func (s *SecureImageStoreService) SecureMount(ctx context.Context, r *api.SecureMountRequest) (response *api.SecureMountResponse, retErr error) {
```
katexochen: s/r/req/

imagestore/internal/service/service.go
```diff
@@ -0,0 +1,57 @@
...
+	log.Info("Handling secure image store mount request")
+	log.Debug("Mount request details", "options", r.Options, "flags", r.Flags)
+
+	defer func() {
+		if retErr != nil {
+			log.Error("Request failed", "err", retErr)
+		}
+	}()
+
+	p, err := getAndVerifyParams(r)
```
katexochen: s/p/params/

imagestore/main.go
```diff
@@ -0,0 +1,102 @@
...
+		SilenceUsage: true,
+		RunE:         run,
+	}
+}
+
+func run(_ *cobra.Command, _ []string) error {
+	log := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
+		Level: slog.LevelDebug,
+	}))
+	ctxSignal, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
```
katexochen: use the context from cobra.Command

katexochen: s/ctxSignal/ctx/

packages/by-name/imagestore/package.nix
```diff
@@ -0,0 +1,26 @@
+# Copyright 2025 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+{ buildGoModule }:
+
+buildGoModule (finalAttrs: {
+  pname = "imagestore";
+  version = builtins.readFile ../../../version.txt;
+
+  src = ../../../.;
```
katexochen: Filter src?

imagestore/internal/service/service.go
charludo: Removed singleflight completely. I initially had trouble with quickly repeating, seemingly simultaneous mount requests from kata, the singleflight was more of a sanity check and way to ensure only one request was fulfilled. Should have been removed before marking as ready for review, sorry.

imagestore/main.go
charludo: Why? 😅 `ctx` is used forther down as the Context of the `errgroup`

packages/by-name/imagestore/package.nix
charludo: Would love to, but I was not able to get this to work unless I included the `go.work` which then in turn requires everything else included in the `go.work` to also be included. If you have an idea I am all ears

packages/by-name/kata/kata-runtime/0026-agent-use-separate-unix-socket-for-image-pulling.patch
charludo: How do you mean?

Edit: ah, I see. Fixed.

imagestore/main.go
katexochen: The top level/generic context should be called ctx. The errgroup context is also derived from ctx, so it is a ctxSignal, too, and the naming doesn't help to differentiate, rather suggests the other ctx wouldn't be targeted by the signals, which is wrong.

charludo: Alright. Makes sense, thanks for the explanation. Will change.

packages/by-name/kata/kata-runtime/0026-agent-use-separate-unix-socket-for-image-pulling.patch
katexochen: Not fixed, now the patch is added twice in the first commit.

renovate.json5
```diff
@@ -97,6 +97,7 @@
       matchPackageNames: [
         "github.com/edgelesssys/contrast",
         "github.com/edgelesssys/contrast/imagepuller",
+        "github.com/edgelesssys/contrast/imagestore",
```
katexochen: Actually needed? imagestore isn't depended on and there is no replacement of it.

internal/kuberesource/parts.go
katexochen, line 1: Why are we adding the volume here to the coordinator? Shouldn't this be injected by the CLI if desired?

imagestore/internal/service/mount_linux.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}()
+
+	device, err := cryptsetup.NewDevice(params.DevicePath, params.KeyFile, params.MapperName)
+	if err != nil {
+		return fmt.Errorf("preparing device: %w", err)
+	}
+	defer func() {
+		if err := device.Close(ctx); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("closing luks device: %w", err))
```
katexochen: Shouldn't this be only done if retErr != nil?

cli/cmd/generate.go
katexochen, line 1: I don't think 5311864263d14fe67ff193bd4468e29c7a9662e4 should be a fixup. Fix the commit message to state that this is a change in the cli package.

internal/kuberesource/mutators.go
```diff
@@ -301,6 +303,92 @@ func AddDmesg(resources []any) []any {
 	return out
 }
 
+// AddImageStore adds a PersistentVolumeClaim, a pod volume for it, and a holder container
+// that attaches the PVC's block device so it appears in the pod VM.
+func AddImageStore(resources []any) []any {
+	getConfigs := func(storeSize string) (*applycorev1.ContainerApplyConfiguration, *applycorev1.VolumeApplyConfiguration) {
+		holderContainer := Container().
+			WithName("pvc-holder").
+			WithImage("ghcr.io/edgelesssys/bash@sha256:cabc70d68e38584052cff2c271748a0506b47069ebbd3d26096478524e9b270b").
```
katexochen: why can't we use a pause image here?

internal/kuberesource/parts.go
katexochen: This is again added in a commit and then removed in 5311864263d14fe67ff193bd4468e29c7a9662e4, please fix.

imagepuller/internal/api/imagepuller.go
katexochen, line 1: Should cbbf71366940a117a8d1028cd7e9c87cdf985e21 be a fixup of 3e576af4c29cccb2765cb0db7ac66381fe52f4cd?

docs/docs/architecture/components/runtime.md
```diff
@@ -70,6 +70,25 @@ For each container scheduled to run in the pod, the kata agent will request the
 The image puller verifies the checksums of both the image manifest, as well as of all subsequently pulled image layers against the provided digest.
 For this reason, unpinned images aren't supported.
 
+### Contrast secure image store
+
+In order to reduce memory requirements, Contrast supports mounting an [ephemeral volume](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/) into each pod-VM.
+Before first use, the underlying block device issued by Kubernetes is LUKS-encrypted and integrity protected with keys generated inside the pod-VM.
```
katexochen, line 76: Is the key actually generated in the pod VM or received from the Coordinator?

docs/docs/architecture/components/runtime.md
```diff
@@ -70,6 +70,25 @@ For each container scheduled to run in the pod, the kata agent will request the
...
+These keys are never persisted or transferred outside of the pod-VM, meaning the volume is only usable from within the specific pod it's attached to, and only for the duration of that pod's lifetime.
+
+:::warning
+
+It's not possible to detect a [specific kind of replay attack][1] by the host system wherein the host replaces the contents of the encrypted block device by a previously captured snapshot of the disk.
+Since the snapshot consists purely of ciphertext generated by the pod-VM, it reads as correctly integrity-protected and the pod-VM can't distinguish it from fresh data.
+
+Importantly, this **doesn't** allow a malicious host to decrypt the ciphertext stored on the device or to inject arbitrary data.
+
+The trade-off between reduced resource requirements and a weakened security posture must be evaluated on a per-use-case basis.
```
katexochen, line 86: This should link to the section that tells a user how to make that decision in practice.

docs/docs/howto/hardening.md
```diff
@@ -83,6 +83,11 @@ Another type of dynamic content are persistent volumes.
 Any volumes mounted to the pod need to be scrutinized, and sensitive data must not be written to unprotected volumes.
 Ideally, a volume is mounted as a raw block device and authenticated encryption is added within the confidential container.
 
+This is also the approach taken by the [secure image store](./secure-image-store.md) feature, which stores image layers outside of memory by setting up encryption and integrity protection on a raw block device.
+Nevertheless, this weakens the overall security posture compared to storing images entirely in memory.
+For further details, refer to the linked documentation section.
```
katexochen, line 89: As this is "how-to harden": add a sentence that images can be pulled into memory by disabling the image store.

docs/docs/getting-started/deployment.md
```diff
@@ -70,8 +70,9 @@ What matters is that the **sum of memory limits across all main containers** in
 
 If `initContainer`s require in sum more memory than the combined usage of all main containers and sidecars, increase the limit of a main container to ensure that the sum of limits for all main containers is **at least as large** as the `initContainer`'s memory requirement.
 
-On bare-metal platforms, the container images are pulled from within the guest CVM and stored in encrypted memory.
-Thus, the uncompressed image size of needs to be added to the memory limits of containers.
+On bare-metal platforms, the container images are pulled from within each pod-VM.
+If the [Contrast secure image store](../howto/secure-image-store.md) feature is disabled, the images are stored in encrypted memory.
```
katexochen: Unclear what the default is.

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,66 @@
+# Secure Image Store
+
+This section provides guidance on how to configure Contrast's secure image store feature to match your needs.
```
katexochen: which needs? this should state which problem we are solving.

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,66 @@
+# Secure Image Store
```
katexochen: no title casing: "Secure image store". I also suggest adding a verb like configure in the beginning.

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,66 @@
...
+## Prerequisites
+
+A running Contrast deployment.
+
+## How-To
+
+The secure image store is enabled by default, providing each pod with `10Gi` of storage.
+This amount can be adjusted on a per-pod basis, the feature can be disabled for individual pods, or it can be disabled globally.
+
+### Adjusting the size of the image store
```
katexochen, line 21: This could describe some situations in which we would want to adjust the size.

docs/docs/howto/troubleshooting.md
```diff
@@ -162,7 +162,7 @@ Upstream backport that's fixing the bug is pending: https://github.com/container
 ## VM runs out of memory
 
 Since pod VMs are statically sized, it's easier to run out of memory due to misconfigurations.
-Setting the right memory limits is even more important on bare metal, where the image layers need to be stored in the guest memory, too.
+Setting the right memory limits is even more important on bare metal, especially in the case that the [secure image store is disabled](./secure-image-store.md), meaning the image layers need to be stored in guest memory, too.
```
katexochen, line 165: We could also suggest in line 173 to check out the image store feature.

docs/docs/howto/workload-deployment/deployment-file-preparation.md
```diff
@@ -147,12 +147,16 @@ spec: # v1.PodSpec
 
 :::note
 
-On bare metal platforms, container images are pulled from within the guest CVM and stored in encrypted memory.
-The images are pulled and uncompressed into encrypted memory, so the uncompressed image size needs to be taken into account when setting the container limits.
+Container images are pulled from within the pod-VMs using the [Contrast image puller](../../architecture/components/runtime#pod-vm-image).
+By default, the images are pulled and uncompressed into an encrypted storage device using [Contrast secure image store](../secure-image-store.md).
+In this case, image sizes don't need to be taken into account when setting the container memory limits.
+
+However, if this feature is disabled, images will instead be pulled into encrypted memory, and the uncompressed image sizes need to be taken into account.
```
katexochen: no new paragraph

packages/by-name/kata/kata-runtime/0027-agent-use-custom-implementation-for-secure-mounting.patch
katexochen, line 1: Patches change again in c97cefe92dfcd9d641b20f66ac4b3e70a74048bf

katexochen (commented):
only some smaller stuff left

imagepuller/internal/api/imagepuller.go
charludo: No, IMO it should be a separate commit. However, I renamed https://github.com/edgelesssys/contrast/commit/cbbf71366940a117a8d1028cd7e9c87cdf985e21 and dropped the "use imagestore" commit, since it was unusable without the changed from https://github.com/edgelesssys/contrast/commit/cbbf71366940a117a8d1028cd7e9c87cdf985e21.

docs/docs/architecture/components/runtime.md
charludo: Yes: https://github.com/edgelesssys/contrast/blob/827cd1081f8efcc5bc36e81310780e4e5315449c/imagestore/internal/service/verify.go#L71-L74

internal/kuberesource/mutators.go
charludo: Using a pause image:

```go
WithImage("registry.k8s.io/pause:3.6@sha256:3d380ca8864549e74af4b29c10f9cb0956236dfb01c40ca076fb6c37253234db").

```

here gives me the error I mentioned a while back: 

```
failed to create containerd task: failed to create shim task: "CreateContainerRequest is blocked by policy: agent_policy:119: allow_create_container_input
```

and I am not sure why.

katexochen: maybe @burgerdev has an idea?

internal/kuberesource/mutators.go
```diff
@@ -301,6 +303,92 @@ func AddDmesg(resources []any) []any {
...
+			WithCommand("/usr/local/bin/bash", "-c", "sleep infinity").
+			WithRestartPolicy(corev1.ContainerRestartPolicyAlways).
+			WithSecurityContext(SecurityContext().
+				WithPrivileged(true).SecurityContextApplyConfiguration).
+			WithResources(ResourceRequirements().
+				WithMemoryLimitAndRequest(50),
+			).
+			WithVolumeDevices(
+				applycorev1.VolumeDevice().
+					WithDevicePath("/dev/trusted_store").
```
katexochen: Is that path required by kata? Otherwise we should change it to image store.

katexochen (approved):
looks good from my side, thanks!

commit message of df5beff4951d1bd20b6f6ffeb18418e3ad65d544 needs to be fixed, still states trusted store.

internal/kuberesource/mutators.go
charludo: Yes, it's their magic device they check for.

I suppose I could include changing its name in the patch, if you'd prefer that?

katexochen: nah, it's ok in that case.

imagepuller/internal/api/imagepuller.go
```diff
@@ -6,4 +6,4 @@ package api
 //go:generate protoc --go_out=. --go_opt=paths=source_relative --go-ttrpc_out=. --go-ttrpc_opt=paths=source_relative imagepuller.proto
 
 // Socket is the unix socket of the imagepuller API.
-const Socket = "/run/confidential-containers/cdh.sock"
+const Socket = "/run/confidential-containers/imagepuller.sock"
```
burgerdev: If we use the socket from https://github.com/kata-containers/kata-containers/pull/11655/files#r2253617153, we might have a chance to not have to change this again in the near future.

imagestore/internal/service/mount_linux.go
```diff
@@ -0,0 +1,84 @@
...
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/edgelesssys/contrast/imagestore/internal/api"
+	"github.com/edgelesssys/contrast/internal/cryptsetup"
+)
+
+func setupLuksAndMount(ctx context.Context, log *slog.Logger, req *api.SecureMountRequest, params *SecureImageStoreParams) (retErr error) {
+	if err := os.WriteFile(params.KeyFile, params.Key, 0o644); err != nil {
```
burgerdev: Permissions should be `0o600`, I suppose?

imagestore/internal/service/mount_linux.go
```diff
@@ -0,0 +1,84 @@
...
+	if err := os.MkdirAll(req.MountPoint, 0o755); err != nil {
+		return fmt.Errorf("failed to create mountpoint: %w", err)
+	}
+
+	var mountArgs []string
+	if len(req.Flags) > 0 {
+		mountArgs = append([]string{"-o", strings.Join(req.Flags, ",")}, params.MapperDevice, req.MountPoint)
+	} else {
+		mountArgs = []string{params.MapperDevice, req.MountPoint}
+	}
```
burgerdev: That's surprising: I'd have expected `req.Options` to be passed with `-o` and flags to be appended to the command. We should document what they are for.

imagestore/internal/service/mount_linux.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return nil
+}
+
+func runCmd(ctx context.Context, name string, args ...string) error {
+	c, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+	cmd := exec.CommandContext(c, name, args...)
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("executing '%s': %w: %s", name, err, out)
```
burgerdev: `args` are not reflected in the error message.

imagestore/internal/service/verify.go
```diff
@@ -0,0 +1,96 @@
...
+	DeviceID      string
+	DevicePath    string
+	DataIntegrity string
+	MapperName    string
+	MapperDevice  string
+	Key           []byte
+	KeyFile       string
+}
+
+func getAndVerifyParams(req *api.SecureMountRequest) (*SecureImageStoreParams, error) {
```
burgerdev, line 28: There are a couple of error returns here where an explicit [status code](https://pkg.go.dev/google.golang.org/grpc@v1.75.0/codes#Code) would make sense. Use with [`status.Errorf`](https://pkg.go.dev/google.golang.org/grpc/status#Errorf).

imagestore/internal/service/verify.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil, fmt.Errorf("generating mapper suffix: %w", err)
+	}
+	mapperName := fmt.Sprintf("secure-%s", hex.EncodeToString(mapperSuffix))
+	mappedDev := filepath.Join("/dev/mapper", mapperName)
+
+	keyFileSuffix := make([]byte, 8)
+	if _, err := rand.Read(keyFileSuffix); err != nil {
+		return nil, fmt.Errorf("generating mapper suffix: %w", err)
+	}
+	keyFile := fmt.Sprintf("/run/key_%s", hex.EncodeToString(keyFileSuffix))
```
burgerdev: You could just use `%x`, which does the right thing for byte slices: https://cs.opensource.google/go/go/+/refs/tags/go1.25.0:src/fmt/doc.go;l=56-61.

imagestore/main.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		return nil
+	})
+
+	eg.Go(func() error {
+		<-ctxEg.Done()
+		if ctx.Err() != nil {
+			log.Info("Received signal, shutting down.")
+		} else {
+			log.Info("Context done, shutting down", "err", ctxEg.Err())
```
burgerdev: This branch is executed when the gRPC server fails, isn't it? The "Context done" is a bit vague.

packages/by-name/kata/kata-runtime/0027-agent-use-custom-implementation-for-secure-mounting.patch
```diff
@@ -0,0 +1,72 @@
...
++        volume_type: volume_type.to_string(),
++        options: options.clone(),
++        flags,
++        mount_point: mount_point.to_string(),
++        ..Default::default()
++    };
++
++    let _ = secure_mount_client
++        .secure_mount(
++            ttrpc::context::with_timeout(AGENT_CONFIG.cdh_api_timeout.as_nanos() as i64),
```
burgerdev, line 53: It looks like this timeout was not used before - is it sufficiently long?

cli/cmd/generate.go
```diff
@@ -403,6 +404,9 @@ func patchTargets(paths []string, imageReplacementsFile, runtimeHandler string,
 				return fmt.Errorf("injecting Service Mesh: %w", err)
 			}
 		}
+		if !skipImageStore {
+			kubeObjs = kuberesource.AddImageStore(kubeObjs)
+		}
```
burgerdev, line 409: I guess @katexochen and @charludo already discussed this at length, but I just wanted to mention that I'm not quite sure about injecting a store by default.

1. What if there is no CSI driver?
2. What if the desired CSI driver is not the default? 
3. What if it needs special configuration?

We might still need to instruct people on how to modify the volume...

internal/kuberesource/mutators.go
burgerdev: If we had the input, we could eyeball which of the rules is failing:

https://github.com/kata-containers/kata-containers/blob/150c90e32aca8f9eadb3a938d798cbdeb1d45fe6/src/tools/genpolicy/rules.rego#L124-L146

internal/kuberesource/mutators.go
```diff
@@ -301,6 +303,92 @@ func AddDmesg(resources []any) []any {
...
+
+		containerExists := false
+		for _, c := range spec.InitContainers {
+			if c.Name != nil && *c.Name == *holderContainer.Name {
+				containerExists = true
+				break
+			}
+		}
+		if !containerExists {
+			spec.InitContainers = append(spec.InitContainers, *holderContainer)
```
burgerdev: Shouldn't the holder container go first, so that all images are pulled into the trusted store?

imagepuller/internal/api/imagepuller.go
```diff
@@ -7,3 +7,11 @@ package api
 
 // Socket is the unix socket of the imagepuller API.
 const Socket = "/run/confidential-containers/imagepuller.sock"
+
+// StorePathMemory is the default dir used for the store cache.
+const StorePathMemory = "/run/kata-containers/image-memory"
+
+// StorePathStorage is the preferred dir for the store cache.
+// The kata-agent uses this dir as the mount point for securely mounted storage.
+// It is only used as store cache when a imagestoreed storage device is actually available.
```
burgerdev: typo

imagepuller/internal/service/service.go
```diff
@@ -43,6 +47,26 @@ func (s *ImagePullerService) PullImage(ctx context.Context, r *api.ImagePullRequ
 		}
 	}()
 
+	var storePath string
+	if len(s.StorePathOverride) != 0 {
```
burgerdev: Nit: `s.StorePathOverride != ""` is often clearer.

e2e/imagestore/imagestore_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+				require.NoError(checkDiskSize(tc.size, diskSize, tc.maxOffset, tc.unit, diskUnit, c.Name))
+			}
+		})
+	}
+}
+
+// extractDiskSize capture the "Size" field from df -h / output.
+func extractDiskSize(logs string) (float64, string, error) {
+	re := regexp.MustCompile(`(?m)^.*\s+([\d.]+)([MG])\s+([\d.]+)[KMGTP]?\s+([\d.]+)[KMGTP]?\s+([\d.]+)%\s+/$`)
```
burgerdev: Why do you use `-h` if you then need to parse it?

e2e/imagestore/imagestore_test.go
```diff
@@ -0,0 +1,168 @@
...
+		"block device size configurable": {
+			name:       "imagestore-configured",
+			size:       2.0,
+			unit:       "G",
+			maxOffset:  1.0,
+			annotation: "2Gi",
+		},
+		"disabled through annotation": {
+			name:       "imagestore-disabled",
+			size:       230.0,
```
burgerdev: Where does this come from? Might make sense to document.

docs/docs/howto/hardening.md
```diff
@@ -60,7 +60,7 @@ If an application authenticates with a certificate signed by the Contrast Mesh C
 2. The command line arguments of containers are exactly the arguments specified in the resource definitions.
 3. All environment variables are either specified in resource definitions, in the container image manifest or in a settings file for the Contrast CLI.
 4. The containers run in a confidential VM that matches the reference values in the manifest.
-5. The containers' root filesystems are mounted in encrypted memory.
+5. The containers' root filesystems are mounted in encrypted memory, or optionally in encrypted and integrity-protected ephemeral storage.
```
burgerdev: This is the wrong way round, because _skipping_ the secure mount is optional.

docs/docs/howto/secure-image-store.md
burgerdev, line 1: The docs should mention that this results in an additional initContainer.

cli/cmd/generate.go
katexochen: Fine for me to switch the flag, docs must be adapted accordingly.

e2e/imagestore/imagestore_test.go
charludo: Readability of the errors. If you want me to change it LMK

imagestore/internal/service/mount_linux.go
charludo: Whoops, yes, looking at the manpage, that's exactly how it should be.

These are *always* empty though. I guess this might change in the future in kata, but for now, I'd say let's completely remove the flag-passing?

Edit: removed it for now.

packages/by-name/kata/kata-runtime/0027-agent-use-custom-implementation-for-secure-mounting.patch
charludo: Default is 50 seconds. Since our new cryptsetup lib doesn't zero out the entire disk, the should be plenty even for large disks IMO.

cli/cmd/generate.go
charludo: Fine by me as well, @katexochen @burgerdev what's the final verdict? (Just to make sure before I invest the time... 😅)

burgerdev: Sorry, I might have expressed myself badly. My concern is mostly about configurability, not so much about the default. Defaulting to on is fine.

Regarding the listed points: let's address that in a follow-up.

e2e/imagestore/imagestore_test.go
burgerdev: You mean errors returned from the test? Those are under your control :) I would not mind comparing two numbers formatted with, say, `%e`.

There's also a potential for edge cases around the unit boundary:

```
-rw-r--r-- 1 root root 1048576 Sep  3 15:45 /tmp/1Mi
-rw-r--r-- 1 root root 1148576 Sep  3 15:45 /tmp/1Mi+x
```
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/loop1      980K   14K  895K   2% /tmp/mnt/1Mi
/dev/loop2      1.1M   14K  981K   2% /tmp/mnt/1Mi+x
```

internal/kuberesource/mutators.go
charludo: Update: renamed it, since it now needs to be mentioned by name in the docs.

imagestore/internal/service/mount_linux.go
burgerdev: I don't think we should ignore them - either we pass them on as expected, or we fail loudly. Since the semantics are not clear to us (and might not be upstream), I'd argue for verifying that they are not set.

e2e/imagestore/imagestore_test.go
charludo: Alright, fair. I'll adjust the test.

docs/docs/howto/secure-image-store.md
charludo: Added a section to the how-to

e2e/imagestore/imagestore_test.go
charludo: Erm. Very good question. It *should* be the sum of memory limits of the pod if I'm not mistaken, but that amounts to only 40MB. Checked manually though, and it is around 225MB.

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+Rerun `contrast generate` and reapply your deployment for the change to take effect:
+
+```bash
+contrast generate resources/
+kubectl apply -f resources/
+```
+
+### Disabling the image store globally
```
katexochen:
```suggestion
### Disabling the image store injection on `generate`
```

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,119 @@
...
+
+To disable the image store globally, specify the `--skip-image-store` flag in the `contrast generate` command, then reapply your deployment:
```
katexochen:
```suggestion
To disable the image store injection, specify the `--skip-image-store` flag in the `contrast generate` command, then reapply your deployment:
```

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,119 @@
...
+
+```bash
+contrast generate --skip-image-store resources/
+kubectl apply -f resources/
+```
+
+### Advanced configuration
+
+If you need further customization - for example because you require the use of a custom [CSI driver](https://kubernetes-csi.github.io/docs/drivers.html) -
```
katexochen: don't use em-dashes (these aren't even real em-dashes).

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,119 @@
...
+you can disable the image store (either [for a single pod](#disabling-the-image-store-for-a-single-pod) or [globally](#disabling-the-image-store-globally)), and instead inject your own image store.
```
katexochen:
```suggestion
you can disable the image store injection (either [for a single pod](#disabling-the-image-store-for-a-single-pod) or [globally](#disabling-the-image-store-globally)), and instead provide your own image store.
```

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,119 @@
...
+
+Internally, whether or not the image store is used depends exclusively on the presence of the magic device `/dev/image_store`.
+If a device with this name is present in the pod-VM, it will be prepared, mounted and used as the secure image store.
+This requires that the device is bound to a [sidecar container](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/), a type of init container which isn't required to exit.
```
katexochen:
```suggestion
This requires that the device is bound to a [sidecar container](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/).
```

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,119 @@
...
+To the definition of the pod for which you wish to handle the secure image store manually, add a sidecar `initContainer` and an ephemeral volume:
+
+```yaml
+spec: # v1.Pod, v1.PodTemplateSpec
+  initContainers:
+    - command:
+        - /usr/local/bin/bash
+        - "-c"
+        - sleep infinity
+      image: "ghcr.io/edgelesssys/bash"
```
katexochen: We shouldn't suggest to use that image. Also, the pinning is missing here.

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,119 @@
...
+### Disabling the image store globally
+
+To disable the image store globally, specify the `--skip-image-store` flag in the `contrast generate` command, then reapply your deployment:
+
+```bash
+contrast generate --skip-image-store resources/
+kubectl apply -f resources/
+```
+
+### Advanced configuration
```
katexochen:
```suggestion
### Manual image store setup
```

docs/docs/howto/secure-image-store.md
```diff
@@ -0,0 +1,119 @@
...
+            accessModes:
+              - ReadWriteOnce
+            resources:
+              requests:
+                storage: 10Gi
+            volumeMode: Block
+      name: image-store
+```
+
+The configuration shown above corresponds to the one generated by `contrast generate` when the image store feature is enabled.
```
katexochen: This won't be kept in sync, so we shouldn't state this. Just describe what it provides (a volume that is used to pull images).

e2e/imagestore/imagestore_test.go
burgerdev: 1. The guest VM should get 512MiB + sum of containers as memory. https://github.com/edgelesssys/contrast/blob/6e1ea4e35166347d8d4271b3cd00915945389ff1/internal/platforms/platforms.go#L102-L113
2. We allocate half of that for `/run`, which is where the container rootfs lives, I assume. https://github.com/edgelesssys/contrast/blob/6e1ea4e35166347d8d4271b3cd00915945389ff1/packages/nixos/kata.nix#L71-L79

That should explain it, right?

charludo: Ah! Forgot about the +512. Yes, that seems correct then.

burgerdev: But on a GPU VM the test would fail, correct?

charludo: I'd assume so, yes

internal/kuberesource/mutators.go
katexochen: How was this resolved?

charludo: We were not able to get this to work with any of a number of pause images. Unclear why. Using the initializer image now, to save on memory and to be able to reuse an existing published image.

katexochen: Please add a comment to document this.

e2e/imagestore/imagestore_test.go
```diff
@@ -0,0 +1,178 @@
...
+	require.True(t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
+	require.True(t.Run("set", ct.Set), "contrast set needs to succeed for subsequent tests")
+	require.True(t.Run("contrast verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
+
+	ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(2*time.Minute))
+	defer cancel()
+	require.NoError(ct.Kubeclient.WaitForStatefulSet(ctx, ct.Namespace, "coordinator"))
+
+	for name, tc := range tests {
+		if platform == platforms.MetalQEMUSNPGPU && !tc.gpu || platform != platforms.MetalQEMUSNPGPU && tc.gpu {
```
katexochen:
```suggestion
		if platforms.IsGPU(platform) != tc.gpu {
```

docs/docs/howto/secure-image-store.md
burgerdev, line 1: This needs to be linked into `docs/sidebars.js`.

internal/kuberesource/mutators.go
burgerdev: For future reference, it looks like k3s/containerd is messing with the images if it sees a known pause container:

```
agent_policy:1110: is_same_image: a = docker.io/rancher/mirrored-pause:3.6 b = registry.k8s.io/
pause:3.6@sha256:3d380ca8864549e74af4b29c10f9cb0956236dfb01c40ca076fb6c37253234db
```

charludo: Thanks, I'll add this to the comment.

PR #1690, opened 2025-08-12
treewide: disallow unknown fields for embedded

katexochen:
We should be aware if we embed structs with unsupported fields.


charludo (approved):
LGTM

PR #1692, opened 2025-08-13
node-installer: target configuration via configMap; remove K3s and RKE2 platforms

katexochen:
This adds a new way to configure the Contrast node installer via ConfigMap. The ConfigMap describes the target system the node installer is installing to and configures things like the containerd config path and the systemd unit name that needs to be restarted on changes to the containerd config. The node installer ships with default vaules, so the ConfigMap is only needed if things differ from those default paths.

One environment where we see these differences from common Kubernetes/containerd setups is K3s. Up to now, Contrast supported K3s as a separate platform. This change removes those platforms and introduces a node installer target ConfigMap for K3s instead.

**[breaking] The following platforms are removed:**

- `K3s-QEMU-SNP`
- `K3s-QEMU-SNP-GPU`
- `K3s-QEMU-TDX`
- `RKE2-QEMU-TDX`

If you previously used one of these platforms you can migrate by

1. Switching to a `Metal-*` platform:
    - `K3s-QEMU-SNP` -> `Metal-QEMU-SNP`
    - `K3s-QEMU-SNP-GPU` -> `Metal-QEMU-SNP-GPU`
    - `K3s-QEMU-TDX` -> `Metal-QEMU-TDX`
2. Installing the node installer config map for K3s, as described in https://docs.edgeless.systems/contrast/howto/cluster-setup/bare-metal#kubernetes-cluster-setup

Note that RKE2 isn't supported anymore as is wasn't properly tested.

### DIFF
--- a/docs/docs/howto/cluster-setup/bare-metal.md
+++ b/docs/docs/howto/cluster-setup/bare-metal.md
@@ -62,12 +62,25 @@ fs.inotify.max_user_instances = 8192
 
 Apply this change by running `systemctl restart systemd-sysctl` and verify it using `sysctl fs.inotify.max_user_instances`.
 
-## K3s setup
+## Kubernetes cluster setup
+
+Contrast can be deployed with different Kubernetes distributions.
+
+### K3s
+
+[K3s](https://k3s.io/) is a lightweight Kubernetes distribution that's easy to set up.
 
 1. Follow the [K3s setup instructions](https://docs.k3s.io/) to create a cluster.
    Contrast is currently tested with K3s version `v1.31.5+k3s1`.
-2. Install a block storage provider such as [Longhorn](https://longhorn.io/docs/latest/deploy/install/install-with-kubectl/) and mark it as the default storage class.
-3. Ensure that a load balancer controller is installed. For development and testing purposes, the built-in [ServiceLB](https://docs.k3s.io/networking/networking-services#service-load-balancer) should suffice.
+2. Install a block storage provider such as [Longhorn](https://longhorn.io/docs/1.9.1/deploy/install/install-with-helm/) and mark it as the default storage class.
+3. Ensure that a load balancer controller is installed.
+   For development and testing purposes, the built-in [ServiceLB](https://docs.k3s.io/networking/networking-services#service-load-balancer) should suffice.
+
+Then, install the ConfigMap to configure the Contrast node-installer for use with K3s:
+
+```sh
+kubectl apply -f https://github.com/edgelesssys/contrast/releases/latest/download/node-installer-target-config-k3s.yml
+```
 
 ## Preparing a cluster for GPU usage
 
--- a/docs/docs/howto/workload-deployment/runtime-deployment.md
+++ b/docs/docs/howto/workload-deployment/runtime-deployment.md
@@ -14,7 +14,7 @@ Required for all Contrast deployments.
 ## How-to
 
 Contrast depends on a [custom Kubernetes `RuntimeClass` (`contrast-cc`)](../../architecture/components/runtime.md), which needs to be installed in the cluster prior to the Coordinator or any confidential workloads.
-This consists of a `RuntimeClass` resource and a `DaemonSet` that performs installation on worker nodes.
+This consists of a `RuntimeClass` resource and the node installer `DaemonSet` that performs installation on worker nodes.
 This step is only required once for each version of the runtime.
 It can be shared between Contrast deployments.
 Also, different Contrast runtime versions can be installed in the same cluster.
--- a/docs/docs/howto/workload-deployment/runtime-deployment.md
+++ b/docs/docs/howto/workload-deployment/runtime-deployment.md
@@ -25,19 +25,19 @@ Also, different Contrast runtime versions can be installed in the same cluster.
 kubectl apply -f https://github.com/edgelesssys/contrast/releases/latest/download/runtime-aks-clh-snp.yml
 ```
 </TabItem>
-<TabItem value="k3s-qemu-snp" label="Bare metal (SEV-SNP)">
+<TabItem value="metal-qemu-snp" label="Bare metal (SEV-SNP)">
 ```sh
-kubectl apply -f https://github.com/edgelesssys/contrast/releases/latest/download/runtime-k3s-qemu-snp.yml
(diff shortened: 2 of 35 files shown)


internal/kuberesource/parts.go
```diff
@@ -204,6 +207,20 @@ func NodeInstaller(namespace string, platform platforms.Platform) (*NodeInstalle
...
+func NodeInstallerTargetConfig(target, namespace string) (*applycorev1.ConfigMapApplyConfiguration, error) {
+	switch target {
+	case "k3s":
+		return applycorev1.ConfigMap("contrast-node-installer-target-config", namespace).
+			WithData(map[string]string{
+				"containerd-config-path": "var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl",
+				"systemd-unit-name":      "k3s.service,k3s-agent.service",
+			}), nil
+	default:
+		return nil, fmt.Errorf("unsupported target %q", target)
```
charludo, line 220: IMO it would be nice to give indication that the target (platform) is either unsupported, or that it is not required to be specified. There's probably not a good way to decide which is the case here, but I could easily see someone passing "metal" or the like here and being confused.

internal/kuberesource/resourcegen/main.go
```diff
@@ -16,12 +16,13 @@ import (
...
 	imageReplacementsPath := flag.String("image-replacements", "", "Path to the image replacements file")
 	namespace := flag.String("namespace", "", "Namespace for namespaced resources")
+	rawPlatform := flag.String("platform", "", "Deployment platform to generate the runtime configuration for")
 	addLoadBalancers := flag.Bool("add-load-balancers", false, "Add load balancers to selected services")
 	addNamespaceObject := flag.Bool("add-namespace-object", false, "Add namespace object with the given namespace")
 	addPortForwarders := flag.Bool("add-port-forwarders", false, "Add port forwarder pods for all services")
 	addLogging := flag.Bool("add-logging", false, "Add logging configuration, based on CONTRAST_LOG_LEVEL and CONTRAST_LOG_SUBSYSTEMS environment variables")
-	rawPlatform := flag.String("platform", "", "Deployment platform to generate the runtime configuration for")
 	addDmesg := flag.Bool("add-dmesg", false, "Add dmesg container")
+	nodeInstallerTargetConfType := flag.String("node-installer-target-conf-type", "", "Type of node installer target configuration to generate (k3s,...)")
```
charludo, line 25: (alternatively: make this "metal" by default and handle appropriately in `NodeInstallerTargetConfig`? Not because it's required from a technical perspective, but for user's clarity)

docs/docs/howto/cluster-setup/bare-metal.md
```diff
@@ -62,12 +62,25 @@ fs.inotify.max_user_instances = 8192
...
 Apply this change by running `systemctl restart systemd-sysctl` and verify it using `sysctl fs.inotify.max_user_instances`.
 
-## K3s setup
+## Kubernetes cluster setup
+
+Contrast can be deployed with different Kubernetes distributions.
+
+### K3s
+
+[K3s](https://k3s.io/) is a lightweight Kubernetes distribution that's easy to set up.
```
charludo, line 71:
```suggestion
[K3s](https://k3s.io/) is a lightweight Kubernetes distribution that is easy to set up.
```
nit: avoid colloquial contractions in written documents like docs

docs/docs/howto/workload-deployment/runtime-deployment.md
```diff
@@ -47,7 +47,7 @@ kubectl apply -f https://github.com/edgelesssys/contrast/releases/latest/downloa
 The Contrast node installer will modify the containerd configuration on the worker nodes to add the runtime class.
 A backup will be created for the original configuration.
 
-Some Kubernetes platforms, for example K3s, use a template for the containerd configuration.
+Some Kubernetes platforms, for example metal, use a template for the containerd configuration.
```
charludo: This might be a search/replace mistake (?)

charludo (commented):
(Note: currently still reviewing the `targetconfig` Go changes)

nodeinstaller/internal/targetconfig/targetconfig.go
```diff
@@ -0,0 +1,129 @@
...
+			if dataStr == "" {
+				return fmt.Errorf("systemd-unit-name cannot be empty")
+			}
+			c.systemdUnitNames = strings.FieldsFunc(dataStr, func(r rune) bool {
+				for _, sep := range []string{",", " ", "\n"} {
+					if string(r) == sep {
+						return true
+					}
+				}
+				return false
```
charludo:
```suggestion
				return slices.Contains([]string{",", " ", "\n"}, string(r))
```

nodeinstaller/internal/targetconfig/targetconfig.go
```diff
@@ -0,0 +1,129 @@
...
+			})
+		case "kata-config-path":
+			c.kataConfigPath = dataStr
```
charludo, line 93: Is it ok if `c.kataConfigPath`is empty / `c.KataConfigPath()` is the `hostMount`?

nodeinstaller/node-installer.go
```diff
@@ -71,19 +68,58 @@ func run(ctx context.Context, fetcher assetFetcher, platform platforms.Platform,
...
+	log.Printf("Using config: %+v\n", config)
 
-	runtimeHandlerName, err := manifest.RuntimeHandler(platform)
+	runtimeHandler, err := manifest.RuntimeHandler(platform)
 	if err != nil {
 		return fmt.Errorf("getting runtime handler name: %w", err)
 	}
+	runtimeBase := filepath.Join("/opt", "edgeless", runtimeHandler)
+
+	// target config, which cna be overridden by the user via configMap.
```
charludo:
```suggestion
	// target config, which can be overridden by the user via configMap.
```

internal/kuberesource/parts.go
katexochen: I think that should be handled by the caller.

docs/docs/howto/cluster-setup/bare-metal.md
katexochen: This is required by the microsoft style guide we are following, and also enforced by ci.

charludo: Oh, wait, really? Interesting.

PR #1696, opened 2025-08-13
cli/verifier: add `NoSharedFSMount` verifier

miampf:
This adds a verifier to be run before `genpolicy` that forbids the following:
1. `volumeMount`s that don't reference a volume of type `configMap`, `downwardAPI`, `emptyDir`, `ephemeral`, `projected` or `secret`
2. `volumeClaimTemplate`s that are not of `volumeMode: Block`

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -111,22 +111,26 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	}
 
 	verifiers := verifier.AllVerifiers()
+	var findings error
 	for _, path := range paths {
 		fileContent, err := os.ReadFile(path)
 		if err != nil {
 			return fmt.Errorf("reading file %q to run generate on: %w", path, err)
 		}
-		resources, err := kubeapi.UnmarshalUnstructuredK8SResource(fileContent)
+		resources, err := kuberesource.UnmarshalApplyConfigurations(fileContent)
 		if err != nil {
 			return fmt.Errorf("parsing file %q to run generate on: %w", path, err)
 		}
 		for _, v := range verifiers {
 			for _, r := range resources {
 				if err := v.Verify(r); err != nil {
-					log.Warn("failed to verify YAML", "path", path, "kind", r.GetKind(), "name", r.GetName(), "error", err)
+					findings = errors.Join(findings, fmt.Errorf("failed to verify YAML %q: %w", path, err))
 				}
 			}
 		}
+	}
+	if findings != nil {
+		return findings
 	}
 
 	// generate a manifest by checking if a manifest exists and using that,
--- a/cli/verifier/no_shared_fs_mount.go
+++ b/cli/verifier/no_shared_fs_mount.go
@@ -1,0 +1,85 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package verifier
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+
+	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+)
+
+const errorMessageTemplate = "volumeMount %q refers to an unsupported %s; supported types are ConfigMap, DownwardAPI, EmptyDir, Ephemeral, Projected and Secret"
+
+// NoSharedFSMount verifies that no filesystems are shared with the root.
+type NoSharedFSMount struct{}
+
+// Verify verifies that no filesystems are shared with the root.
+func (v *NoSharedFSMount) Verify(toVerify any) error {
+	var findings error
+
+	// get all volume mounts that are referenced in containers
+	isNonCC := false
+	kuberesource.MapPodSpec(toVerify, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+		if spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
+			// this isn't a confidential pod so we don't need to check further
+			isNonCC = true
+			return spec
+		}
+		// verify all volume mounts of all containers in pod
+		for _, container := range spec.Containers {
+			for _, mount := range container.VolumeMounts {
+				volumeReferenced := false
+				for _, volume := range spec.Volumes {
+					if *mount.Name != *volume.Name {
+						continue
+					}
+					volumeReferenced = true
+
+					if isSupportedVolume(volume) {
+						continue
+					}
+
+					// a volume with invalid type is referenced
+					findings = errors.Join(findings, fmt.Errorf(errorMessageTemplate, *mount.Name, "volume type"))
+				}
+				if !volumeReferenced {
+					findings = errors.Join(findings, fmt.Errorf(errorMessageTemplate, *mount.Name, "volume claim template"))
+				}
+			}
+		}
+		return spec
+	})
+	if isNonCC {
+		// we don't care about non-confidential pods
+		return nil
+	}
+
+	return findings
+}
+
+func isSupportedVolume(volume applycorev1.VolumeApplyConfiguration) bool {
+	if volume.ConfigMap != nil {
+		return true
+	}
+	if volume.DownwardAPI != nil {
+		return true
+	}
+	if volume.EmptyDir != nil {
+		return true
+	}
+	if volume.Ephemeral != nil {
+		return true
+	}
+	if volume.Projected != nil {
+		return true
+	}
+	if volume.Secret != nil {
+		return true
+	}
+	return false
+}
--- a/cli/verifier/verifiers.go
+++ b/cli/verifier/verifiers.go
@@ -3,24 +3,22 @@
 
 package verifier
 
-import (
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-)
-
 // AllVerifiers returns all verifiers for k8s objects.
 func AllVerifiers() []Verifier {
-	return []Verifier{}
+	return []Verifier{
+		&NoSharedFSMount{},
+	}
 }
 
-// VerificationFunc is a function that verifies a given unstructured object and returns an error if verification fails.
-type VerificationFunc func(toVerify *unstructured.Unstructured) error
+// VerificationFunc is a function that verifies a given apply configuration and returns an error if verification fails.
+type VerificationFunc func(toVerify any) error
 
-// Verify verifies a given k8s object.
-func (f VerificationFunc) Verify(toVerify *unstructured.Unstructured) error {
+// Verify verifies a given k8s object. `toVerify` should be an apply configuration.
+func (f VerificationFunc) Verify(toVerify any) error {
 	return f(toVerify)
 }
 
-// Verifier verifies a given k8s object.
+// Verifier verifies a given k8s object. `toVerify` should be an apply configuration.
 type Verifier interface {
-	Verify(toVerify *unstructured.Unstructured) error
+	Verify(toVerify any) error
 }
--- a/cli/verifier/no_shared_fs_mount_test.go
+++ b/cli/verifier/no_shared_fs_mount_test.go
@@ -1,0 +1,191 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package verifier_test
+
+import (
+	"testing"
+
+	"github.com/edgelesssys/contrast/cli/verifier"
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/stretchr/testify/require"
+)
+
+const podWithEmptyVolume = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+spec:
+  runtimeClassName: contrast-cc
+  volumes:
+    - name: asdf
+      emptyDir: {}
+  containers:
+    - name: test
+      image: bash
+      volumeMounts:
+        - name: asdf
+          mountPath: /tmp
+`
+
+const podWithUnreferencedHostPath = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+spec:
+  runtimeClassName: contrast-cc
+  volumes:
+    - name: unreferenced
+      hostPath:
+        path: /
+  containers:
+    - name: test
+      image: bash
+`
+
+const podWithHostPath = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+spec:
+  runtimeClassName: contrast-cc
+  volumes:
+    - name: asdf
+      hostPath:
+        path: /
+  containers:
+    - name: test
+      image: bash
+      volumeMounts:
+        - name: asdf
+          mountPath: /tmp
+`
+
+const podWithVolumeMountNoVolume = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+spec:
+  runtimeClassName: contrast-cc
+  containers:
+    - name: test
+      image: bash
+      volumeMounts:
+        - name: asdf
+          mountPath: /tmp
+`
+
+const statefulSetWithPodReferencingClaim = `
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: test
+spec:
+  serviceName: "nginx"
+  template:
+    spec:
+      runtimeClassName: contrast-cc
+      containers:
+      - name: nginx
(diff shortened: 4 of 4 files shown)


cli/verifier/verifiers.go
```diff
@@ -4,23 +4,112 @@
...
+				continue
+			}
+			if volume.Secret != nil {
+				continue
+			}
+
+			// a volume with invalid type is referenced
+			findings = errors.Join(findings, fmt.Errorf("the volume %q has an unsupported type, which may cause unexpected behavior, please make sure your volumes are either a configMap, downwardAPI, emptyDir, ephemeral, projected or secret", *volume.Name))
+		}
+	}
```
charludo: I think the nested loop is not be doing anything here, except ignoring volumes with no corresponding `volumeMount` at all. 

Edit: from the `yaml with unreferenced problematic volume` test, that might be intentional? But if so, why?

cli/verifier/verifiers.go
```diff
@@ -4,23 +4,112 @@
...
+
+	// get all stateful sets from the resources to verify
+	unstructuredResources, err := kuberesource.ResourcesToUnstructured([]any{toVerify})
+	if err != nil {
+		return err
+	}
+	var volumeClaims []corev1.PersistentVolumeClaim
+	for _, r := range unstructuredResources {
+		switch r.GetKind() {
+		case "StatefulSet":
```
charludo: I think we also support stand-alone PVC definitions (i.e. not inside a `StatefulSet` at all), those should probably also be checked with the same logic

cli/verifier/verifiers_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"testing"
+
+	"github.com/edgelesssys/contrast/cli/verifier"
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/stretchr/testify/require"
+)
+
+func TestVerifyNoSharedFSMount(t *testing.T) {
```
charludo: PVC inside a `StatefulSet` and standalone-PVC, each of Block/non-Block volume modes are still missing I think

charludo (commented):
I know this is still a draft, I've just been dealing with PVCs for a different PR for the past couple days so this seemed like the perfect review target 😂😄 

One general thing, maybe the `NoSharedFSMount` verifier could already be moved into a separate file/dir, as I imagine the `verifiers.go` and `verifiers_test.go` files will continue growing in the future?

cli/verifier/verifiers.go
miampf: As far as I understood it from the ticket, volumes that aren't referenced in any way are not a problem and can be ignored which would make the verifier more sound since then, less false reports would be created. If I'm wrong with that, please correct me and I'll remove that exception :)

charludo: I see, makes sense. Though then we probably also need to check against `volumeDevices`?

miampf: @burgerdev since you wrote the ticket originally, what do you think? I unfortunately lack the knowledge of how volumes interact with contrast en detail, so having your opinion would be appreciated :D

charludo (approved):
Apart from the open conversation, LGTM!

cli/verifier/no_shared_fs_mount.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			if volume.Ephemeral != nil {
+				continue
+			}
+			if volume.Projected != nil {
+				continue
+			}
+			if volume.Secret != nil {
+				continue
+			}
```
burgerdev: Having this in a helper function `isSupportedVolumeType(volume)` would be clearer, and would probably make the entire Verify logic concise enough to put into one `MapPodSpec`.

cli/verifier/no_shared_fs_mount.go
```diff
@@ -0,0 +1,103 @@
...
+			volumeClaims = append(volumeClaims, statefulSet.Spec.VolumeClaimTemplates...)
+
+		case "PersistentVolumeClaim":
+			var pvc corev1.PersistentVolumeClaim
+			err := runtime.DefaultUnstructuredConverter.FromUnstructured(r.UnstructuredContent(), &pvc)
+			if err != nil {
+				return err
+			}
+			volumeClaims = append(volumeClaims, pvc)
```
burgerdev: I think this is too strict: 

* we do support non-confidential pods 
* the claim might not be used

Would be better to just not have the check.

cli/verifier/no_shared_fs_mount.go
```diff
@@ -0,0 +1,103 @@
...
+// NoSharedFSMount verifies that no filesystems are shared with the root.
+type NoSharedFSMount struct{}
+
+// Verify verifies that no filesystems are shared with the root.
+func (v *NoSharedFSMount) Verify(toVerify any) error {
+	var findings error
+
+	// get all volume mounts that are referenced in containers
+	var volumeMounts []applycorev1.VolumeMountApplyConfiguration
+	kuberesource.MapPodSpec(toVerify, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
```
burgerdev, line 27: We probably need a `runtimeClassName` check here so that we don't complain about non-CC pods.

cli/verifier/verifiers.go
```diff
@@ -3,24 +3,22 @@
...
-func (f VerificationFunc) Verify(toVerify *unstructured.Unstructured) error {
+func (f VerificationFunc) Verify(toVerify any) error {
 	return f(toVerify)
 }
 
 // Verifier verifies a given k8s object.
 type Verifier interface {
-	Verify(toVerify *unstructured.Unstructured) error
+	Verify(toVerify any) error
 }
```
burgerdev: Please document that `any` is expected to be some applyconfig.

cli/verifier/verifiers_test.go
```diff
@@ -0,0 +1,193 @@
...
+	testCases := map[string]struct {
+		k8sObjectYAML string
+		wantErr       bool
+	}{
+		"unproblematic yaml": {
+			k8sObjectYAML: `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test
```
burgerdev: I find this very hard to read. Could we maybe move the yaml strings to constants outside the test body?

cli/verifier/verifiers.go
burgerdev: > As far as I understood it from the ticket, volumes that aren't referenced in any way are not a problem and can be ignored

Kubernetes just ignores these and we won't see them, thus not our problem. Finding such issues would be the job of a k8s yaml linter or something like that.

> Though then we probably also need to check against `volumeDevices`?

If Kubernetes accepts a volume as `volumeDevice`, it should be supported by Kata. This verifier is only about filesystem mounts.

cli/verifier/no_shared_fs_mount.go
charludo: But then it's really simple to circumvent the verification: define a top-level PVC, and create add a `volumeDevice`to a container which uses this PVC.

burgerdev: > add a `volumeDevice` to a container which uses this PVC

As mentioned in another thread, `volumeDevice` is not problematic, but `volumeMount` is (block device vs. filesystem mount). 

An invariant enforced by Kubernetes is that a `volumeMount` has to refer to either a `volumes` entry in the PodSpec, or to a volume claim template in a StatefulSet spec. We checked the volume case above (`persistentVolumeClaim` is not a supported type), so we can just assume that a mount without a volume is a PVC template and reject it.

charludo: > An invariant enforced by Kubernetes is that a `volumeMount` has to refer to either a `volumes` entry in the PodSpec, or to a volume claim template in a StatefulSet spec.

TIL! Thanks for the explanation.

cli/verifier/no_shared_fs_mount.go
```diff
@@ -0,0 +1,105 @@
...
+type NoSharedFSMount struct{}
+
+// Verify verifies that no filesystems are shared with the root.
+func (v *NoSharedFSMount) Verify(toVerify any) error {
+	var findings error
+
+	// get all volume mounts that are referenced in containers
+	isNonCC := false
+	kuberesource.MapPodSpec(toVerify, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+		if spec.RuntimeClassName == nil || !strings.Contains(*spec.RuntimeClassName, "contrast-cc") {
```
burgerdev: Nit: `strings.HasPrefix` would be more precise.

cli/verifier/no_shared_fs_mount.go
```diff
@@ -0,0 +1,105 @@
...
+						continue
+					}
+
+					if isSupportedVolume(volume) {
+						continue
+					}
+
+					// a volume with invalid type is referenced
+					findings = errors.Join(findings, fmt.Errorf("the volume %q has an unsupported type, which may cause unexpected behavior, please make sure your volumes are either a configMap, downwardAPI, emptyDir, ephemeral, projected or secret", *volume.Name))
+				}
```
burgerdev: Need to return a finding if there's no matching volume (because it's most likely a PVC template, or a user error).

cli/verifier/no_shared_fs_mount.go
```diff
@@ -0,0 +1,105 @@
...
+			volumeClaims = append(volumeClaims, statefulSet.Spec.VolumeClaimTemplates...)
+		}
+	}
+
+	// verify all volume claims
+	for _, claim := range volumeClaims {
+		if *claim.Spec.VolumeMode != "Block" {
+			findings = errors.Join(findings, fmt.Errorf("volume claim %q does not have volumeMode=Block, which is unsupported", claim.Name))
+		}
+	}
```
burgerdev: I believe this is unnecessary if we reject volumeMounts without volumes as outlined above.

miampf: I don't see how, for example, a `volumeClaimTemplate` with `volumeMode: Filesystem` would be rejected if we remove this part of the code as long as it is still referenced. Are you sure this code can be removed?

cli/cmd/generate.go
```diff
@@ -111,23 +111,27 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
-		resources, err := kubeapi.UnmarshalUnstructuredK8SResource(fileContent)
+		resources, err := kuberesource.UnmarshalApplyConfigurations(fileContent)
 		if err != nil {
 			return fmt.Errorf("parsing file %q to run generate on: %w", path, err)
 		}
 		for _, v := range verifiers {
 			for _, r := range resources {
 				if err := v.Verify(r); err != nil {
-					log.Warn("failed to verify YAML", "path", path, "kind", r.GetKind(), "name", r.GetName(), "error", err)
+					findings = errors.Join(findings, fmt.Errorf("failed to verify YAML %q: %w", path, err))
```
burgerdev, line 127: Sorry, I missed this in the last round: why are we changing from warning to error?

cli/verifier/no_shared_fs_mount.go
burgerdev: Two cases:

1. There's a template with mode fs, but it is not referenced in any volumeMount: no problem.
2. It is referenced in a volume mount: we won't find the corresponding volume and report a finding.

Am I missing something? Could you just verify this with the unit test?

cli/cmd/generate.go
miampf: No need to apologize :) After a discussion with @katexochen, it seemed more reasonable to error on misconfigurations that could provoke undefined behavior. I could maybe extend the interface of the verifier so that we can distinguish between verifiers that don't need to error and those that should, however, I feel like this would be a separate PR.

burgerdev: Ack, sgtm.

miampf:
For some reason I can't quote reply or comment, this is in answer to https://github.com/edgelesssys/contrast/pull/1696#discussion_r2293801840

Currently, this code grabs all volume claims of `StatefulSet`s and checks if it has a volume mode of `Block`. If that isn't the case, it reports a finding. This is verified in the test cases `"stateful set with block volume mode"` and `"stateful set with problematic volume mode"`. Maybe I am not understanding your question correctly, but this code does not care if the claim is in any further way referenced or not since k8s may dynamically provision a volume if the volume for a claim doesn't exist:

> "When none of the static PVs the administrator created match a user's PersistentVolumeClaim, the cluster may try to dynamically provision a volume specially for the PVC."

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#dynamic

cli/verifier/no_shared_fs_mount.go
miampf: Currently, this code grabs all volume claims of `StatefulSet`s and checks if it has a volume mode of `Block`. If that isn't the case, it reports a finding. This is verified in the test cases `"stateful set with block volume mode"` and `"stateful set with problematic volume mode"`. Maybe I am not understanding your question correctly, but this code does not care if the claim is in any further way referenced or not since k8s may dynamically provision a volume if the volume for a claim doesn't exist:

> "When none of the static PVs the administrator created match a user's PersistentVolumeClaim, the cluster may try to dynamically provision a volume specially for the PVC."

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#dynamic

burgerdev: The issue was the missing metadata on the statefulset, which caused MapPodSpec to skip them. Removing the code as suggested, but adding the metadata, makes the test pass. https://github.com/edgelesssys/contrast/pull/1724.

cli/verifier/verifiers_test.go
burgerdev, line 1: This should probably be `no_shared_fs_mount_test.go`.

miampf:
Re-added some of the stateful set code since the test cases didn't pass because volumeClaims aren't defined in the pod spec. Verification itself is now contained to `MapPodSpec`

cli/verifier/no_shared_fs_mount.go
```diff
@@ -0,0 +1,116 @@
...
+				}
+				for _, volumeClaim := range volumeClaims {
+					if *mount.Name != volumeClaim.Name {
+						continue
+					}
+					volumeReferenced = true
+
+					if *volumeClaim.Spec.VolumeMode == "Block" {
+						continue
+					}
```
burgerdev: I don't understand why you added this again.

1. It's not necessary, as I tried to express in prior comments.
2. The tests were passing on my branch.
3. Referencing a `BLOCK` volume in a `volumeMount` is illegal anyway.

miampf: Then maybe I understood you wrong or not at all. I have removed the volume mode check now. However, I would keep the loop over volume claims with `volumeReferenced = true` since — if this doesn't exist — volume mounts that reference a claim template would error since these aren't created in the pod spec.

burgerdev: Me, on the other thread:

> Two cases:
> 
>     1. There's a template with mode fs, but it is not referenced in any volumeMount: no problem.
> 
>     2. It is referenced in a volume mount: we won't find the corresponding volume and report a finding.

What I meant to say is that 

> volume mounts that reference a claim template would error since these aren't created in the pod spec.

is the expected behaviour: `volumeMounts` that reference volume claim templates are _always_ wrong. Not to be confused with `volumeDevice` references to volume claim templates.

miampf: Oh okay then I was just really not understanding, I'm sorry. Fully removed this code now and assert an error in the unit test.

cli/verifier/no_shared_fs_mount.go
```diff
@@ -0,0 +1,83 @@
...
+
+					if isSupportedVolume(volume) {
+						continue
+					}
+
+					// a volume with invalid type is referenced
+					findings = errors.Join(findings, fmt.Errorf("the volume %q has an unsupported type, which may cause unexpected behavior, please make sure your volumes are either a configMap, downwardAPI, emptyDir, ephemeral, projected or secret", *volume.Name))
+				}
+				if !volumeReferenced {
+					findings = errors.Join(findings, fmt.Errorf("mount %q doesn't referenced any existing volume", *mount.Name))
```
burgerdev: These error messages should be more concise and aligned for consistency, like

```go
const template = "volumeMount %q refers to an unsupported %s; supported types are ..."

// !volumeReferenced
err := fmt.Errorf(template, *mount.Name, "volume claim template")
// else
err := fmt.Errorf(template, *mount.Name, "volume type")
```

burgerdev (approved):
Thanks for persevering - lgtm!

PR #1700, opened 2025-08-15
ci: update systems, use palutena in ci, use hetzner for dev

katexochen:


charludo:
> > LGTM. Only nit I have is that if `get-credentials` does already change one line in `justfile.env`, why not the second as well? I.e., why not do the same thing as in `get-credentials-dev` for all platforms in `get-credentials`?
> 
> Well, how do you choose the platform to get-credentials for then?

via the `platform` arg, I'd assume

IDK, as I said, it's a nit; just seemed a bit inconsistent, changing one value in the envfile and then having the script change the other one.

katexochen:
> changing one value in the envfile and then having the script change the other one.

Feel free to change both from hand, but I'm sure I'll forget. ;)

About invoking just commands with platform as arg, I don't like it and I'm thinking about removing it.

PR #1708, opened 2025-08-18
meshapi: add pod ip to mesh cert SANs

davidweisse:
This adds the pod IP as a SAN to the mesh certificate of the pod, in addition to the already existing DNS name specified in the manifest.

### DIFF
--- a/coordinator/internal/meshapi/meshapi.go
+++ b/coordinator/internal/meshapi/meshapi.go
@@ -10,6 +10,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"log/slog"
+	"net"
 
 	"github.com/edgelesssys/contrast/coordinator/internal/stateguard"
 	"github.com/edgelesssys/contrast/internal/manifest"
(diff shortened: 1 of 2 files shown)


coordinator/internal/meshapi/meshapi.go
```diff
@@ -74,6 +75,12 @@ func (i *Server) NewMeshCert(ctx context.Context, _ *meshapi.NewMeshCertRequest)
 	}
 	dnsNames := entry.SANs
 
+	tcpAddr, ok := p.Addr.(*net.TCPAddr)
+	if !ok {
+		return nil, fmt.Errorf("peer address is not a TCP address")
```
burgerdev: I think skipping the IP would be more prudent in case we don't understand it.

coordinator/internal/meshapi/meshapi.go
```diff
@@ -74,6 +75,12 @@ func (i *Server) NewMeshCert(ctx context.Context, _ *meshapi.NewMeshCertRequest)
 	}
 	dnsNames := entry.SANs
 
+	tcpAddr, ok := p.Addr.(*net.TCPAddr)
```
burgerdev: What if the transport was QUIC? A more robust approach may be

```go
ip, _, err := net.SplitHostPort(p.Addr.String())
```

burgerdev (approved):
Thanks, lgtm!

PR #1716, opened 2025-08-19
e2e/openssl: check for dmesg errors

davidweisse:
In the `openssl` e2e test, we check the `dmesg` logs of the Coordinator pod and fail on any unknown errors.

### DIFF
--- a/e2e/openssl/openssl_test.go
+++ b/e2e/openssl/openssl_test.go
@@ -15,6 +15,8 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"slices"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


e2e/openssl/openssl_test.go
```diff
@@ -282,6 +284,39 @@ func TestOpenSSL(t *testing.T) {
...
+		knownErrors := []string{
+			"Speculative Return Stack Overflow: WARNING: kernel not compiled with MITIGATION_SRSO.",
+			"[Firmware Bug]: Failed to parse event in TPM Final Events Log",
+			"ACPI BIOS Error (bug): Failure creating named object [\\_GPE._HID], AE_ALREADY_EXISTS (20240827/dswload2-326)",
+			"ACPI Error: AE_ALREADY_EXISTS, During name lookup/catalog (20240827/psobject-220)",
+		}
+		for _, line := range strings.Split(strings.TrimSpace(dmesgOutput), "\n") {
+			require.True(slices.ContainsFunc(knownErrors, func(knownError string) bool {
+				return strings.Contains(line, knownError)
+			}), "dmesg line does not contain known error: %s", line)
```
burgerdev, line 317: Could this be `require.Contains`?

davidweisse: Not really, the dmesg error contains a timestamp at the beginning, so we have to do another `strings.Contains()` for each known error.

burgerdev: Understood, thanks.

PR #1707, opened 2025-08-18
kuberesource: change annotation position to pod template

davidweisse:
Previously, we had two places where Contrast annotations could go:
- On the top level resource meta, e.g., service mesh and secure PV annotations.
- On the pod template meta, e.g., pod role, workload secret ID.

With this change, all Contrast annotations are specified in the pod template meta. All annotations are documented in a table in the docs.

This affects all workloads that define one of the following annotations:
- `contrast.edgeless.systems/skip-initializer`
- `contrast.edgeless.systems/servicemesh-ingress`
- `contrast.edgeless.systems/servicemesh-egress`
- `contrast.edgeless.systems/servicemesh-admin-interface-port`
- `contrast.edgeless.systems/secure-pv`

Running `contrast generate` on these workloads will ignore the annotations. To migrate, move the annotations from the top-level resource meta to the Pod Template Meta. The resulting YAML will be generated as before.

PR #1717, opened 2025-08-19
cli/verifier: check servicemesh-egress annotation isn't empty

miampf:
This PR adds a verifier that checks that the `contrast.edgeless.systems/servicemesh-egress` annotation is not empty.

Previously, setting the annotation to the empty string would disable egress but trigger ingress to be activated, which is likely not the intended behavior. The correct way to prevent egress rules setup is to not add the egress annotation.

### DIFF
--- a/cli/verifier/servicemesh_egress.go
+++ b/cli/verifier/servicemesh_egress.go
@@ -1,0 +1,37 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package verifier
+
+import (
+	"errors"
+
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	applymetav1 "k8s.io/client-go/applyconfigurations/meta/v1"
+)
+
+// ServiceMeshEgressNotEmpty verifies that the `contrast.edgeless.systems/servicemesh-egress` annotation
+// isn't empty if it exists.
+type ServiceMeshEgressNotEmpty struct{}
+
+// Verify verifies that the `contrast.edgeless.systems/servicemesh-egress` annotation
+// isn't empty if it exists.
+func (v *ServiceMeshEgressNotEmpty) Verify(toVerify any) error {
+	var findings error
+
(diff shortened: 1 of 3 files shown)


cli/verifier/servicemesh_egress.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	for _, r := range resources {
+		findings = errors.Join(findings, checkUnstructured(r.UnstructuredContent()))
+	}
+
+	return findings
+}
+
+// checkUnstructured recursively checks if the "contrast.edgeless.systems/servicemesh-egress" annotation has content.
+func checkUnstructured(r map[string]any) error {
```
burgerdev: The name should be more descriptive because we're in the shared `verifiers` package.

cli/verifier/servicemesh_egress.go
```diff
@@ -0,0 +1,61 @@
...
+	var findings error
+
+	for k, v := range r {
```
burgerdev: Instead of recursing into the unstructured yaml, you could `MapPodSpecWithMeta` and just work with meta, ignoring the spec. Maybe needs rebasing on #1707.

miampf: Hmm I'm not too sure if this works. Having annotations on a deployment (see `deploymentEmptyEgressAnnotation` in `servicemesh_egress_test.go`) and not in a pod spec causes those annotations to be ignored. Do we only care about pod annotations? If yes, this isn't an issue and `MapPodSpecWithMeta` works.

burgerdev: #1707 moves all annotations to the podspec. Does that answer your question?

miampf: Oh okay yeah it does. Uses `MapPodSpecMeta` now

cli/verifier/verifiers.go
```diff
@@ -7,6 +7,7 @@ package verifier
 func AllVerifiers() []Verifier {
 	return []Verifier{
 		&NoSharedFSMount{},
+		&ServiceMeshEgressNotEmpty{},
```
burgerdev, line 10: This needs a rebase, right? I think this should go into the pre-runs.

burgerdev (approved):

cli/verifier/verifiers.go
miampf: Yeah, I think so too. There was a failing CI run on https://github.com/edgelesssys/contrast/pull/1733 which is why I waited, I'll merge and rebase this PR now tho :)

katexochen:
As this change is user facing, it should be mentioned in the changelog. Please provide a PR title and description meaningful to a user. Also check if the docs need to be updated.

PR #1727, opened 2025-08-25
cli/verifier: ensure that image references are pinned

charludo:
In order to verify that the images specified in deployments actually match the images pulled from the container registry, images need to be pinned by a digest.

This was already being verified during the image pull process; however, `contrast generate` did allow unpinned images. Starting with this PR, `contrast generate` will fail if the digest is missing. This does not change the ultimate behavior of a deployment, but does provide an earlier error and thus faster feedback.

Please ensure that your image refs are of the format

```
image:0.0.1@sha256:0000000000000000000000000000000000000000000000000000000000000000
```
or
```
image@sha256:0000000000000000000000000000000000000000000000000000000000000000
```

### DIFF
--- a/docs/docs/architecture/components/service-mesh.md
+++ b/docs/docs/architecture/components/service-mesh.md
@@ -83,7 +83,7 @@ spec:
       runtimeClassName: contrast-cc
       containers:
         - name: web-svc
-          image: ghcr.io/edgelesssys/frontend:v1.2.3@...
+          image: ghcr.io/edgelesssys/frontend:v1.2.3@sha256:...
           ports:
             - containerPort: 8080
               name: web
@@ -100,7 +100,7 @@ spec:
     runtimeClassName: contrast-cc
     containers:
       - name: web-svc
-        image: ghcr.io/edgelesssys/frontend:v1.2.3@...
+        image: ghcr.io/edgelesssys/frontend:v1.2.3@sha256:...
         ports:
           - containerPort: 8080
             name: web
@@ -167,7 +167,7 @@ spec:
       runtimeClassName: contrast-cc
       containers:
         - name: currency-conversion
-          image: ghcr.io/edgelesssys/conversion:v1.2.3@...
+          image: ghcr.io/edgelesssys/conversion:v1.2.3@sha256:...
 ```
 
 ## Public key Infrastructure
--- a/docs/docs/howto/workload-deployment/deployment-verification.md
+++ b/docs/docs/howto/workload-deployment/deployment-verification.md
@@ -37,7 +37,7 @@ metadata:
 spec:
   containers:
     - name: port-forwarder
-      image: alpine/socat
+      image: "alpine/socat@sha256:..."
       args:
         - -d
         - TCP-LISTEN:1313,fork
--- a/docs/docs/howto/workload-deployment/recover-coordinator.md
+++ b/docs/docs/howto/workload-deployment/recover-coordinator.md
@@ -36,7 +36,7 @@ metadata:
 spec:
   containers:
     - name: port-forwarder
-      image: alpine/socat
+      image: "alpine/socat@sha256:..."
       args:
         - -d
         - TCP-LISTEN:1313,fork
--- a/docs/docs/howto/workload-deployment/set-manifest.md
+++ b/docs/docs/howto/workload-deployment/set-manifest.md
@@ -42,7 +42,7 @@ metadata:
 spec:
   containers:
     - name: port-forwarder
-      image: alpine/socat
+      image: "alpine/socat@sha256:..."
       args:
         - -d
         - TCP-LISTEN:1313,fork


cli/verifier/verifiers.go
```diff
@@ -12,6 +12,7 @@ func AllVerifiersBeforeGenerate() []Verifier {
 func AllVerifiersAfterGenerate() []Verifier {
 	return []Verifier{
 		&NoSharedFSMount{},
+		&ImageRefValid{},
```
katexochen: Shouldn't this be a "before"-verifier? Or is your intention to check whether the images we are injecting are pinned, too?

charludo: > Or is your intention to check whether the images we are injecting are pinned, too?

Yes, essentially. This verifier should work just fine both before and after. Hypothetically it could be a useful sanity check for us to have it run after, but other than that, I don't see any benefits or disadvantages to before/after.

katexochen: disadvantage after: we are pulling images that aren't pinned and that we don't want to use. If a user has many big images, this might be a difference of getting feedback after a sec or after 20 min. We should fail fast where ever possible.

charludo: Had not considered that. But, are images actually being pulled in `generate`? At least for the verifier here, only the manifests are pulled

katexochen: Afaik genpolicy still pulls them.

charludo: Ah, alright. Moved it :)

cli/verifier/image_ref_valid.go
```diff
@@ -0,0 +1,61 @@
...
+
+	var images []string
+	for _, r := range resources {
+		findImageFields(r.Object, &images)
+	}
+
+	for _, img := range images {
+		_, err := name.NewDigest(img)
+		if err != nil {
+			findings = errors.Join(findings, fmt.Errorf("the image reference '%s' is invalid: image references must contain a digest", img))
```
katexochen: I think _invalid_ isn't the right term here. An image name is valid without digest, it is just that we are requiring it.

cli/verifier/image_ref_valid.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+
+	var images []string
+	for _, r := range resources {
+		findImageFields(r.Object, &images)
+	}
+
+	for _, img := range images {
+		_, err := name.NewDigest(img)
```
katexochen: There are other error cases of NewDigest than just the digest missing. If a user runs into them, the error we return doesn't make sense.

cli/verifier/image_ref_valid.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
```
katexochen:
```suggestion
		if _, err := name.NewDigest(img); err != nil {
```

cli/verifier/image_ref_valid.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// findImageFields recursively searches for "image" fields in the unstructured data.
+func findImageFields(data map[string]any, images *[]string) {
+	for key, value := range data {
+		if key == "image" {
+			if img, ok := value.(string); ok {
+				*images = append(*images, img)
+			}
+		} else {
```
katexochen: use continue instead of else to reduce nesting

cli/verifier/image_ref_valid_test.go
```diff
@@ -0,0 +1,143 @@
...
+			wantErr:       true,
+		},
+		"digest missing algorithm, with tag": {
+			k8sObjectYAML: podAlgorithmMissing,
+			wantErr:       true,
+		},
+		"image ref valid": {
+			k8sObjectYAML: podImageRefValid,
+		},
+	}
```
katexochen: use a multi line const ( ) and think about moving this into the test.

katexochen: add tests for other resources than pod

cli/verifier/image_ref_valid_test.go
```diff
@@ -0,0 +1,143 @@
...
+			k8sObjectYAML: podDigestMalformed,
+			wantErr:       true,
+		},
+		"digest missing algorithm, with tag": {
+			k8sObjectYAML: podAlgorithmMissing,
+			wantErr:       true,
+		},
+		"image ref valid": {
+			k8sObjectYAML: podImageRefValid,
+		},
```
katexochen: add a second positive test case without a tag.

katexochen (changes requested):

katexochen:
As this change is user facing, it should be mentioned in the changelog. Please provide a PR title and description meaningful to a user. Also check if the docs need to be updated.

cli/verifier/image_ref_valid.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	var images []string
+	for _, r := range resources {
+		findImageFields(r.Object, &images)
+	}
+
+	for _, img := range images {
+		if _, err := name.NewDigest(img); err != nil {
+			findings = errors.Join(findings, fmt.Errorf("the image reference '%s' could not be verified: %w", img, err))
```
katexochen: now it is unclear what the issue is in case only the digest is missing, isn't it? These checks should also output something that is actionable by a user.

docs/docs/howto/logging.md
```diff
@@ -44,7 +44,7 @@ following variables to your container definition.
 ```yaml
 spec: # v1.PodSpec
   containers:
-    image: "ghcr.io/edgelesssys/contrast/coordinator:latest"
+    image: "ghcr.io/edgelesssys/contrast/coordinator:latest@sha256:dbcc56f4d19882b939f461a71a005cb442e963dc31d59f2e1195d54366575e20"
```
katexochen: This change is wrong. Images here mustn't be pinned, they are replaced on release!

Same for every image with :latest

PR #1731, opened 2025-08-26
cryptsetup: detached header verification, refactor

katexochen:
With this change, we start verifying the cryptsetup header and use it in detached mode afterwards. 

This is a breaking change for cryptsetup volumes that were set up with earlier versions of Contrast or manually with different cryptsetup parameters.

### DIFF
--- a/initializer/main.go
+++ b/initializer/main.go
@@ -93,6 +93,8 @@ func run(cmd *cobra.Command, _ []string) (retErr error) {
 	}
 
 	ctx := cmd.Context()
+	ctx, cancel := signal.NotifyContext(ctx, syscall.SIGTERM, syscall.SIGINT)
+	defer cancel()
 
 	privKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 	if err != nil {
@@ -184,27 +186,35 @@ func run(cmd *cobra.Command, _ []string) (retErr error) {
 	}
 
 	cryptsetupDevicePath := os.Getenv("CRYPTSETUP_DEVICE")
-	if cryptsetupDevicePath == "" {
-		log.Info("Initializer done")
-		return nil
+	if cryptsetupDevicePath != "" {
+		log.Info("Setting up encrypted mount")
+		flags := &cryptsetupFlags{
+			devicePath:       cryptsetupDevicePath,
+			volumeMountPoint: "/state",
+		}
+		if err := setupEncryptedMount(ctx, log, flags); err != nil {
+			return fmt.Errorf("setting up encrypted mount: %w", err)
+		}
+		log.Info("Encrypted mount setup done")
 	}
 
-	log.Info("Setting up encrypted mount")
-
-	// Create tmp dir for cryptsetup lock files.
-	if err := os.MkdirAll("/run/cryptsetup", 0o755); err != nil {
-		return fmt.Errorf("creating cryptestup lock directory: %w", err)
+	if err := os.WriteFile("/done", []byte(""), 0o644); err != nil {
+		return fmt.Errorf("creating startup probe done directory:%w", err)
 	}
+	log.Info("Initializer done")
 
-	flags := &cryptsetupFlags{
-		devicePath:       cryptsetupDevicePath,
-		volumeMountPoint: "/state",
+	if cryptsetupDevicePath != "" {
+		// The device mount is created by the initializer and shared with the application
+		// container through a common emptyDir. We want to avoid mounting on a sub-path of
+		// the emptyDir to make configuration for the application container easier, so we
+		// mount the encrypted volume on top of the emptyDir mount point. The Kata agent
+		// does not know about this and will unmount that mount point when the initializer
+		// exits, which affects the application container, too. We avoid this situation by
+		// keeping the initializer running indefinitely.
+		log.Info("Idling to keep propagated cryptsetup device mounts alive")
+		<-ctx.Done()
 	}
-
-	ctx, cancel := signal.NotifyContext(ctx, syscall.SIGTERM, syscall.SIGINT)
-	defer cancel()
-
-	return setupEncryptedMount(ctx, log, flags)
+	return nil
 }
 
 func printPreface(cmd *cobra.Command, _ []string) {
--- a/initializer/mount.go
+++ b/initializer/mount.go
@@ -7,13 +7,14 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
-	"os/exec"
 	"os/signal"
 	"syscall"
 
+	"github.com/edgelesssys/contrast/internal/cryptsetup"
 	"github.com/edgelesssys/contrast/internal/logger"
 	"github.com/edgelesssys/contrast/internal/mount"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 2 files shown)


charludo (commented):
I will have another look in the morning, but so far everything seems good. Two "features" I think might be nice-to-have:

- for one-off mounts, it would be nice to not have to specify a `headerPath`. A dummy one works as well, of course, but is maybe not as clean.
- closing the device: would it maybe make sense to call `luksClose` on error?

internal/cryptsetup/cryptsetup.go
```diff
@@ -0,0 +1,278 @@
...
+	if digest.Digest == "" {
+		return fmt.Errorf("expected digest to be non-empty")
+	}
+	if len(header.Tokens) != 0 {
+		return fmt.Errorf("expected no tokens, got %d", len(header.Tokens))
+	}
+	return nil
+}
+
+type cryptsetupMetadata struct {
```
burgerdev, line 258: Is there some source for this structure that we could cite here, or do we only know it empirically?

internal/cryptsetup/cryptsetup.go
```diff
@@ -0,0 +1,278 @@
...
+		Segments   []string `json:"segments"`
+		Hash       string   `json:"hash"`
+		Iterations int      `json:"iterations"`
+		Salt       string   `json:"salt"`
+		Digest     string   `json:"digest"`
+	} `json:"digests"`
+	Config struct {
+		JSONSize     string `json:"json_size"`
+		KeyslotsSize string `json:"keyslots_size"`
+	}
```
burgerdev, line 304: `Config` does not appear to be checked, I suppose because it's dynamic of sorts?

internal/cryptsetup/cryptsetup.go
```diff
@@ -0,0 +1,278 @@
...
+	if errors.As(err, &exitErr) {
+		return "", fmt.Errorf("executing '%s': %w, stderr: %s", cmd.String(), err, exitErr.Stderr)
+	} else if err != nil {
+		return "", fmt.Errorf("executing '%s': %w", cmd.String(), err)
+	}
+	return string(output), nil
+}
+
+func ensureBase(p string) error {
+	if err := os.MkdirAll(filepath.Dir(p), 0o700); err != nil {
```
burgerdev: I think inlining this would be more readable.

internal/cryptsetup/cryptsetup.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	args := []string{
+		"luksFormat",
+		"--type=luks2",                          // Use LUKS2 header format.
+		"--cipher=aes-xts-plain64",              // Use AES-XTS cipher.
+		"--pbkdf=argon2id",                      // Use Argon2id as the key derivation function.
+		"--pbkdf-memory=10240",                  // Memory usage for Argon2i, limit to 10 MiB so it won't fail in low-memory pods.
+		"--batch-mode",                          // Suppresses all confirmation questions.
+		fmt.Sprintf("--key-file=%s", d.keyPath), // Path to the key file.
+		d.devicePath,
```
burgerdev, line 69: Since the parameters are fixed, we should be able to tell whether this is compatible or not with prior cryptsetup volumes (referring to the "might be breaking" in the PR description).

internal/cryptsetup/cryptsetup_test.go
```diff
@@ -0,0 +1,79 @@
...
+      "digest":"gzHnbLzZpMe0tMMeP52UgBIjpQDgyP6J6y4St6sX+0M="
+    }
+  },
+  "config":{
+    "json_size":"12288",
+    "keyslots_size":"16744448"
+  }
+}`
+
+func TestVerifyHeader(t *testing.T) {
```
burgerdev, line 73: We don't need to test every single header field, but a handful of negative tests would be reassuring.

initializer/main.go
```diff
@@ -184,27 +186,32 @@ func run(cmd *cobra.Command, _ []string) (retErr error) {
...
 
-	flags := &cryptsetupFlags{
-		devicePath:       cryptsetupDevicePath,
-		volumeMountPoint: "/state",
+	if cryptsetupDevicePath != "" {
+		// The device mount is created by the initializer and consumed by an application container.
+		// This is only possible through mount propagation, from the initializer mount namespace to
+		// the other mount namespaces of the pod. If the initializer exists, its mount namespace is
+		// removed, which also removes the propagated mount. So we must idle instead of exiting.
+		log.Info("Ideling to keep propagated cryptsetup device mounts alive")
```
burgerdev: Idling.

initializer/main.go
```diff
@@ -184,27 +186,32 @@ func run(cmd *cobra.Command, _ []string) (retErr error) {
...
+	log.Info("Initializer done")
 
-	flags := &cryptsetupFlags{
-		devicePath:       cryptsetupDevicePath,
-		volumeMountPoint: "/state",
+	if cryptsetupDevicePath != "" {
+		// The device mount is created by the initializer and consumed by an application container.
+		// This is only possible through mount propagation, from the initializer mount namespace to
+		// the other mount namespaces of the pod. If the initializer exists, its mount namespace is
+		// removed, which also removes the propagated mount. So we must idle instead of exiting.
```
burgerdev: This comment slightly misses the underlying issue, which afaiu is that we're mounting on the tmpfs mountpoint itself. If you change the setup such that `/mnt` holds the tmpfs and `/mnt/state` the LUKS device mount, the initializer does not need to keep running. To be honest, I was surprised that the current approach works at all.

burgerdev (approved):
Very nice package, thanks a lot!

internal/cryptsetup/cryptsetup.go
katexochen: It will break with the follow up PR (integrity).

internal/cryptsetup/cryptsetup_test.go
katexochen: Will do in the follow up.

internal/cryptsetup/cryptsetup.go
katexochen: I think it is. Not sure why you should verify JSONSize (might make sense if you are writing C?).
In general, the current code doesn't verify size/offset values beside `key_size`.

burgerdev:
> > for one-off mounts, it would be nice to not have to specify a headerPath. A dummy one works as well, of course, but is maybe not as clean.
> 
> Thinking about it, this shouldn't be exposed at all.

There are requirements for the detached header placement (i.e., in a secure, writable location). Only the caller knows what these are, although granted we control who's calling it so a sane default would work, too.

katexochen:
> There are requirements for the detached header placement (i.e., in a secure, writable location). Only the caller knows what these are, although granted we control who's calling it so a sane default would work, too.

I moved the location to /dev/shm to guarantee the header is placed in memory. That location should also be writable.

initializer/main.go
burgerdev: Closer to the truth:

> The device mount is created by the initializer and shared with the application container through a common emptyDir. We want to avoid mounting on a sub-path of the emptyDir to make configuration for the application container easier, so we mount the encrypted volume on top of the emptyDir mount point. The Kata agent does not know about this and will unmount that mount point when the initializer exits, which affects the application container, too. We avoid this situation by keeping the initializer running indefinitely.

internal/cryptsetup/cryptsetup.go
```diff
@@ -0,0 +1,287 @@
...
+	headerPath string
+	keyPath    string
+}
+
+// NewDevice creates a new Device instance. It doesn't interact with the device itself.
+func NewDevice(devicePath, keyPath string) (*Device, error) {
+	var randSuffix [4]byte
+	_, err := rand.Read(randSuffix[:])
+	if err != nil {
+		return nil, fmt.Errorf("generating random suffix for key file: %w", err)
```
charludo:
```suggestion
		return nil, fmt.Errorf("generating random suffix for header file: %w", err)
```

burgerdev (approved):

PR #1734, opened 2025-08-27
internal/cryptsetup: use integrity

katexochen:
This PR adds integrity protection to Contrast encrypted volumes.


initializer/mount.go
```diff
@@ -120,13 +129,163 @@ func setupEncryptedMount(ctx context.Context, log *slog.Logger, flags *cryptsetu
...
 
+// isExt4 checks for presence of ext4 magic bytes on a device.
+//
+// Note: We previously used 'blkid' to check the filesystem type, but it tries
+// to read blocks that aren't initialized after dm-verity is set up. We implement
+// this simple manual check, knowing it will only read the expected block which
+// is manually initialized by wipeExt4Blocks before it is used by mkfs.ext4.
+func isExt4(devPath string) (bool, error) {
+	const (
+		magic  = "53ef" // Magic number for ext4 filesystems (little-endian).
```
burgerdev: Maybe it's just me, but I'd find it more natural to deal with bytes here instead of hex.

```go
	const expectMagic = 0xEF53
	// ...
	_, err := f.Seek(offset, 0)
	if err != nil {
		return false, err
	}
	var magic uint16
	err := binary.Read(f, binary.LittleEndian, &magic)
	if err != nil {
		return false, err
	}
	return magic == expectMagic, nil
```

burgerdev (approved):
This is some great hacking, thanks for figuring all this out!

internal/cryptsetup/cryptsetup.go
```diff
@@ -64,6 +64,9 @@ func (d *Device) Format(ctx context.Context) error {
 		"--cipher=aes-xts-plain64",              // Use AES-XTS cipher.
 		"--pbkdf=argon2id",                      // Use Argon2id as the key derivation function.
 		"--pbkdf-memory=10240",                  // Memory usage for Argon2i, limit to 10 MiB so it won't fail in low-memory pods.
+		"--integrity=hmac-sha256",               // Use HMAC-SHA256 for integrity protection via dm-verity.
```
charludo: I'm a bit confused, isn't this using dm-integrity, not dm-verity? AFAIK dm-verity is only for read-only partitions, right?

charludo (approved):
I can only agree with @burgerdev, thank you for figuring this out! 😅

PR #1739, opened 2025-08-28
dev-docs: add guidance for working with GHSAs

burgerdev:


dev-docs/ghsa.md
```diff
@@ -0,0 +1,58 @@
...
+3. Create a patch release for the latest released version.
+4. From the GHSA's drop-down menu, select _Publish Advisory_ and click the button.
+   We don't usually publish CVEs through GitHub.
+
+## FAQ
+
+### What if the vulnerability was already disclosed?
+
+This can happen if the vulnerability is evident in a bug report, for example, or through discussion in open forums.
+It usually still makes sense to use a private fork to avoid further leaks, unless it was the fix itself that accidentally disclosed the vulnerability.
```
katexochen: I think that's a highly controversial opinion..

katexochen (approved):

dev-docs/ghsa.md
burgerdev: Would it help if I reduce the qualifer (usually -> sometimes)? I'm thinking of situations where an attack is discussed on the PR, but it takes a long time to get to a resolution. What would you recommend here?

katexochen: Just saying, I don't care what we write here exactly. I know from the distro/packaging perspective that people don't like half-leaked vulns.

PR #1740, opened 2025-08-28
internal/cryptsetup: provide ext4 setup

katexochen:
Moving the ext4 code a little so it can also be used in #1685. The setup is coupled to the integrity setup anyway and easy to get wrong, so integrating that part into the lib.

### DIFF
--- a/e2e/openssl/openssl_test.go
+++ b/e2e/openssl/openssl_test.go
@@ -311,6 +311,13 @@ func TestOpenSSL(t *testing.T) {
 			"ACPI BIOS Error (bug): Failure creating named object [\\_GPE._HID], AE_ALREADY_EXISTS (20240827/dswload2-326)",
 			"ACPI Error: AE_ALREADY_EXISTS, During name lookup/catalog (20240827/psobject-220)",
 			"NVRM: No NVIDIA GPU found", // openssl test does not use a GPU
+			// This is logged when pages are read from dm-integrity protected, uninitialized/unwiped pages.
+			// This happens in the IsExt4 check of our cryptsetup lib and in other places.
+			"INTEGRITY AEAD ERROR",
+			"Buffer I/O error on dev dm-",
+			// This happens when we mount ext4, just the kernel checking for the right filesystem.
+			"couldn't mount as ext3 due to feature incompatibilities",
+			"couldn't mount as ext2 due to feature incompatibilities",
 		}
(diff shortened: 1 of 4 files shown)


internal/cryptsetup/ext4.go
```diff
@@ -0,0 +1,140 @@
...
+	"strings"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
+// IsExt4 checks for presence of ext4 magic bytes on a device to determine if it is formatted as ext4.
+//
+// Note: We previously used 'blkid' to check the filesystem type, but it tries
+// to read blocks that aren't initialized after dm-verity is set up. We implement
```
charludo:
```suggestion
// to read blocks that aren't initialized after dm-integrity is set up. We implement
```

charludo (approved):
Refactor LGTM. Can confirm works perfectly now with the imagestore.

e2e/openssl/openssl_test.go
```diff
@@ -311,6 +311,13 @@ func TestOpenSSL(t *testing.T) {
 			"ACPI BIOS Error (bug): Failure creating named object [\\_GPE._HID], AE_ALREADY_EXISTS (20240827/dswload2-326)",
 			"ACPI Error: AE_ALREADY_EXISTS, During name lookup/catalog (20240827/psobject-220)",
 			"NVRM: No NVIDIA GPU found", // openssl test does not use a GPU
+			// This is logged when pages are read from dm-integrity protected, uninitialized/unwiped pages.
+			// This happens in the IsExt4 check of our cryptsetup lib and in other places.
+			"trusted_key: device-mapper: crypt: dm-1: INTEGRITY AEAD ERROR",
+			"Buffer I/O error on dev dm-2",
+			// This happens when we mount ext4, just the kernel checking for the right filesystem.
+			"EXT4-fs (dm-2): couldn't mount as ext3 due to feature incompatibilities",
+			"EXT4-fs (dm-2): couldn't mount as ext2 due to feature incompatibilities",
```
burgerdev: The int suffix is probably dynamic, depending on the number of dm volumes, right?


```suggestion
			"Buffer I/O error on dev dm-",
			// This happens when we mount ext4, just the kernel checking for the right filesystem.
			"couldn't mount as ext3 due to feature incompatibilities",
			"couldn't mount as ext2 due to feature incompatibilities",
```

PR #1745, opened 2025-08-29
kata.kata-runtime: pass imagepuller error to kata

charludo:
Previously, when an error occurred during image pulling or unpacking, the relevant error message was not forwarded to the kata-agent and not readable from the pod's logs. Instead, the next step in container initialization failed with an unhelpful error message.

This PR fixes this, and failures in image pulling/unpacking should now be logged using more helpful and failure-specific messages.


burgerdev (approved):
I think this should be labeled as bug fix and get a brief description for users.

PR #1755, opened 2025-09-03
cli/verify: unzip policy files

katexochen:
Policies are compressed in annotations and grpc transport since Contrast v1.12.0. In the verify command, we write the policies to the filesystem so the verifying party can inspect them. With the introduction of the compression, we missed decrompressing before writing the policies to the filesystem.

This change adds decrompression before the files are written, so they are in a human redable form again on disk and ready to inspect.

### DIFF
--- a/cli/cmd/verify.go
+++ b/cli/cmd/verify.go
@@ -4,8 +4,11 @@
 package cmd
 
 import (
+	"bytes"
+	"compress/gzip"
 	"crypto/sha256"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


cli/cmd/verify.go
```diff
@@ -82,7 +85,16 @@ func runVerify(cmd *cobra.Command, _ []string) error {
...
+		gzipReader, err := gzip.NewReader(bytes.NewReader(p))
+		if err != nil {
+			return fmt.Errorf("creating gzip reader: %w", err)
+		}
+		policyDecompressed, err := io.ReadAll(gzipReader)
+		if err != nil {
+			return fmt.Errorf("decompressing policy: %w", err)
+		}
+		gzipReader.Close()
+		filelist[fmt.Sprintf("policy.%s.rego", pHash)] = policyDecompressed
```
burgerdev, line 97: I wonder whether we should write the gzip policy, too. If we decompress in the CLI, we break the verification chain from manifest to policy when only looking at the filesystem. Not sure how bad that is, given that we need to trust the binary anyway to do attestation correctly.

katexochen: You mean users can't hash the policy themselves after this change? I think that is acceptable. The verifier should audit the policy, not redo the work the cli already did.

PR #1758, opened 2025-09-04
docs: refactor encrypted storage tutorial into how-to

charludo:
Dedups the tutorial and turns it into a How-To.

### DIFF
--- a/docs/docs/howto/encrypted-storage.md
+++ b/docs/docs/howto/encrypted-storage.md
@@ -1,6 +1,6 @@
 # Set up encrypted persistent storage
 
-This section provides a tutorial on how to configure encrypted persistent storage for your application.
+This section guides you through the process of configuring encrypted persistent storage for your application.
 
 ## Applicability
 
@@ -9,16 +9,89 @@ This step is recommended for any Contrast deployment that stores sensitive data 
 ## Prerequisites
 
 1. [Set up your cluster](./cluster-setup/bare-metal.md)
+2. [Install CLI](./install-cli.md)
+3. [Deploy the Contrast runtime](./workload-deployment/runtime-deployment.md)
 
 ## How-to
 
-<!-- TODO: I think this has to be changed to really make clear what steps are needed for encrypted persistant storage. Otherwise it is again more or less a tutorial for deploying an already configured Contrast-protected deplyoment. -->
+The following demonstrates how to set up an encrypted LUKS mount for an arbitrary `/mount/path` directory to easily deploy an application with encrypted persistent storage using Contrast.
 
-**This tutorial guides you through deploying a simple application with an
-encrypted MySQL database using the Contrast [workload
-secret](../architecture/secrets.md#workload-secrets).**
+### Configuration
 
-[MySQL](https://mysql.com) is an open-source database used to organize data into
+Add a `volume` of type `emptyDir` to your pod.
+Mount it through a `volumeMount` with `HostToContainer` propagation.
+The specified `mountPath` is where the Contrast Initializer will mount the encrypted volume.
+Use a `volumeClaimTemplate` to make a persistent storage device available to your deployment.
+The Initializer creates an encrypted volume and a filesystem on this device.
+Supported volume types can be found in the [k8s documentation](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support).
+
+Finally, add the [Contrast annotation](../architecture/k8s-yaml-elements.md) `contrast.edgeless.systems/secure-pv` with value `device-name:mount-name`.
+The presence of this annotation instructs an init container running `cryptsetup`
+to use the workload secret at `/contrast/secrets/workload-secret-seed` to generate
+a key and setup the block device as a LUKS partition. Before starting the application container,
+the init container uses the generated key to open the LUKS device, which is then mounted
+by the application container. For the application container, this process is completely
+transparent and works like mounting any other volume. The `cryptsetup` container
+will remain running to provide the necessary decryption context for the workload
+container.
+
+```yaml
+spec: # v1.StatefulSetSpec
+  template:
+    metadata:
+      annotations:
+        contrast.edgeless.systems/secure-pv: device-name:mount-name
+    spec:
+      containers:
+      - volumeMounts:
+        - mountPath: /mount/path
+          mountPropagation: HostToContainer
+          name: mount-name
+      volumes:
+      - emptyDir: {}
+        name: mount-name
+  volumeClaimTemplates:
+  - apiVersion: v1
+    kind: PersistentVolumeClaim
+    metadata:
+      name: device-name
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: 1Gi
+      volumeMode: Block
+```
+
+:::note[Persistent workload secrets]
+
+Secure persistent volumes make use of workload secrets to be able to re-open devices after container restarts.
+Please read about the implications in the [workload secrets](../architecture/secrets.md#workload-secrets) section.
+
+:::
+
+### Deployment
+
+Follow the steps of the generic workload deployment instructions:
+
+- [Add the Coordinator.](./workload-deployment/add-coordinator.md)
+- [Generate the policies.](./workload-deployment/generate-annotations.md)
+- [Apply the resources.](./workload-deployment/deploy-application.md)
+- [Set the manifest.](./workload-deployment/set-manifest.md)
+
+## Example application: MySQL
+
+The above guide can be tried out with a [MySQL](https://mysql.com) sample deployment which is part of the contrast release.
+You can add the deployment file to your resources by running:
+
+```sh
+curl -fLO https://github.com/edgelesssys/contrast/releases/latest/download/mysql-demo.yml --output-dir resources
+```
+
+:::note[MySQL]
+
+MySQL is an open-source database used to organize data into
 tables and quickly retrieve information about its content. All of the data in a
 MySQL database is stored in the `/var/lib/mysql` directory. In this example, we
 use the workload secret to setup an encrypted LUKS mount for the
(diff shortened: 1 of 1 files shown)


docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,117 @@ This step is recommended for any Contrast deployment that stores sensitive data
 ## Prerequisites
 
 1. [Set up your cluster](./cluster-setup/bare-metal.md)
+2. [Install CLI](./install-cli.md)
+3. [Deploy the Contrast runtime](./workload-deployment/runtime-deployment.md)
+4. [Add Coordinator to resources](./workload-deployment/set-manifest.md)
+5. [Prepare deployment files](./workload-deployment/deployment-file-preparation.md)
+6. [Generate annotations and manifest](./workload-deployment/generate-annotations.md)
+7. [Deploy application](./workload-deployment/deploy-application.md)
```
burgerdev: We need to do these during the tutorial, don't we? The prerequisite section should probably be similar to https://github.com/edgelesssys/contrast/blob/main/docs/docs/howto/vault.md#prerequisites.

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,117 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
-  FMC:         None
-```
-
-Use the values from `Platform TCB` to fill in the `MinimumTCB` values in the generated `manifest.json` file.
+The following demonstrates how to set up an encrypted LUKS mount for an arbitrary `/mount/path` directory to easily deploy an application with encrypted persistent storage using Contrast.
+
+### Configuration
+
+Add a `volume` to your pod or stateful set (called `mount-name` in the example configuration below)
+which references your encrypted, persistent storage device either through a [persistent volume claim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/), or an [iSCSI device](https://kubernetes.io/docs/concepts/storage/volumes/#iscsi).
```
burgerdev: We should link to [what volume types support block devices](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support) instead of listing them here individually.

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,117 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
-  Boot Loader: 9
-  FMC:         None
-Platform TCB: TCB Version:
-  Microcode:   72
-  SNP:         23
-  TEE:         0
-  Boot Loader: 9
-  FMC:         None
+contrast generate resources/
+kubectl apply -f resources/
```
burgerdev: This is not enough - the manifest has changed, so we need to `set` again.

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,117 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
 ```
 
-Use the values from `Platform TCB` to fill in the `MinimumTCB` values in the generated `manifest.json` file.
+## Verifying the deployment
 
-:::note[Attention!]
+Users of an app may want to verify its security and identity before sharing data by writing it to the persistent volume.
```
burgerdev: This sentence reads as if the _user_ would be writing something to the volume. Since we do not have the MySQL context we had in the prior version of this section, the scenario may need a bit more details on who verifies what.

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,117 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
+working with persistent storage, regular backups are recommended in order to
+prevent data loss. For confidential applications, please also refer to the
+[security considerations](./hardening.md). Also be
+aware of the differences in security implications of the workload secrets for
+the data owner and the workload owner. For more details, see the [Workload
+Secrets](../architecture/secrets.md#workload-secrets) documentation.
 
-</TabItem>
-</Tabs>
+After downloading the MySQL demo file, follow [the steps given above](#applying-the-configuration) to reapply your configuration.
```
burgerdev: Instead of the callout below, we should rather have a section explaining the demo, like in https://github.com/edgelesssys/contrast/blob/main/docs/docs/howto/vault.md#download-the-deployment-files.

docs/docs/howto/encrypted-storage.md
charludo: The main idea behind the ticket for this is to transform the tutorial into a how-to, i.e. the goal was not to show how to go from zero to encrypted storage, but rather to inform experienced users on how to add encrypted storage to their existing setup.

AFAIK, @katexochen has strong opinions on this, so I'd defer to him on this :)

docs/docs/howto/encrypted-storage.md
charludo: Hm, right, Thinking about it, the verifying section has nothing to do with the encrypted storage, so I'd probably just remove the entire section here.

docs/docs/howto/encrypted-storage.md
burgerdev: That does not change the fact that 4-7 are not prereqs for this, does it? 

I agree that we can assume reader familiarity and hand-wave these parts - as we do in the Vault docs ;)

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,152 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
+
+### Configuration
+
+Add a `volume` to your pod or stateful set (called `mount-name` in the example configuration below)
+which references your encrypted, persistent storage device either through one of the [supported volume types](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support).
+To your application container, add a `volumeMount` with the `mount-name` used for the `volume`.
+This `volumeMount` **must** be of type `emptyDir`.
+
+Additionally, make your persistent volume claim or iSCSI device available to the deployment; the example below uses a PVC `1Gi` in size.
+Give the device a name (the example uses `device-name`).
```
burgerdev: It would make more sense if the second half of this sentence explained that the `volumeClaimTemplate` accomplishes the "making available", and I think we can just drop the `1Gi` and the `device-name` part because they are kinda self-evident.

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,152 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
-  FMC:         None
-```
-
-Use the values from `Platform TCB` to fill in the `MinimumTCB` values in the generated `manifest.json` file.
+The following demonstrates how to set up an encrypted LUKS mount for an arbitrary `/mount/path` directory to easily deploy an application with encrypted persistent storage using Contrast.
+
+### Configuration
+
+Add a `volume` to your pod or stateful set (called `mount-name` in the example configuration below)
+which references your encrypted, persistent storage device either through one of the [supported volume types](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support).
```
burgerdev: I just realized that the purpose of this volume is not clear enough and we're mixing the emptyDir and the block device. We should rearrange the docs like this:

> Add a volume of type emptyDir to your pod. Mount it with HostToCOntainer propagation. This is where the initializer will mount the encrypted volume.
>
> Add a block device volume to the pod. The initializer creates an encrypted volume and a filesystem on this device.

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,152 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
+To your application container, add a `volumeMount` with the `mount-name` used for the `volume`.
+This `volumeMount` **must** be of type `emptyDir`.
+
+Additionally, make your persistent volume claim or iSCSI device available to the deployment; the example below uses a PVC `1Gi` in size.
+Give the device a name (the example uses `device-name`).
+
+Finally, add the [Contrast annotation](../architecture/k8s-yaml-elements.md) `contrast.edgeless.systems/secure-pv` with value `device-name:mount-name`.
```
burgerdev, line 28: We should move the annotation explanation from below the yaml to here.

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,152 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
+container.
 
-:::note[Attention!]
+:::note[Persistent workload secrets]
 
-This must be done on a trusted machine, with a secure and trusted connection to it.
+During the initialization process of the workload pod, the Contrast Initializer
+sends an attestation report to the Coordinator and receives a [workload secret](../architecture/secrets.md#workload-secrets)
+derived from the Coordinator's secret seed and the workload secret ID specified in the
+manifest, and writes it to a secure in-memory `volumeMount`.
```
burgerdev: I think this could just be replaced with a note:

> Secure persistent volumes use workload secrets, see implications at secrets.md

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,152 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
-  FMC:         None
-Platform TCB: TCB Version:
-  Microcode:   72
-  SNP:         23
-  TEE:         0
-  Boot Loader: 9
-  FMC:         None
+contrast generate resources/
+kubectl apply -f resources/
+contrast set resources/
```
burgerdev: This is not executable (missing coordinator address). Instead of writing the commands down here, should we just point to the other how-tos?

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,152 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
+[security considerations](./hardening.md). Also be
+aware of the differences in security implications of the workload secrets for
+the data owner and the workload owner. For more details, see the [Workload
+Secrets](../architecture/secrets.md#workload-secrets) documentation.
 
-</TabItem>
-</Tabs>
+The MySQL deployment is defined as a `StatefulSet`, with a storage setup analogous to what has been described above.
+
+```yaml
```
burgerdev: I don't think we should replicate parts of the published yaml here.

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,152 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
+working with persistent storage, regular backups are recommended in order to
+prevent data loss. For confidential applications, please also refer to the
+[security considerations](./hardening.md). Also be
+aware of the differences in security implications of the workload secrets for
+the data owner and the workload owner. For more details, see the [Workload
+Secrets](../architecture/secrets.md#workload-secrets) documentation.
 
-</TabItem>
-</Tabs>
+The MySQL deployment is defined as a `StatefulSet`, with a storage setup analogous to what has been described above.
```
burgerdev: The MySQL client should be mentioned, too: it is referred to in the callout below.

burgerdev:
```suggestion
The MySQL server is defined as a `StatefulSet`, with a storage setup analogous to what has been described above.
```

docs/docs/howto/encrypted-storage.md
charludo: Ahhh, sorry, I misunderstood what you meant [here](https://github.com/edgelesssys/contrast/pull/1758#discussion_r2348525402).

docs/docs/howto/encrypted-storage.md
```diff
@@ -9,177 +9,112 @@ This step is recommended for any Contrast deployment that stores sensitive data
...
+### Deployment
 
-```sh
-contrast generate --reference-values metal-qemu-snp-gpu resources/
-```
+Follow the steps of the generic workload deployment instructions:
 
-On bare-metal SEV-SNP, `contrast generate` is unable to fill in the `MinimumTCB` values as they can vary between platforms and CPU models.
-They will have to be filled in manually.
+- [Add the Coordinator.](./workload-deployment/set-manifest.md)
```
burgerdev: That's the wrong link.

PR #1768, opened 2025-09-05
packages/contrast-releases: simplify

katexochen:


charludo (approved):
Very nice, this looks great. I noticed that the result is not completely identical to before (e.g. for `contrast-releases.latest`, an additional `coordinator.yml` is placed in the top-level dir) but assume an additional file is not an issue.

charludo (approved):

PR #1771, opened 2025-09-08
cli/initdata: init initdata lib

charludo:
TBH I find it a bit difficult to gauge what we do (not) need without being able to use the lib anywehere.

### DIFF
--- a/internal/initdata/initdata_test.go
+++ b/internal/initdata/initdata_test.go
@@ -1,0 +1,187 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package initdata
+
+import (
+	"encoding/base64"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func buildTemplate(algorithm, version string, data bool) string {
+	var builder strings.Builder
+
+	if algorithm != "" {
+		fmt.Fprintf(&builder, "algorithm = \"%s\"\n", algorithm)
+	}
+	if version != "" {
+		fmt.Fprintf(&builder, "version = \"%s\"\n", version)
+	}
+
+	if data {
+		builder.WriteString(`
+[data]
+somekey = "somevalue"
+"imagepuller.toml" = '''
+socket = 'unix:///run/guest-services/imagepuller.sock'
+credentials = []
+'''
+`)
+	}
+	return builder.String()
+}
+
+func tomlToAnnotation(t *testing.T, toml string) string {
+	t.Helper()
+	compressed, err := compress([]byte(toml))
+	require.NoError(t, err)
+	encoded := base64.StdEncoding.EncodeToString(compressed)
+	return encoded
+}
+
+func annotationToToml(t *testing.T, annotation string) string {
+	t.Helper()
+	decoded, err := base64.StdEncoding.DecodeString(annotation)
+	require.NoError(t, err)
+	decompressed, err := decompress(decoded)
+	require.NoError(t, err)
+	return string(decompressed)
+}
+
+func TestDecodeKataAnnotation(t *testing.T) {
+	tests := map[string]struct {
+		version   string
+		algorithm string
+		data      bool
+		wantErr   error
+	}{
+		"success sha256":       {version: InitdataVersion, algorithm: "sha256", data: true},
+		"success sha-256":      {version: InitdataVersion, algorithm: "sha256", data: true},
+		"success sha384":       {version: InitdataVersion, algorithm: "sha384", data: true},
+		"success sha-384":      {version: InitdataVersion, algorithm: "sha384", data: true},
+		"success sha512":       {version: InitdataVersion, algorithm: "sha512", data: true},
+		"success sha-512":      {version: InitdataVersion, algorithm: "sha512", data: true},
+		"version missing":      {version: "", wantErr: errVersionMismatch},
+		"version unknown":      {version: "1.1.0", wantErr: errVersionMismatch},
+		"algorithm missing":    {version: InitdataVersion, wantErr: errAlgorithmUnknown},
+		"algorithm unknown":    {version: InitdataVersion, algorithm: "sha123", wantErr: errAlgorithmUnknown},
+		"success data missing": {version: InitdataVersion, algorithm: "sha256", data: false},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			require := require.New(t)
+
+			annotation := tomlToAnnotation(t, buildTemplate(tc.algorithm, tc.version, tc.data))
+			i, err := DecodeKataAnnotation(annotation)
+			if tc.wantErr != nil {
+				require.ErrorIs(err, tc.wantErr)
+				return
+			}
+
+			require.NoError(err)
+			assert.Equal(tc.algorithm, i.Algorithm)
+			assert.Equal(InitdataVersion, i.Version)
+
+			if tc.data {
+				assert.Contains(i.Data, "somekey")
+				assert.Contains(i.Data, "imagepuller.toml")
+			}
+		})
+	}
+}
+
+func TestEncodeKataAnnotation(t *testing.T) {
+	tests := map[string]struct {
+		version   string
+		algorithm string
+		data      map[string]string
+		wantErr   error
+	}{
+		"success sha256": {
+			version:   InitdataVersion,
+			algorithm: "sha256",
+			data:      map[string]string{"key": "value"},
+		},
+		"success sha384": {
+			version:   InitdataVersion,
+			algorithm: "sha384",
+			data:      map[string]string{"key": "value"},
+		},
+		"success sha512": {
+			version:   InitdataVersion,
+			algorithm: "sha512",
+			data:      map[string]string{"key": "value"},
+		},
+		"success complex keys": {
+			version:   InitdataVersion,
+			algorithm: "sha256",
+			data:      map[string]string{"key": "value", "imagepuller.toml": "something"},
+		},
+		"version missing": {
+			wantErr: errVersionMismatch,
+		},
+		"version unsupported": {
+			version: "1.1.0",
+			wantErr: errVersionMismatch,
+		},
+		"algorithm not sanitized": {
+			version:   InitdataVersion,
+			algorithm: "sha-256",
+			wantErr:   errAlgorithmUnknown,
+		},
+		"algorithm missing": {
+			version: InitdataVersion,
+			wantErr: errAlgorithmUnknown,
+		},
+		"algorithm unknown": {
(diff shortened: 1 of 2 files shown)


cli/initdata/initdata.go
burgerdev, line 1: I think this should go into package `internal/initdata` instead.

cli/initdata/initdata.go
```diff
@@ -0,0 +1,224 @@
...
+			i.algorithm,
+		)
+	}
+
+	return digest, nil
+}
+
+// Get returns the value associated with the given key, or an error.
+// A "" zero value response with nil error indicates that the field had explicitly been set to "".
+func (i *Initdata) Get(key string) (string, error) {
```
burgerdev: Why not return `(string, bool)`?

cli/initdata/initdata.go
```diff
@@ -0,0 +1,224 @@
...
+	return toml, nil
+}
+
+// Unmarshal attempts to unmarshal TOML into an Initdata struct.
+func (i *Initdata) Unmarshal(raw []byte) error {
+	err := toml.Unmarshal(raw, &i)
+	return err
+}
+
+// Decompress decompresses gzipped bytes.
```
burgerdev: The functions below don't need to be methods. I think it would be better to have a `(En|De)codeAnnotation` function that does base64 and gzip.

cli/initdata/initdata.go
```diff
@@ -0,0 +1,224 @@
...
+		return []byte{}, err
+	}
+	return toml, nil
+}
+
+// Unmarshal attempts to unmarshal TOML into an Initdata struct.
+func (i *Initdata) Unmarshal(raw []byte) error {
+	err := toml.Unmarshal(raw, &i)
+	return err
+}
```
burgerdev: Not sure this is worth having over just `toml.Marshal(i)`.

cli/initdata/initdata.go
```diff
@@ -0,0 +1,224 @@
...
+// Delete removes the given key and its value from the initdata.
+// If the key had not previously been set, the "" zero value is returned.
+func (i *Initdata) Delete(key string) string {
+	old := i.data[key]
+	delete(i.data, key)
+	return old
+}
+
+// FromString parses Initdata from a base64-encoded, gzipped TOML string.
+func (i *Initdata) FromString(raw string) error {
```
burgerdev: This overlaps with `New`. If the goal is to only have a safe API, we should not have `FromString`, otherwise parsing and validating should be separate. I'm leaning towards the former.

cli/initdata/initdata.go
```diff
@@ -0,0 +1,224 @@
...
+	switch i.algorithm {
+	case "sha256":
+		bytes := sha256.Sum256(digestible)
+		digest = string(bytes[:])
+	case "sha384":
+		bytes := sha512.Sum384(digestible)
+		digest = string(bytes[:])
+	case "sha512":
+		bytes := sha512.Sum512(digestible)
+		digest = string(bytes[:])
```
burgerdev: Sigh - looks like the specification is inconsistent.

1. Spec says names follow https://www.iana.org/assignments/hash-function-text-names/hash-function-text-names.xhtml.
2. Examples below use `sha256` (instead of `sha-256`). 
3. Agent uses the example version: https://github.com/kata-containers/kata-containers/blob/fa6e4981a1e9b9e349fae349e6a9b382168742fd/src/agent/src/initdata.rs#L133-L135.

cli/initdata/initdata.go
```diff
@@ -0,0 +1,224 @@
...
+	return old
+}
+
+// Delete removes the given key and its value from the initdata.
+// If the key had not previously been set, the "" zero value is returned.
+func (i *Initdata) Delete(key string) string {
+	old := i.data[key]
+	delete(i.data, key)
+	return old
+}
```
burgerdev: Unsure whether we need the mutators. If you intend to call `Set` or `Delete`, you are going to want a new digest, too, and don't need the safety that the API currently provides. What do you think about changing the API like this?

```go
func New(algorithm string, data map[string]string) (*Initdata, error)
func ParseAndValidate(b []byte) (*Initdata, error)
func (i *Initdata) EncodeAndDigest() ([]byte, string, error)
func (i *Initdata) Data() (map[string]string, error)

func DecodeKataAnnotation(anno string) ([]byte, error)
func EncodeKataAnnotation(tomlBytes []byte) (string, error)
```

cli/initdata/initdata.go
```diff
@@ -0,0 +1,224 @@
...
+
+	encoded, err := i.Encode(compressed)
+	if err != nil {
+		return "", "", err
+	}
+
+	digest, err := i.Digest(encoded)
+	if err != nil {
+		return "", "", err
+	}
```
burgerdev: Digest is calculated on the raw TOML, not on the annotation.

cli/initdata/initdata.go
```diff
@@ -0,0 +1,224 @@
...
+func (i *Initdata) Validate() error {
+	if i.version != InitdataVersion {
+		return fmt.Errorf(
+			"specified initdata version is %s, but supported is only version %s",
+			i.version,
+			InitdataVersion,
+		)
+	}
+
+	if _, err := i.Digest([]byte{}); err != nil {
```
burgerdev: That seems incomplete.

cli/initdata/initdata.go
burgerdev: I think I misunderstood what `Validate` is for. Imho `Parse` should do basic validation (like "known algo" and "expected version"), and return the parsed `*Initdata` and the digest byte slice.

cli/initdata/initdata.go
charludo: I'll just add both

cli/initdata/initdata.go
charludo: In my mind, the order of operations would have been `FromString`->`Validate` -> `[Set|Deleta]*` -> `Encode, Digest`. 

API suggestion LGTM though, will implement; LMK if the set /delete methods should stay or not.

cli/initdata/initdata.go
charludo: Not really, in order to not have to repeat the algorithm switch, this runs `Digest`, and if `Digest` complains about an unsupported algorithm, then an error is returned from Validate. I've added a comment.

cli/initdata/initdata.go
charludo: I've changed the API, see the second commit. I'm not sure if `ParseAndValidate` and `EncodeAndDigest` are really required. What do you think about instead changing 
```go
func DecodeKataAnnotation(anno string) ([]byte, error)
func EncodeKataAnnotation(tomlBytes []byte) (string, error)
```
to
```go
func DecodeKataAnnotation(anno string) (*Initdata, error)
func (i *Initdata) EncodeKataAnnotation() (string, string, error)
```
?

I have a hard tome coming up with scenarios where those two would currently be useful without immediately after/prior calling `ParseAndValidate` / `EncodeAndDigest`.

cli/initdata/initdata.go
charludo: Will do once the comments below are all resolved.

cli/initdata/initdata.go
```diff
@@ -0,0 +1,212 @@
...
+	// Fails with an err if the Initdata was created with an unsupported algorithm.
+	if _, err := i.Digest([]byte{}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Digest calculates the hash of the Initdata.
+func (i *Initdata) Digest(digestible []byte) (string, error) {
```
burgerdev: Do you think this needs to be exported? The interface is a bit odd, so I'd rather keep it internal.

cli/initdata/initdata.go
```diff
@@ -0,0 +1,212 @@
...
+		digest = string(bytes[:])
+	case "sha384", "sha-384":
+		bytes := sha512.Sum384(digestible)
+		digest = string(bytes[:])
+	case "sha512", "sha-512":
+		bytes := sha512.Sum512(digestible)
+		digest = string(bytes[:])
+	default:
+		return "", fmt.Errorf(
+			"algorithm '%s' is not supported. Supported are only sha256, sha384, sha512",
```
burgerdev:
```suggestion
			"algorithm %q is not supported. Supported are only sha256, sha384, sha512",
```

cli/initdata/initdata.go
burgerdev: sgtm. Now that we have the API, we should also add unit tests.

internal/initdata/initdata_test.go
```diff
@@ -0,0 +1,185 @@
...
+	compressed, _ := compress([]byte(toml))
+	encoded := base64.StdEncoding.EncodeToString(compressed)
+	return encoded
+}
+
+func annotationToToml(annotation string) string {
+	decoded, _ := base64.StdEncoding.DecodeString(annotation)
+	decompressed, _ := decompress(decoded)
+	return string(decompressed)
+}
```
burgerdev: In order to make this a bit safer but still convenient, you could 

```go
func annotationToToml(t *testing.T, annotation string) {
  t.Helper()
  decoded, err := base64.StdEncoding.DecodeString(annotation)
  require.NoError(t, err)
  // ...
}
```

internal/initdata/initdata_test.go
```diff
@@ -0,0 +1,185 @@
...
+		"success sha256":       {version: InitdataVersion, algorithm: "sha256", data: true},
+		"success sha-256":      {version: InitdataVersion, algorithm: "sha256", data: true},
+		"success sha384":       {version: InitdataVersion, algorithm: "sha384", data: true},
+		"success sha-384":      {version: InitdataVersion, algorithm: "sha384", data: true},
+		"success sha512":       {version: InitdataVersion, algorithm: "sha512", data: true},
+		"success sha-512":      {version: InitdataVersion, algorithm: "sha512", data: true},
+		"version missing":      {version: "", wantErr: errVersionMismatch},
+		"version unknown":      {version: "1.1.0", wantErr: errVersionMismatch},
+		"algorithm missing":    {version: InitdataVersion, wantErr: errAlgorithmUnknown},
+		"algorithm unknown":    {version: InitdataVersion, wantErr: errAlgorithmUnknown},
```
burgerdev: Did you mean to set an `algorithm` here?

internal/initdata/initdata_test.go
```diff
@@ -0,0 +1,185 @@
...
+		algorithm string
+		data      map[string]string
+		want      string
+		wantErr   error
+	}{
+		"success sha256": {
+			version:   InitdataVersion,
+			algorithm: "sha256",
+			data:      map[string]string{"key": "value"},
+			want:      "a1c4b2afd1fa1c06bab23f0e7d8b6bbc07e25a3d0240e79f0048da0775bae6a5",
```
burgerdev: This is now testing the stability of toml encoding. Maybe a roundtrip test `Decode(Encode(foo)) == foo` would be more robust.

burgerdev (approved):
Thx

PR #1775, opened 2025-09-09
flake: move packages into overlay, scopes, refactor

katexochen:


katexochen:
testing the release on this state: https://github.com/edgelesssys/contrast/actions/runs/17582948020

packages/by-name/contrast/package.nix
```diff
@@ -9,6 +9,7 @@
   contrast,
   installShellFiles,
   calculateSnpIDBlock,
+  contrastPkgs,
```
charludo:
```suggestion
  contrastPkgsStatic,
```

packages/by-name/contrast/package.nix
```diff
@@ -202,7 +203,7 @@ buildGoModule (finalAttrs: {
   subPackages = packageOutputs ++ [ "internal/kuberesource/resourcegen" ];
 
   prePatch = ''
-    install -D ${lib.getExe kata.genpolicy} cli/genpolicy/assets/genpolicy-kata
+    install -D ${lib.getExe contrastPkgs.contrastPkgsStatic.kata.genpolicy} cli/genpolicy/assets/genpolicy-kata
```
charludo:
```suggestion
    install -D ${lib.getExe contrastPkgsStatic.kata.genpolicy} cli/genpolicy/assets/genpolicy-kata
```

packages/scripts.nix
```diff
@@ -74,7 +74,7 @@
     text = ''
       exitcode=0
 
-      tags="${lib.concatStringsSep "," pkgs.contrast.tags}"
+      tags="${lib.concatStringsSep "," pkgs.contrastPkgs.contrast.tags}"
```
charludo:
```suggestion
      tags="${lib.concatStringsSep "," contrastPkgs.contrast.tags}"
```

(same for the other instances of `pkgs.contrastPkgs`)

flake.nix
```diff
@@ -37,23 +37,27 @@
...
 
         checks.formatting = treefmtEval.config.build.check self;
 
-        legacyPackages = pkgs // ourPkgs;
+        legacyPackages = pkgs.contrastPkgs // {
+          nixpkgs = builtins.removeAttrs pkgs [
+            "fenix"
+            "contrastPkgs"
+          ];
+        };
```
charludo, line 60: Is there something preventing

```suggestion
        legacyPackages =
          pkgs.contrastPkgs
          // (builtins.removeAttrs pkgs [
            "fenix"
            "contrastPkgs"
          ]);
```

from working? Tried it just now, and as far as I can see, both `contrastPkgs` and `pkgs` are available prefix-free

packages/default.nix
```diff
@@ -1,32 +1,30 @@
 # Copyright 2024 Edgeless Systems GmbH
 # SPDX-License-Identifier: BUSL-1.1
 
-{ lib, pkgs }:
+{
+  pkgs,
+}:
```
charludo:
```suggestion
{ pkgs }:
```
nit

flake.nix
katexochen: in case of a name collision, packages in nixpkgs will silently overwrite our packages.

charludo: Then how about overwriting nixpkgs with ours instead?

katexochen: I think I'd like to land this as is, we can see if it is too annoying this way and improve later.

charludo: Alright!

PR #1795, opened 2025-09-23
e2e: add memdump test

davidweisse:
This adds the `memdump` E2E test, which verifies the runtime encryption of Contrast deployments. The test setup is as follows:
- `listener` deployment, which listens for a TCP connection via `socat`.
- `sender` deployment, which can send messages via `socat`.
- `memdump` deployment, which is non-cc and run with `hostPID: true`, which can be used to generate a core dump.

The `listener` and `sender` pod are set up with Contrast Service Mesh, so the `sender` can send an encrypted canary string via `socat` to the `listener`, which stores the string in memory. The `memdump` pod then creates a core dump of the qemu process of the `listener` pod and verifies that it does not contain the canary string in plaintext.

E2E run: https://github.com/edgelesssys/contrast/actions/runs/18101832623

### DIFF
--- a/.github/workflows/e2e_nightly.yml
+++ b/.github/workflows/e2e_nightly.yml
@@ -33,6 +33,7 @@ jobs:
           # keep-sorted start
           - atls
           - imagestore
+          - memdump
           - openssl
           - peerrecovery
           - policy
--- a/e2e/memdump/memdump_test.go
+++ b/e2e/memdump/memdump_test.go
@@ -1,0 +1,130 @@
+// Copyright 2024 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build e2e
+
+package memdump
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/edgelesssys/contrast/e2e/internal/contrasttest"
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/platforms"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	listenerDeployment = "listener"
+	senderDeployment   = "sender"
+	memdumpDeployment  = "memdump"
+
+	canaryString = "deadbeafcafebabe0123456789abcdef"
+	nsenterCmd   = `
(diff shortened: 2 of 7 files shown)


.github/workflows/e2e_nightly.yml
burgerdev, line 1: Don't forget to drop this commit.

e2e/memdump/memdump_test.go
```diff
@@ -0,0 +1,132 @@
...
+	require.True(t, t.Run("set", ct.Set), "contrast set needs to succeed for subsequent tests")
+
+	require.True(t, t.Run("contrast verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
+
+	memdumpTester := kuberesource.MemDumpTester()
+	memdumpTester = kuberesource.PatchImages(memdumpTester, ct.ImageReplacements)
+	memdumpTester = kuberesource.PatchNamespaces(memdumpTester, ct.Namespace)
+	memdumpUnstructured, err := kuberesource.ResourcesToUnstructured(memdumpTester)
+	require.NoError(t, err)
+	require.NoError(t, ct.Kubeclient.Apply(t.Context(), memdumpUnstructured...))
```
burgerdev, line 60: Nit: I'd apply this before `ct.Init` so that it can start up in the background.

e2e/memdump/memdump_test.go
```diff
@@ -0,0 +1,132 @@
...
+			"labels.\\\"io.cri-containerd.kind\\\"==sandbox",
+		}, ",")
+		argv = []string{
+			"/bin/sh",
+			"-c",
+			"nsenter --target 1 --mount -- " + ctrCmd + " -n k8s.io c ls -q " + filterStr + " |" +
+				"xargs -I {} pgrep -f sandbox-{} |" +
+				"xargs -I {} gcore -o /memdump {} >/dev/null &&" +
+				"strings /memdump.*",
+		}
```
burgerdev: I wonder if this could be a bit more readable. Maybe make the script a constant somewhere:

```go
const cmd = `
nsenter --target 1 --mount -- %s -n k8s.io c ls -q  'labels."io.kubernetes.pod.namespace"==%q,... |
			xargs -I {} pgrep -f sandbox-{} |
			xargs gcore -o /memdump >/dev/null && 
			strings /memdump.*
`
```

e2e/memdump/memdump_test.go
davidweisse: `ct.Init` creates the namespace, so we have to wait for that, but we can do it before calling `generate`.

burgerdev (approved):

PR #1797, opened 2025-09-24
cli/verifier: add `VersionsMatch`

charludo:
Version mismatches between the Contrast CLI and the Contrast resources used (specifically, the `coordinator`, `initializer` and `service-mesh-proxy`) can lead to hard-to-debug errors.

Starting with this PR, `contrast generate` will fail if a version mismatch is encountered. The error message will inform you which resource needs updating.

### DIFF
--- a/cli/verifier/image_ref_valid_test.go
+++ b/cli/verifier/image_ref_valid_test.go
@@ -19,6 +19,7 @@ kind: Pod
 metadata:
   name: test
 spec:
+  runtimeClassName: contrast-cc
   containers:
     - name: test
       image: "%s"
@@ -31,6 +32,7 @@ metadata:
 spec:
   template:
     spec:
+      runtimeClassName: contrast-cc
       containers:
       - name: test
         image: "%s"
@@ -113,6 +115,53 @@ func TestVerifyImageRef(t *testing.T) {
 						}
 					}
 				})
+			}
+		})
+	}
+}
+
+const (
+	podNoImage = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+spec:
+  runtimeClassName: contrast-cc
+  containers:
+    - name: test
+`
+	statefulSetNoImage = `
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: test
+spec:
+  template:
+    spec:
+      runtimeClassName: contrast-cc
+      containers:
+      - name: test
+`
+)
+
+func TestVerifyImageRefMissing(t *testing.T) {
+	testCases := map[string]string{
+		"pod without image fails":          podNoImage,
+		"stateful set without image fails": statefulSetNoImage,
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			require := require.New(t)
+
+			toVerifySlice, err := kuberesource.UnmarshalApplyConfigurations([]byte(tc))
+			require.NoError(err)
+
+			verifier := verifier.ImageRefValid{}
+
+			for _, toVerify := range toVerifySlice {
+				require.Error(verifier.Verify(toVerify))
 			}
 		})
 	}
--- a/cli/verifier/versions_match.go
+++ b/cli/verifier/versions_match.go
@@ -1,0 +1,77 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package verifier
+
+import (
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/regclient/regclient/types/ref"
+
+	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	applymetav1 "k8s.io/client-go/applyconfigurations/meta/v1"
+)
+
+// VersionsMatch verifies that the cli version matches the version of the used resources.
+type VersionsMatch struct {
+	Version string
+}
+
+// Verify verifies that the cli version matches the version of the used resources.
+func (v *VersionsMatch) Verify(toVerify any) error {
+	// ignore this verifier for preview and dev builds.
+	if strings.Contains(v.Version, "dev") || strings.Contains(v.Version, "pre") {
+		return nil
+	}
+
+	var findings error
+
+	kuberesource.MapPodSpecWithMeta(toVerify, func(
+		meta *applymetav1.ObjectMetaApplyConfiguration,
+		spec *applycorev1.PodSpecApplyConfiguration,
+	) (*applymetav1.ObjectMetaApplyConfiguration, *applycorev1.PodSpecApplyConfiguration) {
+		if spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
+			return meta, spec
+		}
+
+		for _, container := range spec.Containers {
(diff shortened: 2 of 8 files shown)


cli/verifier/versions_match.go
```diff
@@ -0,0 +1,64 @@
...
+
+	kuberesource.MapPodSpec(toVerify, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+		if spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
+			return spec
+		}
+
+		for _, container := range spec.Containers {
+			if !slices.Contains(versionedResources, *container.Name) {
+				continue
+			}
```
burgerdev: This is not precise enough to identify Contrast containers (some other container may be called coordinator by coincidence). I think we need to check the [pod role](https://github.com/edgelesssys/contrast/blob/7de9fa26d744f210171a4a83a832d5e6a0b3f4d5/cli/cmd/generate.go#L35) and make the initializer container name `contrast-initializer`.

cli/verifier/versions_match.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+
+	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+)
+
+var versionedResources = []string{
+	"coordinator",
+	"initializer",
+	"service-mesh-proxy",
```
burgerdev: That's not the right name, is it?

cli/verifier/versions_match.go
```diff
@@ -0,0 +1,83 @@
...
+func matchesResource(image, podRole string) bool {
+	for _, r := range versionedResources {
+		if r.image == image {
+			if podRole == "" || r.podRole == podRole {
+				return true
+			}
+		}
+	}
+	return false
+}
```
burgerdev: These should go to the bottom of the file: https://github.com/edgelesssys/wiki/blob/82b70dabd097b9518f91bd44083c6d721a9e745e/company/gostyle.md#order-of-elements-in-a-source-file

cli/verifier/versions_match.go
```diff
@@ -0,0 +1,83 @@
...
+				continue
+			}
+
+			image := *container.Image
+			if idx := strings.Index(image, "@"); idx != -1 {
+				image = image[:idx]
+			}
+			if idx := strings.Index(image, ":"); idx != -1 {
+				image = image[idx+1:]
+			}
```
burgerdev: It would be safer to use a proper parser here (thinking of registry with host and port, for example) - I used https://pkg.go.dev/github.com/regclient/regclient/types/ref#Ref in the past, but maybe `go-containerregistry` has one, too?

cli/verifier/versions_match.go
```diff
@@ -0,0 +1,83 @@
...
+var versionedResources = []identifier{
+	{image: "coordinator", podRole: "coordinator"},
+	{image: "contrast-initializer"},
+	{image: "contrast-service-mesh"},
+}
+
+func matchesResource(image, podRole string) bool {
+	for _, r := range versionedResources {
+		if r.image == image {
+			if podRole == "" || r.podRole == podRole {
```
burgerdev: If I add a pod with a role other than coordinator (we don't have that right now, but might), it will be skipped, right?

cli/verifier/versions_match.go
burgerdev: We should skip this container if the image ref does not have a tag.

cli/verifier/versions_match.go
```diff
@@ -0,0 +1,83 @@
...
+			image := *container.Image
+			if idx := strings.Index(image, "@"); idx != -1 {
+				image = image[:idx]
+			}
+			if idx := strings.Index(image, ":"); idx != -1 {
+				image = image[idx+1:]
+			}
+
+			if image != v.Version {
+				findings = errors.Join(findings, fmt.Errorf("you are attempting to use Contrast version %q with a %q resource of version %q. Please update the resource",
```
burgerdev: This could be more concise (the CTA should be quite clear if the sentence before says "version mismatch").

cli/verifier/versions_match.go
charludo: Hm, doesn't look like it. There's `Reference`, but that does not provided the wanted functionality. Went with regclient, that works nicely.

cli/verifier/versions_match.go
charludo: You mean: if I add a container with name "coordinator"  in a pod with non-coordinator role, it will be skipped? If so, then: Yes, I thought that was the intention behind your previous comment about the name "coordinator" not being a sufficient indicator

cli/verifier/versions_match.go
```diff
@@ -0,0 +1,84 @@
...
+		if spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
+			return meta, spec
+		}
+
+		for _, container := range spec.Containers {
+			if !matchesResource(*container.Name, meta.Annotations["contrast.edgeless.systems/pod-role"]) {
+				continue
+			}
+
+			r, err := ref.New(*container.Image)
```
burgerdev, line 49: Nit: this will panic if the image is unset. We should report a proper error back, since this is never going to work on any k8s cluster.

cli/verifier/versions_match.go
```diff
@@ -0,0 +1,84 @@
...
+			if err != nil {
+				findings = errors.Join(findings, fmt.Errorf("could not parse image reference %q: %w", *container.Image, err))
+			}
```
burgerdev, line 53: Do we need a `continue` in this block?

cli/verifier/versions_match.go
burgerdev: I find the logic as implemented a bit hard to follow. What we want is this, right?

```go
func needsImageVersionCheck(containerName string, podRole string) {
  if slices.Contains([]string{"contrast-initializer", "contrast-service-mesh"}, containerName) {
    return true
  }
  if containerName == "coordinator" && podRole == "coordinator" {
    return true
  }
  return false
}
```

charludo: Sounds good to me.

cli/verifier/versions_match.go
charludo: oops, yes

cli/verifier/versions_match.go
charludo: Mh, that's true, but this doesn't really seem like the right verifier to handle this.I have reworked the `ImageRefValid` verifier since this seems to fall within its domain, and reordered the verifiers to ensure that one runs first. As a precaution, let's additionally just skip empty images in the `VersionsMatch` verifier.

Is this OK with you?

burgerdev: Yup, sgtm

cli/verifier/image_ref_valid_test.go
```diff
@@ -117,3 +119,56 @@ func TestVerifyImageRef(t *testing.T) {
...
+
+		toVerifySlice, err := kuberesource.UnmarshalApplyConfigurations([]byte(statefulSetNoImage))
+		require.NoError(err)
+
+		verifier := verifier.ImageRefValid{}
+
+		for _, toVerify := range toVerifySlice {
+			require.Error(verifier.Verify(toVerify))
+		}
+	})
```
burgerdev: These should be table driven, like in `TestVerifyImageRef`.

burgerdev (approved):

PR #1798, opened 2025-09-24
draft: internal: use literal style on multi-line yaml strings

charludo:
Draft, because:
- [ ] indentation does not match (4 instead of 2 spaces, changing that is a bit awkward but should be done for readability and consistency with current resources)
- [ ] lots of repetitive code in different places, can definitely be simplified into something reusable
- [ ] tests might need adjustment

@burgerdev I'd be happy to invest the time to fix the above points, if you think it's worth the effort. Otherwise, we could just keep the `yq -P`.


cli/cmd/generate.go
```diff
@@ -529,9 +529,18 @@ func getCombinedYAML(paths []string) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("read %s: %w", path, err)
 		}
+		kubeObjs, err := kuberesource.UnmarshalApplyConfigurations(data)
+		if err != nil {
+			return nil, fmt.Errorf("unmarshal %s: %w", path, err)
+		}
+		resource, err := kuberesource.EncodeResources(kubeObjs...)
```
burgerdev: Does this also reproduce the `---` doc separators?

cli/genpolicy/genpolicy.go
```diff
@@ -81,6 +82,22 @@ func (r *Runner) Run(ctx context.Context, yamlPath string, cmPath []string, logg
...
+		return fmt.Errorf("unmarshal %s: %w", yamlPath, err)
+	}
+	resource, err := kuberesource.EncodeResources(kubeObjs...)
+	if err != nil {
+		return err
+	}
+	if err := os.WriteFile(yamlPath, resource, 0o644); err != nil {
+		return fmt.Errorf("write output file: %w", err)
+	}
+
```
burgerdev: Why do we need the formatting both in genpolicy as well as in generate?

justfile
```diff
@@ -79,7 +79,7 @@ runtime target=default_deploy_target platform=default_platform:
         --add-namespace-object \
         --node-installer-target-conf-type ${node_installer_target_conf_type} \
         --platform {{ platform }} \
-        runtime > ./{{ workspace_dir }}/runtime/runtime.yml
+        runtime | yq -P > ./{{ workspace_dir }}/runtime/runtime.yml
```
burgerdev: Is this still required with the changes to `kuberesource/writer.go`? I'd rather fix this in Go, if possible.

cli/genpolicy/genpolicy.go
charludo: Both write the files out. My suggestion would be to write a wrapper function around this and just call it from both.

cli/cmd/generate.go
charludo: Yes, they are manually added during the execution of `EncodeResources`. (Actually, I'm pretty sure I saw somewhere in the docs for yaml.v3 that this should also work automatically, so the code there can probably be simplified further)

justfile
charludo: Agreed

cli/genpolicy/genpolicy.go
charludo: Edit: ~~oh, you mean why do we need to read, encode, and write again? As far as I can tell, the formatting info is lost in the bytes representation read from the file, quotes and newlines get escaped.~~

Nonesense, sorry. `genpolicy` overwrites the `yamlFile`, and does not care about our intended formatting. Opening, formatting, writing was the simplest way to reformat the file without resorting to external tools.

internal/kuberesource/reader.go
```diff
@@ -30,3 +32,23 @@ func UnmarshalApplyConfigurations(data []byte) ([]any, error) {
...
 }
+
+// YAMLBytesFromFile reads a k8 YAML file and returns a formatting-preserving encoding.
+func YAMLBytesFromFile(yamlPath string) ([]byte, error) {
+	data, err := os.ReadFile(yamlPath)
+	if errors.Is(err, os.ErrNotExist) {
+		// A YAML file not existing should not result in an error; the new contents are simply empty.
+		return nil, nil
+	} else if err != nil {
+		return nil, fmt.Errorf("read %s: %w", yamlPath, err)
```
katexochen:
```suggestion
		return nil, fmt.Errorf("reading %s: %w", yamlPath, err)
```

internal/kuberesource/reader.go
```diff
@@ -30,3 +32,23 @@ func UnmarshalApplyConfigurations(data []byte) ([]any, error) {
...
+	}
+	kubeObjs, err := UnmarshalApplyConfigurations(data)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal %s: %w", yamlPath, err)
```
katexochen:
```suggestion
		return nil, fmt.Errorf("unmarshaling %s: %w", yamlPath, err)
```

internal/kuberesource/reader.go
```diff
@@ -30,3 +32,23 @@ func UnmarshalApplyConfigurations(data []byte) ([]any, error) {
 	}
 	return result, nil
 }
+
+// YAMLBytesFromFile reads a k8 YAML file and returns a formatting-preserving encoding.
+func YAMLBytesFromFile(yamlPath string) ([]byte, error) {
+	data, err := os.ReadFile(yamlPath)
+	if errors.Is(err, os.ErrNotExist) {
+		// A YAML file not existing should not result in an error; the new contents are simply empty.
+		return nil, nil
```
burgerdev: What's the use case for this branch? I'd have expected the caller to care about that.

internal/kuberesource/writer.go
```diff
@@ -5,10 +5,11 @@ package kuberesource
 
 import (
 	"bytes"
+	"strings"
 
+	"gopkg.in/yaml.v3"
```
burgerdev: This is now unmaintained :/

internal/kuberesource/writer.go
```diff
@@ -5,10 +5,11 @@ package kuberesource
...
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/yaml"
```
burgerdev, line 11: I wonder if we're missing out on some special considerations by this package:

> In short, this library first converts YAML to JSON using go-yaml and then uses json.Marshal and json.Unmarshal to convert to or from the struct. This means that it effectively reuses the JSON struct tags as well as the custom JSON methods MarshalJSON and UnmarshalJSON unlike go-yaml.

internal/kuberesource/reader.go
charludo: Mostly for the genpolicy test, which calls genpolicy with non-existent files.

internal/kuberesource/writer.go
charludo: Oh fantastic -.-

charludo: Seems like we should be able to migrate to `go-yaml`; however, we also seem to have a bunch of indirect dependencies on `yam.v3` anyways 😅

charludo:
@burgerdev migrated to `go-yaml`, which fortunately further simplified the yaml encoding.

internal/kuberesource/writer.go
charludo: Hmm. Good question. I wasn't able to find whether this is something added for the users' convenience, or if `k8` actually depends on this...

internal/kuberesource/reader.go
burgerdev: I don't think this needs to be handled in this function. The caller should know whether this file exists, and what should happen if it doesn't. In this function, we can't really tell what the intent was, so we should not swallow errors.

internal/kuberesource/writer.go
burgerdev: The new library has this covered, I believe: https://pkg.go.dev/github.com/goccy/go-yaml#UseJSONMarshaler.

charludo: Oh, very nice!

PR #1799, opened 2025-09-24
imagepuller: reuse cached layers

charludo:
Not sure if a further test is warranted - would need to somehow get the logs from the imagepuller... 😅

### DIFF
--- a/imagepuller/internal/service/stubs_test.go
+++ b/imagepuller/internal/service/stubs_test.go
@@ -4,6 +4,7 @@
 package service
 
 import (
+	"errors"
 	"io"
 
 	"github.com/containers/storage"
@@ -16,17 +17,29 @@ import (
 )
 
 type stubStore struct {
-	putLayerDigest digest.Digest
-	putLayerErr    error
+	putLayerDigest  digest.Digest
+	putLayerErr     error
+	putLayerCount   int
+	lookupID        string
+	lookupSucceeded bool
 
 	storage.Store
 }
 
 func (s *stubStore) PutLayer(_, _ string, _ []string, _ string, _ bool, _ *storage.LayerOptions, _ io.Reader) (*storage.Layer, int64, error) {
+	s.putLayerCount++
 	if s.putLayerErr != nil {
 		return nil, 0, s.putLayerErr
 	}
 	return &storage.Layer{CompressedDigest: s.putLayerDigest}, 0, nil
+}
+
+func (s *stubStore) Lookup(_ string) (string, error) {
+	if s.lookupID != "" {
+		s.lookupSucceeded = true
+		return s.lookupID, nil
+	}
+	return "", errors.New("by default, return a unique error")
 }
 
 type stubRemote struct {
--- a/imagepuller/internal/service/verify.go
+++ b/imagepuller/internal/service/verify.go
@@ -54,7 +54,6 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
 
 		var digestFound *gcr.Hash
 		for _, m := range manifest.Manifests {
-			log.Info("MANIFEST", "name", m.Platform.String())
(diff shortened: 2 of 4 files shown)


imagepuller/internal/service/verify.go
```diff
@@ -92,6 +91,15 @@ func (s *ImagePullerService) storeAndVerifyLayers(log *slog.Logger, remoteImg gc
 
 	previousLayer := ""
 	for idx, layer := range layers {
+		ldManifest := manifest.Layers[idx].Digest.String()
```
burgerdev: `ldManifest` is a layer digest, right? In that case, `layerDigest` would be a more intuitive variable name.

imagepuller/internal/service/verify.go
```diff
@@ -116,12 +124,18 @@ func (s *ImagePullerService) storeAndVerifyLayers(log *slog.Logger, remoteImg gc
...
 		if ldManifest != ld {
 			return "", fmt.Errorf("%w: expected digest '%s' but got digest '%s'", errValidateLayer, ldManifest, ld)
 		}
 
+		if err := s.Store.RemoveNames(putLayer.ID, putLayer.Names); err != nil {
+			return "", fmt.Errorf("removing pre-existing layer names: %w", err)
+		}
+		if err := s.Store.AddNames(putLayer.ID, []string{ldManifest}); err != nil {
+			return "", fmt.Errorf("adding layer digest as layer name: %w", err)
+		}
```
burgerdev: What names is the library using, if not the digest? Could we simply set the name in the `Store.PutLayer` operation?

imagepuller/internal/service/verify_test.go
```diff
@@ -214,6 +221,11 @@ func TestStoreAndVerifyLayers(t *testing.T) {
 			_, err = s.storeAndVerifyLayers(log, tc.stubImg)
 
 			assert.ErrorIs(err, tc.wantErr)
+
+			out := buf.String()
+			if tc.cachedLayerID != "" {
+				assert.Contains(out, tc.cachedLayerID)
+			}
```
burgerdev: This check seems a bit brittle. Can't we count the number of times the stub has been called and work with that instead?

charludo: Good idea, will do.

imagepuller/internal/service/verify.go
charludo: My thinking here was originally that not stripping names might allow a bad actor to write the digest of layer B of image B as a name in layer A from image A, causing the imagepuller to create a hybrid image with some layers from B, some from A if B is pulled before A. If B had e.g. suid binaries in places accessible to an application running in A, that might open a way for privilege escalation.

Buuut your comment prompted me to look into it again, and since my assumption was faulty and the names are not a feature of docker images, but of the store, this all becomes irrelevant 😂 

We're doing a similar thing for wholy-cached images, want me to remove both instances?

burgerdev: Yeah, makes sense.

imagepuller/internal/service/verify_test.go
charludo: Instead of counting, I went with setting a flag when `stubStore.Lookup` is called; let me know if this is OK with you.

Edit: added both.

imagepuller/internal/service/stubs_test.go
```diff
@@ -16,19 +17,39 @@ import (
...
+	return "", errors.New("by default, return a unique error")
+}
+
+func (s *stubStore) AddNames(_ string, _ []string) error {
+	return nil
+}
+
+func (s *stubStore) RemoveNames(_ string, _ []string) error {
+	return nil
+}
```
burgerdev: These are not needed anymore, correct?

burgerdev (approved):
lgtm

PR #1800, opened 2025-09-24
rfc: registry authentication

burgerdev:
How can we add authentication (and more) to guest pulls? Read the [rendered version](https://github.com/edgelesssys/contrast/blob/rfc/011/rfc/011-registry-authentication.md) of this draft RFC.

### DIFF
--- a/rfc/011-registry-authentication.md
+++ b/rfc/011-registry-authentication.md
@@ -1,0 +1,139 @@
+# RFC 011: Registry Authentication
+
+## Background
+
+### Registry credentials
+
+Most container registries used in enterprise environments require authentication to pull images.
+There are several ways in which these credentials can be supplied to the container runtime:
+
+1. Image pull secret referenced directly in the pod spec: <https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/>.
+2. Image pull secret attached to the service account: <https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account>. This indirectly populates the `imagePullSecrets` field of all pod specs using this service account.
+3. Global credentials in the container runtime configuration: <https://github.com/containerd/containerd/blob/5bcf77a/docs/cri/registry.md?plain=1#L37-L40>.
+4. Dedicated image credential providers: <https://kubernetes.io/docs/tasks/administer-cluster/kubelet-credential-provider/>.
+
+Image pull secrets come in [two formats](https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets).
+The `dockercfg` format has been [removed](https://github.com/docker/cli/pull/2504) by Docker itself, but this wasn't too long ago and it's likely still widely used.
+
+### Registry certificates
+
+Internal registries often use certificates that don't chain back to a web PKI root CA, but an internal one.
+Since Kubernetes doesn't have an API for specifying trusted registry CAs, such CA certificates need to be [configured at the container runtime](https://github.com/containerd/containerd/blob/5bcf77a55038ad658c57fdecc48af54935a0d42f/docs/cri/config.md?plain=1#L744).
+
+### Proxies
+
+Many corporate networks require HTTP connections to the public internet to go through a proxy.
+The de-facto standard for specifying such proxies are the environment variables `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY`.
+If Contrast is running in such an environment, it needs to respect these variables in order to pull images from within the guest.
+
+## Requirements
+
+1. Users must be able to guest-pull images from registries that require authentication.
+2. Users must be able to guest-pull images from registries that don't participate in web PKI.
+3. Users must be able to guest-pull images through HTTP proxies.
+
+## Design
+
+**NOTE**: This design depends on the [initdata processor](https://github.com/kata-containers/kata-containers/issues/11532), which isn't merged upstream at the time of writing.
+
+### Changes to the `imagepuller`
+
+We add support for an optional configuration file to the `imagepuller`, with the following structure:
+
+```go
+import "github.com/google/go-containerregistry/pkg/authn"
+
+type Config struct {
+    Auths map[string]authn.AuthConfig // key is the OCI registry
+    CA [][]byte // list of trusted PEM-encoded CA certificates
+    ExtraEnv map[string]string
+}
+```
+
+On startup, the `imagepuller` reads the config file from `/run/measured-cfg/imagepuller.TBD`, where it was provisioned by the initdata processor.
+It takes the `Auths` and stores them in a field, and creates a [`tls.Config`].
+If `CA` isn't empty, it overrides the `RootCAs` field of the [`tls.Config`].
+Each key-value pair in `ExtraEnv` is added to the `imagepuller`s own environment.
+
+When a `PullImage` request comes in, the `imagepuller` extracts the registry part from the image reference and looks it up in `Auths`.
+If present, it wraps the [`authn.AuthConfig`] with [`authn.FromConfig`] and passes it as an option [`remote.WithAuth`].
+The [`tls.Config`] and [`http.ProxyFromEnvironment`] are used to construct an [`http.Transport`] that's passed as an option [`remote.WithTransport`].
+
+[`tls.Config`]: https://pkg.go.dev/crypto/tls#Config
+[`http.ProxyFromEnvironment`]: https://pkg.go.dev/net/http#ProxyFromEnvironment
+[`http.Transport`]: https://pkg.go.dev/net/http#Transport
+[`authn.AuthConfig`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#AuthConfig
+[`authn.FromConfig`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#FromConfig
+[`remote.WithAuth`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithAuth
+[`remote.WithTransport`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithTransport
+
+### Changes to the CLI
+
+During `contrast generate`, the CLI walks over the Kubernetes resources looking for a secret with the appropriate `type` field.
+It combines the individual `dockerconfigjson` messages into the `Auths` field of `imagepuller.Config`.
+Two new repeatable command line flags, `--image-pull-cert` and `--image-pull-env`, populate the remaining fields of the config file.
+The config is then serialized and attached to the initdata produced by `genpolicy`.
+
+### Security considerations
+
+#### Credentials visible in initdata annotation
+
+Credentials transferred via initdata suffer from the resource leak considerations described in [architecture/components/policies](../docs/docs/architecture/components/policies#evaluation).
+This is unavoidable until we've a dedicated channel between guest components and Coordinator (see also [below](#credentials-distributed-by-coordinator)).
+
+#### CA certificates
+
+The CA certificates aren't required for image integrity - that's accomplished by pinning references.
+Rather, the CA certificate option serves two purposes:
+
+1. Allow connecting to a registry that isn't publicly trusted in the web PKI.
+2. Restrict who can ostensibly intercept and log traffic for metadata analysis (for example, what images are loaded).
+
+## Alternatives considered
+
+### Configuration file
+
+The information in an `imagepuller` config file is usually global for a deployment site, and would thus naturally fit into a config file for the CLI.
+However, the question whether we want a config file for the CLI and how we should go about adding one is beyond the scope of this proposal.
+As soon as we've a config file, we can allow registry authentication there while keeping or deprecating the mechanism proposed here.
+
+### Deeper integration into Kubernetes
+
+As outlined in the [Background](#background) section, there are other Kubernetes-native ways to get to registry credentials.
+Implementing these for Contrast would be a large stretch, though:
+
+* In order to retrieve secrets based on service accounts, the CLI would need direct access to the Kubernetes cluster.
+* In order to use credential helpers, there would need to be a communication channel from the guest to the node.
+
+Both seem disproportionately complex to implement, so we settle for the easier solution that still provides a path for users blocked by registry authentication.
+
+#### Extend list of CAs
+
+Instead of overriding `RootCAs`, we could instead extend the list of system roots.
+However, users have good reasons not to trust the system roots, so we would need to add another knob to allow that.
+It's probably easier - to implement and to audit - to let users specify the full list of trusted roots.
+
+### CA per registry
+
+Instead of having a list of globally trusted CAs, we could have CAs per registry.
+This would allow fine-grained control over who can authenticate which image source.
+On the other hand, this introduces complexity for the user, because they need to express this relationship somehow.
+Given that the CA cert is defense in depth and doesn't affect the CC-security (see [CA certificates](#ca-certificates)), it's probably fine to defer such a feature until we've configuration files.
+
+### Credentials distributed by Coordinator
+
+Instead of sending the credentials with initdata, we could attach them to the manifest and hand them to the initializer, like the workload secret.
+In the current architecture, this would result in a chicken-and-egg problem, because the initializer is a container that needs pulling.
+While we could eventually move the initialization workflow into a guest component, this is a larger feature beyond the scope of this proposal.
+But whenever we reach the point where the initializer runs as a guest component, we can change the delivery mechanism for the image puller config without user-visible changes.
+
+#### Encrypted images
+
+A topic that comes up frequently are encrypted images.
+While at first glance it would seem appropriate to include encrypted images in this proposal, there are reasons why they're out of scope.
+
+1. Confidentiality of the keys would require confidential conveyance, likely over an aTLS channel.
+   At the very least, we'd need the initialization workflow changes outlined above.
+2. In the `force_guest_pull` world, the host needs to have access to the layer keys (at least in general, there may be exceptions).
+
+Since encryption in transit is already covered by CA certs and client credentials, encrypted images can be deferred until the necessary prerequisites are in place.
--- a/tools/vale/styles/config/vocabularies/edgeless/accept.txt
+++ b/tools/vale/styles/config/vocabularies/edgeless/accept.txt
@@ -74,6 +74,7 @@ Hetzner
 hostname
 hyperv
 IAM
+initdata
 initializers?
 initramfs
 initrd


rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,139 @@
...
+### Changes to the `imagepuller`
+
+We add support for an optional configuration file to the `imagepuller`, with the following structure:
+
+```go
+import "github.com/google/go-containerregistry/pkg/authn"
+
+type Config struct {
+    Auths map[string]authn.AuthConfig // key is the OCI registry
+    CA [][]byte // list of trusted PEM-encoded CA certificates
```
thomasten: InsecureSkipVerify could be helpful for debugging or setting up test deployments. It could even be sufficient for running a registry mirror inside the cluster in production (although not required).

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,139 @@
...
+[`http.Transport`]: https://pkg.go.dev/net/http#Transport
+[`authn.AuthConfig`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#AuthConfig
+[`authn.FromConfig`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#FromConfig
+[`remote.WithAuth`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithAuth
+[`remote.WithTransport`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithTransport
+
+### Changes to the CLI
+
+During `contrast generate`, the CLI walks over the Kubernetes resources looking for a secret with the appropriate `type` field.
+It combines the individual `dockerconfigjson` messages into the `Auths` field of `imagepuller.Config`.
```
thomasten: The Secret itself may not be part of the K8s resources used for deployment. Should we add a command line flag to add an auth from a file?

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,139 @@
...
+Two new repeatable command line flags, `--image-pull-cert` and `--image-pull-env`, populate the remaining fields of the config file.
+The config is then serialized and attached to the initdata produced by `genpolicy`.
+
+### Security considerations
+
+#### Credentials visible in initdata annotation
+
+Credentials transferred via initdata suffer from the resource leak considerations described in [architecture/components/policies](../docs/docs/architecture/components/policies#evaluation).
```
thomasten:
```suggestion
Credentials transferred via initdata suffer from the resource leak considerations described in [architecture/components/policies](../docs/docs/architecture/components/policies.md#evaluation).
```

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,139 @@
...
+### CA per registry
+
+Instead of having a list of globally trusted CAs, we could have CAs per registry.
+This would allow fine-grained control over who can authenticate which image source.
+On the other hand, this introduces complexity for the user, because they need to express this relationship somehow.
+Given that the CA cert is defense in depth and doesn't affect the CC-security (see [CA certificates](#ca-certificates)), it's probably fine to defer such a feature until we've configuration files.
+
+### Credentials distributed by Coordinator
+
+Instead of sending the credentials with initdata, we could attach them to the manifest and hand them to the initializer, like the workload secret.
```
thomasten: Not sure what exactly you mean with "attach them to the manifest", but the manifest itself should not contain secrets to be publicly sharable.

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,139 @@
...
+[`authn.AuthConfig`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#AuthConfig
+[`authn.FromConfig`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#FromConfig
+[`remote.WithAuth`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithAuth
+[`remote.WithTransport`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithTransport
+
+### Changes to the CLI
+
+During `contrast generate`, the CLI walks over the Kubernetes resources looking for a secret with the appropriate `type` field.
+It combines the individual `dockerconfigjson` messages into the `Auths` field of `imagepuller.Config`.
+Two new repeatable command line flags, `--image-pull-cert` and `--image-pull-env`, populate the remaining fields of the config file.
```
m1ghtym0: Just to recap for myself what the effects are from a UX perspective:

1. People are using pull secrets directly in their k8s resources -> No changes
2. People are attaching pull secrets in other ways -> Need to switch to (1.)
3. People using HTTP proxys or alternative CAs -> Need to pass them through the Contrast CLI flags and other references might become obsolete

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,139 @@
...
+
+As outlined in the [Background](#background) section, there are other Kubernetes-native ways to get to registry credentials.
+Implementing these for Contrast would be a large stretch, though:
+
+* In order to retrieve secrets based on service accounts, the CLI would need direct access to the Kubernetes cluster.
+* In order to use credential helpers, there would need to be a communication channel from the guest to the node.
+
+Both seem disproportionately complex to implement, so we settle for the easier solution that still provides a path for users blocked by registry authentication.
+
+#### Extend list of CAs
```
katexochen:
```suggestion
### Extend list of CAs
```

rfc/011-registry-authentication.md
burgerdev: Agreed, let me add it.

rfc/011-registry-authentication.md
burgerdev: I see the problem that it's inconvenient to remove the secret after generate and before apply, and I agree that the secret is rarely applied together with the resources, so a side input makes sense. I'd still want it to be a dockerconfigjson secret, though, which leaves the options:

1. `kubectl get` the secret directly from the cluster, if they have access.
2. `kubectl create secret dockerregistry --docker-password=`  if they have the credentials in some other form.
3. The secret is  a regular part of the deployment anyway.

I don't want to invite users to use their `~/.docker/config.json` here, because it's unlikely to work (credential helpers) and those are usually not the right credentials anyway.

rfc/011-registry-authentication.md
burgerdev: Point (2) is addressed by the suggestion in https://github.com/edgelesssys/contrast/pull/1800/files#r2378236153. 

> other references might become obsolete

I don't understand what you mean by this.

rfc/011-registry-authentication.md
burgerdev: Good point, seems like I was a bit lazy when constructing this alternative. Credentials could be sent securely with the `SetManifest` call, though.

However, the point remains that for this to happen, we'd need to change the initialization architecture.

rfc/011-registry-authentication.md
m1ghtym0: Yeah, sorry. You can actually just cut the second half of my sentence. All I meant was that the "original" deployment files might contain env variables to configure HTTP proxies or CA configurations which are now passed via the Contrast CLI instead. Hence, you might have redundant/obsolete information in the deployment files and you might want to remove it from there. No big deal. Just a point to remember if people think about having two versions of the same deployment: One with Contrast for production and one without for debugging for example.

burgerdev: The workload's configuration of env vars and CA certs is unrelated to the imagepuller configuration. The imagepuller currently uses the VM's trust store, whereas the containers come with their own. Same goes for the env vars. So, users can keep their deployments as-is, but now have the option to configure the imagepuller, too.

rfc/011-registry-authentication.md
thomasten: reminder to add InsecureSkipVerify

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,163 @@
...
+[`authn.FromConfig`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#FromConfig
+[`remote.WithAuth`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithAuth
+[`remote.WithTransport`]: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithTransport
+
+### Changes to the node-installer
+
+The node-installer gets another optional volume mount, similar to the [target config](https://github.com/edgelesssys/contrast/blob/6ef858031759966dfd3cdeda2b4570bed45fdcda/internal/kuberesource/parts.go#L126-L131) but using a secret.
+The secret contains one key, `imagepuller.TBD`, containing a serialized version of `InsecureConfig` defined above.
+This secret is created by the k8s administrator (or the workload owner) before applying the runtime.
+If the node-installer finds a mounted secret, it writes the content into `/opt/edgeless/contrast-cc-*/etc/host-config/imagepuller.TBD`.
```
thomasten, line 171: Can I update the config? Deleting the nodeinstaller pods to rewrite it should work, right?

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,163 @@
...
+Instead of overriding `RootCAs`, we could instead extend the list of system roots.
+However, users have good reasons not to trust the system roots, so we would need to add another knob to allow that.
+It's probably easier - to implement and to audit - to let users specify the full list of trusted roots.
+
+### CA per registry
+
+Instead of having a list of globally trusted CAs, we could have CAs per registry.
+This would allow fine-grained control over who can authenticate which image source.
+On the other hand, this introduces complexity for the user, because they need to express this relationship somehow.
+Given that the CA cert is defense in depth and doesn't affect the CC-security (see [CA certificates](#wrong-ca-certificates)), it's probably fine to defer such a feature until we've configuration files.
```
thomasten: If I use images from docker.io and images from my private registry, the combination of not extending the list of CAs and not defining CAs per registry would force me to collect the required CAs for the web PKI myself.

rfc/011-registry-authentication.md
burgerdev: Done

rfc/011-registry-authentication.md
burgerdev: Added a paragraph to explain this flow.

rfc/011-registry-authentication.md
burgerdev: Clarified.

rfc/011-registry-authentication.md
burgerdev: This is obsolete after the change to node-installer provisioned config.

rfc/011-registry-authentication.md
burgerdev: Also obsolete after rewrite

rfc/011-registry-authentication.md
```diff
@@ -77,6 +77,9 @@ The secret contains one key, `imagepuller.TBD`, containing a serialized version
 This secret is created by the k8s administrator (or the workload owner) before applying the runtime.
 If the node-installer finds a mounted secret, it writes the content into `/opt/edgeless/contrast-cc-*/etc/host-config/imagepuller.TBD`.
```
katexochen, line 171: Note: This might be good for now, but maybe we need to split this from the actual runtime installation in the future. A runtimeClass can be used by multiple deployments. But maybe you want to use different credentials for different deployments? At the same time, we have plans to allow multiple runtimeClasses (think: gpu/non-gpu) in the same deployment, where we don't want to copy the secrets for each runtime.

rfc/011-registry-authentication.md
burgerdev: Yes, that's inconvenient. Do you think CA per registry is the way to go, possibly with some sort of matching:

```go
type InsecureConfig struct {
    // Mapping of registry globs to trusted PEM-encoded CA certificates.
    // Could be more than one: https://pkg.go.dev/crypto/x509#CertPool.AppendCertsFromPEM.
    CA map[string][]byte 
}

var myAdditiveConfig = InsecureConfig{
    CA: map[string][]byte{
        "*.corp.internal": myCert,
        "*.corp.com": myOtherCert,
        // key not found -> web PKI
    },
}

var myParanoidConfig = InsecureConfig{
    CA: map[string][]byte{
        "*": append(myCert, myOtherCert),
        // all registries match above, so web PKI does not apply anywhere
    },
}
```

rfc/011-registry-authentication.md
burgerdev: > A runtimeClass can be used by multiple deployments. But maybe you want to use different credentials for different deployments? 

We need to adapt the Kata runtime in order to read this, so we could go all the way and implement this as a config entry that can be overridden per deployment with the appropriate annotation? That would leave the option to customize, but would not clutter the common path (where registry configuration is per-cluster). 

> At the same time, we have plans to allow multiple runtimeClasses (think: gpu/non-gpu) in the same deployment, where we don't want to copy the secrets for each runtime.

I don't think this is too bad, since the same secret can be used for separate node-installers.

rfc/011-registry-authentication.md
thomasten: Yes, I'd prefer that over extending. Your suggestion looks good.

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Changes to the node-installer
+
+The node-installer gets another optional volume mount, similar to the [target config](https://github.com/edgelesssys/contrast/blob/6ef858031759966dfd3cdeda2b4570bed45fdcda/internal/kuberesource/parts.go#L126-L131) but using a secret.
+The secret contains one key, `imagepuller.TBD`, containing a serialized version of `InsecureConfig` defined above.
+This secret is created by the k8s administrator (or the workload owner) before applying the runtime.
+If the node-installer finds a mounted secret, it writes the content into `/opt/edgeless/contrast-cc-*/etc/host-config/imagepuller.TBD`.
+
+node-installer operations are intended to be idempotent.
+In order to change the imagepuller configuration, the k8s administrator only needs to change the secret and restart the `DaemonSet`.
```
charludo, line 174: For clarification: this implies that the imagepuller re-reads the configuration upon every pull requests, correct? Just asking because further up, it's explicitly written that the configuration is just read once at startup.

burgerdev: The configuration _inside the guest_ is only provisioned once, and never changes. If the k8s secret is changed, the pod needs to be restarted to receive the updated config.

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,281 @@
...
+rxdeFC7ySuuu95VPjSvO7EUM9qyjUzBRMB0GA1UdDgQWBBTVnuI2o36Mrja3RvwE
+82lWg2m19zAfBgNVHSMEGDAWgBTVnuI2o36Mrja3RvwE82lWg2m19zAPBgNVHRMB
+Af8EBTADAQH/MAoGCCqGSM49BAMCA0gAMEUCIGmEkl8jxjxqyAxs3QoAXeIx++Bz
+Zm9dwbeTbrKysrGXAiEA8ce6iyJUCZCZVVJs/HDLcPbOKc2EPZvdcGGjIlGXulo=
+-----END CERTIFICATE-----
+'''
+
+[registries."."]
+ca-certs = "no PEM here means no CA certificates"
+```
```
charludo, line 146: How do we handle multiple auth configs? I.e. if the `[registries."."]` entry was amended to

```toml
[registries."."]
ca-certs = "no PEM here means no CA certificates"
auth = "conflicting"
```

simply select the first match?

A similar question for the certificates: If registries A and B have differing auth-configs, but share certs, should specifying the cert once for `registries."."` apply it to both, and should multiple matches concatenate all matched certs?

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,281 @@
...
+The internal registry is served from within the corporate firewall, while access to external registries is mediated by a proxy.
+Other registries are allowed, as long as they don't require authentication.
+
+```toml
+[extra-env]
+HTTP_PROXY = "https://proxy.corp"
+HTTPS_PROXY = "https://proxy.corp"
+NO_PROXY = ".corp"
+
+[registries."registry.corp."]
```
charludo, line 92: What is the purpose of the trailing/leading `.`, e.g. in `[registries."registry.corp."]`? If this is meant as a "glob everything", I'd rather we use `[registries."registry.corp.*"]` (as well as `[registries."*"]`, and so on, as there's quite some confusion with the toml key notation to be had.

rfc/011-registry-authentication.md
```diff
@@ -0,0 +1,281 @@
...
+type InsecureConfig struct {
+    Registries map[string]Registry // mapping of domain name patterns to registry configuration
+    ExtraEnv map[string]string
+}
+
+type Registry struct {
+    authn.AuthConfig
+    CACerts string // concatenated, PEM-encoded CA certificates (system certs if nil)
+    InsecureSkipVerify bool
+}
```
charludo, line 56: Using `string` for the certs is a bit inconvenient IMO. Is there anything standing in the way of


```suggestion
type Registry struct {
    authn.AuthConfig
    CACerts []tls.Certificate
    InsecureSkipVerify bool
}
```

instead? With TOML changed to 

```toml
[registries."registry.corp."]
ca-certs = [
    '''
    -----BEGIN CERTIFICATE-----
    MIIBezCCASGgAwIBAgIUUugBbePTzyVApU4DLSMmHnXXjcwwCgYIKoZIzj0EAwIw
    EzERMA8GA1UEAwwIWW91ck5hbWUwHhcNMjUxMDIxMTUwNDI0WhcNMjYxMDIxMTUw
    NDI0WjATMREwDwYDVQQDDAhZb3VyTmFtZTBZMBMGByqGSM49AgEGCCqGSM49AwEH
    A0IABIgsA5IEeiBq6jDpH2ttxrI96beeOqa+EpGqmznQmzpFkPEpLWMUt21Ien71
    rxdeFC7ySuuu95VPjSvO7EUM9qyjUzBRMB0GA1UdDgQWBBTVnuI2o36Mrja3RvwE
    82lWg2m19zAfBgNVHSMEGDAWgBTVnuI2o36Mrja3RvwE82lWg2m19zAPBgNVHRMB
    Af8EBTADAQH/MAoGCCqGSM49BAMCA0gAMEUCIGmEkl8jxjxqyAxs3QoAXeIx++Bz
    Zm9dwbeTbrKysrGXAiEA8ce6iyJUCZCZVVJs/HDLcPbOKc2EPZvdcGGjIlGXulo=
    -----END CERTIFICATE-----
    ''',

    '''
    -----BEGIN CERTIFICATE-----
    MIIBezCCASGgAwIBAgIUUugBbePTzyVApU4DLSMmHnXXjcwwCgYIKoZIzj0EAwIw
    EzERMA8GA1UEAwwIWW91ck5hbWUwHhcNMjUxMDIxMTUwNDI0WhcNMjYxMDIxMTUw
    NDI0WjATMREwDwYDVQQDDAhZb3VyTmFtZTBZMBMGByqGSM49AgEGCCqGSM49AwEH
    A0IABIgsA5IEeiBq6jDpH2ttxrI96beeOqa+EpGqmznQmzpFkPEpLWMUt21Ien71
    rxdeFC7ySuuu95VPjSvO7EUM9qyjUzBRMB0GA1UdDgQWBBTVnuI2o36Mrja3RvwE
    82lWg2m19zAfBgNVHSMEGDAWgBTVnuI2o36Mrja3RvwE82lWg2m19zAPBgNVHRMB
    Af8EBTADAQH/MAoGCCqGSM49BAMCA0gAMEUCIGmEkl8jxjxqyAxs3QoAXeIx++Bz
    Zm9dwbeTbrKysrGXAiEA8ce6iyJUCZCZVVJs/HDLcPbOKc2EPZvdcGGjIlGXulo=
    -----END CERTIFICATE-----
    '''
]
```

of course.

thomasten (approved):
Thank you! Seems easy yet flexible enough for the cases I can think of.

rfc/011-registry-authentication.md
burgerdev: The lookup scheme is designed so that at most one block matches a given registry - this makes it harder to configure something insecure by accident. Note the difference to the k8s auth lookup mentioned in the appendix. While we could do the lookup for each individual key, the user configuring the key would need to be aware of the precedence order. A bit of duplication seems ok to me here, especially because auths are almost always not shareable between registries. It would also be harder to implement a case like example 2, but where the explicit registry uses web pki certs and others should not be used.

rfc/011-registry-authentication.md
burgerdev: The trailing `.` makes this a _fully-qualified domain name_. As mentioned in the appendix, it does not have to be in here, but should be supported for completeness.

The leading `.` indicates "only subdomains should match". We can replace the leading empty label (left of the leading `.`, if you will) with a single asterisk, if that helps readability. Semantics should be the same. What I like about the notation without asterisk is that it does the right thing for the fully-qualified root domain, i.e. `.`.

What I dislike about the asterisk is that it somewhat implies sophisticated matching, and people will try to write regexes or glob expressions like `registry.*.internal` into their config. Working on the suffix label list seems clearer to me.

rfc/011-registry-authentication.md
burgerdev: I wrote this struct with [`CertPool.AppendCertsFromPEM`](https://pkg.go.dev/crypto/x509#CertPool.AppendCertsFromPEM) in mind, which is the easiest way to go from PEM to `tls.Config` imho. This format also supports "comments" out of the box (the text between the PEM blocks). 

Also, I think that for the user it'd be easier to work with a multiline string than with a list of multiline strings.

charludo: > [`CertPool.AppendCertsFromPEM`](https://pkg.go.dev/crypto/x509#CertPool.AppendCertsFromPEM)

That makes sense!

rfc/011-registry-authentication.md
charludo: > The trailing `.` makes this a _fully-qualified domain name_. As mentioned in the appendix, it does not have to be in here, but should be supported for completeness.

I'm sorry, I had not seen the new appendix. That would indeed have clarified this.

With this explanation, the `.` makes sense, no need to switch to `*`.

To summarize: sort registries by `key` length in descending order, use first registry for which `imageRefRegistry` ends in`key[0:-1]` if ending in `.`, `key` otherwise. Seems like a nice and simple solution!

PR #1801, opened 2025-09-25
e2e: add openssl logging for unexpected success

burgerdev:
This happened in a CI flake the other day. Although the issue was likely an earlier failing test, it would be good to have the openssl connection logs for debugging this situation.


charludo (approved):
LGTM, test failure seems to be unrelated

PR #1807, opened 2025-09-26
initdata: add FromDevice function

burgerdev:
This port of [Kata agent code](https://github.com/kata-containers/kata-containers/blob/077aaa6480953de3770b8c3e240dbb0dc44a186e/src/agent/src/initdata.rs#L81-L98) will be used in our initdata processor.

### DIFF
--- a/internal/initdata/initdata.go
+++ b/internal/initdata/initdata.go
@@ -9,10 +9,13 @@ import (
 	"crypto/sha256"
 	"crypto/sha512"
 	"encoding/base64"
+	"encoding/binary"
 	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
+	"os"
+	"slices"
 	"strings"
 
 	"github.com/pelletier/go-toml/v2"
@@ -81,7 +84,11 @@ func DecodeKataAnnotation(annotation string) (*Initdata, error) {
 		return nil, err
 	}
 
-	decompressed, err := decompress(decoded)
+	return decodeGzipped(decoded)
+}
+
+func decodeGzipped(data []byte) (*Initdata, error) {
+	decompressed, err := decompress(data)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


internal/initdata/initdata.go
```diff
@@ -125,10 +128,60 @@ func (i *Initdata) EncodeKataAnnotation() (string, string, error) {
...
+	var i Initdata
+	if err := toml.Unmarshal(decompressed, &i); err != nil {
+		return nil, nil, fmt.Errorf("parsing TOML: %w", err)
+	}
+	digest, err := i.digest(decompressed)
+	if err != nil {
+		return nil, nil, fmt.Errorf("calculating digest: %w", err)
+	}
+
+	return &i, digest, nil
```
charludo: This is almost 1:1 the same as `DecodeKataAnnotation`, minus verifying that a known algorithm is used. A shared helper would seem to make sense.

internal/initdata/initdata.go
```diff
@@ -180,3 +233,8 @@ func compress(decompressed []byte) ([]byte, error) {
 
 	return compressed.Bytes(), nil
 }
+
+var (
+	errWrongMagic = errors.New("wrong magic number")
+	errTooLarge   = errors.New("initdata is too large")
+)
```
charludo: Please move up top, to the other error definitions.

charludo (changes requested):

internal/initdata/initdata.go
burgerdev: Done. I wonder whether we'll need to return the digest from `DecodeKataAnnotation`, too, but I guess we'll see when we implement the CLI changes.

internal/initdata/initdata.go
burgerdev: Done

charludo (approved):

PR #1813, opened 2025-09-30
initdata: add Raw type

burgerdev:
Most of the time, working with initdata requires keeping a reference of the serialized TOML document and the corresponding digest (see [`manifest.Policy`](https://github.com/edgelesssys/contrast/blob/24b5742b6894c1bf5bcc7960d561fe1c4f18c62f/internal/manifest/manifest.go#L280)). It's difficult to work with the current API because the intermediate state is not exposed.

This commit adds the intermediate representation as an exported type and adjusts the existing functions to work with that type. So, instead of `DecodeKataAnnotation()`, one now needs to call `DecodeKataAnnotation()` followed by `Parse()`. The only function that now needs a digest output is the explicit `Raw.Digest` method. Existing use of `FromDevice()`, for example, needs to first call `FromDevice()`, and then `Digest()` and `Parse()` on the Raw object.

### DIFF
--- a/internal/initdata/initdata.go
+++ b/internal/initdata/initdata.go
@@ -10,9 +10,9 @@ import (
 	"crypto/sha512"
 	"encoding/base64"
 	"encoding/binary"
-	"encoding/hex"
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 	"os"
 	"slices"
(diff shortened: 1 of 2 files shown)


charludo (approved):
LGTM. All usecases I could think of seem to still be availble with a roundtrip via `Raw`.

PR #1814, opened 2025-10-01
initdata: add processor package

burgerdev:
This implements the initdata processor as proposed in [1]. It's intended to run to completion before the Kata agent starts, parsing and validating the initdata block device against HOSTDATA/MRCONFIGID.

[1]: https://github.com/kata-containers/kata-containers/issues/11532


initdata-processor/validator/validator.go
```diff
@@ -0,0 +1,69 @@
...
+	digestGetter digestGetter
+}
+
+// New constructs a Validator suitable for the current runtime environment, if available.
+func New() (*Validator, error) {
+	sqp, err := snpclient.GetQuoteProvider()
+	if err == nil {
+		return &Validator{&snpDigestGetter{sqp}}, nil
+	}
+	log.Printf("SNP not supported: %v", err)
```
charludo: Do we need to know this if TDX is supported?

initdata-processor/validator/validator.go
```diff
@@ -0,0 +1,69 @@
...
+	GetDigest() ([]byte, error)
+}
+
+func compareDigests(expected, actual []byte) error {
+	const minDigestLength = 32
+	n := min(len(expected), len(actual))
+	if n < minDigestLength {
+		return fmt.Errorf("%w: expected %x, got %x", errUnexpectedDigestSize, expected, actual)
+	}
+	if slices.Compare(expected[:n], actual[:n]) != 0 {
```
charludo, line 65: Correct me if I'm wrong, but this seems to relate to `adjust_digest` from upstream. There, explicit lengths per platform are given:

```rust
    let required_len = match platform {
        ProtectedPlatform::Tdx => 48,
        ProtectedPlatform::Snp => 32,
        ProtectedPlatform::Cca => 64,
        ProtectedPlatform::Se => 256,
        ProtectedPlatform::NoProtection => digest.len(),
    };
```

would it make sense to do something similar here, for example to prevent 32-sized TDX digests?

Even if not, a comment would be nice, since the only-compare-until-length-n behavior is a bit surprising (at least to me)

internal/initdata/initdata.go
charludo, line 42: Has this changed, or is it just not rebased yet?

burgerdev: The "rebase and merge" action creates a different base commit, so after the other PR was merged Github changed the base automatically, but does not consider the commit on main to be the same as the commit in this branch, so it shows the diff to the common ancestor (I guess) and a merge conflict. Fixed by rebasing.

initdata-processor/validator/validator.go
burgerdev: The `minDigestLength` is actually not strictly necessary, just defense-in-depth:

* The expected digest comes from the report field and is guaranteed to be at least 32 bytes (for the currently supported platforms).
* The actual digest is the output of an initdata `Digest` operation, and thus a SHA-X hash with also at least 32 bytes.
* Whenever the two sizes don't match, we want to compare only the shorter prefix (what else _could_ we compare). 

Note that upstream's `adjust_digest` is only used for _setting_ `HOSTDATA`. 

I'll add a comment explaining this.

initdata-processor/validator/validator.go
burgerdev: Unfortunately, TDX is always supported: https://github.com/google/go-tdx-guest/blob/43efd1ff6ae89a081dbb3cccb6f1fbe6541a0ad6/client/client_linux.go#L129-L131. I'll leave a comment explaining this.

burgerdev: I reworked the order so that we don't log here, but return the joined error.

charludo: > Unfortunately, TDX is always supported

Ah, sneaky.

New version LGTM

initdata-processor/validator/validator.go
charludo: Alright, makes sense. A comment would be nice.

initdata-processor/validator/snp.go
```diff
@@ -0,0 +1,27 @@
...
+
+func (g *snpDigestGetter) GetDigest() ([]byte, error) {
+	reportRaw, err := g.GetRawQuote([64]byte{})
+	if err != nil {
+		return nil, fmt.Errorf("snp: getting raw report: %w", err)
+	}
+	report, err := abi.ReportToProto(reportRaw)
+	if err != nil {
+		return nil, fmt.Errorf("snp: parsing report: %w", err)
+	}
```
katexochen, line 33: Just to be clear, we don't have to verify the report here as it is delivered via a secured channel on both platforms?

initdata-processor/main.go
```diff
@@ -0,0 +1,83 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"slices"
+
+	"github.com/edgelesssys/contrast/initdata-processor/validator"
+	"github.com/edgelesssys/contrast/internal/initdata"
+)
+
+func main() {
```
katexochen, line 20: This should print the preface including the component name and version like every other service.

initdata-processor/validator/snp.go
burgerdev: Yes, let me try to add references.

initdata-processor/validator/validator.go
burgerdev: Comment is already there :)

initdata-processor/main.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+	return nil
+}
+
+func failf(format string, v ...any) {
+	// TODO(burgerdev): this error won't be visible for the runtime!
+	log.Printf(format, v...)
+	// We exit with 0 here so that the Kata agent can spawn and forward errors to the runtime.
+	os.Exit(0)
```
katexochen: It is good practice to only call exit in a top level function. If we always exit 0 anyway, why aren't we just printing the line an using break to stop the loop?

initdata-processor/main.go
```diff
@@ -0,0 +1,83 @@
...
+		if err := os.WriteFile(path, []byte(content), 0o644); err != nil {
+			return fmt.Errorf("writing file %q: %w", path, err)
+		}
+	}
+	return nil
+}
+
+func failf(format string, v ...any) {
+	// TODO(burgerdev): this error won't be visible for the runtime!
+	log.Printf(format, v...)
```
katexochen, line 92: Maybe this should be prefixed with "Error:"?

burgerdev: Thing is, this will not be visible to anybody, unless running a debug image and entering the console quickly enough before the Kata runtime tears it down. Hence the TODO: I'll replace this with something that can propagate the error in a subsequent PR.

charludo (approved):

katexochen (approved):

PR #1816, opened 2025-10-05
dev-docs: update NVIDIA RIM IDs

edgelessci:
Automated updated created by [rim_updates workflow](https://github.com/edgelesssys/contrast/blob/main/.github/workflows/rim_updates.yml).


dev-docs/nvidia/rim-ids.json
```diff
@@ -23,11 +26,13 @@
     "NV_GPU_DRIVER_GH100_570.148.08",
     "NV_GPU_DRIVER_GH100_570.158.01",
     "NV_GPU_DRIVER_GH100_570.172.08",
+    "NV_GPU_DRIVER_GH100_570.195.03",
     "NV_GPU_DRIVER_GH100_570.44",
     "NV_GPU_DRIVER_GH100_570.86.15",
     "NV_GPU_DRIVER_GH100_575.57.08",
     "NV_GPU_DRIVER_GH100_580.65.06",
     "NV_GPU_DRIVER_GH100_580.82.07",
+    "NV_GPU_DRIVER_GH100_580.95.05",
```
katexochen, line 35: @thomasten what was the upgrade process again here? We need to check these versions are supported by NRAS, too, right?

thomasten: In my understanding, NRAS supports it as soon as RIM supports it. (btw, we don't use NRAS in PM anymore and use RIM directly.)

katexochen: So we can bump the driver to the latest version listed by rim (580.95.05) right away?

thomasten: Yes, attestation should work.

PR #1818, opened 2025-10-06
cli: fix combined output yaml

davidweisse:
This changes the `YAMLBytesFromFile` function to accept multiple paths, so we can directly use it to generate the combined YAML output during `generate` when the `--output` flag is set. This also fixes the combined output, which was broken after #1798, and adds a unit test for the `YAMLBytesFromFiles` function.

### DIFF
--- a/internal/kuberesource/reader.go
+++ b/internal/kuberesource/reader.go
@@ -32,15 +32,19 @@ func UnmarshalApplyConfigurations(data []byte) ([]any, error) {
 	return result, nil
 }
 
-// YAMLBytesFromFile reads a k8 YAML file and returns a formatting-preserving encoding.
-func YAMLBytesFromFile(yamlPath string) ([]byte, error) {
-	data, err := os.ReadFile(yamlPath)
-	if err != nil {
-		return nil, fmt.Errorf("reading %s: %w", yamlPath, err)
-	}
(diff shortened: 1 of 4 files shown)


internal/kuberesource/reader.go
```diff
@@ -32,15 +32,19 @@ func UnmarshalApplyConfigurations(data []byte) ([]any, error) {
...
-// YAMLBytesFromFile reads a k8 YAML file and returns a formatting-preserving encoding.
-func YAMLBytesFromFile(yamlPath string) ([]byte, error) {
-	data, err := os.ReadFile(yamlPath)
-	if err != nil {
-		return nil, fmt.Errorf("reading %s: %w", yamlPath, err)
-	}
-	kubeObjs, err := UnmarshalApplyConfigurations(data)
-	if err != nil {
-		return nil, fmt.Errorf("unmarshaling %s: %w", yamlPath, err)
+// YAMLBytesFromFiles reads one or multiple K8s YAML files and returns a formatting-preserving encoding.
```
burgerdev: While we're touching this again: I'm not sure if the docstring here is correct - we're not actually preserving formatting, but rather pretty printing, right? The input could even be pure JSON...

burgerdev (approved):
Thanks for writing the test!

PR #1820, opened 2025-10-07
node-installer: install static runtime

katexochen:
This is a fix for an issue introduced in #1775, where we switched the way we are referencing static packages. Since that point, the node-installer image shipped a dynamically linked runtime, which might not work on clients' host system.


charludo (approved):
Good catch

PR #1822, opened 2025-10-08
manifest: support URI address as SAN

burgerdev:
Before this change, URIs in the manifest were added to the certificate as DNS SANs, which is (a) not allowed and (b) not what the user intended.

This PR changes the CA such that URIs in the policy's SAN field are automatically detected and added to the certificate with the appropriate SAN type. This should allow building integrations for SPIFFE or other authentication schemes with URIs.

### DIFF
--- a/docs/docs/getting-started/deployment.md
+++ b/docs/docs/getting-started/deployment.md
@@ -498,6 +498,29 @@ Add the `frontendIP` to the list of SANs:
      },
 ```
 
+### Adding a URI SAN to the manifest
+
+Some authentication and authorization schemes rely on URI SANs in X.509 certificates.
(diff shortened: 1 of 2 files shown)


docs/docs/getting-started/deployment.md
```diff
@@ -498,6 +498,29 @@ Add the `frontendIP` to the list of SANs:
      },
 ```
 
+### Adding a URI SAN to the manifest
+
```
katexochen, line 502: Could you add a todo that this should be moved to the manifest site once present?

burgerdev: The parent section already has that TODO (line 467).

PR #1824, opened 2025-10-08
e2e: fail tests requiring GPU if sync settings not set

charludo:
E2E tests can specify that they need a GPU. For such cases, the test fails if `SYNC_FIFO_UUID` and `SYNC_ENDPOINT` are unset.

In the `just e2e` recipe, set these envvars before running a gpu test.

### DIFF
--- a/e2e/atls/atls_test.go
+++ b/e2e/atls/atls_test.go
@@ -54,7 +54,7 @@ func TestSNPValidators(t *testing.T) {
 
 	coordinator = kuberesource.AddPortForwarders(coordinator)
 
-	ct.Init(t, coordinator)
+	ct.Init(t, coordinator, platform)
 	require.True(t, t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
 
 	require.True(t, t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
--- a/e2e/internal/contrasttest/contrasttest.go
+++ b/e2e/internal/contrasttest/contrasttest.go
@@ -97,7 +97,7 @@ func New(t *testing.T) *ContrastTest {
 }
 
 // Init patches the given resources for the test environment and makes them available to Generate and Set.
-func (ct *ContrastTest) Init(t *testing.T, resources []any) {
+func (ct *ContrastTest) Init(t *testing.T, resources []any, platform platforms.Platform) {
 	require := require.New(t)
 
 	f, err := os.Open(ct.ImageReplacementsFile)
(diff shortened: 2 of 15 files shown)


e2e/internal/contrasttest/contrasttest.go
```diff
@@ -111,9 +111,12 @@ func (ct *ContrastTest) Init(t *testing.T, resources []any) {
 	// that only one test is using the cluster at a time.
 	var fifo *ksync.Fifo
 	var ticketUUID string
-	if fifoUUID, ok := os.LookupEnv("SYNC_FIFO_UUID"); ok {
+	if withGPU {
```
katexochen: I don't think this is correct, is it? We want to always use the fifo if the env is set. We could fail here if it isn't set and we are on a gpu platform, but this change is not strictly required (and might be wrong for gpu hardware used in the future?).

e2e/atls/atls_test.go
```diff
@@ -54,7 +54,7 @@ func TestSNPValidators(t *testing.T) {
 
 	coordinator = kuberesource.AddPortForwarders(coordinator)
 
-	ct.Init(t, coordinator)
+	ct.Init(t, coordinator, false)
```
katexochen: passing true/false/nil to a function makes the code hard to read on the caller side. You could define aliases in the package defining the function that could be used instead.

e2e/internal/contrasttest/contrasttest.go
charludo: My goal was to go from "if envvars exist, assume GPU and use fifo" to "if GPU is required, ensure envvars exist".

Would you prefer if we checked ` platforms.IsGPU` here and required the envvars to be set in that case? I.e.

```go
if platforms.IsGPU(platform) {
  // require existence of envvars, fail otherwise
}
```

justfile
```diff
@@ -52,6 +52,13 @@ node-installer platform=default_platform:
 e2e target=default_deploy_target platform=default_platform: soft-clean coordinator initializer openssl port-forwarder service-mesh-proxy memdump (node-installer platform)
     #!/usr/bin/env bash
     set -euo pipefail
+    if [[ {{ platform }} == "Metal-QEMU-SNP-GPU" || {{ target }} == "gpu" ]]; then
+        sync_ip=$(kubectl get svc sync -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+        export SYNC_ENDPOINT="http://$sync_ip:8080"
+        sync_uuid=$(kubectl get configmap sync-server-fifo -o jsonpath='{.data.uuid}')
+        export SYNC_FIFO_UUID="$sync_uuid"
+    fi
+
```
katexochen, line 61:
```suggestion
```

justfile
```diff
@@ -52,6 +52,13 @@ node-installer platform=default_platform:
 e2e target=default_deploy_target platform=default_platform: soft-clean coordinator initializer openssl port-forwarder service-mesh-proxy memdump (node-installer platform)
     #!/usr/bin/env bash
     set -euo pipefail
+    if [[ {{ platform }} == "Metal-QEMU-SNP-GPU" || {{ target }} == "gpu" ]]; then
```
katexochen, line 55:
```suggestion
    if [[ {{ platform }} == "Metal-QEMU-SNP-GPU" ]]; then
```
would suggest to go with the same check as in the other location for consistency.

katexochen, line 55: maybe this should be factored into a just function as we are using it in multiple places?

justfile
```diff
@@ -52,6 +52,13 @@ node-installer platform=default_platform:
...
+        sync_ip=$(kubectl get svc sync -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+        export SYNC_ENDPOINT="http://$sync_ip:8080"
+        sync_uuid=$(kubectl get configmap sync-server-fifo -o jsonpath='{.data.uuid}')
+        export SYNC_FIFO_UUID="$sync_uuid"
```
katexochen, line 59: After running the e2e test, the resources will remain until just undeploy is called. We need to ensure the ticket is (only) released at that point, so we need to write it to a file.

katexochen (changes requested):

justfile
charludo: There's a cleanup function used in `contrasttest.go/Init` which releases the ticket. Is this not sufficient?

katexochen: As mentioned, the resources will remain in the cluster, so the ticket shall not be released by the e2e test code but by `undeploy`.

justfile
charludo: I moved the check into a separate function - the actual contents of the if clause don't overlap.

Edit: just saw the newest comment, so my comment is no longer relevant

PR #1833, opened 2025-10-09
kata: upgrade to 3.21, switch to initdata

burgerdev:
This PR updates Kata to version 3.21.

The 3.21 release comes with a major change for confidential VMs: the Kata agent policy is not configured with its own agent method anymore, but is bundled in a structure called `initdata`. For Contrast users, this has the following consequences:

* `contrast verify` now writes the initdata documents instead of the policy files
  - before:
    ```console
    $ tree verify
    verify
    ├── coordinator-root-ca.pem
    ├── manifest.0.json
    └── mesh-ca.pem
    ├── policy.ec26c62dd8ac84ff13085975907d96a2b1638c5c1be803fb34883f86189bf3f3.rego
    ```
  - after:
    ```console
    $ tree verify
    verify
    ├── coordinator-root-ca.pem
    ├── initdata.eae77960e5386c3ec010abb048bf360586a1e64753a4a710180b02f166c7f34e.toml
    ├── manifest.0.json
    └── mesh-ca.pem
    ```
  - the policy is embedded in the initdata document and can be extracted, if needed:
    ```sh
    yq --input-format toml '.data["policy.rego"]'  <initdata.*.toml
* policy errors (as seen, for example, in Kubernetes events) now have a different prefix:
  - before: `agent_policy:117:  allow_anno_key`
  - after: `/run/measured-cfg/policy.rego:117:  allow_anno_key`
* the error message for missing annotations (i.e., resources applied before `contrast generate`) looks slightly different:
  - before: 
    ```
    Warning  FailedCreatePodSandBox  9s (x8 over 94s)   kubelet                  Failed to create pod sandbox: rpc error: code = InvalidArgument desc = failed to create containerd task: failed to create shim task: The policy annotation is missing. Did you forget to run `contrast generate`?: invalid argument 
    ```
  - after: 
    ```
    Warning  FailedCreatePodSandBox  7s                 kubelet                  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: "UpdateInterfaceRequest is blocked by policy: /run/measured-cfg/policy.rego:81: Internal error: no initdata device found": unknown
    ```

### DIFF
--- a/initdata-processor/main.go
+++ b/initdata-processor/main.go
@@ -50,9 +50,7 @@ func main() {
 			break
 		}
 		if err := validator.ValidateDigest(digest); err != nil {
-			// TODO(burgerdev): DANGER!!! this must fail, but I'm returning nil until the digest is initdata, not policy.
-			// failf("Validating initdata digest: %v", err)
-			log.Printf("validation errors:\n%v", err)
+			failf("Validating initdata digest: %v", err)
 			break
 		}
 		i, err := doc.Parse()


cli/cmd/policies.go
```diff
@@ -113,10 +114,18 @@ func policiesFromKubeResources(yamlPaths []string) ([]deployment, error) {
...
 	for _, depl := range policies {
-		if entry, ok := policyHashes[depl.policy.Hash()]; ok {
+		hash, err := depl.initdata.Digest()
+		if err != nil {
+			return nil, fmt.Errorf("digesting initdata for %q: %w", depl.name, err)
+		}
+
+		if err != nil {
+			return nil, fmt.Errorf("encoding initdata for %q: %w", depl.name, err)
+		}
```
charludo: Are you missing a call to an initdata lib func here, or should the second error message be removed?

initdata-processor/main.go
charludo, line 91: Is there something else planned for this? Since it's now just a wrapper around `Printf`

burgerdev: Yes, as the comment implies I'll change this function so that it does propagate the error message to the kata agent, which in turn should forward it to the runtime and k8s. I wanted to keep it out from this PR, as it is already considerably large.

cli/cmd/policies.go
burgerdev: Good catch - looks like a leftover.

packages/by-name/initdata-processor/package.nix
```diff
@@ -1,7 +1,10 @@
 # Copyright 2024 Edgeless Systems GmbH
 # SPDX-License-Identifier: BUSL-1.1
 
-{ lib, buildGoModule }:
+{
+  lib,
+  buildGoModule,
+}:
```
charludo: unrelated diff and not required, please revert

charludo (approved):

packages/by-name/kata/release-tarball/package.nix
```diff
@@ -3,17 +3,31 @@
...
-  hash = "sha256-sEdySOuqf1WUoy+fgQzV4/mZ3zb90lTTY7s43+6oXMM=";
-  stripRoot = false;
+stdenv.mkDerivation {
+  name = "src";
+  src = fetchurl {
+    url = "https://github.com/kata-containers/kata-containers/releases/download/${version}/kata-static-${version}-amd64.tar.zst";
+    hash = "sha256-A8qH1gKcuzoQPWqnTzsg6zn4qLF5Xk4bLY0u4gsr4Ag=";
+  };
+
+  nativeBuildInputs = [ pkgs.zstd ];
```
katexochen: would be cleaner to continue using `fetchzip` and use `nativeBuildInputs`/`postFetch` on that, see https://nixos.org/manual/nixpkgs/unstable/#sec-pkgs-fetchers-fetchzip

packages/by-name/kata/release-tarball/package.nix
```diff
@@ -3,17 +3,31 @@
...
+
+  unpackPhase = ''
+    mkdir -p $out
+    tar --zstd -xvf $src -C $out
+  '';
+
+  dontBuild = true;
+
```
katexochen:
```suggestion
fetchzip {
  url = "https://github.com/kata-containers/kata-containers/releases/download/${version}/kata-static-${version}-amd64.tar.zst";
  hash = "sha256-ktGksilc21qWi3ktg21RZ9I2g4/uOvTk39wRDfRy9pk=";
  stripRoot = false;
  nativeBuildInputs = [ zstd ];
```

packages/by-name/kata/release-tarball/package.nix
```diff
@@ -3,17 +3,31 @@
 
 {
   runtime,
-  fetchzip,
+  pkgs,
```
katexochen: `pkgs` shouldn't usually be used here, as it weakens the callPackage abstraction. Just request `zstd` directly.

cli/cmd/verify.go
```diff
@@ -83,18 +79,12 @@ func runVerify(cmd *cobra.Command, _ []string) error {
...
-			return fmt.Errorf("creating gzip reader: %w", err)
+			return fmt.Errorf("calculating initdata digest: %w", err)
 		}
-		policyDecompressed, err := io.ReadAll(gzipReader)
-		if err != nil {
-			return fmt.Errorf("decompressing policy: %w", err)
-		}
-		gzipReader.Close()
-		filelist[fmt.Sprintf("policy.%s.rego", pHash)] = policyDecompressed
+		filelist[fmt.Sprintf("initdata.%x.toml", digest)] = initdata
```
katexochen, line 87: IMO this is a regression in the user interface. There is really no way I can audit the initdata without further steps:

image

Maybe we could write the policy as additional file und the same initdata hash?

katexochen, line 87: We also need to adapt the dev scripts that work with policies/initdata, specifically extract-policies, maybe other.

burgerdev: You are referring to the policy being a single-line string? How about I fix that instead?

burgerdev: Is it important to do that in this PR?

packages/by-name/kata/release-tarball/package.nix
burgerdev: Pleasant surprise, thanks!

cli/cmd/verify.go
katexochen: That would be even better, yes.

katexochen: Or in a quick follow up.

burgerdev: https://github.com/kata-containers/kata-containers/pull/11944

initdata-processor/main.go
burgerdev: The real error message comes with https://github.com/edgelesssys/contrast/pull/1844.

cli/cmd/verify.go
burgerdev: Backported here: https://github.com/edgelesssys/contrast/pull/1845.

burgerdev: https://github.com/edgelesssys/contrast/pull/1846

PR #1834, opened 2025-10-09
treewide: move fifo interaction into justfile/scripts

katexochen:
Previously, we had this all around, especially there was code for fifo handling in the e2e Go code. However, the e2e test would release the fifo ticket at the end of the run, before the resources were actually cleaned up (and we collected the logs etc.). So there was a window where the previous run had released the ticket, but the resources were still present, and the next run would start up and didn't get the GPU because it was still in use.
Further, this was even more broken when executing  e2e tests from the justfile (where we didn't use the fifo at all) or when setting the skip-undeploy option in CI (where we would release the ticket and keep the resources around).

### DIFF
--- a/.github/workflows/asciinema.yml
+++ b/.github/workflows/asciinema.yml
@@ -42,22 +42,16 @@ jobs:
       - name: Get credentials for CI cluster
         run: |
           just get-credentials
-      - name: Get ticket for the CI cluster
+      - name: Request fifo ticket
         run: |
-          syncIP="$(kubectl get svc sync -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')"
-          fifo="$(kubectl get configmap sync-server-fifo -o jsonpath='{.data.uuid}')"
-          ticket="$(curl -fsSL "${syncIP}:8080/fifo/${fifo}/ticket" | jq -r '.ticket')"
-          echo "SYNC_IP=$syncIP" | tee -a "$GITHUB_ENV"
-          echo "SYNC_FIFO=$fifo" | tee -a "$GITHUB_ENV"
-          echo "SYNC_TICKET=$ticket" | tee -a "$GITHUB_ENV"
-          curl -fsSL "${syncIP}:8080/fifo/${fifo}/wait/$ticket"
+          just request-fifo-ticket
       - name: Update asciinema screencast
         working-directory: tools/asciinema
         run: |
           ./generate-screencasts.sh
-      - name: Mark ticket done
+      - name: Release fifo ticket
         run: |
-          curl -fsSL "${SYNC_IP}:8080/fifo/${SYNC_FIFO}/done/${SYNC_TICKET}"
+          just release-fifo-ticket
       - name: Create PR
         uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
         with:
--- a/justfile
+++ b/justfile
@@ -52,6 +52,9 @@ node-installer platform=default_platform:
 e2e target=default_deploy_target platform=default_platform: soft-clean coordinator initializer openssl port-forwarder service-mesh-proxy memdump (node-installer platform)
     #!/usr/bin/env bash
     set -euo pipefail
+    if [[ {{ platform }} == "Metal-QEMU-SNP-GPU" ]] ; then
+        just request-fifo-ticket 90m
+    fi
     nix shell .#contrast.e2e --command {{ target }}.test -test.v \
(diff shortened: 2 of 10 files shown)


.github/workflows/asciinema.yml
```diff
@@ -42,22 +42,16 @@ jobs:
...
-          echo "SYNC_FIFO=$fifo" | tee -a "$GITHUB_ENV"
-          echo "SYNC_TICKET=$ticket" | tee -a "$GITHUB_ENV"
-          curl -fsSL "${syncIP}:8080/fifo/${fifo}/wait/$ticket"
+          just request-fifo-ticket
       - name: Update asciinema screencast
         working-directory: tools/asciinema
         run: |
           ./generate-screencasts.sh
-      - name: Mark ticket done
+      - name: Release fifo ticket
```
charludo, line 52: Is the `if: always()` plus conditions not needed here?

justfile
```diff
@@ -280,6 +282,26 @@ wait-for-workload target=default_deploy_target:
...
+
+release-fifo-ticket:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if [[ ! -f ./{{ workspace_dir }}/just.sync-ticket ]]; then
+        echo "No sync ticket found, nothing to release."
+    fi
+    ticket=$(cat ./{{ workspace_dir }}/just.sync-ticket)
+    nix run .#scripts.release-sync-ticket ${ticket}
+    rm -f ./{{ workspace_dir }}/just.sync-ticket
```
charludo: How would you feel about additionally aborting `soft-clean` when the ticket file exists? Just to make sure no unreleased ticket is lost.

packages/scripts.nix
```diff
@@ -660,11 +663,17 @@ lib.makeScope pkgs.newScope (scripts: {
...
+      echo "Releasing fifo ticket $ticket" >&2
       sync_ip=$(kubectl get svc sync -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+      echo "Sync server IP: $sync_ip" >&2
       sync_uuid=$(kubectl get configmap sync-server-fifo -o jsonpath='{.data.uuid}')
-      ticket=$1
-      curl -fsSL "$sync_ip:8080/fifo/$sync_uuid/done/$ticket" || true
-      echo "Released ticket $ticket from fifo $sync_uuid" >&2
+      echo "Sync fifo UUID: $sync_uuid" >&2
+      if ! curl -fsSL "$sync_ip:8080/fifo/$sync_uuid/done/$ticket"; then
+        echo "Failed to release fifo $sync_uuid with ticket $ticket" >&2
```
burgerdev, line 673: Should we `exit 1` here?

katexochen: I think the reason we don't is that we assume the ticket was released via timeout or cleanup job instead. The sync server hasn't got any notation of tickets that were valid before but now deleted, you will just get a 404 if you try to release an already released ticket. If we fail here, we have to do the cleanup manually in those case, i.e. delete the ticket file by hand in case the namespace of a developer was cleaned up by the job.

justfile
katexochen: Ah, I forgot to use the release-fifo-ticket target in soft-clean, should make more sense now.

charludo: LGTM

.github/workflows/asciinema.yml
davidweisse, line 1: Just want to mention that this workflow is still based on AKS and won't work.

justfile
```diff
@@ -159,10 +162,9 @@ apply target=default_deploy_target platform=default_platform:
         ;;
         *)
             if [[ {{ platform }} == "Metal-QEMU-SNP-GPU" ]] ; then
-                sync_ticket=$(nix run .#scripts.get-sync-ticket 90m)
-                echo $sync_ticket > ./{{ workspace_dir }}/just.sync-ticket
-                trap 'nix run .#scripts.release-sync-ticket $sync_ticket' ERR
-                kubectl label ns $(cat ./{{ workspace_dir }}/just.namespace) contrast.edgeless.systems/sync-ticket=$sync_ticket --overwrite
+                just request-fifo-ticket 90m
+                trap 'just release-fifo-ticket' ERR
+                kubectl label ns $(cat ./{{ workspace_dir }}/just.namespace) contrast.edgeless.systems/sync-ticket=$(cat ./{{ workspace_dir }}/just.sync-ticket) --overwrite
```
katexochen, line 167: Actually, we should do this for the `just e2e` case, too, but it has to happen in the Go code, as the namespace is only created by the contrasttest library in this case. Should be done in a follow up (@davidweisse).

PR #1835, opened 2025-10-09
internal: use CDI instead of guest-hook for GPU support 

charludo:
With this PR, the way Contrast handles GPU support is changed from the legacy, guest-hook based flow to the CDI method embraced upstream.

For users, no change is required.

### DIFF
--- a/docs/docs/howto/workload-deployment/GPU-configuration.md
+++ b/docs/docs/howto/workload-deployment/GPU-configuration.md
@@ -33,7 +33,7 @@ If the cluster is [configured for GPU usage](../../howto/cluster-setup/bare-meta
 To do so, a CDI annotation needs to be added, specifying to use the `pgpu` (passthrough GPU) mode. The `0` corresponds to the PCI device index.
 
 - For nodes with a single GPU, this value is always `0`.
-- For nodes with multiple GPUs, the value needs to correspond to the device's order as enumerated on the PCI bus. You can identify this order by inspecting the `/var/run/cdi/nvidia.com-pgpu.yaml` file on the specific node.
+- For nodes with multiple GPUs, the value needs to correspond to the device's order as enumerated on the PCI bus. You can identify this order by inspecting the `/etc/cdi/nvidia.com-pgpu.yaml` file on the specific node. On some hosts, and depending on the initial setup, this file might also be located at `/var/run/cdi/nvidia.com-pgpu.yaml`.
 
 This process ensures the correct GPU is allocated to the workload.
 


internal/platforms/platforms.go
```diff
@@ -103,7 +103,7 @@ func DefaultMemoryInMegaBytes(p Platform) int {
 	switch p {
 	case MetalQEMUSNPGPU:
 		// Guest components contribute around 600MiB with GPU enabled.
-		return 1024
+		return 2048
```
katexochen: Why does this change adopting CDI?

packages/nixos/gpu.nix
```diff
@@ -46,74 +46,7 @@ let
...
-      };
-    };
-    nvidia-container-runtime-hook = {
-      path = lib.getExe' pkgs.nvidia-container-toolkit.tools "nvidia-container-runtime-hook";
-      skip-mode-detection = true;
-    };
-    nvidia-ctk.path = lib.getExe' pkgs.nvidia-container-toolkit "nvidia-ctk";
-  };
 in
-
```
katexochen, line 116: keep the empty line

packages/nixos/gpu.nix
```diff
@@ -46,74 +46,7 @@ let
           ]
         );
       });
-
-  # nix-store-mount-hook mounts the VM's nix store into the container.
-  # TODO(burgerdev): only do that for containers that actually get a GPU device.
-  nix-store-mount-hook = pkgs.writeShellApplication {
```
katexochen, line 52: I really would like to understand why this isn't needed anymore..

packages/nixos/gpu.nix
```diff
@@ -151,34 +84,13 @@ in
     systemd.services."nvidia-container-toolkit-cdi-generator".wantedBy = [ "kata-containers.target" ];
 
     hardware.nvidia-container-toolkit.enable = true;
+    hardware.nvidia-container-toolkit.mount-nvidia-docker-1-directories = true;
```
katexochen, line 95: Ah, this is where the magic happens. This defaults to true on unstable, maybe we should bump system.StateVersion from time to time.

internal/platforms/platforms.go
charludo: Ah, sorry, I think that can be reverted. From the kata logs:
```
level=warning msg="VM memory (1024MB) smaller than image \"/opt/edgeless/contrast-cc-metal-qemu-snp-gpu-2f01066d/share/kata-containers.img\" size (2016MB)"
```
I changed this when trying to debug the setup. But looks like the warning is indeed just a warning.

burgerdev: That warning is misguided, let's try to remove it upstream.

packages/nixos/gpu.nix
charludo: Would you rather have me do that here?

packages/nixos/gpu.nix
charludo: Good thing we don't, otherwise we'd have to keep hook support 😅

internal/platforms/platforms.go
charludo: Seems they are aware, but disagree:
```go
// Images do not need to be fully read
// into memory, but it would be highly
// unusual to have an image larger
// than the amount of memory assigned
// to the VM.
kataUtilsLogger.Warn(msg)
```

burgerdev: Afaict that's only true for initrds, not for images that are mounted as block devices.

packages/nixos/gpu.nix
katexochen: No, out of scope of this PR.

internal/platforms/platforms.go
charludo: Sorry, should have copied the full context. For initrds, an error is returned just above this snippet. 

I'd be happy to open a PR removing this anyways; let me know if you think it's worth the effort to get it approved ^^

packages/nixos/gpu.nix
```diff
@@ -82,7 +82,14 @@ in
...
     # legacy mode as of now, but will be once we switch to CDI.
-    systemd.services."nvidia-container-toolkit-cdi-generator".wantedBy = [ "kata-containers.target" ];
+    systemd.services."nvidia-container-toolkit-cdi-generator" = {
+      wantedBy = [ "kata-containers.target" ];
+      before = [ "kata-agent.service" ];
+    };
+    systemd.services."kata-agent" = {
+      after = [ "nvidia-container-toolkit-cdi-generator.service" ];
+      requires = [ "nvidia-container-toolkit-cdi-generator.service" ];
+    };
```
burgerdev, line 92: How does this interact with the attributes in `kata.nix`?

burgerdev (approved):
Thanks!

packages/nixos/gpu.nix
charludo: The lists are additive, entries from both files get combined.

charludo: (well. additive is probably the wrong word. But duplicate list definitions in separate nix modules get merged)

katexochen (approved):

PR #1837, opened 2025-10-10
docs: dedup 'connect to coordinator' sections

charludo:
Removed three instances of the same section, moved it to a separate file/step.

### DIFF
--- a/docs/sidebars.js
+++ b/docs/sidebars.js
@@ -149,19 +149,19 @@ const sidebars = {
           id: "howto/manifest-update",
         },
         {
-          type: 'doc',
-          label: 'Coordinator high-availability',
-          id: 'howto/coordinator-ha',
+          type: "doc",
+          label: "Coordinator high-availability",
+          id: "howto/coordinator-ha",
         },
         {
-          type: 'doc',
-          label: 'Vault',
-          id: 'howto/vault'
+          type: "doc",
+          label: "Vault",
+          id: "howto/vault",
         },
         {
           type: "doc",
           label: "Secure image store",
-          id: "howto/secure-image-store"
+          id: "howto/secure-image-store",
         },
         {
           type: "doc",


docs/docs/howto/workload-deployment/connect-to-coordinator.md
```diff
@@ -0,0 +1,56 @@
+# Connect to the Contrast Coordinator
+
+This step describes how to connect to a running Contrast Coordinator.
+
+## Applicability
+
+This step is necessary for some of the following steps, particularly for [setting the manifest](./set-manifest.md) and for [verifying the deployment](./deployment-verification.md).
```
katexochen: Hm this could go a bit more into detail why this is an issue and what we are trying to archive here?

docs/docs/howto/workload-deployment/connect-to-coordinator.md
```diff
@@ -0,0 +1,56 @@
...
+## How-to
+
+The released Coordinator resource includes a `LoadBalancer` definition we can use.
+
+```sh
+coordinator=$(kubectl get svc coordinator -o=jsonpath='{.status.loadBalancer.ingress[0].ip}')
+```
+
+:::info[Port-forwarding of Confidential Containers]
+
```
katexochen: right now this how-to has a single sentence in its main body, which looks a bit weird. Maybe we move the port forwarding into a section instead of a box and describe a bit when you would use which option?

docs/sidebars.js
```diff
@@ -149,19 +149,19 @@ const sidebars = {
...
-          id: 'howto/vault'
+          type: "doc",
+          label: "Vault",
+          id: "howto/vault",
         },
         {
           type: "doc",
           label: "Secure image store",
-          id: "howto/secure-image-store"
+          id: "howto/secure-image-store",
```
katexochen, line 164: unrelated change?

charludo: Yes... Unrelated but not wrong, I'd say 😅 Mind if I put this in a separate commit?

PR #1840, opened 2025-10-13
docs: add 'immutable deployment' howto

charludo:


docs/docs/howto/immutable-deployments.md
```diff
@@ -0,0 +1,75 @@
...
+2. [Install CLI](./install-cli.md)
+3. [Deploy the Contrast runtime](./workload-deployment/runtime-deployment.md)
+4. [Add Coordinator to resources](./workload-deployment/add-coordinator.md)
+5. [Prepare deployment files](./workload-deployment/deployment-file-preparation.md)
+6. [Generate annotations and manifest](./workload-deployment/generate-annotations.md)
+
+## How-To
+
+After running the [`contrast generate`](./workload-deployment/generate-annotations.md) command, the manifest contains fields `WorkloadOwnerKeyDigest` and `SeedshareOwnerPubKeys`.
+Deleting these fields before calling [`contrast set`](./workload-deployment/set-manifest.md) will render the deployment immutable.
```
katexochen: There is a `--disable-updates` flag on `contrast generate`.

charludo: Yes, I mentioned it below. But that only affects the presence of `WorkloadOwnerKeyDigest`, so mentioning it here in the "summary" would complicate the text a bit, to no real benefit

docs/docs/howto/immutable-deployments.md
```diff
@@ -0,0 +1,75 @@
+# Immutable deployments with Contrast
+
+This section provides guidance on how to setup an immutable deployment with Contrast, that is, a deployment which can't be changed after the initial setup.
```
burgerdev:
```suggestion
This section explains how to setup a Contrast deployment that can't be changed after the initial setup.
```

docs/docs/howto/immutable-deployments.md
burgerdev: Imho the explanation should go into https://docs.edgeless.systems/contrast/architecture/components/coordinator#the-manifest (until we have a dedicated page for manifest ops). The how-to could then reference that for details and just explain how to run generate and the risks posed by this configuration.

docs/docs/howto/immutable-deployments.md
```diff
@@ -0,0 +1,75 @@
...
+    "<remove this key to prevent state decryption and coordinator recovery>"
+  ]
+}
+```
+
+### Preventing manifest updates
+
+To prevent manifest updates, delete the `WorkloadOwnerKeyDigest`.
+Now [set the manifest](./workload-deployment/set-manifest.md) through `contrast set`.
+Once completed, the Contrast Coordinator will no longer accept the workload owner key for future manifest changes.
```
burgerdev: The seedshare owner can still force manifest changes (through recovery), at least if they have access to the Coordinator's ConfigMaps.

docs/docs/howto/immutable-deployments.md
```diff
@@ -0,0 +1,75 @@
...
+
+**It's highly unlikely that you want to use this feature in conjunction with the encrypted persistent storage feature.**
+
+:::
+
+### Verifying immutability
+
+[Deployment verification](./workload-deployment/deployment-verification.md) works as usual.
+The `contrast verify` step for a manifest with removed `WorkloadOwnerKeyDigest` and/or `SeedshareOwnerPubKeys` will only succeed if the manifest set on the Coordinator is also missing the corresponding keys,
+thereby guaranteeing no party can update and/or recover the deployment and its state.
```
burgerdev: We should mention that all this needs to happen before the first `set`, because it won't be possible (or safe) afterwards, and verification needs to check the history, too.

https://github.com/edgelesssys/contrast/blob/8a86dae81f7c0438fd621c51af191bf1803b9b55/coordinator/internal/userapi/userapi.go#L104-L107

docs/docs/howto/immutable-deployments.md
charludo: > Imho the explanation should go into https://docs.edgeless.systems/contrast/architecture/components/coordinator#the-manifest (until we have a dedicated page for manifest ops). The how-to could then reference that for details and just explain how to run generate and the risks posed by this configuration.

What part exactly? Just the `Deleting these fields before calling [`contrast set`](./workload-deployment/set-manifest.md) will render the deployment immutable.` sentence (modified)?

burgerdev: This entire subsection, essentially.

What we should do here:

* Tell people to use `--disable-updates`
* Explain that this removes the workload/seedshare owner keys

What we should do over there:

* Explain what the keys are for and what it means to remove them

docs/docs/howto/immutable-deployments.md
```diff
@@ -0,0 +1,68 @@
...
+2. [Install CLI](./install-cli.md)
+3. [Deploy the Contrast runtime](./workload-deployment/runtime-deployment.md)
+4. [Add Coordinator to resources](./workload-deployment/add-coordinator.md)
+5. [Prepare deployment files](./workload-deployment/deployment-file-preparation.md)
+6. [Generate annotations and manifest](./workload-deployment/generate-annotations.md)
+
+## How-To
+
+After running the [`contrast generate`](./workload-deployment/generate-annotations.md) command, the manifest contains fields `WorkloadOwnerKeyDigest` and `SeedshareOwnerPubKeys`.
+Deleting these fields before calling [`contrast set`](./workload-deployment/set-manifest.md) for the first time will render the deployment immutable and/or make Coordinator recovery impossible.
```
burgerdev:
```suggestion
Deleting these fields before calling [`contrast set`](./workload-deployment/set-manifest.md) for the first time will render the deployment immutable and make Coordinator recovery impossible.
```
"Deleting these fields" does both.

docs/docs/howto/immutable-deployments.md
```diff
@@ -0,0 +1,68 @@
...
+:::
+
+### Verifying immutability
+
+[Deployment verification](./workload-deployment/deployment-verification.md) works as usual.
+The `contrast verify` step for a manifest with removed `WorkloadOwnerKeyDigest` and/or `SeedshareOwnerPubKeys` will only succeed if the manifest set on the Coordinator is also missing the corresponding keys,
+thereby guaranteeing no party can update and/or recover the deployment and its state.
+
+Additionally, deletion of the `WorkloadOwnerKeyDigest` and `SeedshareOwnerPubKeys` needs to occur before the first use of `contrast set`.
+Setting the manifest and verifying the manifest history will both fail if these keys have been changed.
```
burgerdev: * Only set will fail, verify is not relevant for this detail.
* Workload owner keys can be changed after the first `set`, but we should still recommend to start with none if immutability is desired.

tools/vale/styles/config/vocabularies/edgeless/accept.txt
```diff
@@ -136,6 +134,7 @@ runtime
 runtimes?
 safeguardable
 SBOM
+seedshare
```
burgerdev: I believe we're spelling it "seed share" in other places.

docs/docs/howto/immutable-deployments.md
charludo: Thanks. Since it then no longer really makes sense to have this in the verification section, I split the explanation by key type and moved them into their respective sections.

docs/docs/howto/immutable-deployments.md
```diff
@@ -0,0 +1,67 @@
...
+
+**It's highly unlikely that you want to use this feature in conjunction with the encrypted persistent storage feature.**
+
+:::
+
+### Verifying immutability
+
+[Deployment verification](./workload-deployment/deployment-verification.md) works as usual.
+The `contrast verify` step for a manifest with removed `WorkloadOwnerKeyDigest` and/or `SeedshareOwnerPubKeys` will only succeed if the manifest set on the Coordinator is also missing the corresponding keys,
+thereby guaranteeing no party can update and/or recover the deployment and its state.
```
burgerdev:
```suggestion
Verifying that no manifest in the manifest history contained workload owner keys or seed share owner keys guarantees that no party can update and/or recover the deployment and its state.
```

PR #1842, opened 2025-10-14
cli/generate: ignore unsupported resources

davidweisse:
This refactors how the `generate` command handles the input YAML files, so that resources not supported by the `genpolicy` tool are ignored during processing. For this, the CLI first reads and unmarshals all resources containing at least one pod controller with the `contrast-cc*` runtime class and saves them in a map `map[string][]*unstructured.Unstructured`, containing all objects per filename. This includes custom resource definitions, cannot be decoded into apply configurations. This can later be used to write the resources back to their original position. To avoid unnecessary marshalling/unmarshalling of YAML files, we now keep all unstructured objects in memory.

While the resources are being read from the files, all ConfigMaps/Secrets are gathered and written to a temporary file. This is used as extra input for `genpolicy`. Further, `genpolicy` does not work on the full YAML anymore, but rather on one `contrast-cc` pod controller at a time, provided via `stdin`. The annotation is then read from `stdout` and added to the apply configuration manually. By doing this, `genpolicy` only processes the resources it needs, while all other resources (possibly not supported by `genpolicy`, but also not necessary for policy generation) are ignored.

While working with the `kubeapi`/`kuberesource/reader` packages, I noticed that they provide very similar functionality, so I moved the `kubeapi` into the `kuberesource` package for consistency.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -20,11 +20,14 @@ import (
 
 	"github.com/edgelesssys/contrast/cli/genpolicy"
 	"github.com/edgelesssys/contrast/cli/verifier"
+	"github.com/edgelesssys/contrast/internal/initdata"
 	"github.com/edgelesssys/contrast/internal/kuberesource"
 	"github.com/edgelesssys/contrast/internal/manifest"
 	"github.com/edgelesssys/contrast/internal/platforms"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	applyappsv1 "k8s.io/client-go/applyconfigurations/apps/v1"
 	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	applymetav1 "k8s.io/client-go/applyconfigurations/meta/v1"
 
 	"github.com/spf13/cobra"
 )
@@ -96,21 +99,28 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	paths, cmPaths, err := findGenerateTargets(args, log)
+	paths, err := findYamlFiles(args)
 	if err != nil {
 		return err
 	}
-	if flags.outputFile != "" {
-		tmpDir, newPaths, err := getTmpPaths(paths)
-		if err != nil {
-			return fmt.Errorf("get temporary paths: %w", err)
-		}
-		defer os.RemoveAll(tmpDir)
-		paths = newPaths
+
+	extraFile, err := os.CreateTemp("", "contrast-generate-extra-*.yml")
+	if err != nil {
+		return fmt.Errorf("create temp file for configmaps/secrets: %w", err)
+	}
+	defer os.Remove(extraFile.Name())
+
+	fileMap, err := extractTargets(paths, extraFile, log)
+	closeErr := extraFile.Close()
+	if err != nil {
+		return fmt.Errorf("extracting targets: %w", err)
+	}
+	if closeErr != nil {
+		return fmt.Errorf("closing temp file for configmaps/secrets: %w", closeErr)
 	}
 
 	verifiers := verifier.AllVerifiersBeforeGenerate()
-	if err := runVerifiers(paths, verifiers); err != nil {
+	if err := runVerifiers(fileMap, verifiers); err != nil {
 		return err
 	}
 
@@ -142,28 +152,18 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("get runtime handler: %w", err)
 	}
 
-	if err := patchTargets(paths, flags.imageReplacementsFile, runtimeHandler, flags.skipInitializer, flags.skipServiceMesh, flags.skipImageStore, log); err != nil {
+	if err := patchTargets(fileMap, flags.imageReplacementsFile, runtimeHandler, flags.skipInitializer, flags.skipServiceMesh, flags.skipImageStore); err != nil {
 		return fmt.Errorf("patch targets: %w", err)
 	}
 	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Patched targets")
 
-	if err := generatePolicies(cmd.Context(), flags, paths, cmPaths, log); err != nil {
+	if err := generatePolicies(cmd.Context(), flags, fileMap, extraFile.Name(), log); err != nil {
 		return fmt.Errorf("generate policies: %w", err)
-	}
-
-	if flags.outputFile != "" {
-		combinedYAML, err := kuberesource.YAMLBytesFromFiles(paths...)
-		if err != nil {
-			return fmt.Errorf("get combined YAML: %w", err)
-		}
-		if err := os.WriteFile(flags.outputFile, combinedYAML, 0o644); err != nil {
-			return fmt.Errorf("write output file: %w", err)
-		}
 	}
 
 	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Generated workload policy annotations")
 
-	policies, err := policiesFromKubeResources(paths)
+	policies, err := policiesFromKubeResources(fileMap)
 	if err != nil {
 		return fmt.Errorf("find kube resources with policy: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


internal/kuberesource/mutators.go
```diff
@@ -528,7 +522,7 @@ func MapPodSpecWithMeta(
 	f func(
 		meta *applymetav1.ObjectMetaApplyConfiguration,
 		spec *applycorev1.PodSpecApplyConfiguration,
-	) (*applymetav1.ObjectMetaApplyConfiguration, *applycorev1.PodSpecApplyConfiguration),
+	),
```
burgerdev: We briefly talked about the alternative: actually use the return values to update the meta/spec. That would allow setting an existing pointer to nil, which the version without return value does not allow. Do you still think we should be removing the return value?

internal/kubeapi/kubeapi_test.go
burgerdev, line 1: Shouldn't we keep the test?

internal/kuberesource/mutators.go
```diff
@@ -515,14 +515,16 @@ func PatchCoordinatorMetrics(resources []any) []any {
 	return resources
 }
 
-// MapPodSpecWithMeta applies a function to a PodSpec in a Kubernetes resource,
+// MapPodSpecWithEnsuredMeta applies a function to a PodSpec in a Kubernetes resource,
 // and its corresponding object metadata.
-func MapPodSpecWithMeta(
+// If ensureMetaExists is set, it adds the metadata field if it does not exist.
```
burgerdev: This seems to be a niche use case - why can't the caller check for nil?

cli/cmd/policies.go
```diff
@@ -37,53 +32,28 @@ func policiesFromKubeResources(yamlPaths []string) ([]deployment, error) {
...
-		case *kubeapi.ReplicationController:
-			name = obj.Name
-			annotation = obj.Spec.Template.Annotations[initdata.InitdataAnnotationKey]
-			role = manifest.Role(obj.Spec.Template.Annotations[contrastRoleAnnotationKey])
-			workloadSecretID = obj.Spec.Template.Annotations[workloadSecretIDAnnotationKey]
-		}
+		kuberesource.MapPodSpecWithMeta(objAny, func(meta *applymetav1.ObjectMetaApplyConfiguration, spec *applycorev1.PodSpecApplyConfiguration) (*applymetav1.ObjectMetaApplyConfiguration, *applycorev1.PodSpecApplyConfiguration) {
+			annotation = meta.Annotations[initdata.InitdataAnnotationKey]
+			role = manifest.Role(meta.Annotations[contrastRoleAnnotationKey])
+			workloadSecretID = meta.Annotations[workloadSecretIDAnnotationKey]
```
burgerdev: Can we be certain that `meta != nil` at this point? Looks like we did not check before, but now it would be easy to add the check.

internal/kuberesource/reader.go
```diff
@@ -4,32 +4,67 @@
...
+func UnmarshalUnstructuredK8SResource(data []byte) ([]*unstructured.Unstructured, error) {
+	documentsData, err := splitYAML(data)
+	if err != nil {
+		return nil, fmt.Errorf("splitting YAML into multiple documents: %w", err)
+	}
+
+	var objects []*unstructured.Unstructured
+	for idx, documentData := range documentsData {
+		obj := &unstructured.Unstructured{}
+		decoder := serializer.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
```
burgerdev: Could we create the decoder outside the loop?

cli/verifier/versions_match.go
```diff
@@ -34,12 +34,16 @@ func (v *VersionsMatch) Verify(toVerify any) error {
...
+		if spec == nil || spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
 			return meta, spec
 		}
 
 		for _, container := range spec.Containers {
-			if !needsImageVersionCheck(*container.Name, meta.Annotations["contrast.edgeless.systems/pod-role"]) {
+			var podRole string
+			if meta != nil && meta.Annotations != nil {
+				podRole = meta.Annotations["contrast.edgeless.systems/pod-role"]
+			}
```
burgerdev: Could be done outside the loop.

internal/kuberesource/wrappers.go
```diff
@@ -31,15 +33,29 @@ func (t *PodTemplate) GetObjectMeta() *applymetav1.ObjectMetaApplyConfiguration
 	if t.PodTemplateSpecApplyConfiguration != nil && t.ObjectMetaApplyConfiguration != nil {
 		return t.ObjectMetaApplyConfiguration
 	}
-	return &applymetav1.ObjectMetaApplyConfiguration{}
+	return nil
```
burgerdev, line 36: Nit: the second clause in the if statement can be removed.

internal/kuberesource/wrappers.go
```diff
@@ -31,15 +33,29 @@ func (t *PodTemplate) GetObjectMeta() *applymetav1.ObjectMetaApplyConfiguration
...
+}
+
+// SetObjectMeta sets the ObjectMeta of the Pod template.
+func (t *PodTemplate) SetObjectMeta(meta *applymetav1.ObjectMetaApplyConfiguration) {
+	if t.PodTemplateSpecApplyConfiguration != nil {
+		t.ObjectMetaApplyConfiguration = meta
+	}
```
burgerdev: It's a bit unexpected that `SetObjectMeta` does nothing if the wrapped template is nil. Maybe we should initialize it? Applies to other methods below.

cli/cmd/generate.go
```diff
@@ -96,21 +99,25 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
-			return fmt.Errorf("get temporary paths: %w", err)
-		}
-		defer os.RemoveAll(tmpDir)
-		paths = newPaths
+
+	extraFile, err := os.CreateTemp("", "contrast-generate-extra-*.yml")
+	if err != nil {
+		return fmt.Errorf("create temp file for configmaps/secrets: %w", err)
+	}
+	defer extraFile.Close()
```
burgerdev: Doesn't this need to be closed after `extractTargets`, if only to force flushing?

cli/cmd/generate.go
```diff
@@ -223,29 +220,24 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
-		}
-		for _, v := range verifiers {
-			for _, r := range resources {
-				if err := v.Verify(r); err != nil {
-					findings = errors.Join(findings, fmt.Errorf("failed to verify YAML %q: %w", path, err))
-				}
+	for _, v := range verifiers {
+		for _, r := range resources {
+			if err := v.Verify(r); err != nil {
+				path, name := getPathAndName(r, fileMap)
```
burgerdev: Why are we not iterating over `fileMap` directly?

cli/cmd/generate.go
```diff
@@ -306,35 +292,54 @@ pathLoop:
...
-		if err != nil {
-			logger.Warn("read file", "path", path, "err", err)
-			continue
-		}
-		if !bytes.Contains(data, []byte(runtimeClassNamePrefix)) {
-			logger.Info("Ignoring non-CoCo runtime", "className", runtimeClassNamePrefix, "path", path)
-			continue
+func isCCWorkload(resource any) (ret bool) {
+	kuberesource.MapPodSpec(resource, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+		if spec.RuntimeClassName != nil && strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
```
burgerdev: Need to check for `spec == nil`.

cli/cmd/generate.go
```diff
@@ -382,77 +415,51 @@ func patchTargets(paths []string, imageReplacementsFile, runtimeHandler string,
...
-			continue
-		}
-		_, err := kuberesource.AddServiceMesh(resource, kuberesource.ServiceMeshProxy())
-		if err != nil {
-			return err
-		}
+func injectServiceMesh(resource any) error {
+	r, ok := resource.(*applyappsv1.StatefulSetApplyConfiguration)
+	if ok && r.Spec != nil && r.Spec.Template != nil && r.Spec.Template.ObjectMetaApplyConfiguration != nil && r.Spec.Template.Annotations != nil &&
+		r.Spec.Template.Annotations[contrastRoleAnnotationKey] == "coordinator" {
```
burgerdev: While you're touching this, please use the constant `manifest.RoleCoordinator`.

cli/genpolicy/genpolicy_test.go
```diff
@@ -43,6 +46,10 @@ while [ $# -gt 0 ]; do
   shift
 done
 
+while IFS= read -r line; do
+  echo "$line" >>"stdin.yaml"
+done < /dev/stdin
```
burgerdev: Could this just be `cat >"stdin.yaml"`?

internal/kuberesource/wrappers.go
davidweisse: If the wrapped template is nil, we can initialize it here. However, this is not represented in the parent K8s object if the `PodTemplate` is constructed via a `PodTemplate()` function of, e.g., a `Deployment`. Rather, we can change the `PodTemplate()` functions to return a `PodSpecAccessor` interface. If the return value is nil, it is not processed by `MapPodSpec`. This should be intended, since `MapPodSpec` does not make sense if the resource has no pod template.

cli/cmd/generate.go
```diff
@@ -224,39 +223,61 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	fmt.Fprintf(cmd.OutOrStdout(), "✔️ Updated manifest %s\n", flags.manifestPath)
 
 	verifiers = verifier.AllVerifiersAfterGenerate()
-	if err := runVerifiers(paths, verifiers); err != nil {
+	if err := runVerifiers(fileMap, verifiers); err != nil {
 		return err
 	}
 
+	if err = writeOutputFiles(fileMap, flags.outputFile); err != nil {
```
burgerdev:
```suggestion
	if err := writeOutputFiles(fileMap, flags.outputFile); err != nil {
```

cli/cmd/generate.go
```diff
@@ -224,39 +223,61 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
-		}
-		resources, err := kuberesource.UnmarshalApplyConfigurations(fileContent)
-		if err != nil {
-			return fmt.Errorf("parsing file %q to run generate on: %w", path, err)
-		}
-		for _, v := range verifiers {
-			for _, r := range resources {
-				if err := v.Verify(r); err != nil {
-					findings = errors.Join(findings, fmt.Errorf("failed to verify YAML %q: %w", path, err))
+func mapCCWorkloads(fileMap map[string][]*unstructured.Unstructured, f func(res any, path string, idx int) error) error {
```
burgerdev, line 240: Please document the function.

cli/cmd/generate.go
```diff
@@ -275,67 +296,64 @@ func findGenerateTargets(args []string, logger *slog.Logger) ([]string, []string
...
-			continue
-		}
-		filtered = append(filtered, path)
+	extraData, err := kuberesource.EncodeUnstructured(extraResources)
+	if err != nil {
+		return nil, fmt.Errorf("encoding configmaps/secrets: %w", err)
 	}
-	return filtered
+	_, err = configFile.Write(extraData)
+	if err != nil {
```
burgerdev:
```suggestion
	if _, err := configFile.Write(extraData); err != nil {
```

cli/genpolicy/genpolicy.go
```diff
@@ -53,18 +52,22 @@ func New(rulesPath, settingsPath, cachePath string, bin []byte) (*Runner, error)
 // Run runs the tool on the given yaml.
 //
 // Run can be called more than once.
-func (r *Runner) Run(ctx context.Context, yamlPath string, cmPath []string, logger *slog.Logger) error {
+func (r *Runner) Run(ctx context.Context, res any, extraPath string, logger *slog.Logger) (string, error) {
```
burgerdev, line 55: Output should be mentioned in the docstring.

PR #1844, opened 2025-10-15
initdata: forward errors through policy

burgerdev:
This PR ensures that we see initdata processing errors on the host side.

If an error occurs, the initdata-processor writes a special policy to `/run/measured-cfg/policy.rego`, which answers every runtime request with the error message observed inside the VM. These errors are then visible on the host, for example in k8s events:

```
  Warning  FailedCreatePodSandBox  14s                kubelet                  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: "UpdateInterfaceRequest is blocked by policy: /run/measured-cfg/policy.rego:81: Internal error: handling initdata: writing file \"/run/measured-cfg/foo/bar.json\": open /run/measured-cfg/foo/bar.json: no such file or directory": unknown
```

(triggered by an initdata file with a `/` in the filename)


initdata-processor/policy/policy.go
```diff
@@ -0,0 +1,56 @@
...
+// DenyWithMessage returns a Rego policy that answers all Kata agent requests with the given message.
+func DenyWithMessage(format string, params ...any) []byte {
+	buf := &bytes.Buffer{}
+	buf.Write(denyWithMessagePolicy)
+	buf.WriteString("\n\nmessage := ")
+	fmt.Fprint(buf, regoEscape(fmt.Sprintf(format, params...)))
+	return buf.Bytes()
+}
+
+// regoEscape returns a double-quoted Rego literal for s.
```
katexochen, line 26: Is there any source for this that states who strings should be escaped in rego?

Another way might be to call json.Marshal on the string?

packages/by-name/kata/genpolicy/package.nix
```diff
@@ -114,6 +114,41 @@ rustPlatform.buildRustPackage rec {
...
+            # implementation of print_message (`message := "foo"` needs to be appended by the user)
+            END { print "\nprint_message {\n  print(\"Internal error:\", message)\n  false\n}" }
+            ' \
+            $src/src/libs/protocols/protos/agent.proto >deny-with-message.rego
+      '';
+      installPhase = ''
+        runHook preInstall
+        install -D deny-with-message.rego $out/genpolicy-rules.rego
+        runHook postInstall
+      '';
```
katexochen: Why is this part of the genpolicy package?

And what is the relationship to deny-with-message.rego in initdata-processor/policy/assets? Looks like this is duplicated.

burgerdev: I thought it would make sense to put it next to the existing policy-generating derivations. Maybe it doesn't? `deny-with-message.rego` is supposed to overwrite `initdata-processor/policy/assets/deny-with-message.rego` (which I thought needs to be checked in anyway, like [genpolicy settings](https://github.com/edgelesssys/contrast/blob/5497a7d6f1a902cced5ce87bf31e3d62e6cedebb/cli/genpolicy/assets/genpolicy-settings-kata.json)). Should I put it into the initdata-processor Nix package?

initdata-processor/policy/policy.go
burgerdev: Thanks, that's a good suggestion. I did not realize that the spec actually says strings are either

1. JSON strings
2. Raw strings 
   ```ebnf
   raw-string = "`" { CHAR-"\`" } "`"
   (* CHAR   Unicode character *)
   ```

I don't quite trust the raw-string definition, but JSON seems straightforward - let me change that.

packages/by-name/kata/genpolicy/package.nix
katexochen: Yes, I think I'd just include it in the initdata-processor package via src. The overwrite is confusing.

burgerdev: As discussed elsewhere, I'm using a structure like this:

* a pseudo-file `initdata-processor/policy/assets/deny-with-message.rego` in tree, just to not trip up the linter
* a derivation that produces the `deny-with-message` policy directly from the list of messages in `agent.proto`
* the file is overwritten in pre-patch, similar to how we include policy settings etc. in package contrast

PR #1853, opened 2025-10-20
imagepuller: use InsecureConfig, if available

charludo:
First step towards implementing registry auth. If available, reads the `InsecureConfig` toml stored in the hardcoded location. `AuthTransportFor(imageRef)` constructs the transport/authenticator for the given image's registry. Falls back to anonymous, unproxied, default-certs config.

### DIFF
--- a/imagepuller/internal/auth/auth.go
+++ b/imagepuller/internal/auth/auth.go
@@ -1,0 +1,109 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package auth
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"os"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	"github.com/edgelesssys/contrast/imagepuller/internal/api"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+)
+
+// Config represents the imagepuller's registry authentication configurations.
+type Config struct {
+	Registries map[string]Registry `toml:"registries"`
+	ExtraEnv   map[string]string   `toml:"extra-env"`
+}
+
+// Registry represents authentication configuration for a single registry.
+type Registry struct {
+	authn.AuthConfig
+	CACerts            string `toml:"ca-certs"`
+	InsecureSkipVerify bool   `toml:"insecure-skip-verify"`
+}
+
+// ReadInsecureConfig reads the auth config from the specified TOML file.
+// No integrity checks are performed on the device from which the initdata-processor originally read this config.
+// An attacker with k8s admin privileges could thus change the config.
+func ReadInsecureConfig(path string, log *slog.Logger) (*Config, error) {
+	data, err := os.ReadFile(path)
+	if errors.Is(err, os.ErrNotExist) {
+		log.Info("Imagepuller auth config file does not exist. Authenticated pulls are not available")
+		return &Config{}, nil
+	} else if err != nil {
+		return nil, fmt.Errorf("reading insecure config file %q: %w", api.InsecureConfigPath, err)
+	}
+
+	var cfg Config
+	if err := toml.Unmarshal(data, &cfg); err != nil {
+		return nil, fmt.Errorf("parsing insecure config TOML: %w", err)
+	}
+	log.Info("Found and parsed imagepuller auth config")
+
+	return &cfg, nil
+}
+
+var errUnparseableRef = errors.New("could not parse image ref")
+
+// AuthTransportFor constructs the appropriate http.Transport and authn.Authenticator for the given image's registry.
+func (c *Config) AuthTransportFor(imageRef string) (*authn.Authenticator, *http.Transport, error) {
+	// Note: this does no check image pinning.
+	ref, err := name.ParseReference(imageRef)
+	if err != nil {
+		return nil, nil, fmt.Errorf("%w: %w", errUnparseableRef, err)
+	}
+	registry := c.registryFor(ref.Context().RegistryStr())
+
+	authenticator := authn.Anonymous
+	if registry.AuthConfig != (authn.AuthConfig{}) {
+		authenticator = authn.FromConfig(registry.AuthConfig)
+	}
+
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: registry.InsecureSkipVerify},
+		Proxy:           http.ProxyFromEnvironment,
+	}
+	if registry.CACerts != "" {
+		certpool := x509.NewCertPool()
+		certpool.AppendCertsFromPEM([]byte(registry.CACerts))
+		transport.TLSClientConfig.RootCAs = certpool
+	}
+
+	return &authenticator, transport, nil
+}
+
+// ApplyEnvVars applies the envvar-based proxy configuration in ExtraEnv.
+func (c *Config) ApplyEnvVars() {
+	allowedEnvVars := []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+	for _, env := range allowedEnvVars {
+		if value, ok := c.ExtraEnv[env]; ok {
+			os.Setenv(env, value)
+		}
+	}
+}
+
+// registryFor returns the registry, if any, for the given registry name.
+func (c *Config) registryFor(name string) Registry {
+	var registry Registry
+	maxMatchingLabels := -1 // "." has 0 matching labels.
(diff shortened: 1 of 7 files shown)


charludo:
@burgerdev this is ready for review now. One thing I wasn't sure about: should auth configs also match glob patterns? I.e. a user might specify a single auth config for `*.example.com`?

I don't know how much practical sense this would make. Implementation-wise, the first match should probably just be selected. On the upside, this would make combining `Auths` and `CA` into a single map trivial.

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,99 @@
...
+	Registries map[string]Registry
+	ExtraEnv   map[string]string
+}
+
+// Registry represents authentication configuration for a single registry.
+type Registry struct {
+	authn.AuthConfig
+	CACerts            string
+	InsecureSkipVerify bool
+}
```
burgerdev, line 33: Please add some toml marshalling hints (I believe the convention is `kebab-case`) and verify that the JSON marshalling hints for [`AuthConfig`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#AuthConfig) are respected when unmarshalling TOML.

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ReadInsecureConfig reads the auth config from the specified TOML file.
```
burgerdev, line 35: Please add a sentence explaining why this is insecure and what that means for the caller.

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,99 @@
...
+func ReadInsecureConfig(path string, log *slog.Logger) (*Config, error) {
+	data, err := os.ReadFile(path)
+	if errors.Is(err, os.ErrNotExist) {
+		log.Info("Imagepuller auth config file does not exist. Authenticated pulls are not available")
+		return &Config{}, nil
+	} else if err != nil {
+		return nil, fmt.Errorf("reading insecure config file %s: %w", api.InsecureConfigPath, err)
```
burgerdev:
```suggestion
		return nil, fmt.Errorf("reading insecure config file %q: %w", api.InsecureConfigPath, err)
```

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,99 @@
...
+	var registry Registry
+	maxSpecificity := -1 // "." has specificity 0.
+	for fqdn, registryCandidate := range c.Registries {
+		fqdn = strings.TrimSuffix(fqdn, ".")
+		specificity := strings.Count(fqdn, ".")
+		if specificity > maxSpecificity && strings.HasSuffix(registryName, fqdn) {
+			maxSpecificity = specificity
+			registry = registryCandidate
+		}
+	}
```
burgerdev: If this were a standalone function, testing it would be easier.

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,99 @@
...
+var errUnparseableRef = errors.New("could not parse image ref")
+
+// AuthTransportFor constructs the appropriate http.Transport and authn.Authenticator for the given image's registry.
+func (c *Config) AuthTransportFor(imageRef string) (*authn.Authenticator, *http.Transport, error) {
+	// Note: this does no check image pinning.
+	ref, err := name.ParseReference(imageRef)
+	if err != nil {
+		return nil, nil, fmt.Errorf("%w: %w", errUnparseableRef, err)
+	}
+	registryName := ref.Context().RegistryStr()
```
burgerdev: We need to handle a port, if present.

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,99 @@
...
+
+	var registry Registry
+	maxSpecificity := -1 // "." has specificity 0.
+	for fqdn, registryCandidate := range c.Registries {
+		fqdn = strings.TrimSuffix(fqdn, ".")
```
burgerdev: `registryName` could also have a trailing dot.

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,99 @@
...
+func (c *Config) AuthTransportFor(imageRef string) (*authn.Authenticator, *http.Transport, error) {
+	// Note: this does no check image pinning.
+	ref, err := name.ParseReference(imageRef)
+	if err != nil {
+		return nil, nil, fmt.Errorf("%w: %w", errUnparseableRef, err)
+	}
+	registryName := ref.Context().RegistryStr()
+
+	var registry Registry
+	maxSpecificity := -1 // "." has specificity 0.
```
burgerdev: Maybe it's just me, but I'd find "suffix length" more natural than "number of dots".

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,99 @@
...
+		transport.TLSClientConfig.RootCAs = certpool
+	}
+
+	return &authenticator, transport, nil
+}
+
+// ApplyEnvVars applies the envvar-based proxy configuration in ExtraEnv.
+func (c *Config) ApplyEnvVars() {
+	for k, v := range c.ExtraEnv {
+		os.Setenv(k, v)
```
burgerdev: On second thought, we should check the keys against an allowlist `{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", }` for now.

imagepuller/internal/auth/auth.go
charludo: Hm. Longer matching suffix implies higher specificity here, sure, but isn't that just a result of the underlying idea that a lower level of subdomain is more specific?

But it ultimately doesn't matter, I do not have a strong opinion on this and am happy to change it.

burgerdev: You're right. wdyt about renaming `maxSpecificity` to `maxMatchingLabels` to make this more self-explanatory?

charludo: sounds good to me!

imagepuller/internal/auth/auth.go
charludo: Is that actually a requirement? In the RFC, using IPs and ports was [explicitly ignored](https://github.com/edgelesssys/contrast/blob/03cb64f007c8fbbf355555699491cb7de765b114/rfc/011-registry-authentication.md?plain=1#L265). 

Or do you just mean, strip the port away from the ref in case someone does e.g. `registry.com:8000/user/repo/image:tag`, so that we use the config for `registry.com`?

burgerdev: I think for now I only care that ports can be specified in both image ref and config. I don't care too much about whether we compare them, but it's probably easy to do that. IPs should still be ignored for now.

charludo: Specifying an IP+-Port in ref/config does not cause any issues. I've added two testcases.

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,107 @@
...
+// Registry represents authentication configuration for a single registry.
+type Registry struct {
+	authn.AuthConfig
+	CACerts            string `toml:"ca-certs"`
+	InsecureSkipVerify bool   `toml:"insecure-skip-verify"`
+}
+
+// ReadInsecureConfig reads the auth config from the specified TOML file.
+// No integrity checks are performed on the device from which the initdata-processor originally read this config.
+// An attacker with k8n admin privileges could thus change the config..
```
burgerdev:
```suggestion
// An attacker with k8s admin privileges could thus change the config.
```

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return &authenticator, transport, nil
+}
+
+// ApplyEnvVars applies the envvar-based proxy configuration in ExtraEnv.
+func (c *Config) ApplyEnvVars() {
+	allowedEnvVars := []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+	for _, env := range allowedEnvVars {
+		os.Setenv(env, c.ExtraEnv[env])
```
burgerdev: This potentially overrides variables with the empty string that might have been set non-empty in the systemd unit file (for example). This is not breaking anything right now and might never be relevant, but can we check, just for completeness, whether the variable even is in `ExtraEnv` before applying it?

imagepuller/internal/auth/auth.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+// registryFor returns the registry, if any, for the given registry name.
+func (c *Config) registryFor(name string) Registry {
+	var registry Registry
+	maxMatchingLabeld := -1 // "." has 0 matching labels.
```
burgerdev: Typo: `maxMatchingLabels`?

PR #1856, opened 2025-10-21
imagepuller: migrate to container-libs

charludo:
Context: https://blog.podman.io/2025/08/upcoming-migration-of-three-containers-repositories-to-monorepo/


burgerdev (approved):
Thanks!

PR #1862, opened 2025-10-22
e2e: add fifo label to namespace in contrasttest

davidweisse:
If an E2E test is run via `nix shell .#contrast.e2e --command {{ target }}.test` as in `just e2e` or in GitHub workflows, the namespace for the test is created by `contrasttest`, so we need to add the `contrast.edgeless.systems/sync-ticket` label here. If a test is cleaned up by the maintenance namespace cleanup job after an hour, the fifo ticket is then also released.

### DIFF
--- a/e2e/internal/contrasttest/contrasttest.go
+++ b/e2e/internal/contrasttest/contrasttest.go
@@ -44,6 +44,7 @@ type testFlags struct {
 	NamespaceFile               string
 	NamespaceSuffix             string
 	NodeInstallerTargetConfType string
+	SyncTicketFile              string
 }
 
 // RegisterFlags registers the flags that are shared between all tests.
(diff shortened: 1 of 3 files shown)


e2e/internal/contrasttest/contrasttest.go
```diff
@@ -105,14 +107,23 @@ func (ct *ContrastTest) Init(t *testing.T, resources []any) {
 	require.NoError(err, "Parsing image replacements from %s failed", ct.ImageReplacementsFile)
 
 	// Create namespace
-	namespace, err := kuberesource.ResourcesToUnstructured([]any{kuberesource.Namespace(ct.Namespace)})
+	namespace := kuberesource.Namespace(ct.Namespace)
+
+	// Add sync ticket label if provided
+	if ticket, _ := os.ReadFile(Flags.SyncTicketFile); len(ticket) > 0 {
```
burgerdev: Please don't swallow the error here - we should check for `ENOENT` and fail on other errors.

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -105,14 +107,25 @@ func (ct *ContrastTest) Init(t *testing.T, resources []any) {
...
+	namespace := kuberesource.Namespace(ct.Namespace)
+
+	// Add sync ticket label if provided
+	if ticket, err := os.ReadFile(Flags.SyncTicketFile); err == nil {
+		namespace.WithLabels(map[string]string{
+			"contrast.edgeless.systems/sync-ticket": strings.TrimSpace(string(ticket)),
+		})
+	} else if !errors.Is(err, os.ErrNotExist) {
+		require.NoError(err, "Reading sync ticket from %s failed", Flags.SyncTicketFile)
+	}
```
burgerdev: Nit: you could just

```go
} else {
  require.ErrorIs(//...
}
```

PR #1863, opened 2025-10-23
docs: describe initdata flow

burgerdev:


katexochen:
> I think we need to update the runtime graphic under https://edgelesssys.github.io/contrast/pr-preview/pr-1863/architecture/components/runtime#node-level-components


The initdata processor should also have its own section in that document.

docs/docs/_media/runtime.drawio.svg
burgerdev, line 1: I need to figure out how to edit without applying my desktop color scheme :/

PR #1865, opened 2025-10-23
packages/contrast-docs: yarn logging for ci

katexochen:
This should significantly reduce the amount of output by telling yarn it's running in CI.

### DIFF
--- a/.github/workflows/docs_publish.yml
+++ b/.github/workflows/docs_publish.yml
@@ -9,7 +9,7 @@ on:
       - docs/**
       - "!docs/README.md"
       - .github/workflows/docs_publish.yml
-      - packages/contrast-docs/**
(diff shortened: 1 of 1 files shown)


packages/by-name/contrast-docs/package.nix
```diff
@@ -33,6 +33,7 @@ mkYarnPackage rec {
 
   buildPhase = ''
     export HOME=$(mktemp -d)
+    export CI=true
```
burgerdev, line 36: It would be clearer if we used `yarn --log-level` here. Can we link documentation for yarn that explains how `CI` affects the build?

katexochen: I think the point is that this will also affect webpack and stuff? Didn't found any docs for yarn classic about this.

PR #1868, opened 2025-10-23
docs: remove all references to the coco project

katexochen:
We are not using things from CoCo project and have diverged in what we do, too. So removing these to for reduced confusion. Leaving some occurrences without title casing to refer to the concept, not the project.

### DIFF
--- a/docs/docs/howto/encrypted-storage.md
+++ b/docs/docs/howto/encrypted-storage.md
@@ -120,7 +120,7 @@ Follow [the steps given above](#deployment) to reapply your configuration.
 
 The deployment YAML shipped for this demo is already configured to be used with Contrast.
 A [runtime class](../architecture/components/runtime) `contrast-cc`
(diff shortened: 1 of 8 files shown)


docs/docs/howto/encrypted-storage.md
```diff
@@ -120,7 +120,7 @@ Follow [the steps given above](#deployment) to reapply your configuration.
 
 The deployment YAML shipped for this demo is already configured to be used with Contrast.
 A [runtime class](../architecture/components/runtime) `contrast-cc`
-was added to the pods to signal they should be run as Confidential Containers. During the generation process,
+was added to the pods to signal they should be run as confidential container. During the generation process,
```
charludo:
```suggestion
was added to the pods to signal they should be run as confidential containers. During the generation process,
```

or


```suggestion
was added to each pod to signal it should be run as a confidential container. During the generation process,
```

PR #1874, opened 2025-10-23
docs: update resources that influence policy

katexochen:
The only resources that influence the policy beside pod-generating controllers are ConfigMaps and Secrets. All other resources are ignored by genpolicy (`contrast generate`) and won't affect the policy. With #1842, the Contrast CLI won't fail anymore if those resources are present.

### DIFF
--- a/docs/docs/architecture/components/policies.md
+++ b/docs/docs/architecture/components/policies.md
@@ -104,16 +104,8 @@ Contrast policies can be generated for all Kubernetes resource kinds that spawn 
 The following resource kinds influence pod configuration and are also taken into account for policy generation:
 
 <!-- keep-sorted start by_regex=`(\w+)` -->
-- `ClusterRole`
-- `ClusterRoleBinding`
 - `ConfigMap`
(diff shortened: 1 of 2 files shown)


docs/docs/architecture/components/policies.md
```diff
@@ -104,16 +104,8 @@ Contrast policies can be generated for all Kubernetes resource kinds that spawn
...
-- `PodDisruptionBudget`
-- `Role`
-- `RoleBinding`
 - `Secret`
-- `Service`
-- `ServiceAccount`
 <!-- keep-sorted end -->
 
-All other resource kinds are unsupported and can't be passed to `contrast generate`.
+All other resource kinds are ignored by `contrast generate` and don't influence the generated policy.
```
burgerdev, line 111: If we change this here, we should also remove https://github.com/edgelesssys/contrast/blob/8ce91e4b6bb601641388f9ef02c7b7290e45e6e6/docs/docs/architecture/features-limitations.md?plain=1#L23-L24.

PR #1885, opened 2025-10-28
treefmt: docs-selfref-lint

katexochen:


packages/scripts.nix
```diff
@@ -701,4 +701,22 @@ lib.makeScope pkgs.newScope (scripts: {
...
     '';
   };
+
+  docs-selfref-lint = writeShellApplication {
+    name = "docs-selfref-lint";
+    runtimeInputs = with pkgs; [ gnugrep ];
+    text = ''
+      exitcode=0
+      for file in "$@"; do
+        if grep -qE 'https://docs\.edgeless\.systems' "$file"; then
```
burgerdev: Should this match only `docs.edgeless.systems/contrast`?

burgerdev: I'd say this calls for `-F` instead of `-E`.

packages/scripts.nix
```diff
@@ -701,4 +701,22 @@ lib.makeScope pkgs.newScope (scripts: {
       exit $exitcode
     '';
   };
+
+  docs-selfref-lint = writeShellApplication {
+    name = "docs-selfref-lint";
+    runtimeInputs = with pkgs; [ gnugrep ];
+    text = ''
+      exitcode=0
+      for file in "$@"; do
```
burgerdev: Nit: You could also do `grep -l ... "$@"` instead of the for loop.

burgerdev (approved):

PR #1886, opened 2025-10-29
e2e: add layer caching regression test

burgerdev:
* test failure before 568133061e1c6128bfbdcb4381f79a5f23a75373: https://github.com/edgelesssys/contrast/actions/runs/18901503381/job/53949621539


charludo (commented):
Thank you for the test, this is a really nice reproducer. The only thing I noticed is that this would (as far as I can tell for the first) time introduce dependence on images hosted on a "personal" registry. Not sure if that's an issue though.

charludo (approved):
> Moved the images to the edgelesssys registry.

Thank you!

PR #1887, opened 2025-10-29
imagepuller-benchmark: run with local pulls

burgerdev:
Change the benchmark workflow to first pull the images into a local registry and then run the benchmark against that registry. This should prevent network speed influences on the benchmark timings.

### DIFF
--- a/.github/workflows/imagepuller-benchmark.yml
+++ b/.github/workflows/imagepuller-benchmark.yml
@@ -33,11 +33,14 @@ jobs:
       - name: Download baseline JSON from S3
         run: |
           aws s3 cp "s3://contrast-public/imagepuller-benchmark/baseline.json" "baseline.json"
+      - name: pre-pull images to local registry
+        run: |
+          nix run .#scripts.local-registry -- localhost:30000 < tools/imagepuller-benchmark/images.txt > local-images.txt
       - name: Run imagepuller-benchmark script
         run: |
           imagepullereval="$(nix build .#imagepuller-benchmark --print-out-paths)/bin/imagepuller-benchmark"
           imagepuller="$(nix build .#imagepuller --print-out-paths)/bin/imagepuller"
-          sudo "$imagepullereval" "$imagepuller" -c baseline.json -o baseline-new.json -i tools/imagepuller-benchmark/images.txt
+          sudo "$imagepullereval" "$imagepuller" -c baseline.json -o baseline-new.json -i local-images.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/imagepuller-benchmark.yml
```diff
@@ -33,11 +33,21 @@ jobs:
       - name: Download baseline JSON from S3
         run: |
           aws s3 cp "s3://contrast-public/imagepuller-benchmark/baseline.json" "baseline.json"
+      - name: pre-pull images to local registry
+        run: |
+          nix build nixpkgs#crane
+          nix shell nixpkgs#crane --command crane registry serve --address localhost:30001 &
```
charludo:
```suggestion
          nix shell nixpkgs#crane --command crane registry serve --address localhost:30001 &
```

(or, rather: what's this needed for?)

.github/workflows/imagepuller-benchmark.yml
```diff
@@ -33,11 +33,21 @@ jobs:
...
+          sleep 2 # wait for registry to become ready
+          while read -r line; do
+            localImg="localhost:30001/${line#*/}"
+            skopeo copy --all --dest-tls-verify=false "docker://$line" "docker://$localImg"
+            echo "$localImg" >> local-images.txt
+          done < tools/imagepuller-benchmark/images.txt
```
charludo: Could you move this to `scripts.nix`? It would be nice to start the local registry in a single command for local testing.

.github/workflows/imagepuller-benchmark.yml
burgerdev: If I don't build before spawning the shell, I can't tell when the registry is ready (note that the registry is forked off).

charludo: Ah, didn't make the connection. Makes sense!

.github/workflows/imagepuller-benchmark.yml
burgerdev: Yeah, also makes sense due to the tool deps.

burgerdev: Moved, ptal

burgerdev:
Ok, this turned out to be really unstable after a couple of test runs. I'll do something different: add ghcr credentials to the imagepuller config and move the tensorflow image there, too. That way, we should at least not run into rate limits or quotas.

PR #1889, opened 2025-10-29
nodeinstaller: remove serviceaccount and clusterrole

davidweisse:
This removes the service account from the node installer daemon set, as well as the associated `ServiceAccount`, `ClusterRole` and `ClusterRoleBinding` resources from the runtime YAML.

These resources were added in #1103 for the nydus pull container, which was removed in #1434. Since there is no other use, we can safely remove the resources here.

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -36,16 +36,8 @@ func ContrastRuntimeClass(platform platforms.Platform) (*RuntimeClassConfig, err
 	return &RuntimeClassConfig{r}, nil
 }
 
-// NodeInstallerConfig wraps a DaemonSetApplyConfiguration for a node installer.
-type NodeInstallerConfig struct {
-	*applyappsv1.DaemonSetApplyConfiguration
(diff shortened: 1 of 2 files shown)


internal/kuberesource/parts.go
```diff
@@ -39,9 +39,6 @@ func ContrastRuntimeClass(platform platforms.Platform) (*RuntimeClassConfig, err
 // NodeInstallerConfig wraps a DaemonSetApplyConfiguration for a node installer.
 type NodeInstallerConfig struct {
 	*applyappsv1.DaemonSetApplyConfiguration
-	*applycorev1.ServiceAccountApplyConfiguration
-	*applyrbacv1.ClusterRoleApplyConfiguration
-	*applyrbacv1.ClusterRoleBindingApplyConfiguration
```
katexochen, line 44: Can we also remove the NodeInstallerConfig as wrapper type and just use the ApplyConfig?

katexochen:
Can you cleanup the remaining resources in the CI by hand please?

katexochen (approved):

PR #1890, opened 2025-10-29
ci: add docs backport reminder workflow

davidweisse:


.github/workflows/docs_backport_reminder.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Add reminder comment
+        if: steps.find-comment.outputs.comment-id == ''
+        uses: peter-evans/create-or-update-comment@e8674b075228eee787fea43ef493e45ece1004c9 # v5.0.0
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            ### Does this change need to be backported?
+
+            - [ ] Yes, this should be backported.
+            - [ ] No, backporting is not needed.
```
katexochen:
```suggestion
            ### Does the documentation changes need to be backported?
            
            Changes to `/doc/docs` won't be visible until the next release.
            If you are fixing something in the docs that should be immediately visible, the change needs to be made to both `/docs/docs` and /`docs/versioned_docs/version-X.Y`, where X.Y is the version of the latest minor release. This can be done in this same PR.

            - [ ] Yes, this should be backported to the currently version of the docs.
            - [ ] No, the PR only contains docs changes relevant for future versions.
```

.github/workflows/docs_backport_reminder.yml
```diff
@@ -0,0 +1,35 @@
...
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Check if comment already exists
+        id: find-comment
+        uses: peter-evans/find-comment@b30e6a3c0ed37e7c023ccd3f1db5c6c0b0c23aad # v4.0.0
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body-includes: "Do the documentation changes need to be backported?"
+      - name: Add reminder comment
```
katexochen, line 35: We can add a step to check if docs/versioned_docs/.. was modified. In this case, we don't need to print the reminder.

PR #1882, opened 2025-10-28
initdata: handle imagepuller config

charludo:
If `/etc/host-config/imagepuller.toml` exists in the runtime's dir, the kata runtime will pack it into a device, similar to the initdata device; the initdata-processor will then see the device, read its contents and write the original file back to the `/run/insecure-config/imagepuller.toml` file expected by the imagepuller.

PR #1893, opened 2025-10-30
nodeinstaller: write imagepuller config secret to host

charludo:
Wrote this as part of the e2e for the imagepuller auth, but seemed like it should rather be in its own, narrowly defined PR.

## How to test:

1. deploy some workload:
    ```bash
    just
    ```
2. take note of namespace:
    ```bash
    export ns=$(cat workspace/just.namespace)
    ```
3. create a file `contrast-imagepuller.toml` to be used as a secret. For testing, a comment is enough:
    ```toml
    # Hello from the other side!
    ```
4. create the secret in kubernetes:
    ```bash
    kubectl create secret generic contrast-node-installer-imagepuller-config --from-file contrast-imagepuller.toml=contrast-imagepuller.toml -n $ns
    ```
5. restart the nodeinstaller daemonset:
    ```bash
    kubectl rollout restart ds contrast-cc-metal-qemu-snp-be766437-nodeinstaller -n $ns
    ```
6. on the host, find your file in `/opt/edgeless/contrast-cc-metal-qemu-snp-be766437/etc/host-config/contrast-imagepuller.toml`
7. if you rebase on #1882 before all of this, you can also kill/restart a pod and then find `/run/insecure-config/imagepuller.toml` and a mention of its usage in the imagepuller logs.

### DIFF
--- a/nodeinstaller/node-installer.go
+++ b/nodeinstaller/node-installer.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


nodeinstaller/node-installer.go
```diff
@@ -364,3 +369,41 @@ type assetFetcher interface {
...
+	defer inputFile.Close()
+
+	outputFile, err := os.Create(imagepullerDest)
+	if err != nil {
+		return fmt.Errorf("opening imagepuller config destination file %q: %w", imagepullerDest, err)
+	}
+	defer outputFile.Close()
+
+	_, err = io.Copy(outputFile, inputFile)
+	if err != nil {
```
burgerdev:
```suggestion
	if _, err := io.Copy(outputFile, inputFile); err != nil {
```

burgerdev (approved):
lgtm

PR #1894, opened 2025-10-30
cli/generate: add flag for insecure debug shell access

katexochen:
This adds a new flag to the generate command of the Contrast CLI, enabling access to the pod VM via SSH shell for debugging. See the added documentation section for details.


internal/initdata/initdata.go
```diff
@@ -164,17 +164,15 @@ func FromDevice(devicePath string) (Raw, error) {
 		return nil, err
 	}
 	buf := make([]byte, 8)
-	_, err = f.ReadAt(buf, 0)
-	if err != nil {
+	if _, err = f.ReadAt(buf, 0); err != nil {
```
burgerdev: In the cleanup commit: should be `:=`.

internal/initdata/initdata.go
```diff
@@ -164,17 +164,15 @@ func FromDevice(devicePath string) (Raw, error) {
...
 		return nil, fmt.Errorf("reading magic number: %w", err)
 	}
 	const magic = "initdata"
 	if slices.Compare(buf, []byte(magic)) != 0 {
 		return nil, fmt.Errorf("%w: expected %x, got %x", errWrongMagic, magic, buf)
 	}
 	buf = make([]byte, 8)
-	_, err = f.ReadAt(buf, 8)
-	if err != nil {
+	if _, err = f.ReadAt(buf, 8); err != nil {
```
burgerdev: In the cleanup commit: should be `:=`.

internal/initdata/initdata.go
```diff
@@ -183,9 +181,7 @@ func FromDevice(devicePath string) (Raw, error) {
 		return nil, fmt.Errorf("%w: expected at most 128MiB, got %d byte", errTooLarge, size)
 	}
 	buf = make([]byte, size)
-
-	_, err = f.ReadAt(buf, 16)
-	if err != nil {
+	if _, err = f.ReadAt(buf, 16); err != nil {
```
burgerdev: In the cleanup commit: should be `:=`.

go.work
```diff
@@ -1,13 +1,16 @@
...
 	./imagestore
 	./initdata-processor
 	./service-mesh
+	./tools/debugshell
 	./tools/igvm
 	./tools/imagepuller-benchmark
-	./tools/tdx-measure
 	./tools/snp-id-block-generator
+	./tools/tdx-measure
+// keep-sorted end
```
burgerdev, line 15: Should this be indented?

packages/by-name/debugshell/package.nix
```diff
@@ -0,0 +1,69 @@
...
+        (path.append root "go.mod")
+        (path.append root "go.sum")
+        (fileset.fileFilter (file: hasSuffix ".go" file.name) root)
+      ];
+    };
+
+  proxyVendor = true;
+  vendorHash = "sha256-gZcbNCevRe79wazZBMOJPexLG49lw25cOIa2UjGP/Cs=";
+
+  subPackages = [ "." ];
```
burgerdev, line 33: is this necessary if we only have one subpackage at the src root?

packages/by-name/debugshell/package.nix
```diff
@@ -0,0 +1,69 @@
...
+
+  postInstall =
+    let
+      debugshell = writeShellApplication {
+        name = "debugshell";
+        runtimeInputs = [ openssh ];
+        text = ''
+          if [[ ! -f /etc/passwd ]]; then
+              echo "root:x:0:0:root:/root:/bin/bash" > /etc/passwd
+          fi
```
burgerdev, line 50: Can we fix this in the rootfs instead?

packages/by-name/debugshell/package.nix
```diff
@@ -0,0 +1,69 @@
...
+          if [[ ! -f ./id_ed25519 ]]; then
+              ssh-keygen -t ed25519 -f ./id_ed25519 -N ""
+          fi
```
burgerdev, line 53: Do we even need to generate a public key, given that we accept all passwords in the server?

tools/debugshell/main.go
```diff
@@ -0,0 +1,143 @@
...
+	ptyReq, winCh, isPty := s.Pty()
+	if !isPty {
+		if len(s.Command()) == 0 {
+			log.Printf("No pty or command requested from %s", s.RemoteAddr())
+			cmd := exec.CommandContext(s.Context(), bashPath, "-c", "echo 'No pty requested'; exit")
+			cmd.Stdout = s
+			cmd.Stderr = s
+			if err := cmd.Run(); err != nil {
+				fmt.Fprintf(s, "cmd run error: %v\n", err)
+			}
```
burgerdev: Could we just write directly to `s`?

tools/debugshell/main.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		log.Printf("Executing non-interactive command for %s", s.RemoteAddr())
+		cmd := exec.CommandContext(s.Context(), bashPath, "-lc", strings.Join(s.Command(), " "))
```
burgerdev: This sounds like it would be prone to quoting issues. I believe we should either pass only `s.Command()` as argv for `CommandContext`, or execute the `s.RawCommand()` as a bash script.

tools/debugshell/main.go
```diff
@@ -0,0 +1,143 @@
...
+			// send SIGWINCH to the process so it updates children
+			_ = cmd.Process.Signal(syscall.SIGWINCH)
+		}
+	}()
+
+	// Copy between the pty and SSH session
+	go func() {
+		_, _ = io.Copy(ptmx, s) // session → pty (stdin)
+	}()
+	_, _ = io.Copy(s, ptmx) // pty → session (stdout & stderr)
```
burgerdev, line 135: The logic for one in background and one in foreground is "after ptmx closed its output, it won't accept anything on its input anyway", correct? Might make sense to document that.

packages/containers.nix
```diff
@@ -169,6 +169,25 @@ let
...
+        busybox
+        bash
+        coreutils
+        ncurses
+        contrastPkgs.debugshell
+        openssh
+      ];
+      config = {
+        Cmd = [
+          "/bin/debugshell"
```
burgerdev: Should this be the `Entrypoint`?

packages/containers.nix
```diff
@@ -169,6 +169,25 @@ let
...
+          "journalctl && journalctl -f"
```
burgerdev: This could be just `journalctl --no-tail --no-pager -f`.

internal/kuberesource/parts.go
```diff
@@ -470,3 +470,13 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
 			"-l", "debug",
 		)
 }
+
+// DebugShell creates a new debug shell container.
+func DebugShell() *applycorev1.ContainerApplyConfiguration {
+	return applycorev1.Container().
+		WithName("debug-shell").
```
burgerdev: The other containers have a `contrast-` prefix.

docs/docs/howto/debugshell.md
```diff
@@ -0,0 +1,88 @@
...
+## Collecting pod VM logs
+
+The `debug-shell` container exposes the journal logs of the pod VM into Kubernetes pod logs by default.
+These logs can then be collected via standard Kubernetes tooling:
+
+```sh
+kubectl logs <pod-name> -c debug-shell
+```
+
+Ensure the `settings.json` used for generate includes the following settings so that access to the pod logs are allowed:
```
burgerdev: Might want to highlight that this is the default.

docs/docs/howto/debugshell.md
```diff
@@ -0,0 +1,88 @@
...
+{
+    ...
+    "request_defaults": {
+        ...
+        "ReadStreamRequest": true,
+```
+
+## Accessing the pod VM via debug shell
+
+Ensure the `settings.json` used for generate includes the following settings to allow interactive exec in Contrast pods:
```
burgerdev, line 56: We should make clear that generate needs to be called again after modifying `settings.json`.

justfile
```diff
@@ -10,11 +10,17 @@ push target:
     printf "ghcr.io/edgelesssys/contrast/%s:latest=%s\n" "{{ target }}" "$pushedImg" >> {{ workspace_dir }}/just.containerlookup
 
 coordinator: (push "coordinator")
+
```
burgerdev, line 13: The whitespace was removed a few commits earlier and is reintroduced here.

go.work
katexochen: Whatever formatter is running on that file thinks so. :sweat_smile:

packages/by-name/debugshell/package.nix
katexochen: Not necessary but future safe. Setting subPackages is generally recommended.

packages/by-name/debugshell/package.nix
katexochen: The idea was to be able to install the package into different images. For example, the way this is currently working you can add the debugshell package to the coordinator image if you are trying to debug something specific there. You think this could cause issues?

packages/by-name/debugshell/package.nix
katexochen: There is no way passing a password via ssh cli flags. It is possible to enter with an empty password interactively. I experimented a bit and found this to be the most convenient way for both interactive and non-interactive exec.

packages/by-name/debugshell/package.nix
burgerdev: No, it's probably fine.

go.work
burgerdev: Disagrees, you mean? Ok then.

PR #1895, opened 2025-10-30
imagepuller: cache individual layers

burgerdev:
Before this PR, the imagepuller was only caching full images. If two different images shared layers, these would be downloaded two times.

With this change, we're switching the underlying layer storage engine (from podman to umoci), such that we have finer control over the storage layout. This allows storing the layers in independent, content-addressed directories, which can then be reused when assembling images.

---

Note to developers: 
 
* I removed the digest checks in the layer pull paths, because they are already implemented in go-containerregistry, as demonstrated by the evil registry test.
* Using `compress/gzip` instead of `pgzip` causes a pull time regression north of 30%.
* Contexts are not propagated, which causes the pull to continue even if the agent gave up. This is unchanged from before, but we should tackle this soon.
* Before this change, manifests were cached implicitly, because the images were tagged with the manifest digest. Now, we are pulling the manifest each time, even if we are pulling the same image twice. This is not critical, as manifests are on the order of kB, but we should eventually cache them, too.
* [2/3 benchmark runs show no regression](https://github.com/edgelesssys/contrast/actions/runs/18943296218)


imagepuller/internal/service/service.go
```diff
@@ -57,58 +63,53 @@ func (s *ImagePullerService) PullImage(ctx context.Context, r *api.ImagePullRequ
...
+		return nil, fmt.Errorf("listing layers: %w", err)
 	}
-	if err := s.Store.AddNames(newImg.ID, []string{r.ImageUrl}); err != nil {
-		return nil, fmt.Errorf("adding image url as image name: %w", err)
+	for i, l := range remoteLayers {
+		digest, err := l.Digest()
+		if err != nil {
+			return nil, fmt.Errorf("getting digest of layer %d: %w", i, err)
+		}
+		layers = append(layers, digest.String())
```
charludo: It appears to me that this might reasonably be moved into the `storeAndVerifyLayers` func, i.e. store layers, then return the list of layers. That would also make it easier to add some unittests to the existing ones for this function which are checking that all wanted layers are being returned, regardless of whether they were pulled just now or in a previous image's pull.

imagepuller/internal/service/verify.go
```diff
@@ -84,52 +85,32 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
 	return remoteImg, nil
 }
 
-func (s *ImagePullerService) storeAndVerifyLayers(log *slog.Logger, remoteImg gcr.Image) (string, error) {
+func (s *ImagePullerService) storeAndVerifyLayers(log *slog.Logger, remoteImg gcr.Image) error {
```
charludo: The name is not really fitting anymore, since we don't do manual digest checks any longer.

Maybe this should even be moved into the new `store` package, as `store.Store(gcr.Image)` or `store.StoreDeduplicated(...)`?

imagepuller/internal/service/verify.go
```diff
@@ -84,52 +85,32 @@ func (s *ImagePullerService) getAndVerifyImage(ctx context.Context, log *slog.Lo
...
+		return fmt.Errorf("obtaining image manifest: %w", err)
 	}
 
-	previousLayer := ""
-	for idx, layer := range layers {
-		rc, err := layer.Compressed()
-		if err != nil {
-			return "", fmt.Errorf("reading layer %d: %w", idx, err)
-		}
+	// TODO(burgerdev): pass a context here?
```
charludo, line 94: I might be missing something here, but it seems like we should quite easily be able to pass in the context from `PullImage` here?

imagepuller/internal/store/store.go
```diff
@@ -0,0 +1,132 @@
...
+	return nil
+}
+
+// getDecompressedReader returns a reader for the uncompressed layer tarball.
+// Its Close method must be called after unpacking finished.
+//
+// While we could be using remoteLayer.Uncompressed(), the gcr implementation currently uses gzip
+// from the stdlib, which causes a significant performance hit compared to pgzip. This is why we
+// decompress here.
+func getDecompressedReader(remoteLayer gcr.Layer) (io.ReadCloser, error) {
```
charludo, line 90: I'd really appreciate some unittests for this package. Seems to me it's otherwise easy to get something wrong here in future iterations 😅

imagepuller/internal/service/verify.go
burgerdev: I'd prefer to do this in a separate PR.

imagepuller/internal/service/service.go
burgerdev: Done

imagepuller/internal/service/verify.go
burgerdev: I'd like to keep the `Store` layer-based, so I'm keeping this function for dealing with a list of layers, but with a more appropriate name.

imagepuller/internal/store/store.go
burgerdev: Implemented

PR #1896, opened 2025-11-01
dev-docs: how does CDI work?

burgerdev:


dev-docs/nvidia/cdi.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Kata runtime
+
+When the Kata runtime receives a `CreateSandbox` request, it [looks specifically for CDI annotations](https://github.com/kata-containers/kata-containers/blob/c75a46d17f800e1d825aca31c62c7bf3f44ca8b1/src/runtime/pkg/containerd-shim-v2/create.go#L111-L125).
+This results in the pause container receiving the same device edits as the main container did, but from the CDI spec instead of the device plugin.
+With this information, Kata can set up the VM with the GPU plugged in.
+It also [stores a list of devices attached to the VM](https://github.com/kata-containers/kata-containers/blob/c75a46d17f800e1d825aca31c62c7bf3f44ca8b1/src/runtime/virtcontainers/container.go#L1059), linking the host device path with its corresponding guest PCI port.
+
+For every regular container, the Kata agent checks the devices in the OCI spec against the list of attached PCIe devices.
+If there's a match, it [adds a _sibling annotation_](https://github.com/kata-containers/kata-containers/blob/c75a46d17f800e1d825aca31c62c7bf3f44ca8b1/src/runtime/virtcontainers/container.go#L1098-L1111).
```
burgerdev, line 132: @charludo: If we want to go back to the original device assignment, we could replace this check with something that applies more broadly. Maybe not assign the device to every container, but according to env vars (`NVIDIA_VISIBLE_DEVICES` comes to mind).

PR #1898, opened 2025-11-03
imagepuller: swap out toml libraries (BurntSushi -> pelletier)

charludo:
Just noticed that #1853 introduced an (unnecessary) new dependency on `BurntSushi/toml`, where everywhere else we seem to be using `pelletier/go-toml`. As far as the  imagepuller usecase goes, this is a drop-in replacement.

### DIFF
--- a/imagepuller/go.mod
+++ b/imagepuller/go.mod
@@ -3,10 +3,10 @@ module github.com/edgelesssys/contrast/imagepuller
 go 1.25.0
 
 require (
-	github.com/BurntSushi/toml v1.5.0
 	github.com/containerd/ttrpc v1.2.7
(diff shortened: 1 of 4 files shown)


burgerdev (approved):

imagepuller/go.mod
```diff
@@ -16,6 +16,7 @@ require (
 )
 
 require (
+	github.com/BurntSushi/toml v1.5.0 // indirect
```
katexochen, line 19: looks like it is still used transitively, so not much benefit. Good to keep it consistent anyway.

charludo: True. Seems to come from `go.podman.io/storage/types`.

PR #1899, opened 2025-11-03
cli/generate: refactor resource mapping

katexochen:
To be able to make changes to make changes to the resources while mapping, the map* functions must return/write back the modified resources.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -237,18 +237,27 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 // mapCCWorkloads applies the given function to all workloads with the 'contrast-cc' runtime class.
 // The callback receives an apply configuration together with the file path and index the unstructured object has in the file map.
 // Changes to the apply configuration are not applied to the original unstructured object.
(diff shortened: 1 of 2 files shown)


davidweisse:
This fixes the regression tests on `contrast generate`. The test data for the `Job` and `CronJob` resources do not include a metadata in the `PodTemplate`. Missing metadata should be added by the `MapPodSpecWithMeta` function after db0b189acdc388cfb971612df0ac88c086667fc2, this however only applies *after* the `MapPodSpecWithMeta` returns, i.e., marshalling the resource and calling `fileMap[path][idx] = resUnstructured[0]` inside `MapPodSpecWithMeta` does not apply any newly created metadata object. With this change, the logic is moved into the `mapCCWorkloads` function, so changes to the file map are only applied after the resource has been changed.

davidweisse (commented):
https://github.com/edgelesssys/contrast/blob/6de61fa2f8ec2f6f9eda3e4c04cbf91d53616c59/cli/cmd/generate.go#L402-L410
This should be removed from `generatePolicies`.

PR #1903, opened 2025-11-05
kata/runtime: assign GPU devices to multiple containers

charludo:
When we switched from NVIDIA CTK hooks to CDI specs in #1835, we lost the ability to mount the GPU into containers that did not request a GPU with extended resources. While this is the intended behavior of CDI and extended resources (see [cdi.md](https://github.com/edgelesssys/contrast/blob/607e0000fdd8367a79f51ff6eb43696e83745565/dev-docs/nvidia/cdi.md)), we are implementing a workaround here that restores the old hook behavior: all containers with an environment variable `NVIDIA_VISIBLE_DEVICES=all` will receive all GPUs and their corresponding library mounts.

---

Original PR description:

https://github.com/edgelesssys/contrast/pull/1896#discussion_r2483852706 was surprisingly simple to implement. From my testing, both all pod-known GPU(s) and the corresponding Nvidia libs are handed to any container in the pod with
```yaml
- env:
  - name: NVIDIA_VISIBLE_DEVICES
    value: all
```
set.

We could potentially expand on this by parsing the actual value of that envvar and using it to only hand through specific GPUs, but
- that would (afaict) require specifying the VFIO num from the host CDI config, so not super portable,
- and there's probably not really a usecase for this (?)

I am also, unfortunately, pretty sure that this is a "forever-patch" and we will *not* get this merged into upstream kata.

### DIFF
--- a/internal/kuberesource/sets.go
+++ b/internal/kuberesource/sets.go
@@ -671,17 +671,31 @@ func GPU() []any {
 				WithSpec(PodSpec().
 					WithContainers(
 						Container().
-							WithName("gpu-tester").
+							WithName("gpu-tester-direct"). // This container directly requests the H100 resource.
+							WithImage("ghcr.io/edgelesssys/contrast/ubuntu:24.04@sha256:0f9e2b7901aa01cf394f9e1af69387e2fd4ee256fd8a95fb9ce3ae87375a31e6").
+							WithCommand("/bin/sh", "-c", "sleep inf").
+							WithResources(ResourceRequirements().
+								WithMemoryLimitAndRequest(400). // This accounts for nvidia-smi and the guest pull overhead.
+								WithLimits(corev1.ResourceList{
+									corev1.ResourceName("nvidia.com/GH100_H100_PCIE"): resource.MustParse("1"),
+								}),
+							),
+						Container().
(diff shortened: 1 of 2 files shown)


packages/by-name/kata/runtime/package.nix
thomasten, line 1: Can we cover this feature in the gpu e2e test?

internal/kuberesource/sets.go
```diff
@@ -671,18 +671,25 @@ func GPU() []any {
...
 								WithLimits(corev1.ResourceList{
 									corev1.ResourceName("nvidia.com/GH100_H100_PCIE"): resource.MustParse("1"),
 								}),
 							),
+						Container().
+							WithName("gpu-tester-indirect"). // This container indirectly shares the H100 through the NVIDIA_VISIBLE_DEVICES env var.
+							WithImage("ghcr.io/edgelesssys/contrast/ubuntu:24.04@sha256:0f9e2b7901aa01cf394f9e1af69387e2fd4ee256fd8a95fb9ce3ae87375a31e6").
+							WithCommand("/bin/sh", "-c", "sleep inf").
+							WithEnv(EnvVar().
+								WithName("NVIDIA_VISIBLE_DEVICES").WithValue("all"),
```
katexochen, line 688: Can we add a third container that has the env not set and check that it has no access to the GPU?

katexochen (approved):

PR #1909, opened 2025-11-10
kata.runtime-rs: init

katexochen:
- [ ] [regression, TDX](https://github.com/edgelesssys/contrast/actions/runs/25442516130) (fails, will be addressed in a follow-up)
- [x] [regression, SNP](https://github.com/edgelesssys/contrast/actions/runs/25495778305)


packages/by-name/kata/runtime-rs/package.nix
```diff
@@ -0,0 +1,83 @@
...
+
+  cargoTestFlags = [ "--bins" ];
+
+  checkFlags = [
+    "--skip=device::device_manager::tests::test_new_block_device"
+    "--skip=network::endpoint::endpoints_test::tests::test_ipvlan_construction"
+    "--skip=network::endpoint::endpoints_test::tests::test_macvlan_construction"
+    "--skip=network::endpoint::endpoints_test::tests::test_vlan_construction"
+    "--skip=test::test_new_hypervisor"
+  ];
```
burgerdev, line 65: Maybe add a comment that these require root privs (or why they can't be run otherwise).

packages/by-name/kata/runtime-rs/package.nix
```diff
@@ -0,0 +1,83 @@
...
+
+  # This is a placeholder to make this package compatible with the Go runtime,
+  # as the node-installer is configured to install this file.
+  # TODO(katexochen): Remove when switching to runtime-rs.
+  postInstall = ''
+    touch $out/bin/kata-runtime
```
burgerdev: I will probably be wondering what that 0-length file is supposed to be a few months down the road. How about populating it with the reason for its absence?

packages/by-name/kata/runtime/package.nix
```diff
@@ -132,6 +132,8 @@ buildGoModule (finalAttrs: {
       # This is unlikely to be fixed in Kata upstream, but rather in the NVIDIA components.
       # Upstream issue: https://github.com/NVIDIA/sandbox-device-plugin/issues/46
       ./0019-shim-guess-CDI-devices-without-direct-match.patch
+
+      ./0020-runtime-rs-allow-initrd-AND-image-to-be-set.patch
```
burgerdev: This patch is referenced in 61b7843, but the patch file is only added in a later commit, and with a different index.

packages/by-name/kata/runtime/package.nix
```diff
@@ -134,6 +134,9 @@ buildGoModule (finalAttrs: {
       # Upstream issue: https://github.com/NVIDIA/sandbox-device-plugin/issues/46
       ./0020-shim-guess-CDI-devices-without-direct-match.patch
 
+      # Deny unknown fields where possible to ease migration. This isn't possible where flatten is used.
+      # TODO: upstream.
```
burgerdev: I suppose you could link https://github.com/kata-containers/kata-containers/pull/12756 now?

packages/by-name/kata/runtime/package.nix
```diff
@@ -137,6 +137,12 @@ buildGoModule (finalAttrs: {
       # Deny unknown fields where possible to ease migration. This isn't possible where flatten is used.
       # TODO: upstream.
       ./0021-runtime-rs-deny-unknown-fields-in-config.patch
+
+      # Use virtio-blk with serial name for initdata
+      # Our initdata-processor expects the initdata device to be present at /dev/disks/by-label/initdata,
+      # which requires the device to have a stable name. Using virtio-blk with a serial number achieves this.
+      # TODO: check if we can improve the situation upstream or implement a fallback in the initdata-processor.
```
burgerdev, line 161: I created https://github.com/kata-containers/kata-containers/issues/12764 for this.

nodeinstaller/internal/kataconfig/runtime_rs.go
```diff
@@ -20,10 +24,15 @@ var (
...
-func extraRuntimeConfig(config Config) Config {
-	// runtime-rs specific configuration.
+func extraRuntimeConfig(config Config, platform platforms.Platform) Config {
 	config.Runtime["name"] = "virt_container"
 	config.Runtime["hypervisor_name"] = "qemu"
 	config.Runtime["agent_name"] = "kata"
+	config.Runtime["experimental"] = []string{"force_guest_pull"}
+
+	config.Agent["kata"]["dial_timeout_ms"] = 1000
+	config.Agent["kata"]["reconnect_timeout_ms"] = 600000
```
burgerdev: This is likely a debug leftover, right? If not, please leave a comment for why it needs to be that high.

.github/workflows/e2e_runtime_rs.yml
```diff
@@ -0,0 +1,48 @@
...
+      fail-fast: false
+    name: "${{ matrix.platform.name }}"
+    uses: ./.github/workflows/e2e.yml
+    with:
+      skip-undeploy: false
+      test-name: ${{ matrix.test-name }}
+      platform: ${{ matrix.platform.name }}
+      runner: ${{ matrix.platform.runner }}
+      self-hosted: ${{ matrix.platform.self-hosted }}
+      debug-shell: true
```
burgerdev: Don't we need to change the `SET`?

https://github.com/edgelesssys/contrast/blob/be62ea5888eed10ab654dba6a96632740c1605df/.github/workflows/e2e.yml#L54-L59

nodeinstaller/internal/kataconfig/runtime_rs.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func extraRuntimeConfig(config Config, platform platforms.Platform) Config {
+	config.Runtime["name"] = "virt_container"
+	config.Runtime["hypervisor_name"] = "qemu"
+	config.Runtime["agent_name"] = "kata"
+	config.Runtime["experimental"] = []string{"force_guest_pull"}
+
+	config.Agent["kata"]["dial_timeout_ms"] = 1000
+	config.Agent["kata"]["reconnect_timeout_ms"] = 60000
```
charludo, line 34: [seems still applicable](https://github.com/edgelesssys/contrast/pull/1909/#discussion_r3020601294)

overlays/sets/runtime-rs.nix
charludo, line 1: Would you like me to start working on the multi-set thing we talked about in the retro? Seems to become relevant once this is merged

nodeinstaller/internal/kataconfig/runtime_rs.go
burgerdev: I need to check the semantics - if it dials in a loop and each dial times out quickly that could be entirely correct, but that would be a significant change between runtime-go and runtime-rs.

overlays/sets/runtime-rs.nix
burgerdev: You mean for combining `debug` and `runtime-rs`? I don't think it's urgent, until we want to actually change the default runtime. From what I've seen during test runs, we're not exactly there yet.

charludo (approved):
- c45af0d8c0f940e05ec4517385090ce662b0e554 LGTM; 
- 92492110eea34bbc332c7db11621efcfdd16e698 LGTM;
- aac487d3607a6f53e845dbe3fe968225a8565ce8 LGTM;
- f85e5d1a2307d189588b1190c8803e5ea8a74517 LGTM.

nodeinstaller/internal/kataconfig/runtime_rs.go
burgerdev: Yep, the semantics changed such that the `reconnect_timeout_ms` defines how long individual connection attempts with `dial_timeout_ms` are tried, so this is actually the same (or close enough) as runtime-go.

burgerdev (approved):
I did some very minor changes in @katexochen's commits which are not worthy of re-review by a second team member.

burgerdev:
Current status: working on SNP, not working on TDX due to RTMR mismatch.

PR #1910, opened 2025-11-10
ci: collect discrete events, not pretty aggregates

burgerdev:
The default output of `kubectl events` is aggregated over similar kinds of events. This loses information about the exact timing of an event, which makes the default rendering pretty much useless for debugging. If the output format is yaml, all individual events will be reported back.


katexochen (approved):
noice!

PR #1912, opened 2025-11-11
kata.runtime: support full DeploymentSpec, JobSpec

katexochen:


e2e/regression/regression_test.go
```diff
@@ -119,30 +119,31 @@ func TestRegression(t *testing.T) {
...
+					t.Cleanup(func() {
+						if err := cleanupResource(t.Context(), resource, ct); err != nil {
+							t.Logf("failed to delete resource: %s:", err)
+						}
+					})
+					ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(3*time.Minute))
+					defer cancel()
+					switch r := resource.(type) {
+					case *applyappsv1.DeploymentApplyConfiguration:
+						require.NoError(c.WaitForDeployment(ctx, ct.Namespace, *r.Name))
```
burgerdev: Now that this is wrapped in its own subtest, the `require.NoError`s here apply to the wrong (outer) test.

PR #1925, opened 2025-11-14
docs: use reported TCB in manifest

katexochen:
In case provisional firmware is deployed, snphost will show a higher platform TCB than reported TCB. As the platform operator can rollback from provisional firmware anytime to the committed firmware version, Contrast relies on the committed firmware version at the time of guest launch (launch TCB) in verification. Reported TCB should point to a committed firmware version and is used to derive the key that signs the report, so we should be using that instead of the actual platform TCB.


burgerdev (approved):
thx

PR #1931, opened 2025-11-17
e2e: add debugshell to e2e tests

davidweisse:
This allows setting the `--insecure-enable-debug-shell-access` option for e2e tests, which gets passed through to `contrast generate`. This respects the `debug` variable when using `just e2e` or the associated option when running a manual e2e test. Additionally, the test matrices for the nightly and PR tests are extended to include all tests with and without the debug option.

### DIFF
--- a/e2e/internal/kubeclient/kubeclient.go
+++ b/e2e/internal/kubeclient/kubeclient.go
@@ -21,6 +21,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_on_pull_request.yml
```diff
@@ -23,6 +23,8 @@ jobs:
   test_matrix:
     strategy:
       matrix:
+        platform: [Metal-QEMU-SNP, Metal-QEMU-SNP-GPU, Metal-QEMU-TDX]
+        debug-shell: [true, false]
```
katexochen: Not sure we want to add this dimension on every PR, thoughts @burgerdev ?

.github/workflows/e2e.yml
```diff
@@ -43,6 +46,7 @@ jobs:
       PLATFORM: ${{ inputs.platform }}
       NODE_INSTALLER_TARGET_CONF_TYPE: ${{ inputs.platform == 'Metal-QEMU-SNP' && 'none' || 'k3s' }}
       TEST_NAME: ${{ inputs.test-name }}
+      DEBUG_SHELL: ${{ inputs.debug-shell }}
```
katexochen, line 52: we need to add this to the name of the artifact upload so the names won't clash, see https://github.com/edgelesssys/contrast/actions/runs/19429399776/job/55584462487?pr=1931#step:12:81

.github/workflows/e2e_on_pull_request.yml
katexochen: Ok, we decided to use only **true** on PR.

e2e/internal/kubeclient/kubeclient.go
```diff
@@ -165,33 +166,19 @@ func (c *Kubeclient) Exec(ctx context.Context, namespace, pod string, argv []str
...
-	})
-
-	return buf.String(), errBuf.String(), err
+	if len(podList.Items) == 0 {
+		return "", "", fmt.Errorf("pod not found: %s/%s", namespace, pod)
+	}
+	if len(podList.Items[0].Spec.Containers) == 0 {
+		return "", "", fmt.Errorf("pod %s/%s has no containers", namespace, pod)
+	}
+	return c.ExecContainer(ctx, namespace, pod, podList.Items[0].Spec.Containers[0].Name, argv)
```
burgerdev, line 181: Couldn't we just 

```go
c.ExecContainer(ctx, namespace, pod, "", argv)
```

burgerdev (approved):
lgtm % comment, but it's anyway good to consolidate the two execs.

e2e/internal/kubeclient/kubeclient.go
davidweisse: Isn't this essentially the same thing as before? Same issue here.

burgerdev: I wrongly  assumend that "pick the first main container" is the API server default, but it seems to be implemented in kubectl only.

PR #1932, opened 2025-11-17
tools/imagepuller-benchmark: set fixed limits

charludo:
Instead of reading/writing the moving baseline of the last successful benchmark to S3, set fixed limits for resource usage in the tool itself.

Values for time and memory were taken [from this failed run](https://github.com/edgelesssys/contrast/actions/runs/19421964985/job/55560871657), then increased somewhat generously, with the understanding that these values are not targets we aim for, but the limits beyond which we want to check what went wrong.

The storage values are generated by the new `imagepuller-benchmark-update-sizes` script, which `docker pulls` and `docker inspects` the images, then adds a 20% margin to the uncompressed size of the image in MB. 20% is very generous here.

### DIFF
--- a/.github/workflows/imagepuller-benchmark.yml
+++ b/.github/workflows/imagepuller-benchmark.yml
@@ -26,11 +26,6 @@ jobs:
         with:
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
-      - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@00943011d9042930efac3dcd3a170e4273319bc8 # v5.1.0
-        with:
-          role-to-assume: arn:aws:iam::795746500882:role/ContrastPublicBucketRW
-          aws-region: eu-central-1
       - name: Log in to ghcr.io Container registry
         uses: docker/login-action@5e57cd118135c172c3672efd75eb46360885c0ef # v3.6.0
         with:
@@ -44,15 +39,11 @@ jobs:
           [registries."ghcr.io."]
           auth = "$auth"
           EOF
-      - name: Download baseline JSON from S3
-        run: |
-          aws s3 cp "s3://contrast-public/imagepuller-benchmark/baseline.json" "baseline.json"
       - name: Run imagepuller-benchmark script
         run: |
-          imagepullereval="$(nix build .#imagepuller-benchmark --print-out-paths)/bin/imagepuller-benchmark"
+          imagpullerbenchmark="$(nix build .#imagepuller-benchmark --print-out-paths)/bin/imagepuller-benchmark"
           imagepuller="$(nix build .#imagepuller --print-out-paths)/bin/imagepuller"
-          sudo "$imagepullereval" "$imagepuller" -c baseline.json -o baseline-new.json -i tools/imagepuller-benchmark/images.txt -a imagepuller.toml
-      - name: Upload new baseline.json to S3
-        if: github.event_name == 'push' && github.ref_name == 'main'
-        run: |
-          aws s3 cp "baseline-new.json" "s3://contrast-public/imagepuller-benchmark/baseline.json"
+          sudo "$imagpullerbenchmark" \
+            "$imagepuller" \
+            ./tools/imagepuller-benchmark/benchmark.json \
+            -a imagepuller.toml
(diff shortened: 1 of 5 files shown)


.github/workflows/imagepuller-benchmark.yml
```diff
@@ -44,15 +39,8 @@ jobs:
...
         run: |
-          imagepullereval="$(nix build .#imagepuller-benchmark --print-out-paths)/bin/imagepuller-benchmark"
+          imagpullerbenchmark="$(nix build .#imagepuller-benchmark --print-out-paths)/bin/imagepuller-benchmark"
           imagepuller="$(nix build .#imagepuller --print-out-paths)/bin/imagepuller"
-          sudo "$imagepullereval" "$imagepuller" -c baseline.json -o baseline-new.json -i tools/imagepuller-benchmark/images.txt -a imagepuller.toml
-      - name: Upload new baseline.json to S3
-        if: github.event_name == 'push' && github.ref_name == 'main'
-        run: |
-          aws s3 cp "baseline-new.json" "s3://contrast-public/imagepuller-benchmark/baseline.json"
+          sudo "$imagpullerbenchmark" "$imagepuller" -l tools/imagepuller-benchmark/limits.json -o baseline-new.json -i tools/imagepuller-benchmark/images.txt -a imagepuller.toml
```
katexochen: nit: break long line

tools/imagepuller-benchmark/limits.json
katexochen, line 1: Should we use the same document for images & limits?

.github/workflows/imagepuller-benchmark.yml
katexochen: I think we can remove `-o baseline-new.json` here and in the go code?

tools/imagepuller-benchmark/limits.json
```diff
@@ -0,0 +1,22 @@
+{
+  "continuous": {
+    "time": 120,
+    "memory": 100,
+    "storage": 8000
```
katexochen: Please document where these values are coming from, ideally with how we can derive them from the image size.

.github/workflows/imagepuller-benchmark.yml
charludo: I think I'd like to keep it in the go code, to have a quick and easy way to generate a limits file. But yes, that it's still in the yml is an oversight.

tools/imagepuller-benchmark/limits.json
charludo: Done. Also updated the description of the PR with more details.

packages/scripts.nix
```diff
@@ -730,4 +730,36 @@ lib.makeScope pkgs.newScope (scripts: {
       fi
     '';
   };
+
+  imagepuller-benchmark-update-sizes = writeShellApplication {
```
katexochen, line 736: Please add a comment what this does and how it is intended to be used.

packages/scripts.nix
```diff
@@ -730,4 +730,36 @@ lib.makeScope pkgs.newScope (scripts: {
...
+
+      keys=$(jq -r 'keys[]' "$TMP_FILE")
+      for key in $keys; do
+          has_image=$(jq -r --arg k "$key" '.[$k] | has("image")' "$TMP_FILE")
+          if [[ "$has_image" == "true" ]]; then
+              image=$(jq -r --arg k "$key" '.[$k].image' "$TMP_FILE")
+              docker pull "$image" > /dev/null
+              size=$(docker inspect -f "{{ .Size }}" "$image")
+              size_mb=$(( size / 1024 / 1024 ))
+              size_mb_with_margin=$(( (size_mb * 120) / 100 ))
```
katexochen: 120 here is what we have as tolerance? It should be a constant at the beginning of the script.

charludo:
@katexochen fixed the review comments. Also noticed that I forgot to update the "continuous" storage limit, fixed as well.

Tested the storage requirements with ext4 and btrfs, storage limits are correct for both of those, looks like this is just an xfs issue 😅

tools/imagepuller-benchmark/benchmark.json
```diff
@@ -0,0 +1,25 @@
+{
+  "continuous": {
+    "time": 120,
```
burgerdev: Shouldn't `continuous.time` be the sum of the individual `$img.time` values?

PR #1937, opened 2025-11-18
just: inject imagepuller config

charludo:
Reads GH token from GCP, and uses it to create the secrets file in `just populate`.

We could just store the entire config in gcp, but this way it's easy to adjust and extend with additional configuration.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,13 +106,19 @@ If you are struggling with your git history during the review process, please as
 
    Additionally, you may want to add the [VSCode extension](https://github.com/direnv/direnv-vscode).
 
-4. Execute and follow instructions of
+4. Log into GCP.
+
+    ```sh
+    nix shell nixpkgs#google-cloud-sdk -c gcloud auth login
+    ```
+
+5. Execute and follow instructions of
 
     ```sh
     just onboard
     ```
 
-5. Provision a CoCo enabled AKS cluster with
+6. Provision a CoCo enabled AKS cluster with
 
     ```sh
     just create
--- a/packages/scripts.nix
+++ b/packages/scripts.nix
@@ -402,8 +402,7 @@ lib.makeScope pkgs.newScope (scripts: {
     ];
     text = ''
       set -euo pipefail
-      # this is the officially recommended way to get the raw output...
-      gcloud secrets versions access "$1" --format='get(payload.data)' | tr '_-' '/+' | base64 -d
+      gcloud secrets versions access "$1"
     '';
   };
 


packages/scripts.nix
```diff
@@ -394,6 +394,19 @@ lib.makeScope pkgs.newScope (scripts: {
     '';
   };
 
+  # Usage: get-read-token $gcloudSecretRef
+  get-read-token = writeShellApplication {
```
katexochen:
```suggestion
  get-ghcr-read-token = writeShellApplication {
```

justfile
```diff
@@ -103,6 +103,16 @@ populate target=default_deploy_target platform=default_platform:
     if [[ -f ./{{ workspace_dir }}/deployment/deployment.yml ]]; then
         echo "---" >> ./{{ workspace_dir }}/deployment/deployment.yml
     fi
+    if [ -n "$CONTRAST_GHCR_READ" ]; then
```
katexochen:
```suggestion
    if [[ -n "$CONTRAST_GHCR_READ" ]]; then
```

justfile
```diff
@@ -395,6 +413,8 @@ namespace_suffix=""
 CONTRAST_CACHE_DIR="./workspace.cache"
 # Log level for the CLI.
 CONTRAST_LOG_LEVEL=""
+# A Github token with read access to Contrast's ghcr.io packages.
+CONTRAST_GHCR_READ=""
```
katexochen: should be lower case

justfile
```diff
@@ -335,12 +345,20 @@ get-credentials platform=default_platform:
...
 # Load the kubeconfig from the dev cluster.
 get-credentials-dev:
+    #!/usr/bin/env bash
+    set -euo pipefail
     nix run -L .#scripts.get-credentials "projects/796962942582/secrets/hetzner-ax162-snp-kubeconfig/versions/latest"
     sed -i 's/^default_platform=.*/default_platform="Metal-QEMU-SNP"/' justfile.env
     sed -i 's/^node_installer_target_conf_type=.*/node_installer_target_conf_type="k3s"/' justfile.env
 
+    token=$(nix run -L .#scripts.get-read-token "projects/796962942582/secrets/ghcr-read-token/versions/latest")
+    sed -i "s/^CONTRAST_GHCR_READ=.*/CONTRAST_GHCR_READ=\"${token}\"/" justfile.env
```
katexochen: this looks like the wrong place for it. Suggesting to move it to its own target, and the use that in onboarding.

justfile
```diff
@@ -395,6 +413,8 @@ namespace_suffix=""
 CONTRAST_CACHE_DIR="./workspace.cache"
 # Log level for the CLI.
 CONTRAST_LOG_LEVEL=""
+# A Github token with read access to Contrast's ghcr.io packages.
```
katexochen: Should state this is filled automatically and what target to call to refresh it.

justfile
```diff
@@ -103,6 +103,16 @@ populate target=default_deploy_target platform=default_platform:
...
     fi
+    if [ -n "$CONTRAST_GHCR_READ" ]; then
+        cat > "./{{ workspace_dir }}/contrast-imagepuller.toml" <<EOF
+    [registries]
+    [registries."ghcr.io."]
+    auth = "$(printf "user-not-required-here:%s" "$CONTRAST_GHCR_READ" | base64 -w0)"
+    EOF
+        kubectl create secret generic contrast-node-installer-imagepuller-config \
+            --from-file "contrast-imagepuller.toml"="./{{ workspace_dir }}/contrast-imagepuller.toml" \
+            --namespace {{ target }}${namespace_suffix-}
```
katexochen: This shouldn't happen in populate but in apply (runtime).

justfile
charludo: What exactly? github / contrast?

katexochen: s/CONTRAST_GHCR_READ/ghcr_read_token/

uppercase should be used for stuff that is consumed as env, this is rather a variable you are processing in justfile steps.

justfile
charludo: Moved to own target, but not sure if calling it in `just onboard` is worth it: `gcloud` would need to be added to the devshell to be able to test if the person is even logged into gcloud yet, and to the login flow otherwise. I'd rather add a sentence to the onboarding docs.

katexochen: Mentioning in the onboarding guide that you should login to gcloud sounds like a good idea. I don't want to have this in the get-credentials targets as these are called frequently and it will lead to a further slow down.

charludo: > I don't want to have this in the get-credentials targets as these are called frequently and it will lead to a further slow down.

Understood, removed it from there.

justfile
charludo: Seems like there's currently no access to the namespace in apply (just.namespace does not exist yet). What's the reason this should not happen in populate?

katexochen: Separation of concerns mostly. I don't expect populate to interact with kubernetes, and I want to be able to re-run apply runtime without re-running populate, as that would overwrite changes made to the resources in the workspace.

CONTRIBUTING.md
```diff
@@ -112,7 +112,21 @@ If you are struggling with your git history during the review process, please as
...
+    nix shell nixpkgs#google-cloud-sdk -c gcloud auth login
+    ```
+
+6. Retrieve the GitHub token allowing you to read private Contrast packages from ghcr.io.
+
+    ```sh
+    just get-ghcr-read-token
+    ```
+
+    The token will be saved into your `justfile.env` file.
```
katexochen: How is this better than log into gcp first and then do just onboard and include get-ghcr-read-token there?

justfile
charludo: Makes sense, thx. Moved the writing of the namespace to a separate recipe, secret creation is in apply runtime now.

packages/scripts.nix
```diff
@@ -394,6 +394,19 @@ lib.makeScope pkgs.newScope (scripts: {
...
+  # Usage: get-ghcr-read-token $gcloudSecretRef
+  get-ghcr-read-token = writeShellApplication {
+    name = "get-ghcr-read-token";
+    runtimeInputs = with pkgs; [
+      google-cloud-sdk
+    ];
+    text = ''
+      set -euo pipefail
+      # this is the officially recommended way to get the raw output...
+      gcloud secrets versions access "$1" --format='get(payload.data)' | tr '_-' '/+' | base64 -d
```
burgerdev: Invoking without a format flag seems to work fine?

```console
$ gcloud secrets versions access projects/796962942582/secrets/ghcr-read-token/versions/latest
ghp_[redacted}
```

charludo: Indeed. Looks like I misinterpreted the docs after getting a "Take the gcloud survey today!  prompt after the ouput and was looking to avoid that.

katexochen (approved):

PR #1941, opened 2025-11-18
treefmt: ensure docs don't contain non-ASCII chars

burgerdev:


treefmt.nix
```diff
@@ -62,11 +62,10 @@
       command = "${lib.getExe pkgs.contrastPkgs.scripts.check-sidebar}";
       includes = [ "docs/sidebars.js" ];
     };
-    em-dash-lint = {
-      command = "${lib.getExe pkgs.contrastPkgs.scripts.em-dash-lint}";
+    ascii-lint = {
+      command = "${lib.getExe pkgs.contrastPkgs.scripts.ascii-lint}";
       includes = [
         "docs/docs/*.md"
-        "README.md"
```
katexochen, line 69: Why replace the emojis in the readme _and_ remove it here from includes?

burgerdev: That was a mistake, I first removed the README but then decided to fix it instead. Do you have a preference either way?

katexochen: better include it.

burgerdev: done

katexochen (approved):

PR #1945, opened 2025-11-19
renovate: trigger via dashboard, disable some updates

katexochen:


renovate.json5
```diff
@@ -1,15 +1,15 @@
 {
   $schema: "https://docs.renovatebot.com/renovate-schema.json",
   extends: [
+    "config:recommended",
```
burgerdev: The `config:recommended` moves around between commits.

katexochen: Yes, reordered it as that is the base, and the following stuff is modification of it.

burgerdev: I meant to say that 1ef51159a68dc9f96f24c7f36d633df024bec4a7 moves it up while d5852f4ca365bb872584fbc6bc95d18831ab40ba moves it down again. Is that what you intended?

katexochen: Ahh, got it. I wanted to move it up, but I guess the formatter is sorting the lists.

katexochen: not rebasing to fix this to save the ci cycles for the release

PR #1948, opened 2025-11-19
rfc: application level verification

burgerdev:
Rendered: https://github.com/edgelesssys/contrast/blob/rfc/012/rfc/012-application-level-verification.md

### DIFF
--- a/rfc/012-application-level-verification.md
+++ b/rfc/012-application-level-verification.md
@@ -9,6 +9,7 @@ This requires users to be able to establish a direct TCP connection to the Coord
    - Coordinator doesn't benefit from protocol-level mitigations and access controls.
 2. Can't be used with a TLS-intercepting box-in-the-middle.
 3. Might trip systems that analyze TLS handshakes due to unconventional use of ALPN and certificate extensions.
+4. Can't be used in environments that don't offer flexible TLS handshake APIs, such as a web browser.
 
 ## Requirements
 


rfc/012-application-level-verification.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Background
+
+Contrast's `verify` subcommand is currently implemented as an RPC method served over aTLS.
+This requires users to be able to establish a direct TCP connection to the Coordinator, which comes with the following drawbacks:
+
+1. Contrast Coordinator needs to be exposed directly to the user, likely over the public internet.
+   - Coordinator doesn't benefit from protocol-level mitigations and access controls.
+2. Can't be used with a TLS-intercepting box-in-the-middle.
+3. Might trip systems that analyze TLS handshakes due to unconventional use of ALPN and certificate extensions.
```
msanft, line 11:
```suggestion
3. Might trip systems that analyze TLS handshakes due to unconventional use of ALPN and certificate extensions.
4. Can't be used in environments that don't offer flexible TLS handshake APIs, such as a web browser.
```

rfc/012-application-level-verification.md
```diff
@@ -0,0 +1,112 @@
...
+```go
+type VerifyResponse struct {
+    RawAttestationDoc []byte // serialized TDX or SNP report (the same format as currently provided to validators)
+
+    // The following fields match the fields of userapi.GetManifestsResponse
+
+    Manifests [][]byte // Manifest history, the current manifest being last.
+    Policies  map[manifest.HexString][]byte // Policies referred to by manifests.
+    RootCA    []byte // PEM-encoded certificate
+    MeshCA    []byte // PEM-encoded certificate
```
burgerdev, line 53: We also have 

https://github.com/edgelesssys/contrast/blob/dda33d507293c2291e1c9072b3554434d2f29909/sdk/verify.go#L99-L109

Might be good to consolidate this into a shared type.

rfc/012-application-level-verification.md
```diff
@@ -0,0 +1,113 @@
...
+## Alternatives considered
+
+### Expected history instead of expected manifest
+
+Some clients may be interested in validating the entire history instead of just the current manifest and _some_ history.
+However, some clients may also not care about the history, or it may be hard to communicate the entire history out of band.
+This could be a future _addition_ to the SDK, but is mostly unrelated to the proposal made here.
+
+As a workaround for now, the history check can be implemented outside the SDK (since it's just a byte-for-byte comparison of slice items).
+This check can be made even simpler if the Contrast deployment is only ever expected to have one manifest in the history (Privatemode.ai, for example).
```
thomasten, line 102: If I `--disable-updates`, I don't need to check anything myself, right?

rfc/012-application-level-verification.md
```diff
@@ -0,0 +1,113 @@
...
+Then the client creates a list of validators from the expected manifest, validates the attestation report with the expected `REPORTDATA`.
+If validation passes, the client knows:
+
+- The attestation report was created for them, due to the nonce being included in `REPORTDATA` (requirement 3.2).
+- The Coordinator runs the expected software in a TEE, because it passed validation according to the manifest (requirement 3.1).
+
+After that, the client checks that the last manifest in the response is exactly equal to the expected manifest.
+Since we already know that the Coordinator is running the correct software, the client can implicitly assume that the Coordinator
+
+- sent the manifest history up to including the currently enforced manifest (requirements 3.3, 3.4)
```
katexochen, line 84: Can't the client verify this via the transition chain?

rfc/012-application-level-verification.md
burgerdev: `--disable-updates` results in a manifest without workload owner keys. However, that state can also be reached by just setting a manifest without workload owner keys on an existing coordinator history. If you care about that (which you only need to if you're relying on workload secrets, imho), you should validate that the history has only one manifest.

PR #1950, opened 2025-11-20
fix: always attach imagepuller-config device; empty if no config specified

charludo:
- only set imgpuller config path in runtime config if the secret/file exist
- always attach (empty) imagepuller device
- update rtmr_0 input hashes

### DIFF
--- a/tools/tdx-measure/rtmr/rtmr.go
+++ b/tools/tdx-measure/rtmr/rtmr.go
@@ -221,9 +221,9 @@ func CalcRtmr0(firmware []byte) ([48]byte, error) {
 	// These are the hashes for some fw_cfg/ACPI related measurements.
 	// TODO(freax13): Don't hard-code these, calculate them instead.
 	configHashes := []string{
-		"dbb32c5a195f60df71564cdb220d8058b096f91c962a8cce7d0653686124c4ba3a7f3a67cfd60246fce8372cbcb5ecfb",
-		"91cde1273d7cd58821d8a1e70ca10daa30f068c371447800ee18131f180661df3589e5fa259bb3b1b40e162444c341e2",
-		"b0d0bc2648a85f47642e9e4ad20a966ac723b36e16f297dd2cce70b19a717855636086db513ce6ed56ffc42ffb13b5bb",
+		"2685cabde42188ea89029e3f717eee76108b4251a80ad011e5ebeaccc62fdd6636d64ee3ae6b0da7e374e379578c4e06",
+		"c40177551aa9c6034f27ae8764a4b19a4689f2d9ab0f15650f95a4742f1ffa3a87ba247f1dda0cf583603ec56aeda8e9",
+		"bae56cb4bba81b167ccb2c9a5a442b0f0bdcd894dad5c22a61b717e66db39fe863eb97c7cc07803008ba37e65b74c255",
 	}
 	for _, hash := range configHashes {
 		var buffer [48]byte


nodeinstaller/node-installer.go
```diff
@@ -91,17 +91,24 @@ func run(ctx context.Context, fetcher assetFetcher, platform platforms.Platform)
...
+	if err := writeContainerdRuntimeConfig(targetConf.KataConfigPath(), kataRuntimeConfig); err != nil {
+		return fmt.Errorf("writing kata runtime configuration: %w", err)
 	}
 
-	if err := installImagepullerConfig(hostMount, runtimeHandler); err != nil {
+	imagpullerConfigPath, err := installImagepullerConfig(hostMount, runtimeHandler)
+	if err != nil {
 		return fmt.Errorf("installing imagepuller config: %w", err)
 	}
+	kataRuntimeConfig.Hypervisor["qemu"]["contrast_imagepuller_config"] = imagpullerConfigPath
```
katexochen: I don't think this should be exposed at this level. Can't you just pass the `imagepullerConfigPath` into the `containerdRuntimeConfig` func and skip adding it to the kata config if it is the empty string?

charludo: SGTM

packages/by-name/kata/runtime/0020-runtime-pass-imagepuller-config-device-to-vm.patch
```diff
@@ -70,23 +70,26 @@ index 1bcbb03e6f1a166c092c1a3a9b180b617a7eeb8b..f3f9869bce2ace9375017a61ce0f8421
...
-+	} else if err != nil {
-+		return err
-+	}
++		logger.Info("No imagepuller config provided. Writing empty config.")
++	} else {
++		if _, err := os.Stat(config.ImagepullerConfigPath); errors.Is(err, os.ErrNotExist) {
++			logger.
++				WithField("imagepuller_config_path", config.ImagepullerConfigPath).
++				Info("Imagepuller config file does not exist. Skipping prepare imagepuller config device")
++			return nil
```
katexochen: This case doesn't make sense anymore. We should return the error in this case, as we expect the file to be present if the annotation is set.

charludo: Yeah makes sense.

nodeinstaller/internal/kataconfig/config.go
```diff
@@ -60,7 +61,7 @@ func KataRuntimeConfig(
 		config.Hypervisor["qemu"]["image"] = filepath.Join(baseDir, "share", "kata-containers.img")
 		config.Hypervisor["qemu"]["rootfs_type"] = "erofs"
 		config.Hypervisor["qemu"]["valid_hypervisor_paths"] = []string{filepath.Join(baseDir, "tdx", "bin", "qemu-system-x86_64")}
-		config.Hypervisor["qemu"]["contrast_imagepuller_config"] = filepath.Join(baseDir, "etc", "host-config", "contrast-imagepuller.toml")
+		config.Hypervisor["qemu"]["contrast_imagepuller_config"] = imagepullerConfigPath
```
katexochen, line 64:
```suggestion
    if imagepullerConfigPath != "" {
		    config.Hypervisor["qemu"]["contrast_imagepuller_config"] = imagepullerConfigPath
		}
```

charludo: Why? `imgpullerConfigPath` is `""` unless the file exists, and we check for string-emptyness in the runtime patch

charludo: This way the `config.Hypervisor["qemu"]["contrast_imagepuller_config"]` also turns up in the expected unittest data

katexochen (approved):

burgerdev (approved):

PR #1954, opened 2025-11-20
docs: point out limitations with registry auth

charludo:


docs/docs/howto/registry-authentication.md
```diff
@@ -72,6 +73,13 @@ This may result in an unexpected leak of sensitive information.
 
 :::
 
+#### Limitations
+
+Currently, it's only possible to specify a single image puller auth configuration, since only a single secret named `contrast-node-installer-imagepuller-config` can exist.
```
katexochen: the limitation here is a single config _per runtime class_, right? For the broader case, you got the instructions further down, so its not a limitation.

docs/docs/howto/registry-authentication.md
```diff
@@ -72,6 +73,13 @@ This may result in an unexpected leak of sensitive information.
...
+Conversely, if this secret is present on a node, every node install will use it, granting deployments access to the configured registries.
+This also extends to every newly installed runtime class.
+If you wish to prevent this for a runtime class, you can change the `secretName` of the `DaemonSet` configuration in `runtime/runtime.yml` to a different name before `apply`-ing the node installer.
```
katexochen:
```suggestion
If you wish to use a different secret for a specific runtime class, change the `secretName` of the `DaemonSet` configuration in `runtime/runtime.yml` to a different name before applying the node installer, and put the image puller auth configuration into a secret with that name.
```

docs/docs/howto/registry-authentication.md
charludo: I mean, you can technically go ahead and write the secret file into the runtime dir manually, in which case, yes. 

But since the secret needs to go into the `kube-system` namespace for the "automatic" placement of the secrets file in the runtime dir, only a single one can exist on the node / if you re-run the nodeinstaller, your manually placed file will be overridden by the nodeinstaller if the `contrast-node-installer-imagepuller-config` exists.

charludo: Edit: ah, I see what you mean. If you additionally change the secret name, then one per runtime class is supported, yes.

katexochen: But you describe further down how you can have different secrets _for different runtime classes_.

docs/docs/howto/registry-authentication.md
```diff
@@ -72,6 +73,15 @@ This may result in an unexpected leak of sensitive information.
 
 :::
 
+#### Limitations
+
+Currently, it's only possible to specify a single image puller auth configuration per runtime class, since only a single secret named `contrast-node-installer-imagepuller-config` can exist.
```
katexochen: Still wrong.

```suggestion
Currently, it's only possible to specify a single image puller auth configuration per runtime class, since only a single secret path can be configured for a runtime class.
```

PR #1956, opened 2025-11-20
history: share outside of coordinator

burgerdev:
This PR moves the `history` package into the common internal directory so that the SDK can make use of the definitions, in particular the `Transition` object. What I'm trying to enable is something like the following code for #1948: 

```go
func buildTransitionChain(manifests [][]byte) []*history.Transition {
	transitions := make([]*history.Transition, 0, len(manifests))
	lastTransitionHash := [history.HashSize]byte{}
	for _, m := range manifests {
		t := &history.Transition{
			PreviousTransitionHash: lastTransitionHash,
			ManifestHash: history.Digest(m),
		}
		transitions = append(transitions, t)
		lastTransitionHash = t.Digest()
	}
	return transitions
}
```

### DIFF
--- a/internal/history/history.go
+++ b/internal/history/history.go
@@ -25,11 +25,12 @@ const (
 	HashSize = sha256.Size
 )
 
+var hashFun func() hash.Hash = sha256.New
+
 // History is the history of the Coordinator.
 type History struct {
-	store   Store
-	hashFun func() hash.Hash
-	log     *slog.Logger
+	store Store
+	log   *slog.Logger
 }
 
 // New creates a new History that uses the default storage backend.
@@ -56,12 +57,8 @@ func New(log *slog.Logger) (*History, error) {
 // NewWithStore creates a new History with the given storage backend.
 func NewWithStore(log *slog.Logger, store Store) *History {
 	h := &History{
-		store:   store,
-		hashFun: sha256.New,
-		log:     log,
-	}
-	if HashSize != h.hashFun().Size() {
-		panic("mismatch between hashSize and hash function size")
+		store: store,
+		log:   log,
 	}
 	return h
 }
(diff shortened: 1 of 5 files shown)


internal/history/history.go
```diff
@@ -209,24 +206,24 @@ func (h *History) getContentaddressed(pathFmt string, hash [HashSize]byte) ([]by
...
 	hashStr := hex.EncodeToString(hash[:])
 	if err := h.store.Set(fmt.Sprintf(pathFmt, hashStr), data); err != nil {
 		return [HashSize]byte{}, err
 	}
 	return hash, nil
 }
 
-func (h *History) hash(in []byte) [HashSize]byte {
-	hf := h.hashFun()
+func Digest(in []byte) [HashSize]byte {
```
charludo, line 226:
```suggestion
// Digest calculates the Digest of a given slice in the same way as used for (Latest)Transitions..
func Digest(in []byte) [HashSize]byte {
```

internal/history/history.go
```diff
@@ -256,6 +253,10 @@ func (t *Transition) marshalBinary() []byte {
 	return data
 }
 
+func (t *Transition) Digest() [HashSize]byte {
```
charludo, line 258:
```suggestion
// Digest returns the digest of the Transition.
func (t *Transition) Digest() [HashSize]byte {
```

internal/history/history.go
```diff
@@ -283,6 +284,10 @@ func (l *LatestTransition) marshalBinary() []byte {
 	return data
 }
 
+func (l *LatestTransition) Digest() [HashSize]byte {
```
charludo, line 290:
```suggestion
// Digest returns the digest of the LatestTransition.
func (l *LatestTransition) Digest() [HashSize]byte {
```

PR #1958, opened 2025-11-21
coordinator: expose `verify` HTTP endpoint

charludo:


coordinator/internal/verify/verify.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Handler implements http.Handler for POST /verify.
+type Handler struct {
+	Issuer     atls.Issuer
+	StateGuard *stateguard.Guard
+}
+
+func (h *Handler) getResponse(ctx context.Context, req *Request) (*sdk.CoordinatorState, error) {
+	// state knows the lates transition
```
burgerdev:
```suggestion
	// state knows the latest transition
```

coordinator/internal/verify/verify.go
```diff
@@ -0,0 +1,131 @@
...
+	Nonce []byte `json:"nonce"`
+}
+
+// Handler implements http.Handler for POST /verify.
+type Handler struct {
+	Issuer     atls.Issuer
+	StateGuard *stateguard.Guard
+}
+
+func (h *Handler) getResponse(ctx context.Context, req *Request) (*sdk.CoordinatorState, error) {
```
burgerdev: Since we're separating request validation from request processing, I think it would make sense to pass only the nonce as input here.

coordinator/internal/verify/verify.go
```diff
@@ -0,0 +1,131 @@
...
+	if len(req.Nonce) != 32 {
+		log.Printf("invalid nonce length: %d", len(req.Nonce))
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	ctx := r.Context()
+	resp, err := h.getResponse(ctx, &req)
+	if err != nil {
+		log.Printf("calculating response: %v", err)
```
burgerdev: We want to fail the request at this point.

coordinator/internal/verify/verify.go
```diff
@@ -0,0 +1,131 @@
...
+	Issuer     atls.Issuer
+	StateGuard *stateguard.Guard
+}
+
+func (h *Handler) getResponse(ctx context.Context, req *Request) (*sdk.CoordinatorState, error) {
+	// state knows the lates transition
+	state, err := h.StateGuard.GetState(ctx)
+	switch {
+	case errors.Is(err, stateguard.ErrNoState):
+		return nil, status.Error(codes.FailedPrecondition, userapi.ErrNoManifest.Error())
```
burgerdev: Instead of the `status.Error`s, which are for gRPC, we should work with HTTP error codes here.

coordinator/internal/verify/verify.go
```diff
@@ -0,0 +1,131 @@
...
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	// Limit size to a small value to avoid abuse (nonce only expected).
+	defer r.Body.Close()
+	body, err := io.ReadAll(io.LimitReader(r.Body, 1024))
+	if err != nil {
+		log.Printf("reading request body: %v", err)
+		w.WriteHeader(http.StatusBadRequest)
```
burgerdev: I think it would be helpful if we sent the error message along with the response code.

sdk/verify.go
```diff
@@ -99,11 +99,13 @@ func (Client) Verify(expectedManifest []byte, manifestHistory [][]byte) error {
...
-	Policies [][]byte
+	Policies [][]byte `json:"policies"`
 	// PEM-encoded certificate of the deployment's root CA.
-	RootCA []byte
+	RootCA []byte `json:"root_ca"`
 	// PEM-encoded certificate of the deployment's mesh CA.
-	MeshCA []byte
+	MeshCA []byte `json:"mesh_ca"`
+	// RawAttestationDoc is a raw attestation report.
+	RawAttestationDoc []byte `json:"raw_attestation_doc"`
```
burgerdev: This does not belong into the `CoordinatorState`, as it's specific to the application-level validation. Instead, the response object should probably be a type defined in the `verify` package, holding both the attestation as well as a `CoordinatorState`.

charludo: Am I correct in the assumption, that in the type

```go
type Response struct {
	// RawAttestationDoc is a raw attestation report.
	RawAttestationDoc []byte `json:"raw_attestation_doc"`

	sdk.CoordinatorState
}
```

I should be able to access 
```
resp := &Response{}
resp.Manifest = manifest
```
? I thought the embedded struct's keys were promoted to the parent struct. But this doesn't seem to be working in this case (?), so I guess we'd have to use `resp.CoordinatorState.Manifests`?

coordinator/internal/verify/verify.go
```diff
@@ -0,0 +1,138 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	enc := json.NewEncoder(w)
+	enc.SetEscapeHTML(false)
+	if err := enc.Encode(resp); err != nil {
+		writeJSONError(w, http.StatusInternalServerError, "encoding response", err)
+	}
+}
+
+func writeJSONError(w http.ResponseWriter, status int, msg string, err error) {
+	errMsg := fmt.Sprintf("%s: %v", msg, err)
```
burgerdev: The `msg` argument does not contribute an awful lot, so maybe just use `err.Error()` as the field value.

PR #1964, opened 2025-11-24
scripts: do not flag newlines as non-ASCII

charludo:
Not sure if I'm the only one experiencing this, but since #1941, `nix fmt` / `just fmt` flags *every single line in the docs` as non-ASCII, because it contains a newline. ~~Switched the grep to use a Perl regex. Range should be as it previously was, plus the newline.~~
~~Using ripgrep also seems to fix the issue.~~
It's a locale issue.

### DIFF
--- a/packages/scripts.nix
+++ b/packages/scripts.nix
@@ -720,6 +720,8 @@ lib.makeScope pkgs.newScope (scripts: {
     name = "ascii-lint";
     runtimeInputs = [ pkgs.gnugrep ];
     text = ''
+      # The range expression below is locale-dependent, we want it to work in standard ASCII.
+      export LC_ALL="C"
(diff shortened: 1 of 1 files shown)


katexochen:
Can't reproduce the issue.

packages/scripts.nix
```diff
@@ -720,6 +720,7 @@ lib.makeScope pkgs.newScope (scripts: {
     name = "ascii-lint";
     runtimeInputs = [ pkgs.gnugrep ];
     text = ''
+      export LC_ALL="C"
```
burgerdev, line 724:
```suggestion
      # The range expression below is locale-dependent, we want it to work in standard ASCII.
      export LC_ALL="C"
```

PR #1966, opened 2025-11-25
coordinator: keep HTTP API objects internal

burgerdev:
* Reword  endpoint `/verify` to `/attest`, because the endpoint itself does not verify anything. Rename packages and structs accordingly.
* Introduces the notion of an HTTP API and moves the packages accordingly. For now, the only API is "attest".
* Move HTTP API objects to an internal, shareable location (these should not be exposed to SDK users). Note that this should not affect intended use of the SDK, because the internal objects can still be obtained with SDK functions. SDK users will not be able to create their own API objects, though.
* Same for the reportdata constructor, which is an implementation detail that should not be exposed.
* Add a version field to response objects and a custom unmarshalling function to aid debugging unparseable responses.

### DIFF
--- a/internal/httpapi/attest.go
+++ b/internal/httpapi/attest.go
@@ -4,6 +4,10 @@
 package httpapi
 
 import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
 	"github.com/edgelesssys/contrast/internal/history"
 )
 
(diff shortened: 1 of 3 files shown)


internal/httpapi/attest.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, errors.Join(
+			origErr,
+			fmt.Errorf("determining server version failed: %w", err),
+		)
+	}
+	switch version := unstructured["version"].(type) {
+	case string:
+		return nil, errors.Join(
+			origErr,
+			fmt.Errorf("server reports Contrast version: %q", version),
```
charludo: might be nice to directly give the client's Contrast version, and/or not return this eror if they match, as to not confuse the user with an unrelated error.

internal/httpapi/attest_test.go
```diff
@@ -0,0 +1,78 @@
...
+			wantErr: "999.0.1",
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			resp, err := UnmarshalAttestationResponse([]byte(tc.resp))
+
+			if tc.wantErr != "" {
+				assert.ErrorContains(err, tc.wantErr)
```
charludo: A custom error type would be nicer than checking for contained string

internal/httpapi/attest.go
burgerdev: That's not going to be too helpful, since the version is fixed in the module as seen by Go, and only replaced in our Nix builds:

https://github.com/edgelesssys/contrast/blob/4919ebe854e512772a6b92d24c3d0e3098952278/internal/constants/constants.go#L12

I'd expect such a parsing failure to be extremely rare - since we're allowing unknown fields, the only way we can get it is if we're carelessly reusing fields with different semantics. However, what I just realized is that the user might be better off just having the full body in the error message - wdyt?

burgerdev: > However, what I just realized is that the user might be better off just having the full body in the error message - wdyt?

I'm going to implement that at as logs at the call site.

charludo: sounds good to me

internal/httpapi/attest_test.go
burgerdev: Done

charludo (approved):
🎉

PR #1967, opened 2025-11-25
packages: don't overwrite main program

katexochen:
In https://github.com/NixOS/nixpkgs/commit/befb2beebd2382aa9a6274b0fd609fbd872d8cc8, upstream started to include meta.mainProgram in the derivation as env, breaking the contract that meta attributes don't affect the build output. This caused Contrast packages that were built as part of contrast.contrast to rebuild (and run the unit test) separately every time something changed, instead of just overwriting the mainProgram so it could be invoked directly. Thus removing the mainProgram overwrite. We should reduce dependency on that mechanism further in the future.

### DIFF
--- a/.github/actions/release_artifacts/action.yml
+++ b/.github/actions/release_artifacts/action.yml
@@ -126,7 +126,7 @@ runs:
       run: |
         mkdir -p workspace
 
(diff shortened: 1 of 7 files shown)


charludo (approved):
Interesting upstream change.

PR #1977, opened 2025-11-25
tdx: support mainline host kernel / QEMU 

msanft:
This PR implements support for using Contrast on TDX machines with a mainline (6.16+) host kernel and QEMU package (v10.1+). ~~Consequentially, it allows us to drop the TEE-platform-specific `qemu-{tdx,snp}` packages and switch to a single, unified `qemu-cc` package.~~

Please see the individual commit messages for details on the changes made in this PR. I recommend to take a commit-by-commit review approach here.

Since this change requires users to run TDX with a 6.16+ host kernel (e.g., Ubuntu 25.10), a follow-up PR for the user-facing documentation will be made explaining the setup process for such a machine. (which is far less involved than the current one relying on the `canonical/tdx` repository)

### DIFF
--- a/packages/by-name/qemu-tdx/0001-avoid-duplicate-definitions.patch
+++ b/packages/by-name/qemu-tdx/0001-avoid-duplicate-definitions.patch
@@ -1,6 +1,6 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Tom Dohrmann <erbse.13@gmx.de>
-Date: Fri, 20 Sep 2024 16:01:15 +0200
+Date: Wed, 10 Jul 2024 13:30:51 +0200
 Subject: [PATCH] avoid duplicate definitions
 
 Another library already defines crc32c and this causes linker errors.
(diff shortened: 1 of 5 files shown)


tools/tdx-measure/eventlog/eventlog.go
```diff
@@ -0,0 +1,25 @@
...
+package eventlog
+
+import (
+	"fmt"
+	"os"
+)
+
+// Dir is the directory where event logs are stored.
+const Dir = "eventlogs"
+
```
burgerdev: This should be an optional flag input, writing nothing if not present.

tools/tdx-measure/tdvf/tdvf.go
```diff
@@ -121,22 +122,20 @@ func CalculateMrTd(firmware []byte) ([48]byte, error) {
 	for _, section := range sections {
 		data := firmware[section.DataOffset:][:section.RawDataSize]
 		shouldExtend := section.Attributes&mrExtend != 0
+		fmt.Fprintf(os.Stderr, "Adding section: type=%d, gpa=0x%x, size=%d, extend=%v\n",
+			section.Type, section.MemoryAddress, section.MemoryDataSize, shouldExtend)
```
burgerdev, line 126: Should we keep this although we have the eventlog?

dev-docs/coco/tdx-measurements.md
```diff
@@ -129,3 +129,58 @@ This contains a measurement for the kernel command line and the initrd, in that
...
+kprobe:tdh_mr_extend { printf("TDH_MR_EXTEND gpa=0x%llx\n", arg1); }
+kprobe:tdh_mr_finalize { printf("TDH_MR_FINALIZE\n"); }
+' > mrtd.log
+```
+
+Note that this, to keep the event log at a manageable size, omits the actual data
+measured in the `TDH_MR_EXTEND` calls. To debug a mismatch there, it's best to
+compare the full-on TDVF blob.
+
+The generated event log has the exact same format than the one produced by the
```
burgerdev: typo: `the exact same format as`.

dev-docs/coco/tdx-measurements.md
```diff
@@ -129,3 +129,58 @@ This contains a measurement for the kernel command line and the initrd, in that
 ### RTMR 3
 
 Reserved, all-0 at the moment of writing.
+
+## MRTD
+
+The MRTD is a special case, since it's a static measurement. It can only be populated
+from the TDX module *while* a TD (TDX VM) is being created, that is, in the time span
+between VM creation and VM launch.
```
burgerdev: Nit: this addition uses different indentation settings than the existing contents. My personal preference would be one sentence per line, because that is the easiest to keep intact and to review changes to.

tools/tdx-measure/tdvf/tdvf.go
msanft: I purposefully opted for keeping this, as it puts the MRTD calculation in line with the RTMR calculation, which also emits coarse logs on what it's measuring to stderr (i.e., the build output). This allows for debugging trivial issues, such as a completely wrong TDVF blob at a glance, e.g. only by looking at the output of a CI job.

tools/tdx-measure/tdvf/tdvf.go
```diff
@@ -121,22 +122,20 @@ func CalculateMrTd(firmware []byte) ([48]byte, error) {
...
 		}
+	}
 
-		if section.Attributes&mrExtend != 0 {
-			data := firmware[section.DataOffset:][:section.RawDataSize]
-			err = launchContext.ExtendRegion(section.MemoryAddress, data)
-			if err != nil {
-				return [48]byte{}, fmt.Errorf("can't extend region: %w", err)
-			}
-		}
```
katexochen, line 133: where did this go? Is that related to the debugging?

msanft: No, it's the actual `TDH_MR_EXTEND` call. This is the two-pass change I've mentioned:
https://github.com/edgelesssys/contrast/blob/ba1c60603eae18403963bb9615b3cec54809ba73/tools/tdx-measure/tdvf/tdvf.go#L123-L128
https://github.com/edgelesssys/contrast/blob/ba1c60603eae18403963bb9615b3cec54809ba73/tools/tdx-measure/mrtd/mrtd.go#L119-L127

katexochen: Maybe it landed in the wrong commit then?

msanft: Yes, sorry. It should actually have been removed in 78360551c7f5966647c9822367c1677289a10cbc. Will fix this

dev-docs/coco/tdx-measurements.md
msanft: Tried to find a good middle ground between keeping column length equal to the other stuff in this document while trying to stay at one sentence per line.

msanft:
Reverted the QEMU merge and bumped `qemu-tdx` instead. PTAL @katexochen @burgerdev

PR #1978, opened 2025-11-26
e2e: adjust memory requirements

charludo:
For the `bash` image, we expect <= 15MB in storage usage. Seems the previous limits were just a bit to tight.

### DIFF
--- a/e2e/imagestore/imagestore_test.go
+++ b/e2e/imagestore/imagestore_test.go
@@ -21,42 +21,26 @@ import (
 	"github.com/edgelesssys/contrast/internal/platforms"
 
 	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestImageStore(t *testing.T) {
 	tests := map[string]struct {
 		name       string
-		size       int
-		maxOffset  int
 		annotation string
-		gpu        bool
 	}{
 		"enabled by default": {
-			name:      "imagestore-default",
-			size:      10000000,
-			maxOffset: 500000,
+			name: "imagestore-default",
 		},
 		"block device size configurable": {
 			name:       "imagestore-configured",
-			size:       2000000,
-			maxOffset:  500000,
 			annotation: "2Gi",
 		},
 		"disabled through annotation": {
-			name: "imagestore-disabled",
-			// (512MB + 2 * 20MB) / 2: half of the memory is available in the /run tmpfs
-			size:       226000,
-			maxOffset:  20000,
+			name:       "imagestore-disabled",
 			annotation: "0",
-		},
-		"disabled through annotation (GPU)": {
-			name: "imagestore-disabled-gpu",
-			// roughly (1024MB + 2 * 20MB) / 2
-			size:       500000,
-			maxOffset:  40000,
-			annotation: "0",
-			gpu:        true,
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


e2e/regression/testdata/job/job.yml
```diff
@@ -25,4 +25,9 @@ spec:
         - name: test
           image: ghcr.io/edgelesssys/bash@sha256:cabc70d68e38584052cff2c271748a0506b47069ebbd3d26096478524e9b270b
           command: ["echo", "Hello World!"]
+          resources:
+            requests:
+              memory: 40Mi
+            limits:
+              memory: 40Mi
```
burgerdev, line 32: There are more regression test pods that don't have resource limits, could you please fix them, too?

charludo: done

e2e/imagestore/imagestore_test.go
```diff
@@ -45,16 +45,16 @@ func TestImageStore(t *testing.T) {
 		},
 		"disabled through annotation": {
 			name: "imagestore-disabled",
-			// (512MB + 2 * 20MB) / 2: half of the memory is available in the /run tmpfs
-			size:       226000,
-			maxOffset:  20000,
+			// (512MB + 2 * 40MB) / 2: half of the memory is available in the /run tmpfs
+			size:       246000,
+			maxOffset:  30000,
```
burgerdev: As discussed privately, we should calculate the size here in kiB, so that we can compare it with the `df` output. 

Another thing is that we probably need to account for the actual RAM available to the VM is considerably lower than what's given to qemu.

We can follow these two things in separate PRs, though.

burgerdev (approved):

e2e/imagestore/imagestore_test.go
charludo: Adjusted to be the expected KiB. Also adjusted the maxOffsets... 👀

e2e/imagestore/imagestore_test.go
```diff
@@ -45,16 +45,16 @@ func TestImageStore(t *testing.T) {
 		},
 		"disabled through annotation": {
 			name: "imagestore-disabled",
-			// (512MB + 2 * 20MB) / 2: half of the memory is available in the /run tmpfs
-			size:       226000,
-			maxOffset:  20000,
+			// (512MB + 2 * 40MB) / 2 / 1024 * 1000: half of the memory is available in the /run tmpfs, in KiB
```
burgerdev: In order to get from MiB to kiB, we need to multiply by 1024, don't we? And the actual value seems to be computed as `(512 + 2 * 40) / 2 / 1024 * 1000^2`.

charludo: The conversion must go from Megabyte to Kibibyte though, right? and 1MB = ~976KiB

charludo: Ah, no, the comment is just wrong about the unit! The actual resource limits are Mebibyte as well. Then, yes, I'll adjust the values and the comments.

burgerdev:
One more thing: I think it would be best if we actually did the calculation in the test, instead of just plugging in the calculated values. There is `platforms.DefaultMemoryInMegaBytes`, at least.

e2e/imagestore/imagestore_test.go
```diff
@@ -33,31 +35,16 @@ func TestImageStore(t *testing.T) {
 		gpu        bool
```
burgerdev: This is now unused, I believe.

e2e/imagestore/imagestore_test.go
```diff
@@ -90,16 +77,31 @@ func TestImageStore(t *testing.T) {
...
+			switch tc.annotation {
+			case "0":
+				defaultMemory := platforms.DefaultMemoryInMebiBytes(platform) * 1024 * 1024
+				for _, c := range pod.Spec.Containers {
+					memory := c.Resources.Limits[corev1.ResourceMemory]
+					defaultMemory += int(memory.Value())
+				}
+				// When the imagestore is disabled, the expectedKibiBytes are (the total VM memory in KiB - internal VM overhead) / 2
+				// For more information on the internal VM overhead, see https://github.com/edgelesssys/contrast/pull/1196
+				expectedKibiBytes = ((defaultMemory / 1024) - 99) / 2
```
burgerdev: This assumes a  _99kiB_ overhead, right? Shouldn't we use something closer to 200MiB?

charludo: Argh! Yes, that was supposed to be 99MiB. THen the 10% will also do just fine (actually, the match will be pretty close)

Why 200? From the linked PR, I took the "Overhead VM" from the TDX table for 50 MiB Container limit, that seemed to be the closest match.

charludo: Just did a test run with a 1% tolerance and the test passed. I think the 99MiB are right on the money :D

burgerdev: Note that the default memory in that table was 256MiB, but now is 512MiB, so I was looking at the rows near 1G total VM memory.

edgelessci:
Successfully created backport PR for `release/v1.15`:
- #1982

PR #1979, opened 2025-11-26
sdk: allow offline verification of evidence

burgerdev:
The SDK can now be used to first fetch a self-contained set of attestation evidence from a plain HTTP server, and later validate that evidence without a direct connection to the Coordinator (possibly even on another machine). The main use case for this is relaying attestation evidence through proxies (corporate firewall) and reverse proxies (Coordinator does not need to be exposed directly to verifiers).

**NOTE: the Contrast SDK is still considered an unstable API and not recommended for general use.**

Usage example (most error handling omitted for clarity):

```go
client := sdk.New()

// Get attestation evidence.

nonce, _ := crypto.GenerateRandomBytes(32)
serializedAttestation, _ := client.GetAttestation(context.TODO(), "http://coordinator:1314", nonce)

// Validate attestation evidence.
// In order to validate on a different machine, only the nonce and the serializedAttestation are required.

kdsCacheDir, _ := os.MkdirTemp("", "contrasttest-kds-cache-dir-")
state, err := client.ValidateAttestation(context.TODO(), kdsCacheDir, nonce, serializedAttestation)
if err != nil {
	log.Fatalf("Validation failed! %w", err)
}

// Verify that the manifest in use matches expectations.

// This should come from a trusted channel.
var expectedManifest []byte
manifest := state.Manifests[len(state.Manifests)-1]
if !bytes.Equal(expectedManifest, manifest) {
	log.Fatalf("Unexpected manifest!")
}
```

---

Note for developers:

This implements the client side of [RFC 012](https://github.com/edgelesssys/contrast/blob/9c04b31c7a4c0efd8666413b85db3f8fb16387b1/rfc/012-application-level-verification.md). Although it's a couple of commits, the diff happens to come in a convenient order so that it should be reviewable in one go. What you're looking at:

1. Two prerequisite fixes to make it work:
   1. The server side implementation was slightly wrong, using the latest digest instead of the transition digest latest points to.
   2. There was a small bug in contrasttest, where the CA certificates loaded after recovery were not the ones that were _recovered to_, but the ones from the last set. I'm fixing this by loading the certs from the `verify` subdir.
2. Integration of HTTP attestation into the standard verify flow. This allows comparing the results of both verification types, spares us another e2e test binary and switching to HTTP verification by default should be easier. **The code in contrasttest also serves as an SDK usage reference, for now.**
3. The user-facing sdk functions. 
   1. The new functions are designed to be opaque to the caller, so that we have more flexibility extending the API. This is why we're passing around byte slices instead of proper structs.
   2. In order to be able to request the attestation in one system and validate it in another, we're providing two separate functions and leave nonce handling to the caller.
   3. We want to allow flexibility for fetching the attestation: sdk users can bring their own HTTP client to account for proxies, TLS settings, etc. 
   4. The actual verification of the manifest (history) is left to the sdk user, which should allow them more flexibility. The prior validation can be implemented with a simple equality check on the expected manifest.

### DIFF
--- a/sdk/verify.go
+++ b/sdk/verify.go
@@ -174,7 +174,7 @@ func (c Client) GetAttestation(ctx context.Context, url string, nonce []byte) ([
 // Note: this function does not verify manifest content! It's the callers responsibility to compare
 // the latest manifest with an expected manifest, if that exists, or verify that all manifest
 // fields match their expectations.
-func (c Client) ValidateAttestation(ctx context.Context, kdsDir string, nonce []byte, attestation []byte) (*httpapi.CoordinatorState, error) {
+func (c Client) ValidateAttestation(ctx context.Context, kdsDir string, nonce []byte, attestation []byte) (*CoordinatorState, error) {
 	if len(nonce) != contrastcrypto.RNGLengthDefault {
 		return nil, fmt.Errorf("wrong nonce length: got %d, want %d", len(nonce), contrastcrypto.RNGLengthDefault)
 	}
@@ -224,7 +224,25 @@ func (c Client) ValidateAttestation(ctx context.Context, kdsDir string, nonce []
 	if !validated {
 		return nil, fmt.Errorf("validation failed:\n%w", errors.Join(errs...))
 	}
-	return &resp.CoordinatorState, nil
+	state := CoordinatorState{
+		Manifests: resp.Manifests,
+		Policies:  resp.Policies,
+		RootCA:    resp.RootCA,
+		MeshCA:    resp.MeshCA,
+	}
+	return &state, nil
+}
+
+// CoordinatorState represents the state of the Contrast Coordinator at a fixed point in time.
+type CoordinatorState struct {
+	// Manifests is a slice of manifests. It represents the manifest history of the Coordinator it was received from, sorted from oldest to newest.
+	Manifests [][]byte
+	// Policies contains all policies that have been referenced in any manifest in Manifests. Used to verify the guarantees a deployment had over its lifetime.
+	Policies [][]byte
+	// PEM-encoded certificate of the deployment's root CA.
+	RootCA []byte
+	// PEM-encoded certificate of the deployment's mesh CA.
+	MeshCA []byte
 }
 
 func buildTransitionChain(manifests [][]byte) []*history.Transition {
--- a/sdk/verify_test.go
+++ b/sdk/verify_test.go
@@ -198,7 +198,15 @@ func TestValidateAttestation(t *testing.T) {
 				return
 			}
 			assert.NoError(err)
-			assert.Equal(&tc.resp.CoordinatorState, state)
+
+			expected := &CoordinatorState{
+				Manifests: tc.resp.Manifests,
+				Policies:  tc.resp.Policies,
+				RootCA:    tc.resp.RootCA,
+				MeshCA:    tc.resp.MeshCA,
+			}
+
+			assert.Equal(expected, state)
 		})
 	}
 }


sdk/common.go
```diff
@@ -18,11 +18,17 @@ import (
...
+var validatorsFromManifestOverride func(*certcache.CachedHTTPSGetter, *manifest.Manifest, *slog.Logger) ([]atls.Validator, error)
+
 // ValidatorsFromManifest returns a list of validators corresponding to the reference values in the given manifest.
 // Originally an unexported function in the contrast CLI.
 // Can be made unexported again, if we decide to move all userapi calls from the CLI to the SDK.
 // Validators MUST NOT be used concurrently.
 func ValidatorsFromManifest(kdsGetter *certcache.CachedHTTPSGetter, m *manifest.Manifest, log *slog.Logger) ([]atls.Validator, error) {
+	if validatorsFromManifestOverride != nil {
+		return validatorsFromManifestOverride(kdsGetter, m, log)
+	}
```
katexochen: Would be cleaner to transform the function into a method to perform the injection.

sdk/verify.go
```diff
@@ -96,3 +105,129 @@ func (Client) Verify(expectedManifest []byte, manifestHistory [][]byte) error {
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewBuffer(body))
+	if err != nil {
+		return nil, fmt.Errorf("constructing HTTP request: %w", err)
+	}
+	req.GetBody = func() (io.ReadCloser, error) { return io.NopCloser(bytes.NewBuffer(body)), nil }
+
+	client := http.DefaultClient
+	if c.HTTPClient != nil {
+		client = c.HTTPClient
+	}
```
katexochen, line 138: nit
```suggestion
	if c.HTTPClient == nil {
		c.HTTPClient = http.defaultClient
	}
```

sdk/verify.go
```diff
@@ -96,3 +105,129 @@ func (Client) Verify(expectedManifest []byte, manifestHistory [][]byte) error {
...
+	body, err := json.Marshal(&httpapi.AttestationRequest{Nonce: nonce})
+	if err != nil {
+		return nil, fmt.Errorf("creating request body: %w", err)
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewBuffer(body))
+	if err != nil {
+		return nil, fmt.Errorf("constructing HTTP request: %w", err)
+	}
+	req.GetBody = func() (io.ReadCloser, error) { return io.NopCloser(bytes.NewBuffer(body)), nil }
```
katexochen: Why is this needed?

burgerdev: To support HTTP redirects: https://cs.opensource.google/go/go/+/master:src/net/http/client.go;l=531-537.

sdk/verify.go
burgerdev: My thinking here was to avoid surprises:

* We could be surprised if the client field is modified during execution of this method.
* The caller could be surprised that they left the field nil, but now it's non-nil.

sdk/common.go
burgerdev: You mean something like 

```go
func (c Client) getValidators(kdsGetter *certcache.CachedHTTPSGetter, m *manifest.Manifest, log *slog.Logger) ([]atls.Validator, error) {
  if c.overrideValidator != nil {
    return c.overrideValidator()
  }
  return ValidatorsFromManifest(kdsGetter, m, log)
}
```
?

katexochen: Usually we do this by having a constructor that sets the real thing and override the field in the test, don't we?

sdk/verify.go
katexochen: Docs sound like this should be automatically set: https://pkg.go.dev/net/http#Client.Do (second last par)

sdk/common.go
burgerdev: I guess I was confused by "transform the function into a method". Is that what you mean?

```go
type Client struct {
  validatorGetter func(kdsGetter *certcache.CachedHTTPSGetter, m *manifest.Manifest, log *slog.Logger)
}

func New() Client {
  return &Client{validatorGetter: ValidatorsFromManifest}
}
```

If not, I'd appreciate if you could propose a change.

sdk/verify.go
burgerdev: Thanks, I did not realize that NewRequest inspects types!

sdk/common.go
katexochen: Exactly, I think that's the pattern we usually use.

sdk/verify.go
```diff
@@ -96,3 +105,129 @@ func (Client) Verify(expectedManifest []byte, manifestHistory [][]byte) error {
...
+// The input for this function should be the nonce passed into GetAttestation and the byte slice
+// returned by it.
+//
+// If this function returns nil, validation passed and the caller can rely on the state.MeshCA
+// issuing certificates according to the last entry of state.Manifests.
+//
+// Note: this function does not verify manifest content! It's the callers responsibility to compare
+// the latest manifest with an expected manifest, if that exists, or verify that all manifest
+// fields match their expectations.
+func (c Client) ValidateAttestation(ctx context.Context, kdsDir string, nonce []byte, attestation []byte) (*httpapi.CoordinatorState, error) {
```
thomasten: It's inconvenient for users of this func that it returns a struct from an internal package.

burgerdev: Yeah, let me fix that.

sdk/common.go
burgerdev: Done

sdk/verify.go
burgerdev: Done

sdk/verify.go
burgerdev: Done

PR #1985, opened 2025-12-01
kuberesource: fix AddImageStore with absent metadata

burgerdev:
Automatic injection of the image pull store was not working correctly without a pod-level metadata object in the Kubernetes resource definition. This PR fixes the bug and adds a regression test for common example resources.

* [x] [regression test](https://github.com/edgelesssys/contrast/actions/runs/19815771040)

### DIFF
--- a/tools/lychee/config.toml
+++ b/tools/lychee/config.toml
@@ -14,6 +14,7 @@ exclude = [
     '^http(s)?://localhost',
     '^https://github\.com/edgelesssys/contrast/(settings|deployments)', # not public
     '^https://.*\.hashicorp\.com', # strict bot protection
+    '^https://.*\.medium\.com', # strict bot protection
 ]
 exclude_path = [
     "dev-docs/frozen",


internal/kuberesource/mutators_test.go
```diff
@@ -416,3 +430,27 @@ func TestAddServiceMesh(t *testing.T) {
...
+		"pod":        {resource: pod},
+	} {
+		t.Run(name, func(t *testing.T) {
+			require := require.New(t)
+
+			res, err := UnmarshalApplyConfigurations(tc.resource)
+			require.NoError(err)
+			res = PatchRuntimeHandlers(res, "contrast-cc")
+			res = AddImageStore(res)
+			encoded, err := EncodeResources(res...)
```
charludo, line 451: Am I correct in assuming the encoding is done because converting the `[]any` to the different resource types is a hassle?

burgerdev: Exactly. I don't mean for this to be an exhaustive test (we should still add that), but rather a smoke detector.

charludo: Makes sense!

PR #1988, opened 2025-12-01
e2e: regression test for containerd bug

charludo:
I'm really sorry this took so long, but the reproducer is finally reliable.

Some notes:
- the reproducer container is not built through nix. We *need* impurity here, and none of the hacks I considered to get this working were worth it. Instead, the new script does essentially the same as `pushOCIDir ( toOciImage ( dockerTools.buildImage ) )`, but causes a new digest and tag on every run
- one image is sufficient to trigger the bug: when the image is first pulled by tag in the `runc` context, containerd remembers the image by its tag. When we then pull by digest, containerd correctly identifies the image as already being present, but uses the old, tag-based name for it, causing the policy failure shown below (see `"io.kubernetes.cri.image-name": "ghcr.io/charludo/contrast/containerd-reproducer:1764579373"`)
- not all hosts are containerd >= 2.0 yet I believe, so some test failures are expected

```
failed to create containerd task: failed to create shim task: "CreateContainerRequest is blocked by policy: /run/measured-cfg/policy.rego:119: allow_create_container_input: input = {\"OCI\": {\"Annotations\": {\"io.katacontainers.pkg.oci.bundle_path\": \"/run/k3s/containerd/io.containerd.runtime.v2.task/k8s.io/f30eb8920d04dce75766bc5318be61282191384d9bdb5d33389ebc5462f09cd1\", \"io.katacontainers.pkg.oci.container_type\": \"pod_container\", \"io.kubernetes.cri.container-name\": \"cc-by-digest\", \"io.kubernetes.cri.container-type\": \"container\", \"io.kubernetes.cri.image-name\": \"ghcr.io/charludo/contrast/containerd-reproducer:1764579373\", \"io.kubernetes.cri.sandbox-id\": \"a9c04fb64800ec3bce3b80b08390549c6423715ca5b50bc2b5847203ea9af621\", \"io.kubernetes.cri.sandbox-name\": \"containerd-digest-pinning-cc-67fdf47b68-n9fl9\", \"io.kubernetes.cri.sandbox-namespace\": \"testcontainerddigestpinning-e30f1510-ch\", \"io.kubernetes.cri.sandbox-uid\": \"b97016e1-6053-4719-9924-55cbc4302120\"}, ...
```

### DIFF
--- a/e2e/containerd-digest-pinning/containerd-digest-pinning_test.go
+++ b/e2e/containerd-digest-pinning/containerd-digest-pinning_test.go
@@ -1,0 +1,73 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build e2e
+
+package containerddigestpinning
+
+import (
+	"context"
+	"flag"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/edgelesssys/contrast/e2e/internal/contrasttest"
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/platforms"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestContainerdDigestPinning(t *testing.T) {
+	platform, err := platforms.FromString(contrasttest.Flags.PlatformStr)
+	require.NoError(t, err)
+	ct := contrasttest.New(t)
+
+	runtimeHandler, err := manifest.RuntimeHandler(platform)
+	require.NoError(t, err)
+
+	deploymentName := "containerd-digest-pinning"
+	runcTester, ccTester := kuberesource.ContainerdDigestPinningTesters(deploymentName)
+	ctx, cancel := context.WithTimeout(t.Context(), 2*time.Minute)
+	t.Cleanup(cancel)
+
+	// Start the runcTester outside the CC context.
+	ct.Init(t, []any{runcTester})
+	_, err = ct.Kubeclient.Client.AppsV1().
+		Deployments(ct.Namespace).
+		Apply(
+			ctx,
+			runcTester,
+			metav1.ApplyOptions{
+				FieldManager: "e2e-test",
+			},
+		)
+	require.NoError(t, err)
+	err = ct.Kubeclient.WaitForDeployment(ctx, ct.Namespace, *runcTester.Name)
+	require.NoError(t, err)
+
+	resources := kuberesource.CoordinatorBundle()
+	resources = append(resources, ccTester)
+	resources = kuberesource.PatchRuntimeHandlers(resources, runtimeHandler)
+	resources = kuberesource.AddPortForwarders(resources)
+
+	ct.Init(t, resources)
+
+	require.NoError(t, err)
+	require.True(t, t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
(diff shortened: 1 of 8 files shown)


e2e/containerd-digest-pinning/containerd-digest-pinning_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	ct.Init(t, resources)
+
+	require.NoError(t, err)
+	require.True(t, t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
+	require.True(t, t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
+	require.True(t, t.Run("set", ct.Set), "contrast set needs to succeed for subsequent tests")
+	require.True(t, t.Run("contrast verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
+
+	t.Cleanup(cancel)
```
burgerdev: This should go right after where `cancel` is returned.

e2e/containerd-digest-pinning/containerd-digest-pinning_test.go
```diff
@@ -0,0 +1,77 @@
...
+	ccPods, err := ct.Kubeclient.PodsFromDeployment(ctx, ct.Namespace, *ccTester.Name)
+	require.NoError(t, err)
+	require.Len(t, ccPods, 1)
+	require.NoError(t, ct.Kubeclient.WaitForPod(ctx, ct.Namespace, ccPods[0].Name))
```
burgerdev: Why not use `ct.Kubeclient.WaitForDeployment`? (same above)

packages/scripts.nix
```diff
@@ -784,4 +784,25 @@ lib.makeScope pkgs.newScope (scripts: {
...
+      skopeo
+    ];
+    text = ''
+      registry="$1"
+      tmpdir=$(mktemp -d)
+      timestamp=$(date +%s)
+      cat e2e/containerd-digest-pinning/Dockerfile
+      docker build --build-arg TIMESTAMP="$timestamp" -t containerd-reproducer:"$timestamp" \
+        ./e2e/containerd-digest-pinning
+      docker save containerd-reproducer:"$timestamp" -o "$tmpdir/image.tar"
```
burgerdev: I would prefer if we were able to do this without a docker daemon, for example with umoci: https://umo.ci/quick-start/workflow/.

packages/scripts.nix
```diff
@@ -784,4 +784,25 @@ lib.makeScope pkgs.newScope (scripts: {
...
+      skopeo copy docker-archive:"$tmpdir/image.tar" oci:"$tmpdir/oci-image" --insecure-policy
+      crane push "$tmpdir/oci-image" "$registry/contrast/containerd-reproducer:$timestamp"
```
burgerdev: You can get the digest directly from the OCI layout:

```
jq -r '.manifests[0].digest' </tmp/img/index.json
```

That should help avoid the regexp in the justfile and the dependency on crane (push with skopeo, too).

.github/workflows/e2e_regression.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           - name: Metal-QEMU-TDX
             runner: TDX
             self-hosted: true
-        test-name: [genpolicy-unsupported, regression]
+        test-name: [genpolicy-unsupported, regression, containerd-digest-pinning]
```
burgerdev: I think running this in nightly is sufficient.

charludo:
Succeeding as expected:
- [x] [palutena](https://github.com/edgelesssys/contrast/actions/runs/19849072582/job/56872064628)
- [x] [olimar](https://github.com/edgelesssys/contrast/actions/runs/19849077532/job/56872078809)

Failing as expected:
- [ ] [discovery](https://github.com/edgelesssys/contrast/actions/runs/19849080601/job/56872739986)

PR #1991, opened 2025-12-01
overlays: unpin edk2 for OVMF-TDX, update tdx-measure

katexochen:
This is possible as we previously unpinned qemu-tdx in #1977.

The newer OVMF measures additional events, so we adapt the tdx-measure tool and extend with the new events.

Open questions:

- [x] Why is the measurement of the kernel in RTMR 1 not reflecting the patches that were previously applied by QEMU?
- [ ] Why did the SMBIOS measurement only turn up now?


jkr0103:
>For some reason, the measurement now contains the unpatched kernel.

Reason: Qemu use to patch kernel setup header but with 10.1.0 version onwards, it's not done anymore, check the diference in file
`qemu/hw/i386/x86-common.c` between 10.1.0 and 9.xx or earlier.

I have added a scanshot which shows that difference.
x86-common

PR #1993, opened 2025-12-02
kuberesource: make debug-shell a sidecar

burgerdev:
If the debug shell is a regular container it will prevent the pod from completing, making our regression/job test fail.

* [x] [regression test w/ debug shell](https://github.com/edgelesssys/contrast/actions/runs/19852713312)


katexochen (approved):
Nice!

PR #1999, opened 2025-12-02
e2e: add kds-pcs-downtime test

davidweisse:
This E2E test tests the different scenarios for when the KDS/PCS is unavailable, documented in the dev-docs. We can block the KDS/PCS connection from the CLI by setting up an HTTPS proxy and blocking specific domains. To simulate unavailability on the issuer side (i.e., Coordinator), we override the `/etc/hosts` file with, e.g., `127.0.0.1 kdsintf.amd.com`. Simulating PCS unavailability for the issuer side is not possible, since all the interaction with the PCS/PCCS happens on the host side, not in the Go code. So for TDX, we only test PCS unavailability on the CLI side.

All the tests currently only test CLI <-> Coordinator connections. In the future, we may also want to test Coordinator <-> Initializer connections, where the Coordinator acts as the client.

SNP run: https://github.com/edgelesssys/contrast/actions/runs/19864351537/job/56922742153
TDX run: https://github.com/edgelesssys/contrast/actions/runs/19864358538/job/56922766667

### DIFF
--- a/e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
+++ b/e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
@@ -1,0 +1,214 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build e2e
+
+package kdspcsdowntime
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/edgelesssys/contrast/e2e/internal/contrasttest"
+	"github.com/edgelesssys/contrast/e2e/internal/kubeclient"
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/platforms"
+	"github.com/elazarl/goproxy"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	kdsAddr = "kdsintf.amd.com:443"
+	pcsAddr = "api.trustedservices.intel.com:443"
+)
+
+func TestKDSPCSDowntime(t *testing.T) {
+	platform, err := platforms.FromString(contrasttest.Flags.PlatformStr)
+	require.NoError(t, err)
+	ct := contrasttest.New(t)
+
+	runtimeHandler, err := manifest.RuntimeHandler(platform)
+	require.NoError(t, err)
+	resources := kuberesource.CoordinatorBundle()
+	resources = kuberesource.PatchRuntimeHandlers(resources, runtimeHandler)
+	resources = kuberesource.AddPortForwarders(resources)
+	ct.Init(t, resources)
+
+	proxy := goproxy.NewProxyHttpServer()
+	server := http.Server{Handler: proxy}
+	errCh := make(chan error)
+
+	// If set to true, connections to KDS and PCS will be blocked by the proxy.
+	var blockKDSPCS atomic.Bool
+	// connectionProxied will be set to true if the proxy performs an HTTP CONNECT to the address of KDS or PCS.
+	var connectionProxied atomic.Bool
+	proxy.ConnectDial = func(network string, addr string) (net.Conn, error) {
+		t.Logf("Proxying connection: %q", addr)
+		if (addr == kdsAddr || addr == pcsAddr) && blockKDSPCS.Load() {
+			t.Logf("Blocking connection to KDS/PCS %q", addr)
+			connectionProxied.Store(true)
+			return nil, fmt.Errorf("connection to KDS/PCS %q blocked by test proxy", addr)
+		}
+		return (&net.Dialer{}).DialContext(t.Context(), network, addr)
+	}
+
+	proxyListener, err := (&net.ListenConfig{}).Listen(t.Context(), "tcp", "127.0.0.1:")
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		require.NoError(t, server.Close())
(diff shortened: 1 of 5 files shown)


dev-docs/endorsement-caching.md
```diff
@@ -51,6 +51,17 @@ The following gives an overview over the request and caching structure we implem
...
+
+| Issuer can't reach KDS | Validator can't reach KDS | Issuer cache | Validator cache | Validation |
+| :--------------------: | :-----------------------: | :----------: | :-------------: | :--------: |
+|                        |                           | *            | *               | Success    |
+| X                      |                           |              | *               | Success    |
+|                        | X                         | *            |                 | Success    |
+| X                      | X                         |              |                 | Failure    |
+| X                      | X                         | CRL + VCEK   | *               | Success    |
+| X                      | X                         | *            | CRL + VCEK      | Success    |
+
```
burgerdev, line 69: In the `cache` columns, I don't quite understand the difference between empty, asterisk and CRL+VCEK. Maybe worthy of an explanation? Same for PCS.

e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	t.Setenv("https_proxy", proxyListener.Addr().String())
+
+	require.True(t, t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
+	require.True(t, t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
+
+	t.Run("kds downtime", func(t *testing.T) {
+		if !strings.Contains(ct.RuntimeClassName, "snp") {
+			t.Skip("KDS downtime test is only applicable to SEV-SNP workloads")
+		}
```
burgerdev: There's `platforms.Is(TDX|SNP)` for this.

e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+		require := require.New(t)
+
+		ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(3*time.Minute))
+		defer cancel()
+
+		c := kubeclient.NewForTest(t)
```
burgerdev: Do you need this when there is `ct.Kubeclient`?

e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
```diff
@@ -0,0 +1,209 @@
...
+		coordinatorPods, err = ct.Kubeclient.PodsFromOwner(ctx, ct.Namespace, "StatefulSet", "coordinator")
+		require.NoError(err)
+		require.NotEmpty(coordinatorPods, "pod not found: %s/%s", ct.Namespace, "coordinator")
+
+		// Block coordinator access to KDS.
+		_, stderr, err = c.Exec(ctx, ct.Namespace, coordinatorPods[0].Name, []string{"/bin/sh", "-c", "echo 127.0.0.1 kdsintf.amd.com >> /etc/hosts"})
+		require.NoError(err, "updating /etc/hosts: stderr: %q", stderr)
+
+		// Unblock CLI access to KDS.
+		blockKDSPCS.Store(false)
```
burgerdev, line 150: Is this actually necessary? I thought the Coordinator was sending its own VCEK with the certificate. Or is this due to the CRL?

e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
```diff
@@ -0,0 +1,209 @@
...
+		blockKDSPCS.Store(true)
+
+		// Set should fail because neither coordinator nor CLI can reach KDS and there is no cached data.
+		// Set loop considers context deadline exceeded from KDS as a retriable error.
+		// Lower the timeout so the set loop doesn't exceed the test timeout.
+		setCtx, setCancel := context.WithTimeout(ctx, ct.FactorPlatformTimeout(1*time.Minute))
+		defer setCancel()
+		err = ct.RunSet(setCtx)
+		t.Logf("Set error: %v", err)
+		require.Error(err)
```
burgerdev: I think we should be more specific about the error we're expecting here (same for TDX).

e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
```diff
@@ -0,0 +1,209 @@
...
+	server := http.Server{Handler: proxy}
+	errCh := make(chan error)
+
+	var blockKDSPCS atomic.Bool
+	proxy.ConnectDial = func(network string, addr string) (net.Conn, error) {
+		t.Logf("Proxying connection: %q", addr)
+		if (addr == kdsAddr || addr == pcsAddr) && blockKDSPCS.Load() {
+			t.Logf("Blocking connection to KDS/PCS %q", addr)
+			return nil, fmt.Errorf("connection to KDS/PCS %q blocked by test proxy", addr)
+		}
```
burgerdev, line 59: Did you consider storing the fact that a request was blocked and verifying it in the test body?

e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
davidweisse: When the VCEK is sent in the handshake, the CLI does not cache it, since the cache is tied to the `HTTPSGetter`. Not great, but I don't see an immediate way around that.

e2e/kds-pcs-downtime/kds-pcs-downtime_test.go
davidweisse: Since this is just a context deadline exceeded error, it is not very meaningful, but I think checking both for `context deadline exceeded` and if the connection was proxied as suggested below, this should be good.

PR #2000, opened 2025-12-02
ci: use hostpath instead of longhorn

burgerdev:
Longhorn has many moving parts and features (replication, durable storage, extra iscsid pod, ...) we don't need for our CI. Switching to a simpler CSI plugin should reduce the amount of sporadic interference by volume provisioning.

The `hostpath` driver is documented as not ready for production, but mainly because it's a single node system, which is not an issue for us. To the contrary, this driver is used by Kubernetes' own CI, so I'd expect it to be quite stable.

Observation on `discovery`, where the driver is currently active, show that volume provisioning (measured from event `Waiting for a volume to be created` to `MapVolume.MapPodDevice succeeded`) takes only 2s, compared to 11s with Longhorn.

### DIFF
--- a/dev-docs/e2e/csi-hostpath/csi-storageclass.yaml
+++ b/dev-docs/e2e/csi-hostpath/csi-storageclass.yaml
@@ -2,6 +2,8 @@ apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
   name: csi-hostpath-sc
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
 provisioner: hostpath.csi.k8s.io
 reclaimPolicy: Delete
 volumeBindingMode: Immediate


dev-docs/e2e/csi-hostpath/csi-attacher-rbac.yaml
katexochen, line 1: I think we have other resources rather laying around in /tools than under /dev-docs? See /tools/bm-maintenance.

dev-docs/e2e/csi-hostpath/README.md
katexochen, line 1: We should apply this automatically as part of the maintenance workflow.

dev-docs/e2e/csi-hostpath/README.md
```diff
@@ -0,0 +1,25 @@
...
+kubectl apply -k .
+```
+
+## Provenance
+
+The official deployment flow is a [shell script] and there are no stable resource definitions.
+Therefore, we observed the rendered resources once, at [this commit], and are tracking them in this directory instead.
+
+If necessary, this can be reproduced by logging the directory content of `TEMP_DIR` whenever `kubectl apply` is executed.
+However, we don't expect too many changes here, since the driver isn't released in lockstep with Kubernetes and releases are infrequent.
```
katexochen, line 22: The upgrade path still isn't very clear to me. Is this a "never gonna happen anyway" type of upgrade procedure?

burgerdev: Ok, let me try to package that correctly.

dev-docs/e2e/csi-hostpath/csi-storageclass.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-hostpath-sc
```
katexochen, line 4:
```suggestion
metadata:
  name: csi-hostpath-sc
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
```

katexochen (approved):
Open comments can be done in a follow up

PR #2001, opened 2025-12-03
coordinator/httpapi: add unittests for httpapi

charludo:
Unittests for the http api handler. Require to define some custom interfaces which the `AttestationHandler` now takes instead of `atls.Guard`/`stateguard.Guard` directly.

### DIFF
--- a/coordinator/internal/httpapi/attest.go
+++ b/coordinator/internal/httpapi/attest.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"mime"
 	"net/http"
 
 	"github.com/edgelesssys/contrast/coordinator/internal/stateguard"
@@ -17,12 +18,27 @@ import (
 	"github.com/edgelesssys/contrast/internal/atls"
 	"github.com/edgelesssys/contrast/internal/constants"
 	"github.com/edgelesssys/contrast/internal/httpapi"
+	"github.com/edgelesssys/contrast/internal/manifest"
 )
+
+var (
+	errContentType        = errors.New("invalid Content-Type")
+	errNonceLength        = errors.New("invalid nonce length")
+	errGettingState       = errors.New("getting state")
+	errGettingHistory     = errors.New("getting history")
+	errGettingAttestation = errors.New("getting attestation report")
+)
+
+// StateGuard is a stateguard.Guard at runtime, but can be stubbed in tests.
+type StateGuard interface {
+	GetState(context.Context) (*stateguard.State, error)
+	GetHistory(ctx context.Context) ([][]byte, map[manifest.HexString][]byte, error)
+}
 
 // AttestationHandler handles POST requests to /attest.
 type AttestationHandler struct {
 	Issuer     atls.Issuer
-	StateGuard *stateguard.Guard
+	StateGuard StateGuard
 }
 
 func (h *AttestationHandler) getResponse(ctx context.Context, nonce []byte) (*httpapi.AttestationResponse, int, error) {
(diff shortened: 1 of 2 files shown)


coordinator/internal/httpapi/attest_test.go
```diff
@@ -0,0 +1,182 @@
...
+		},
+		"stale state": {
+			nonce:     nonce,
+			guard:     &stubGuard{getStateErr: stateguard.ErrStaleState},
+			expStatus: http.StatusPreconditionFailed,
+			expErr:    userapi.ErrNeedsRecovery,
+		},
+		"unknown error during GetState": {
+			nonce:     nonce,
+			guard:     &stubGuard{getStateErr: errors.New("")},
```
burgerdev: Instead of `errors.New("")`, you could use `assert.AnError`, which you could also `assert.ErrorIs`.

coordinator/internal/httpapi/attest_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/edgelesssys/contrast/internal/httpapi"
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/internal/testkeys"
+	"github.com/stretchr/testify/require"
+)
+
+var nonce = make([]byte, 32)
+
+func TestAttestationHandler(t *testing.T) {
+	testCases := map[string]struct {
```
burgerdev, line 31: I think we should have a test case for a bad (or absent, or empty) body, and maybe one for `Content-Type`, too. Might make sense to do these in a separate test, as they may not fit into this table neatly.

coordinator/internal/httpapi/attest.go
```diff
@@ -14,15 +14,33 @@ import (
...
+	errNonceLength        = errors.New("invalid nonce length")
+	errGettingState       = errors.New("getting state")
+	errGettingHistory     = errors.New("getting history")
+	errGettingAttestation = errors.New("getting attestation report")
+)
+
+// Issuer is an atls.Issuer at runtime, but can be stubbed in tests.
+type Issuer interface {
+	Issue(ctx context.Context, reportData [64]byte) (quote []byte, err error)
+}
```
burgerdev: Why can't you just use `atls.Issuer` directly? The stub even implements it already, afaict.

coordinator/internal/httpapi/attest_test.go
```diff
@@ -0,0 +1,182 @@
...
+	policyHashHex := manifest.NewHexString(policyHash[:])
+	m.Policies = map[manifest.HexString]manifest.PolicyEntry{
+		policyHashHex: {
+			SANs:             []string{"test"},
+			WorkloadSecretID: "test",
+		},
+	}
+	meshKey := testkeys.New[ecdsa.PrivateKey](s.t, testkeys.ECDSAP384Keys[1])
+	rootKey := testkeys.New[ecdsa.PrivateKey](s.t, testkeys.ECDSAP384Keys[2])
+	ca, _ := ca.New(rootKey, meshKey)
```
burgerdev: I think I'd create this in the test prelude and replace the `t` field with `ca`.

coordinator/internal/httpapi/attest.go
charludo: Argh, yes, that's an interface already 🤦🏼‍♀️

coordinator/internal/httpapi/attest_test.go
charludo: > `assert.AnError`

Done

> `assert.ErrorIs`

What did you have in mind? since the resp is always plain json, `require.ErrorIs()` can't workk on it; and seomthing like 
```go
require.ErrorIs(errors.New(apiErr.Err), tc.expErr)
```
also fails:
```
    --- FAIL: TestAttestationHandler/unable_to_get_attestation (0.00s)
        attest_test.go:133: 
                Error Trace:    /home/charlotte/Projekte/contrast/coordinator/internal/httpapi/attest_test.go:133
                Error:          Target error should be in err chain:
                                expected: "getting attestation report"
                                in chain: "getting attestation report: assert.AnError general error for testing"
                Test:           TestAttestationHandler/unable_to_get_attestation
```

burgerdev: Sorry, you're right.

coordinator/internal/httpapi/attest_test.go
charludo: Eh, added them here.

PR #2002, opened 2025-12-03
e2e/imagestore: take into account sidecar containers

charludo:


charludo:
- [ ] https://github.com/edgelesssys/contrast/actions/runs/19886534157

e2e/imagestore/imagestore_test.go
```diff
@@ -88,6 +88,13 @@ func TestImageStore(t *testing.T) {
 					memory := c.Resources.Limits[corev1.ResourceMemory]
 					defaultMemory += int(memory.Value())
 				}
+				for _, c := range pod.Spec.InitContainers {
+					if *c.RestartPolicy != corev1.ContainerRestartPolicyAlways {
```
burgerdev:
```suggestion
					if c.RestartPolicy == nil || *c.RestartPolicy != corev1.ContainerRestartPolicyAlways {
```

e2e/imagestore/imagestore_test.go
```diff
@@ -88,6 +88,13 @@ func TestImageStore(t *testing.T) {
...
+						continue
+					}
+					memory := c.Resources.Limits[corev1.ResourceMemory]
+					defaultMemory += int(memory.Value())
```
burgerdev: ... and to be precise, we actually need to calculate

```
memory := defaultMemory + max(max(real initcontainers), sum(main containers)+sum(sidecar containers))
```

charludo: I don't quite get this - so if there's a single init container which, for whatever reason, has more memory requirred than the sum of all main/sidecar container memory limits, k8 will bump the memory assigned to the pod vm accordingly...?

burgerdev: It will assign the largest amount of memory expected to be used by the ~container~ pod over its lifetime. Since non-sidecar initcontainers run sequentially, they consume at most what the largest one of them consumes. Sidecars and main containers run concurrently, so their requests need to be summed up. So th final value is the maximum of either.

https://github.com/kubernetes/kubernetes/blob/v1.31.5/pkg/api/v1/resource/helpers.go#L152

charludo: Thanks for clarifying. My confusion was because I could have sworn our docs said somewhere that `InitContainers` don't influence total assigned memory, but I must be misremembering because I can't find anything like that in the docs now, plus your link clearly states something different.

Question is, do we need to consider this in this test? This doesn't seem to affect the current setup

burgerdev: You remember correctly, but the docs were wrong and we removed that part in #1773.  

> Question is, do we need to consider this in this test? This doesn't seem to affect the current setup

Right now, it doesn't - but if things change (like the root cause for needing this PR)?

charludo: done

burgerdev (approved):
Thanks!

PR #2004, opened 2025-12-03
qemu-cc: init from qemu-{tdx,snp}

msanft:
This creates a unified `qemu-cc` package for SNP and TDX platforms by combining the `qemu-snp` and `qemu-tdx` packages.
To enable GPU support without degrading current TDX security due to skipping ACPI measurements, it conditionally applies a patch to omit the unneeded PCI-related ACPI tables only if no GPU support is required.


katexochen:
needs rebase

PR #2005, opened 2025-12-03
imagestore: make opt-in instead of opt-out

charludo:
Starting with Contrast 1.16.0, the [secure image store feature](https://docs.edgeless.systems/contrast/next/howto/secure-image-store) will be opt-in, instead of opt-out.

- If you were previously using the `--skip-image-store` flag with your deployments, you have to remove this flag, as it is now deprecated. No further action is required.
- If you were previously **not** using the  `--skip-image-store` flag, the following actions are required:
    - If you want to continue using the secure image store, use `contrast generate` with the new `--inject-image-store` flag in the future. No further action is required.
    - If you want to stop using the secure image store, no additional flag is required. However, please consider [the documentation](https://docs.edgeless.systems/contrast/next/howto/workload-deployment/deployment-file-preparation#pod-resources) to determine whether you might need to increase memory requirements and limits of your deployments. Especially in cases where very large images need to be pulled, current memory limits might be set too low.

### DIFF
--- a/docs/docs/howto/hardening.md
+++ b/docs/docs/howto/hardening.md
@@ -84,8 +84,7 @@ Any volumes mounted to the pod need to be scrutinized, and sensitive data must n
 Ideally, a volume is mounted as a raw block device and authenticated encryption is added within the confidential container.
 
 This is also the approach taken by the [secure image store](./secure-image-store.md) feature, which stores image layers outside of memory by setting up encryption and integrity protection on a raw block device.
-Nevertheless, this weakens the overall security posture compared to storing images entirely in memory.
-To pull images into memory instead, disable the image store feature.
+Nevertheless, if enabled, this weakens the overall security posture compared to storing images entirely in memory.
 For further details, refer to the linked documentation section.
 
 
--- a/docs/docs/howto/secure-image-store.md
+++ b/docs/docs/howto/secure-image-store.md
@@ -13,10 +13,18 @@ A running Contrast deployment.
 
 ## How-To
 
-The secure image store is enabled by default, providing each pod with `10Gi` of storage.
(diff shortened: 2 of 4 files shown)


charludo:
- [ ] [SNP](https://github.com/edgelesssys/contrast/actions/runs/19921504269)
- [ ] [SNP-GPU](https://github.com/edgelesssys/contrast/actions/runs/19921508184)
- [ ] [TDX](https://github.com/edgelesssys/contrast/actions/runs/19921511508)

docs/versioned_docs/version-1.15/howto/hardening.md
```diff
@@ -84,8 +84,7 @@ Any volumes mounted to the pod need to be scrutinized, and sensitive data must n
 Ideally, a volume is mounted as a raw block device and authenticated encryption is added within the confidential container.
 
 This is also the approach taken by the [secure image store](./secure-image-store.md) feature, which stores image layers outside of memory by setting up encryption and integrity protection on a raw block device.
-Nevertheless, this weakens the overall security posture compared to storing images entirely in memory.
-To pull images into memory instead, disable the image store feature.
+Nevertheless, if enabled, this weakens the overall security posture compared to storing images entirely in memory.
```
katexochen: This change is wrong, isn't it? This shouldn't be changed for the released version I'd say.

charludo: Oh, whoops, yes! Sorry about that.

burgerdev (approved):
> would it be a good idea to add a new arg to pass to generate which sets the CSI driver directly?

If we still had it on by default I'd say yes, but at this point I have doubts whether we should further build out this feature. Seems easy enough to configure manually, if needed.

PR #2009, opened 2025-12-04
packages: package csi-driver-host-path with nix

burgerdev:


packages/by-name/csi-driver-host-path/package.nix
katexochen, line 1: file should be named like the package: csi-driver-host-path. Use kebab case for packages, camel is usually used for functions/builders etc.

burgerdev: Thanks, fixed.

packages/by-name/csi-driver-host-path/package.nix
```diff
@@ -0,0 +1,74 @@
...
+    sha256 = "sha256:1rxb0rdqqsw2mfc419b36sy99w04dzsgw28zlyrpgi3d5vcxjsdw";
+  };
+  snapshotterRBAC = builtins.fetchurl {
+    url = "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/v8.2.0/deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml";
+    sha256 = "sha256:164wvh5i017wb8223fbp6z0x86a3xmznaks9d2canzzi29jgcryy";
+  };
+in
+stdenvNoCC.mkDerivation {
+  pname = "csi-driver-host-path";
+  version = "v1.17.0";
```
burgerdev: This is wrong.

PR #2012, opened 2025-12-05
ci: run maintenance before nightly

burgerdev:


.github/workflows/bm_maintenance.yml
```diff
@@ -12,9 +18,6 @@ on:
     paths:
       - ".github/workflows/bm_maintenance.yml"
       - "tools/bm-maintenance/**"
-  schedule:
-    - cron: "0 2 * * 0,1,2,3,4,6" # 2 AM every day except Friday
-    - cron: "0 2 * * 5" # 2 AM on Friday
```
katexochen, line 17: The original idea here was to keep some cache for performance reasons, instead of cleaning up every night. Just for documentation, I don't mind if we change it.

.github/workflows/bm_maintenance.yml
```diff
@@ -1,6 +1,12 @@
 name: bare metal maintenance
 
 on:
+  workflow_call:
+    inputs:
+      containerd:
+        description: "Cleanup containerd store"
+        required: true
+        type: boolean
```
katexochen: This is not used, containerd still checks for the cron tab. If we run the containerd cleanup every night anyway, I suggest to remove the input and the condition on the job altogether.

katexochen (changes requested):

.github/workflows/bm_maintenance.yml
burgerdev: Thanks, done

PR #2013, opened 2025-12-05
internal/kuberesource: add gpu target deployable with just

charludo:


katexochen:
why not use `just e2e gpu`?

katexochen (changes requested):
```
Please verify the manifest history and policies
Please register workloads of new targets in wait-for-workload
error: Recipe `wait-for-workload` failed with exit code 1
```

burgerdev (commented):
@katexochen: my main motivation for this was to be able to generate yaml with resourcegen, being able to `just default gpu` is a nice side effect.

PR #2017, opened 2025-12-08
e2e: rename containerd-digest-pinning to containerd-11644-reproducer

charludo:


e2e/containerd-11644-reproducer/containerd-11644-reproducer_test.go
```diff
@@ -20,16 +20,19 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func TestContainerdDigestPinning(t *testing.T) {
+// TestContainerd11644Reproducer is a reproducer and regression test for https://github.com/containerd/containerd/pull/11644.
+// In containerd versions <2.0, pulling an image by tag first, then by digest second will cause continerd to re-use the tag-based image name,
+// leading to a policy failure.
```
katexochen, line 25:
```suggestion
// leading to a policy failure.
//
// If the test fails, we either managed to install an old containerd version in our CI cluster
// or containerd regressed upstream on this behavior.
```

PR #2019, opened 2025-12-08
ci: refactor link checking

katexochen:


justfile
```diff
@@ -393,7 +390,7 @@ lint:
 
 # Check links.
 check-links:
-    nix run .#nixpkgs.lychee -- --config tools/lychee/config.toml .
+    nix run .#nixpkgs.lychee -- --config tools/lychee/config-external.toml .
```
charludo, line 393: Hm, I would expect this to do both (or be named with `-external`)

tools/lychee/config-external.toml
```diff
@@ -19,6 +19,7 @@ exclude = [
 exclude_path = [
     "dev-docs/frozen",
```
charludo, line 20: Probably a good idea to exclude `versioned_docs` here as well, if only to avoid getting ratelimited

katexochen: I disagree. versioned docs aren't archived and expected to work.

charludo: Then maybe we can cache just `200`s again to avoid hitting the same URLs over and over again?

Though to be fair, I guess we could also just see *if* we get ratelimited first 😄

katexochen: This is running once a week, the cache previously had a ttl of a day, that's why I removed it.

justfile
charludo: Did this get lost in the rebase?

katexochen: I kept this as -external and added a name stating the workflow checks external links.

charludo: Ah, alright

charludo (approved):

PR #2020, opened 2025-12-08
tools/ci-watch: init

katexochen:
A simple tool to fetch all the logs from a CI run so you can ripgrep it.

### DIFF
--- a/tools/ci-watch/main.go
+++ b/tools/ci-watch/main.go
@@ -1,0 +1,345 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package main
+
+import (
+	"archive/zip"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/google/go-github/v72/github"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Println("Usage: ci-watch <run-url>")
+		os.Exit(1)
+	}
+
+	runURL := os.Args[1]
+	u, err := url.Parse(runURL)
+	if err != nil {
+		fmt.Printf("Invalid URL: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Expected format: /owner/repo/actions/runs/runID
+	parts := strings.Split(u.Path, "/")
+	if len(parts) < 6 {
+		fmt.Println("Invalid URL format. Expected: https://github.com/owner/repo/actions/runs/runID")
+		os.Exit(1)
+	}
+
+	var owner, repo, runID string
+	// Find "actions" and "runs" to locate IDs
+	for i, part := range parts {
+		if part == "actions" && i+2 < len(parts) && parts[i+1] == "runs" {
+			runID = parts[i+2]
+			if i >= 2 {
+				repo = parts[i-1]
+				owner = parts[i-2]
+			}
+			break
+		}
+	}
+
+	if owner == "" || repo == "" || runID == "" {
+		fmt.Println("Could not parse owner, repo, or runID from URL")
+		os.Exit(1)
+	}
+
+	targetDir := "logs"
+	fmt.Printf("Fetching logs for %s/%s run %s to %s\n", owner, repo, runID, targetDir)
+
+	if err := fetchGithubWorkflowRun(context.Background(), owner, repo, runID, targetDir); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
+	fmt.Println("Done")
+}
+
+// fetchGithubWorkflowRun fetches the logs and artifacts of a GitHub Actions workflow run.
+func fetchGithubWorkflowRun(ctx context.Context, owner, repo, runID, targetDir string) error {
+	token := os.Getenv("GITHUB_TOKEN")
+	if token == "" {
+		return fmt.Errorf("GITHUB_TOKEN environment variable is not set")
+	}
+
+	runIDInt, err := strconv.ParseInt(runID, 10, 64)
+	if err != nil {
+		return fmt.Errorf("invalid run ID %q: %w", runID, err)
+	}
+
+	client := github.NewClient(nil).WithAuthToken(token)
+
+	// Create run directory
+	runDir := filepath.Join(targetDir, runID)
+	if err := os.MkdirAll(runDir, 0o755); err != nil {
+		return fmt.Errorf("failed to create run directory: %w", err)
+	}
+
+	var errs []error
+
+	fmt.Println("Fetching job console logs...")
+	if err := fetchJobConsoleLogs(ctx, client, owner, repo, runIDInt, runDir); err != nil {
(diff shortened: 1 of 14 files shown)


tools/ci-watch/main.go
```diff
@@ -0,0 +1,343 @@
...
+		fmt.Fprintf(os.Stderr, "Error fetching job logs: %v\n", err)
+	}
+
+	fmt.Println("Fetching artifacts...")
+	if err := fetchRunArtifacts(ctx, client, owner, repo, runIDInt, runDir); err != nil {
+		errs = append(errs, fmt.Errorf("failed to fetch artifacts: %w", err))
+		fmt.Fprintf(os.Stderr, "Error fetching artifacts: %v\n", err)
+	}
+
+	return errors.Join(errs...)
```
charludo, line 104: For general failures this will still iterate over everything and return the same error over and over

katexochen:
> * please add `logs/` to the gitignore
> * rmdir the logs before getting them anew

I wouldn't say this is intended to be used in-tree.

> * I think a `just last-night` would be awesome, e.g. with:
>   ```
>   url=$(gh run list --workflow "e2e_nightly.yml" --limit 1 --json url -q '.[0].url)'
>   ```

Nice idea, I have plans that that go into that direction, but will be a bit more sophisticated. But nothing for this PR.

> * then it would probably also make sense to reuse the token from the envfile (needs new permission though...), or to create an additional one for that purpose

I'll keep that in mind for future improvements. I'm planning to actually run that command in the CI next, so things will change a bit in the future.

tools/ci-watch/main.go
katexochen: sounds okay for now, this is just dev tooling.

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+	"strings"
+
+	"github.com/google/go-github/v72/github"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Println("Usage: ci-watch <run-url>")
+		os.Exit(1)
+	}
```
charludo, line 26: Please use cobra even in tools

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+		fmt.Printf("Invalid URL: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Expected format: /owner/repo/actions/runs/runID
+	parts := strings.Split(u.Path, "/")
+	if len(parts) < 6 {
+		fmt.Println("Invalid URL format. Expected: https://github.com/owner/repo/actions/runs/runID")
+		os.Exit(1)
+	}
```
charludo, line 40: What's this check supposed to accomplish exactly? The check below this seems more useful, but will behave correctly even if this check is absent and would fail.

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+				owner = parts[i-2]
+			}
+			break
+		}
+	}
+
+	if owner == "" || repo == "" || runID == "" {
+		fmt.Println("Could not parse owner, repo, or runID from URL")
+		os.Exit(1)
+	}
```
charludo, line 58: Alternatively, and IMO easier to read:


```suggestion
	// Expected format: /owner/repo/actions/runs/runID
	parts := strings.Split(strings.TrimSuffix(u.Path, "/"), "/")
	partsCount := len(parts)

	if partsCount < 5 || parts[partsCount-2] != "runs" || parts[partsCount-3] != "actions" {
		fmt.Println("Invalid URL format. Expected: https://github.com/owner/repo/actions/runs/runID")
		os.Exit(1)
	}

	owner := parts[partsCount-5]
	repo := parts[partsCount-4]
	runID := parts[partsCount-1]
```

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+	runDir := filepath.Join(targetDir, runID)
+	if err := os.MkdirAll(runDir, 0o755); err != nil {
+		return fmt.Errorf("failed to create run directory: %w", err)
+	}
+
+	var errs []error
+
+	fmt.Println("Fetching job console logs...")
+	if err := fetchJobConsoleLogs(ctx, client, owner, repo, runIDInt, runDir); err != nil {
+		errs = append(errs, fmt.Errorf("failed to fetch job logs: %w", err))
```
charludo, line 94:
```suggestion
		errs = append(errs, fmt.Errorf("fetching job logs: %w", err))
```
Also not sure why we're logging something different here than we are returning, seems that could be unified. Same for everything else, error message styling is all over the place.

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+	fmt.Println("Done")
+}
+
+// fetchGithubWorkflowRun fetches the logs and artifacts of a GitHub Actions workflow run.
+func fetchGithubWorkflowRun(ctx context.Context, owner, repo, runID, targetDir string) error {
+	token := os.Getenv("GITHUB_TOKEN")
+	if token == "" {
+		return fmt.Errorf("GITHUB_TOKEN environment variable is not set")
+	}
```
charludo, line 75: Please also document what permissions the token needs.

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+		return nil
+	}
+
+	artName := *art.Name
+	// Strip "e2e_pod_logs-" prefix to match job directories
+	artName = strings.TrimPrefix(artName, "e2e_pod_logs-")
+	artName = sanitizeName(artName)
+
+	// Create dir for artifact
+	// We use the artifact name as the directory name within the run directory.
```
charludo, line 230:
```suggestion
```

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+
+	if err := downloadAndExtractZip(ctx, url.String(), artDir); err != nil {
+		return fmt.Errorf("downloading/extracting zip: %w", err)
+	}
+
+	fmt.Printf("Downloaded and extracted artifact: %s\n", *art.Name)
+	return nil
+}
+
+func sanitizeName(name string) string {
```
charludo, line 252: I guess this works, but maybe something like https://github.com/gosimple/slug makes more sense for future robustness?

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected status: %s", resp.Status)
+	}
+
+	// Read body to memory (artifacts shouldn't be massive for logs, usually)
+	// If they are large, we should use a temp file.
+	// Given they are logs, let's use a temp file to be safe.
```
charludo, line 284: I can't tell if this is an ironic joke 😅

tools/ci-watch/main.go
```diff
@@ -0,0 +1,345 @@
...
+		return fmt.Errorf("creating temp file: %w", err)
+	}
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
+
+	if _, err := io.Copy(tmpFile, resp.Body); err != nil {
+		return fmt.Errorf("writing to temp file: %w", err)
+	}
+
+	// Open zip
```
charludo, line 296:
```suggestion
```

PR #2023, opened 2025-12-09
e2e: append to namespace file

davidweisse:
This allows running E2E tests in multiple namespaces, even in parallel. Running `just undeploy` will automatically delete all namespaces in the file (one for each line). For the normal `just` flow we don't need to change anything, since we only ever have one namespace. The only exception is when running `just e2e` on an E2E test which includes multiple namespaces and then trying to run something like `just verify`. This will fail because we are trying to read one namespace but get two. However, running these commands will most likely fail anyway, since, e.g., we have no manifest in our workspace. "Solving" this by just running `tail -1 just.namespace` for example is also not helpful, since we cannot really predict which namespace will be returned if the tests run in parallel. So I left the `justfile` untouched for now.

Example:

```go
t.Run("1", func(t *testing.T) {
    t.Parallel()
    ct := contrasttest.New(t)
    ct.Init(t, resources1)
})

t.Run("2", func(t *testing.T) {
    t.Parallel()
    ct := contrasttest.New(t)
    ct.Init(t, resources2)
})
```

For each `contrasttest.New(t)` a new namespace with a unique name will be created and appended to the `just.namespace` file. Running `just undeploy` manually or in the cleanup step will delete both namespaces.

### DIFF
--- a/justfile
+++ b/justfile
@@ -133,7 +133,7 @@ populate target=default_deploy_target platform=default_platform:
 
 # Write the namespace so it can be read by other scripts
 write-namespace target=default_deploy_target:
-    echo "{{ target }}${namespace_suffix-}" > ./{{ workspace_dir }}/just.namespace
+    echo "{{ target }}${namespace_suffix-}" >> ./{{ workspace_dir }}/just.namespace
 
 # Generate policies, update manifest.
 generate cli=default_cli platform=default_platform:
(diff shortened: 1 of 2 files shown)


justfile
katexochen: I think this will fail if one of multiple namespaces isn't present, won't it?

Maybe we should implement this a bit cleaner using a loop instead?

PR #2029, opened 2025-12-15
e2e: fix unreachable services

charludo:


e2e/internal/kubeclient/kubeclient.go
```diff
@@ -290,3 +291,17 @@ func (c *Kubeclient) GetContainerLogs(ctx context.Context, resource ResourceWait
...
+func (c *Kubeclient) FirstPodConnection(ctx context.Context, t *testing.T, namespace, deployment string, port int) string {
+	t.Helper()
+	require := require.New(t)
+
+	pods, err := c.PodsFromDeployment(ctx, namespace, deployment)
+	require.NoError(err)
+	require.Len(pods, 1)
+	ip := pods[0].Status.PodIP
+	require.NotEmpty(ip)
+	return fmt.Sprintf("%s:%d", ip, port)
```
burgerdev: This should use `net.JoinHostPort`, but I think it would make more sense to do that at the call site and only return the IP from this function.

e2e/internal/kubeclient/kubeclient.go
```diff
@@ -290,3 +291,17 @@ func (c *Kubeclient) GetContainerLogs(ctx context.Context, resource ResourceWait
...
+
+// FirstPodConnection returns the IP:Port of the first pod found for a given deployment.
+// Useful to connect directly to a pod, avoiding possible race conditions with k8s when trying to connect to a service.
+func (c *Kubeclient) FirstPodConnection(ctx context.Context, t *testing.T, namespace, deployment string, port int) string {
+	t.Helper()
+	require := require.New(t)
+
+	pods, err := c.PodsFromDeployment(ctx, namespace, deployment)
+	require.NoError(err)
+	require.Len(pods, 1)
```
burgerdev: Do we need to require _exactly_ one, or only _at least_ one?

burgerdev (commented):
`e2e/servicemesh/servicemesh_test.go` has a couple of service connections, too.

`e2e/memdump/memdump_test.go` uses a service, too, but replacing it is a bit more involved.

e2e/servicemesh/servicemesh_test.go
```diff
@@ -110,16 +110,19 @@ func TestIngressEgress(t *testing.T) {
 		require.NoError(err)
 		require.Len(frontendPods, 1, "pod not found: %s/%s", ct.Namespace, "web")
 
+		emojiConnection := net.JoinHostPort(c.FirstPodIP(ctx, t, ct.Namespace, "emoji"), "8801")
+		emojiSvcConnection := net.JoinHostPort(c.FirstPodIP(ctx, t, ct.Namespace, "emoji"), "8801")
```
charludo: Scratching my head a bit here. Seems like this can only make sense if it goes via a service.

burgerdev: The `--connect-to` was introduced because the SAN in the manifest reads `emoji` (derived from the deployment name) while the service is called `emoji-svc`. 

```
       --connect-to <HOST1:PORT1:HOST2:PORT2>

              For a request to the given HOST1:PORT1 pair, connect to
              HOST2:PORT2 instead.  This option is suitable to direct
              requests at a specific server, e.g. at a specific cluster
              node in a cluster of servers. This option is only used to
              establish the network connection. It does NOT affect the
              hostname/port that is used for TLS/SSL (e.g. SNI,
              certificate verification) or for the application protocols.
```

I don't know right now why we did not just rename the service, but anyway this should not be required anymore when we're using the pod IP directly.

charludo: Ahhhh thanks for the explanation. That clears it up.

burgerdev:
There's a failing test showing the exact symptom we wanted to fix: https://github.com/edgelesssys/contrast/actions/runs/20298983785/job/58303806739?pr=2029#step:11:198. At first, I thought our fix was wrong or insufficient, but then I realized that this is due to the prior test failure:

1. `TestOpenSSL/certificate_rotation_and_openssl-backend_restart` fails because `openssl-backend` does not become ready.
   https://github.com/edgelesssys/contrast/blob/e0f125233de643e04739862bec87b4d770b38a70/e2e/openssl/openssl_test.go#L204
2. Then we run `TestOpenSSL/certificate_rotation_and_openssl-frontend_restart`. This test **only waits for the `deploymentToRestart`**, i.e. the frontend. The backend is probably still unhealthy!

So, we need to go through all the tests and make sure that they start from a clean slate: all pods are ready, regardless of whether we touched them in this test or not.

PR #2030, opened 2025-12-15
Support TDX platforms with GPUs

msanft:
This adds experimental, purposefully undocumented support for GPU-enabled TDX platforms. See the individual commit messages for details on the technical implementation.

Open questions:
- Should we add a warning about the security implications when using the `Metal-TDX-QEMU-GPU` platform?

### DIFF
--- a/packages/scripts.nix
+++ b/packages/scripts.nix
@@ -615,8 +615,8 @@ lib.makeScope pkgs.newScope (scripts: {
     name = "nix-gc";
     runtimeInputs = with pkgs; [ busybox ];
     text = ''
-      total=$(df /host/nix | tail -1 | awk '{print $2}')
-      avail=$(df /host/nix | tail -1 | awk '{print $4}')
+      total=$(df -P /host/nix | tail -1 | awk '{print $2}')
+      avail=$(df -P /host/nix | tail -1 | awk '{print $4}')
       over=$((total * 1 / 4 - avail)) # Keep 25% of the disk space free
       over=$((over < 0 ? 0 : over * 1024)) # Convert to bytes
       echo "Running nix garbage collection, deleting $over Bytes of store paths"


tools/tdx-measure/rtmr/rtmr.go
```diff
@@ -215,6 +216,13 @@ func CalcRtmr0(firmware []byte) ([48]byte, error) {
 	// - 0x0000: disabled
 	// - 0x0001: enabled
 	rtmr.extendVariableValue([]byte{0x00, 0x00})
+	if withGPU {
+		// QEMU FW CFG.opt/ovmf/X-PciMmio64Mb (GPU only, set by Kata to a static value)
+		// See:
+		// - https://github.com/kata-containers/kata-containers/blob/c7d0c270ee7dfaa6d978e6e07b99dabdaf2b9fda/src/runtime/virtcontainers/qemu.go#L817-L827
+		// - https://github.com/kata-containers/kata-containers/blob/c7d0c270ee7dfaa6d978e6e07b99dabdaf2b9fda/src/runtime/virtcontainers/qemu_arch_base.go#L877-L908
+		rtmr.extendVariableValue([]byte("524288"))
```
burgerdev: This is the memory size in MiB, right? We could make that an input, instead of the bool. On the command line, we could pass the model and look up its memory.

packages/by-name/kata/runtime/package.nix
```diff
@@ -146,6 +146,12 @@ buildGoModule (finalAttrs: {
       # and is going to be yanked.
       # Upstream PR: https://github.com/kata-containers/kata-containers/pull/12151.
       ./0021-agent-Bump-CDI-rs-to-latest.patch
+
+      # With recent versions of the sandbox-device-plugin, a /dev/iommu device is added
+      # to the container spec for GPU-enabled containers.
+      # Since the same thing is done by the CTK within the PodVM, and we only want this
+      # to influence VM creation, we remove this device from the container spec in the agent.
+      ./0022-runtime-remove-iommu-device.patch
```
burgerdev, line 155: This needs an upstream tracking bug.

tools/tdx-measure/rtmr/rtmr.go
msanft: It's a little more contrived, I think. The BAR size is a somewhat scaled-up version of the actual GPU memory size. Will investigate tomorrow.

packages/by-name/kata/runtime/package.nix
msanft: Do we consider it a bug at all? If so, who would you file the bug against? genpolicy, for not considering this device?

burgerdev: Imho this is a runtime bug, because the device should not be just propagated to the agent. There is a precedent with VFIO devices in `constrainGRPCSpec`. But at the end of the day, we need input by NVIDIA for this.

internal/kuberesource/resourcegen/main.go
```diff
@@ -68,7 +68,24 @@ func main() {
...
+				log.Fatalf("Error parsing platform: %v", err)
+			}
+			var deviceURI string
+			switch platform {
+			case platforms.MetalQEMUTDXGPU:
+				deviceURI = "nvidia.com/GB100_B200"
+			case platforms.MetalQEMUSNPGPU:
+				deviceURI = "nvidia.com/GH100_H100_PCIE"
+			default:
+				log.Fatalf("platform %s does not support GPU tests", platform)
```
charludo: This should probabkly move from here and the GPU test to `internal/platforms/platforms.go` or into `e2e/internal`

charludo (approved):
Code LGTM, but I lack the knowledge to judge the implementation details re: TDX.

internal/kuberesource/resourcegen/main.go
msanft: This is very specific to our current CI setup. Probably, we should just put a placeholder here in resourcegen? What do you think @burgerdev?

burgerdev: A placeholder here sgtm, we are not using `resourcegen gpu` anywhere right now.

docs/docs/howto/cluster-setup/bare-metal.md
```diff
@@ -189,7 +189,7 @@ helm repo add nvidia https://helm.ngc.nvidia.com/nvidia && helm repo update
 helm install --wait --generate-name \
    -n gpu-operator --create-namespace \
    nvidia/gpu-operator \
-   --version=v25.3.0 \
+   --version=v25.10.1 \
```
katexochen, line 192: Did you also update the GPU operator on the SNP CI machine to this version? If not, please do so, they should be in sync.

packages/by-name/OVMF-TDX/package.nix
```diff
@@ -6,6 +6,7 @@
   edk2,
   nasm,
   acpica-tools,
+  verifyACPIInsecure ? false,
```
katexochen, line 9: such options should be of style `withXXX`

packages/by-name/kata/calculateTdxLaunchDigests/package.nix
```diff
@@ -25,6 +26,10 @@ let
     baseCmdline
     os-image.cmdline
   ];
+  # Hardcode this to the B200 for now, since we only have a testing system with this GPU.
+  # When we get more heterogenous test systems, or when TDX-GPU goes into production use,
+  # this needs to be made configurable.
+  gpuFlag = if withGPU then "-g b200" else "";
```
katexochen, line 32: use lib.optionalString

msanft:
> We should document this. 

With "this", are you referring to TDX-GPU support on a general level?
Because if so, I don't really understand this part:

> It has the same security implications as the SNP-GPU platform

This rather sounds like it addresses this part of the PR description:

> Should we add a warning about the security implications when using the `Metal-TDX-QEMU-GPU` platform? 

Could you elaborate on what you would like to have documented exactly?

internal/kuberesource/resourcegen/main.go
msanft: This is not the case unfortunately. `just default gpu` uses it. Fixed in https://github.com/edgelesssys/contrast/pull/2035/commits/b5c90642f46d581b2777fd9d09e9a0ae5a46ff71

PR #2034, opened 2025-12-18
flake.nix: build all packages though a single evaluation

charludo:
Adds new flake outputs `.#<set>.matrix`. Would not recommend building these on your local machine... 👀 

Of course, since we are now able to build/develop on Darwin as well, each individual machine can only build (their architectures') subset of the packages we ship. The CI job builds on both, and pushes both to the cache.

Apropros CI: I have commented out building the debug set as well. Thoughts? The tradeoff here is us not having to build debug versions, and the cache filling up faster.

Edit: almost forgot: This also wires in a custom lib for us to use. Could help clean up the flake, and is the right place for reused nix snippets. Available everywhere in the tree as `lib.contrast`.

### DIFF
--- a/.github/workflows/cache_matrix.yml
+++ b/.github/workflows/cache_matrix.yml
@@ -1,0 +1,60 @@
+name: cache matrix
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch: {}
+
+jobs:
+  linux:
+    concurrency:
+      group: cache-matrix-x86_64-linux
+      cancel-in-progress: true
+    runs-on: ubuntu-24.04
+    timeout-minutes: 240
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/setup_nix
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - name: Build base matrix
+        run: |
+          attr=".#legacyPackages.x86_64-linux.base.matrix"
+          if nix build --dry-run "$attr" 2>&1 | grep -q "will be built"; then
+            nix build --no-link --keep-going "$attr"
+          else
+            echo "Nothing to build. Cache already up to date."
+          fi
+
+  darwin:
+    # Runs after the linux job so the x86_64-linux runtime closure can be substituted here.
+    needs: linux
+    concurrency:
+      group: cache-matrix-aarch64-darwin
+      cancel-in-progress: true
+    runs-on: macos-latest
+    timeout-minutes: 240
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/setup_nix
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - name: Build base matrix
+        run: |
+          attr=".#legacyPackages.aarch64-darwin.base.matrix"
+          if nix build --dry-run "$attr" 2>&1 | grep -q "will be built"; then
+            nix build --no-link --keep-going "$attr"
+          else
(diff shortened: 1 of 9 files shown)


.github/workflows/cache_matrix.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: ./.github/actions/setup_nix
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - name: Build base matrix
+        run: |
+          nix build --no-link ".#legacyPackages.${{ matrix.system }}.base.matrix"
+    # - name: Build debug matrix
+    #   run: |
+    #     nix build --no-link ".#legacyPackages.${{ matrix.system }}.debug.matrix"
```
burgerdev: imho we can leave the debug builds out. Not sure about others' workflows, but I rarely use the debug build these days.

lib/matrix.nix
```diff
@@ -0,0 +1,66 @@
...
+    skip: attrs:
+    let
+      go =
+        path: v:
+        let
+          eval = builtins.tryEval v;
+          val = eval.value;
+        in
+        if !eval.success then
+          [ ]
```
burgerdev: What kind of cases does this catch?

packages/scripts.nix
```diff
@@ -768,4 +768,66 @@ lib.makeScope pkgs.newScope (scripts: {
       mdsh --inputs "$@"
     '';
   };
+
+  # show-input-diff shows which input derivations differ between two builds of the matrix output.
+  # LEFT and RIGHT pick flake refs. Defaults to upstream vs. local checkout.
+  show-input-diff = writeShellApplication {
```
burgerdev: I believe complicated shell scripts should go into `packages/by-name/scripts`.

burgerdev: What are we using this for? I don't quite understand the docstring.

.github/workflows/cache_matrix.yml
charludo: Hm, might just be me, I didn't make the jump to the debugshell 😄 Maybe this is the required incentive 😄

packages/scripts.nix
charludo: This is somewhat tangential to the rest of the PR, but only gets enabled thanks to the matrix output. Running this after updating nixpkgs (for example) gives the subset of packages that we actually consume from those that were updated upstream. IMO pretty useful

burgerdev: Thanks - sounds indeed helpful!

lib/matrix.nix
charludo: E.g. wrong `meta.platforms` (although we filter those out already), `throws` and `aborts` in packages, brokenness guards, `knownVulnerabilities`, unfree licenses.

Buuuuut.... This currently evaluates cleanly even without this guard, so let's just remove it and be informed of that list of things by the CI failing 😄

charludo:
This is finally ready for review again @burgerdev (starting Monday, of course). I will drop the final commit, just left it in to ttrigger the run here. However... I'm wondering if we shouldn't just enable this on all pushes.

overlays/nixpkgs.nix
```diff
@@ -22,6 +22,10 @@ final: prev:
       };
     }
   );
+  # Overridden go version makes helm rebuild. Helm has a flaky (possibly network dependent?) test.
+  kubernetes-helm = prev.kubernetes-helm.overrideAttrs (_: {
+    doCheck = false;
+  });
```
burgerdev, line 28: This should probably be in a standalone commit.

packages/by-name/boot-image/package.nix
```diff
@@ -9,7 +9,7 @@
 
 writeShellApplication {
   name = "boot-image";
-  runtimeInputs = [ qemu ];
+  runtimeInputs = [ (qemu.override { minimal = true; }) ];
```
burgerdev: Are you going to drop this from here in favour of #2445?

.github/workflows/cache_matrix.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - ch/linkfarm # let me teeeeest
```
burgerdev: Don't forget to drop this.

burgerdev (approved):
lgtm overall

packages/by-name/boot-image/package.nix
charludo: Yes, let's merge that one first. Applied both your + Spyros suggestions, so IMO is good to go.

PR #2176, opened 2026-02-10
oak.stage0{,-tdx}-bin: init 

katexochen:


sespiros (commented):
Are you going to add more commits to this or do you want to merge just the packaging for stage0 so that we can experiment?

In any case, some information about how these binaries were built would be helpful as I couldn't find the hashes in oak's repo `git log`. Do we have custom patches on top somewhere? Also it's not very clear to me how the imposter attribute is meant to be used. My guess is as drop-in replacement to wherever we are pulling OVMF-TDX/SNP?

PR #2278, opened 2026-03-24
release: add cli for aarch64-darwin

### LINKED ISSUE #874: Release contrast cli for Mac OS on Apple Silicon (nix: aarch64-darwin)
It would be very appreciated if you guys could release CLIs for other CPU/OS architectures too so that we do not need to run contrast CLI virtualized (emulated) environment. Apart from better UX, running virtualisation/emulation bloats the TCB significantly.

sespiros:
This is rebased on top of https://github.com/edgelesssys/contrast/pull/2260

Fixes #874 

- [x] https://github.com/edgelesssys/contrast/actions/runs/24193431515/job/70616660561

### DIFF
--- a/.github/actions/release_artifacts/action.yml
+++ b/.github/actions/release_artifacts/action.yml
@@ -192,19 +192,18 @@ runs:
       id: upload-artifacts
       shell: bash
       env:
-        S3_BUCKET_PATH: ${{ inputs.s3-bucket-path }}
+        S3_BUCKET_PATH: ${{ inputs.s3-bucket-path }}/${{ github.run_id }}/${{ github.run_attempt }}
         FILES: ${{ steps.artifact-paths.outputs.paths }}
       run: |
         set -u
         shopt -s nullglob
-        unix=$(date +%s)
         readarray -t patterns <<< "$FILES"
         for pattern in "${patterns[@]}"; do
           matches=($pattern)
           for file in "${matches[@]}"; do
-            aws s3 cp "$file" "s3://contrast-public/${S3_BUCKET_PATH}/$unix/"
+            aws s3 cp "$file" "s3://contrast-public/${S3_BUCKET_PATH}/"
           done
         done
-        url="https://contrast-public.s3.eu-central-1.amazonaws.com/${S3_BUCKET_PATH}/$unix/"
+        url="https://contrast-public.s3.eu-central-1.amazonaws.com/${S3_BUCKET_PATH}/"
         echo "Artifacts available at $url" | tee -a "$GITHUB_STEP_SUMMARY"
         echo "url=$url" | tee -a "$GITHUB_OUTPUT"
--- a/.github/actions/release_artifacts/action.yml
+++ b/.github/actions/release_artifacts/action.yml
@@ -113,7 +113,7 @@ runs:
       run: |
         cat << 'EOF' | tee -a "${GITHUB_OUTPUT}"
         paths<<EOM
-        workspace/contrast-cli/bin/contrast
+        workspace/contrast-x86_64-linux
         workspace/coordinator*.yml
         workspace/runtime-*.yml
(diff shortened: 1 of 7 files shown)


.github/actions/release_artifacts/action.yml
```diff
@@ -182,6 +182,8 @@ runs:
         SET: base
       run: |
         nix build -L ".#${SET}.contrast.cli-release" --out-link workspace/contrast-cli
+        SYSTEM=$(nix eval --raw --impure --expr 'builtins.currentSystem')
```
charludo: I'd rather pass the platform explicitly to the composite action

.github/actions/release_artifacts/action.yml
charludo, line 45: The "build CLI" step below is apparently also intended to build the Darwin CLI. That means that these (linux-only) container builds also happen in the same step. Best case, I'd assume this currently downloads from the cache (IF the x84_64-linux variant ran prior), or fails?

Would make sense to either split up the composite action into linux only / linux+darwin, or to use a passed-in system var to determine if steps should run.

.github/workflows/release.yml
```diff
@@ -121,7 +121,7 @@ jobs:
   update-main:
     name: Update main branch
     runs-on: ubuntu-24.04
-    needs: [process-inputs, release]
+    needs: [process-inputs, release, release-aarch64-darwin]
```
charludo:
```suggestion
    needs: [process-inputs, release-x86_64-linux, release-aarch64-darwin]
```

Nit, but IMO should be consistent with the naming

.github/workflows/test-darwin-cli.yml
```diff
@@ -0,0 +1,36 @@
+name: Test darwin CLI build
+
+on:
+  push:
+    branches:
+      - sse/darwin-cli
```
charludo, line 6: Is this workflow just used for testing? In that case remember to remove this file, otherwise the condition needs an update.

.github/workflows/test-darwin-cli.yml
charludo, line 1: I wonder if we could also ensure that some regular e2e tests are run *from* darwin. But ttbh, not sure how expensive/worth it that would be, since the unit tests run during CLI build...

docs/docs/howto/install-cli.md
```diff
@@ -12,14 +12,33 @@ Required for deploying with Contrast.
...
+<TabItem value="macos" label="macOS (Apple Silicon)">
 
 ```bash
+curl --proto '=https' --tlsv1.2 -fLo contrast https://github.com/edgelesssys/contrast/releases/latest/download/contrast-aarch64-darwin
 sudo install contrast /usr/local/bin/contrast
 ```
+
+:::note
+If you download the binary via a web browser instead of `curl`, macOS may show a warning
+that the software can't be verified. Remove the quarantine attribute to resolve this:
```
charludo:
```suggestion
that the software can't be verified. Remove the quarantine attribute before installing the binary:
```

(I assume? Or does this handle PATH entries as well?)

packages/update-contrast-releases.sh
```diff
@@ -31,7 +31,8 @@ target_configs=(
 # declare an associative array that pairs the field name
 # in ./packages/versions.json with the path to the file
 declare -A fields
-fields["contrast"]="./workspace/contrast-cli/bin/contrast"
+fields["contrast"]="./workspace/contrast-x86_64-linux"
```
charludo, line 34: Unsure if we shouldn't also change to the explicit name here as well. But that might take a good deal of additional work in the docs and so on :/

charludo (approved):
Actions look good to me - thanks for doing the restructuring, as well!

Let's run the `pr release artifacts` workflow on this branch to test everything out again before merging. And we should probably do a release soon-ish, while all of these changes are still fresh in mind, in case something needs fixing.

PR #2301, opened 2026-04-07
treewide: decouple vCPU count from number of SNP `ReferenceValues` in manifest

charludo:
## `TrustedMeasurements`

Since version `1.19.0`, Contrast supports multi-CPU pods. As part of that change, manifests significantly increased in size, since an additional `ReferenceValues` block was included for each possible vCPU count.

With this PR, this is again reduced to a single `ReferenceValues` block, just as pre-`1.19.0`. At the same time, the `TrustedMeasurement` field has been converted to `TrustedMeasurements` (plural), mapping each supported vCPU count to the appropriate launch measurement.

A manifest previously containing

```json
  "ReferenceValues": {
    "snp": [
      {
        "TrustedMeasurement": "<launch measurement for 1 vCPU>",
        "CPUs": 1
      },
     {
        "TrustedMeasurement": "<launch measurement for 2 vCPUs>",
        "CPUs": 2
      }
    ]
  }
```
needs to be changed to follow the new format shown below:
```json
  "ReferenceValues": {
    "snp": [
      {
        "TrustedMeasurements": {
          "1": "<launch measurement for 1 vCPU>",
          "2": "<launch measurement for 2 vCPUs>"
        }
      }
    ]
  }
```
Only the measurements for the actually required vCPU counts need to be included. `contrast generate` automatically filters for these required counts when generating the manifest.
 
Additionally, the maximum amount of supported vCPUs has been bumped from 8 to 220.
Please see the documentation for further details.

## Backward- and forward-compatibility 

For compatibility of older CLI versions with the new manifest layout, the old `TrustedMeasurement` field will, for the time being, still be populated with the measurement corresponding to the number of vCPUs the Coordinator is assigned. This allows older CLI versions to `verify` newer Contrast deployments.

Conversely, manifests generated with an older versions of the CLI will still be accepted by newer versions of the CLI, which fall back to the old `TrustedMeasurement` field if the new `TrustedMeasurements` are not present.

The `TrustedMeasurement` field will be deprecated at an as of yet undetermined point in time. Please ensure your Contrast version stays up to date.

### DIFF
--- a/internal/manifest/referencevalues.go
+++ b/internal/manifest/referencevalues.go
@@ -195,6 +195,7 @@ type SNPReferenceValues struct {
 	Platform                string
 	ProductName             ProductName
 	TrustedMeasurements     map[string]HexString
+	TrustedMeasurement      HexString
 	MinimumTCB              SNPTCB
 	GuestPolicy             abi.SnpPolicy
 	PlatformInfo            abi.SnpPlatformInfo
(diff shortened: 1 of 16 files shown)


internal/manifest/referencevalues.go
```diff
@@ -194,16 +194,12 @@ func (r *ReferenceValues) Patch(patches ReferenceValuePatches) error {
 type SNPReferenceValues struct {
 	Platform                string
 	ProductName             ProductName
-	TrustedMeasurement      HexString
+	TrustedMeasurements     map[string]HexString
```
thomasten, line 197: To not break old clients verifying the Coordinator, I'm going to have both `TrustedMeasurement` (with the measurement matching the Coordinator CPU count) and `TrustedMeasurements` (with multiple measurements) in my `manifest.json`. This should work and the client will just ignore the field it doesn't know, right?

charludo: Yes, that's what I was referring to yesterday with the backwards compatibility this change (currently) brakes 😅 

The intention is as you said, support both fields for now, prefer the `s` variant and fall back to the old field.

thomasten: Got it. I was more concerned about forward compatibility though, i.e., using a new manifest with an old client.

charludo: Oh sorry, yes. Unmarshalling the manifest in the client will ignore the new field, and as long as the correct "old" `TrustedMeasurement` is also present, there shouldn't be an issue.

charludo:
Closed in favor of #2344

PR #2303, opened 2026-04-08
cli: allow signed manifest updates

davidweisse:
This PR allows setting a manifest without the CLI having access to the workload owner key. A user can instead pass a signature to the CLI which, when verified by the Coordinator using one of the workload owner keys from current manifest, authorizes the manifest update.

The user has to use the `--prepare` flag of `contrast sign` to obtain the blob to sign:

```sh
contrast sign --prepare
```

The user can then use some other tool with potentially HSM managed keys to generate a signature over the transition blob:

```sh
openssl dgst -sha256 -sign <workload-owner-key> -out transition.sig next-transition
```

The signature is then passed to the CLI instead of a workload owner key:

```sh
contrast set --signature transition.sig resources/
```

- [x] [E2E Coordinator](https://github.com/edgelesssys/contrast/actions/runs/24663103215)

### DIFF
--- a/cli/cmd/set.go
+++ b/cli/cmd/set.go
@@ -58,6 +58,8 @@ issuer certificates.`,
 	cmd.Flags().String("workload-owner-key", workloadOwnerPEM, "path to workload owner key (.pem) file")
 	cmd.Flags().Bool("atomic", false, "only set the manifest if the coordinator's state matches the latest transition hash")
 	cmd.Flags().String("latest-transition", "", "latest transition hash set at the coordinator (hex string)")
+	cmd.Flags().StringP("signature", "s", "", "path to a detached transition signature (DER) file")
+	must(cmd.MarkFlagFilename("signature"))
 
 	return cmd
 }
@@ -90,6 +92,13 @@ func runSet(cmd *cobra.Command, args []string) error {
 		workloadOwnerKey = nil
 	} else if err != nil {
 		return fmt.Errorf("loading workload owner key: %w", err)
+	}
+	var signatureBytes []byte
+	if flags.signaturePath != "" {
+		signatureBytes, err = os.ReadFile(flags.signaturePath)
+		if err != nil {
+			return fmt.Errorf("reading signature file: %w", err)
+		}
 	}
 
 	paths, err := findYamlFiles(args)
@@ -153,6 +162,7 @@ func runSet(cmd *cobra.Command, args []string) error {
 		Manifest:               manifestBytes,
 		Policies:               getInitdataDocuments(policies),
 		PreviousTransitionHash: previousTransitionHash,
+		Signature:              signatureBytes,
 	}
 	resp, err := setLoop(cmd.Context(), client, cmd.OutOrStdout(), req)
 	if err != nil {
@@ -160,7 +170,9 @@ func runSet(cmd *cobra.Command, args []string) error {
 		if ok {
 			if grpcSt.Code() == codes.PermissionDenied {
 				msg := "Permission denied."
-				if workloadOwnerKey == nil {
+				if signatureBytes != nil {
+					msg += " Ensure the signature is valid and corresponds to the latest transition hash."
+				} else if workloadOwnerKey == nil {
 					msg += " Specify a workload owner key with --workload-owner-key."
 				} else {
 					msg += " Ensure you are using a trusted workload owner key."
@@ -204,6 +216,7 @@ type setFlags struct {
 	workloadOwnerKeyPath string
 	atomic               bool
 	latestTransition     string
+	signaturePath        string
 	workspaceDir         string
 }
 
@@ -231,11 +244,13 @@ func parseSetFlags(cmd *cobra.Command) (*setFlags, error) {
 	if err != nil {
 		return nil, fmt.Errorf("getting latest-transition flag: %w", err)
 	}
-
 	if !flags.atomic && flags.latestTransition != "" {
 		return nil, fmt.Errorf("\"latest-transition\" flag cannot be set without \"atomic\" flag")
 	}
-
+	flags.signaturePath, err = cmd.Flags().GetString("signature")
+	if err != nil {
+		return nil, fmt.Errorf("getting signature flag: %w", err)
+	}
 	flags.workspaceDir, err = cmd.Flags().GetString("workspace-dir")
 	if err != nil {
 		return nil, fmt.Errorf("getting workspace-dir flag: %w", err)
--- a/cli/cmd/sign.go
+++ b/cli/cmd/sign.go
@@ -1,0 +1,176 @@
+// Copyright 2026 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package cmd
+
+import (
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/edgelesssys/contrast/internal/history"
+	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/spf13/cobra"
+)
+
+// NewSignCmd creates the contrast sign subcommand.
+func NewSignCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sign [flags]",
+		Short: "Sign the given manifest and transition hash",
+		Long: `Sign the given manifest and transition hash.
+
+This will read the manifest and previous transition hash and sign them using
+the provided workload owner key. The resulting signature can be used for a
+signed manifest update at the Coordinator without providing the workload owner
+key to the CLI setting the manifest.
+
+Using the prepare flag, the CLI will compute the next transition hash and
+output it to a file so it can be signed using an external tool like an HSM.`,
+		RunE: withTelemetry(runSign),
+	}
+	cmd.SetOut(commandOut())
+
+	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
+	cmd.Flags().String("workload-owner-key", workloadOwnerPEM, "path to workload owner key (.pem) file")
+	cmd.Flags().String("latest-transition", "", "latest transition hash set at the coordinator (hex string)")
+	cmd.Flags().Bool("prepare", false, "prepare the next transition hash for signing without signing it")
+	cmd.Flags().String("out", "", "output file for the signature (or next transition hash when using --prepare)")
+	must(cmd.MarkFlagRequired("out"))
+	must(cmd.MarkFlagFilename("manifest", "json"))
+
+	return cmd
+}
+
+func runSign(cmd *cobra.Command, _ []string) error {
+	flags, err := parseSignFlags(cmd)
+	if err != nil {
+		return fmt.Errorf("failed to parse flags: %w", err)
+	}
+
+	log, err := newCLILogger(cmd)
+	if err != nil {
+		return err
+	}
+
+	manifestBytes, err := os.ReadFile(flags.manifestPath)
+	if err != nil {
+		return fmt.Errorf("failed to read manifest file: %w", err)
+	}
+	var m manifest.Manifest
+	if err := json.Unmarshal(manifestBytes, &m); err != nil {
+		return fmt.Errorf("failed to unmarshal manifest: %w", err)
+	}
+	if err := m.Validate(); err != nil {
+		return fmt.Errorf("validating manifest: %w", err)
+	}
+
+	if flags.latestTransition == "" {
+		data, err := os.ReadFile(filepath.Join(flags.workspaceDir, verifyDir, latestTransitionHashFilename))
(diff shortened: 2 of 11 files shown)


cli/cmd/set.go
```diff
@@ -126,6 +134,20 @@ func runSet(cmd *cobra.Command, args []string) error {
...
+		tr := &history.Transition{
+			ManifestHash:           history.Digest(manifestBytes),
+			PreviousTransitionHash: [history.HashSize]byte(previousTransitionHash),
+		}
+		transitionHash := tr.Digest()
+		transitionHashHex := hex.EncodeToString(transitionHash[:])
+		if err := os.WriteFile(filepath.Join(flags.workspaceDir, nextTransitionHashFilename), []byte(transitionHashHex), 0o644); err != nil {
+			return fmt.Errorf("writing transition hash: %w", err)
+		}
+		fmt.Fprintf(cmd.OutOrStdout(), "✔️ Transition hash %s written to %s\n", transitionHashHex, filepath.Join(flags.workspaceDir, nextTransitionHashFilename))
```
davidweisse: Do we want to write to a file here or to stdout? I've written it to a file for now since we compute the signature over a file when using `openssl`.

cli/cmd/set.go
davidweisse, line 176: Should we change something here? I'm not sure if this still makes sense considering an error when validating the manifest signature.

coordinator/internal/userapi/userapi.go
davidweisse, line 118: I'm also not sure if we need to consider a signed manifest update on a first set manifest call. It doesn't really make sense but it is possible to pass a signature here. The only thing we could do though is check if the signature is valid for a key in the new manifest. For atomic updates we check if the provided previous hash is empty, but here the previous hash would be integrated into the signature...

cli/cmd/set.go
davidweisse: Expanded this to include an error message if the signature is set.

coordinator/internal/userapi/userapi.go
davidweisse: Added the validation using the keys in the provided manifest on a first set manifest call.

cli/cmd/set.go
```diff
@@ -91,6 +93,12 @@ func runSet(cmd *cobra.Command, args []string) error {
 	} else if err != nil {
 		return fmt.Errorf("loading workload owner key: %w", err)
 	}
+	signatureBytes, err := os.ReadFile(flags.signaturePath)
+	if errors.Is(err, os.ErrNotExist) {
+		signatureBytes = nil
+	} else if err != nil {
+		return fmt.Errorf("reading signature file: %w", err)
+	}
```
burgerdev, line 101: What if there happens to be a signature file lying around, but the user did not mean to use it?

cli/cmd/set.go
burgerdev: File is probably safer, but then it should probably be configurable? And if it were configurable, users could provide `/dev/stdout` if they prefer, but then we'd need to write log messages to `cmd.ErrOrStderr` (which is the right thing to do anyway).

cli/cmd/set.go
```diff
@@ -52,12 +52,14 @@ issuer certificates.`,
 	}
 	cmd.SetOut(commandOut())
 
+	cmd.Flags().Bool("dry-run", false, "compute the next transition hash without setting the manifest at the coordinator")
 	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
 	cmd.Flags().StringP("coordinator", "c", "", "endpoint the coordinator can be reached at")
-	must(cobra.MarkFlagRequired(cmd.Flags(), "coordinator"))
+	cmd.MarkFlagsOneRequired("dry-run", "coordinator")
```
burgerdev, line 58: Not so sure about the `dry-run` anymore, since it's also associated with tests or validation. Validation happens, sorta, but additional data is written, so this may be misleading and it's not really a speaking name. Alternatives that come to mind are `--offline`, `--prepare-transition-hash`. 

Considering future use cases, such as multi-party updates, it might make sense to introduce a new `sign` subcommand that computes the signature directly, and optionally takes a `--dry-run` for signing with something else.

coordinator/internal/userapi/userapi.go
```diff
@@ -297,4 +337,6 @@ var (
 	ErrAlreadyRecovered = errors.New("coordinator is already recovered")
 	// ErrNeedsRecovery is returned if state exists, but no secrets are available, e.g. after restart.
 	ErrNeedsRecovery = errors.New("coordinator is in recovery mode")
+	// ErrNoSignature is returned if a manifest update request does not contain a signature.
+	ErrNoSignature = errors.New("manifest signature is empty")
```
burgerdev, line 341: Afaict this is not a user-facing error at the moment. Should we unexport it?

coordinator/internal/userapi/userapi_test.go
```diff
@@ -251,6 +253,42 @@ func TestSetManifest(t *testing.T) {
 		_, err = coordinator.SetManifest(ctx, req)
 		require.Error(err)
 	})
+
+	t.Run("signed manifest update", func(t *testing.T) {
+		require := require.New(t)
+
+		coordinator := newCoordinator()
+		ctx := rpcContext(t.Context(), trustedKey)
```
burgerdev, line 261: We should test without a client cert, too.

e2e/coordinator/coordinator_test.go
```diff
@@ -107,6 +110,57 @@ func TestCoordinator(t *testing.T) {
...
+		sig, err := ecdsa.SignASN1(rand.Reader, key, transitionHashShaSum[:])
+		require.NoError(err)
+		require.NoError(os.WriteFile(filepath.Join(ct.WorkDir, "transition.sig"), sig, 0o644))
+
+		// Remove the key so that it is not passed to the CLI.
+		require.NoError(os.Remove(filepath.Join(ct.WorkDir, "workload-owner.pem")))
+
+		require.NoError(ct.RunSet(ctx, "--signature", filepath.Join(ct.WorkDir, "transition.sig")))
+
+		require.True(t.Run("verify", ct.Verify), "contrast verify needs to succeed for subsequent tests")
```
burgerdev, line 149: We should also test without key and without signature.

e2e/coordinator/coordinator_test.go
```diff
@@ -107,6 +110,57 @@ func TestCoordinator(t *testing.T) {
 		require.NoError(err)
 		require.NoError(ct.RunSet(ctx, "--atomic", "--latest-transition", string(transitionHash)))
 	})
+
+	t.Run("signed manifest update", func(t *testing.T) {
```
burgerdev, line 114: It would be really cool if we could sign both with go and with openssl in this test.

cli/cmd/set.go
davidweisse: The signature file path has to be provided. Since it is user generated it has no fixed filename that can be read automatically.

If we were to introduce a `sign` subcommand, we could have a mandatory `--out` flag or specify some default here. But the signature during `set` is not read automatically either way.

e2e/coordinator/coordinator_test.go
davidweisse: I don't think I have access to `openssl` in the test environment, since this would be on the CLI side, right?

cli/cmd/set.go
burgerdev: Sorry, I mixed that up - empty by default is fine.

e2e/coordinator/coordinator_test.go
burgerdev: The e2e binaries are nix-built, so you can just have openssl in `runtimeInputs`?

cli/cmd/set.go
```diff
@@ -58,6 +58,7 @@ issuer certificates.`,
 	cmd.Flags().String("workload-owner-key", workloadOwnerPEM, "path to workload owner key (.pem) file")
 	cmd.Flags().Bool("atomic", false, "only set the manifest if the coordinator's state matches the latest transition hash")
 	cmd.Flags().String("latest-transition", "", "latest transition hash set at the coordinator (hex string)")
+	cmd.Flags().StringP("signature", "s", "", "path to a signed transition (DER) file")
```
burgerdev, line 61:
```suggestion
	cmd.Flags().StringP("signature", "s", "", "path to a detached transition signature (DER) file")
```
... right?

burgerdev, line 61: Should this also `MarkFlagFileName`?

cli/cmd/sign.go
```diff
@@ -0,0 +1,173 @@
...
+Using the prepare flag, the CLI will compute the next transition hash and
+output it to a file so it can be signed using an external tool like an HSM.`,
+		RunE: withTelemetry(runSign),
+	}
+	cmd.SetOut(commandOut())
+
+	cmd.Flags().StringP("manifest", "m", manifestFilename, "path to manifest (.json) file")
+	cmd.Flags().String("workload-owner-key", workloadOwnerPEM, "path to workload owner key (.pem) file")
+	cmd.Flags().String("latest-transition", "", "latest transition hash set at the coordinator (hex string)")
+	cmd.Flags().Bool("prepare", false, "prepare the next transition hash for signing without signing it")
```
burgerdev, line 45: Should `workload-owner-key` and `prepare` be marked mutually exclusive?

cli/cmd/sign.go
```diff
@@ -0,0 +1,173 @@
...
+		flags.latestTransition = string(data)
+	}
+	previousTransitionHash, err := hex.DecodeString(flags.latestTransition)
+	if err != nil {
+		return fmt.Errorf("decoding latest transition hash: %w", err)
+	}
+
+	tr := &history.Transition{
+		ManifestHash:           history.Digest(manifestBytes),
+		PreviousTransitionHash: [history.HashSize]byte(previousTransitionHash),
```
burgerdev, line 92: Since this is user-provided, we should probably verify the size.

coordinator/internal/userapi/userapi.go
```diff
@@ -110,6 +116,12 @@ func (s *Server) SetManifest(ctx context.Context, req *userapi.SetManifestReques
 		}
 	} else {
 		// First SetManifest call, initialize seed engine.
+		if req.Signature != nil {
+			if err := validateSignature(m.WorkloadOwnerPubKeys, [history.HashSize]byte{}, req); err != nil {
+				s.logger.Warn("SetManifest signature validation failed for initial manifest", "err", err)
+				return nil, status.Errorf(codes.PermissionDenied, "validating manifest signature: %v", err)
+			}
```
burgerdev, line 123: I forgot one requirement: https://docs.edgeless.systems/contrast/howto/immutable-deployments#preventing-manifest-updates. That will probably fail in the first line of `validateSignature`.

cli/cmd/sign.go
davidweisse: Not sure if mutually exclusive is correct here, since they are not conflicting and we can just ignore the key on `prepare`.

cli/cmd/set.go
davidweisse: Interestingly, my shell (Bash) autocompletes every flag to a file name no matter what, and trying to limit the extensions using `MarkFlagFileName("manifest", "json")` for example has no effect.

coordinator/internal/userapi/userapi.go
davidweisse: Sending a signature on a first set manifest call, I would expect it to fail, since there is no key to verify the signature with. The problem I see is, if the user wants to do a subsequent manifest update to remove the workload owner keys, but in reality the Coordinator has no state. We get an error that the signature could not be verified, when the expected error should be that the transition hash does not match.

Since a signature does not really make sense on a first set manifest call and only accomplishes what an atomic update would do, maybe we should disallow signatures on first set manifest calls and advertise atomic updates in that case? I currently don't have any better ideas.

burgerdev: Yes, I did not really think that through, signature without workload owner keys does not make sense to begin with. However, I do think that signatures for the first SetManifest call could make sense, although in fringe cases: data owner trusts a workload owner _quorum_ (i.e., multi party update), but not necessarily an individual workload owner - something like that.

cli/cmd/sign.go
burgerdev: True.

cli/cmd/set.go
burgerdev: Do you actually configure the completions? I'd expect shells to suggest file names by default.

davidweisse: I think so too, but then what effect does `MarkFlagFileName` even have? In the description it says that completion is limited to the specified extensions.

coordinator/internal/userapi/userapi.go
davidweisse: Should we then just leave it as it is? For the error, we would get `setting manifest is disabled` plus `Ensure the signature is valid and corresponds to the latest transition hash`. This tells us that there are no workload owner keys with which to verify the signature with, so either it was disabled to begin with and our expected transition hash is wrong, or we are trying to set a first manifest with a signature corresponding to no trusted keys.

burgerdev (approved):
Thanks!

PR #2309, opened 2026-04-10
just: fix location of containerd-reproducer push script

burgerdev:
The installation target for the containerd-reproducer was accidentally changed in #2260, although the script for pushing the reproducer is different from other containers. 

This should fix the error here: https://github.com/edgelesssys/contrast/actions/runs/24218822301/job/70706785086. 

- [x] [test run](https://github.com/edgelesssys/contrast/actions/runs/24238629092)


charludo (commented):
Argh, sorry, that was my suggestion 😂😬 

> Since it's platform independent, it does not need to move to containers.nix.

Agreed, but is there a reason it should not be moved to `packages/by-name/scripts/containers/package.nix`? Seems that would be the more appropriate place. I probably originally put it into `scripts.nix` because it seemed to pollute `containers.nix` a bit, but since #2260, all the `push-` scripts have moved anyways.

charludo (approved):
Thank you!

PR #2310, opened 2026-04-10
cli: add `CPUCountValid` verifier

charludo:


cli/verifier/cpu_count_valid.go
```diff
@@ -0,0 +1,36 @@
...
+	var findings error
+
+	kuberesource.MapPodSpec(toVerify, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+		if spec == nil || spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
+			return spec
+		}
+
+		cpuCount := kuberesource.GetPodCPUCount(spec)
+		if cpuCount < 1 || cpuCount > 8 {
+			findings = errors.Join(findings, fmt.Errorf("the pod failed verification. Currently only 0-7 additional vCPUs are supported (1-8 vCPUs in total), but %d were requested", cpuCount))
```
burgerdev: Could we somehow get the deployment/... name into the error message?

internal/kuberesource/parts.go
```diff
@@ -519,3 +519,34 @@ func getPorts(podSpec *applycorev1.PodSpecApplyConfiguration) (ports []applycore
...
+// GetPodCPUCount computes the total CPU count (in whole CPUs) required by a pod, adding hypervisor overhead.
+func GetPodCPUCount(spec *applycorev1.PodSpecApplyConfiguration) uint64 {
+	var regularContainersCPU int64
+	for _, container := range spec.Containers {
+		regularContainersCPU += GetCPUCount(container.Resources)
+	}
+	var initContainersCPU int64
+	for _, container := range spec.InitContainers {
+		cpuCount := GetCPUCount(container.Resources)
+		initContainersCPU += cpuCount
```
burgerdev: Looking at the resource calculation logic again, I fear we made a few mistakes. 

1. Sidecar containers should not be added to the initcontainers here.
2. Initcontainers should not be summed, only the biggest initcontainer should be considered.
3. Shouldn't the pod-level resources be added to the container-level resources, instead of `max`?

charludo: Yeah... I'll rework it. #2301 does something similar, but the function used there can and should be swapped out with the one here after this PR is merged, so I'd prefer to fix it here, and not bother there.

cli/verifier/cpu_count_valid.go
charludo: done

internal/kuberesource/parts.go
charludo: done

cli/verifier/cpu_count_valid.go
```diff
@@ -19,17 +20,22 @@ type CPUCountValid struct{}
 func (v *CPUCountValid) Verify(toVerify any) error {
 	var findings error
 
-	kuberesource.MapPodSpec(toVerify, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
+	kuberesource.MapPodSpecWithMeta(toVerify, func(meta *v1.ObjectMetaApplyConfiguration, spec *applycorev1.PodSpecApplyConfiguration) (*v1.ObjectMetaApplyConfiguration, *applycorev1.PodSpecApplyConfiguration) {
```
burgerdev: I'm afraid that this is not the right meta (we would be interested in the top-level resource, but this is the meta of the `podTemplate` for everything but simple pods. Sorry, thought that would be easier - let's just log a TODO and move on.

burgerdev (approved):

PR #2312, opened 2026-04-10
containers: build utilities with musl

burgerdev:
e2fsprogs and iptables depend on glibc, which in turn comes with megabytes of locale data and other cruft. Building the utilities against musl saves 12MiB (compressed) in both the coordinator and the initializer.

### DIFF
--- a/packages/containers.nix
+++ b/packages/containers.nix
@@ -4,6 +4,7 @@
 {
   lib,
   pkgs,
+  pkgsMusl,
   contrastPkgs,
   dockerTools,
 }:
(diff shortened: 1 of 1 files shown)


katexochen:
Building the used packages under my local nixpkgs ref fails wit some test failures. Seems to fail on the SNP runner, too, for different reasons. Notice that there is no guarantee all packages in pkgsMusl will just work, this is mostly untested upstream. Also adds quite some build costs (~190 pkgs), but that only matters nixpkgs updates I guess. Change looks good otherwise.

PR #2316, opened 2026-04-13
cli: configurable guest policy

charludo:
Starting with this PR, the SNP 'GuestPolicy' is no longer hard-coded in Contrast, but instead configurable. `contrast generate` reads the 'GuestPolicy' from an existing manifest, using it to calculate the SNP ID block that is injected as a pod annotation. For new manifest creation, the same default values as before are used.

**Note:** after changing the 'GuestPolicy', `contrast generate` needs to be re-run.


cli/cmd/generate.go
```diff
@@ -752,10 +759,57 @@ func patchIDBlockAnnotation(res any) error {
...
-		if snpIDBlocks[platform] == nil || snpIDBlocks[platform][cpuCount] == nil ||
-			snpIDBlocks[platform][cpuCount][genoa].IDBlock == "" || snpIDBlocks[platform][cpuCount][milan].IDBlock == "" {
-			return meta, spec, fmt.Errorf("missing ID block configuration for runtime %s with %s CPUs", platform, cpuCount)
+		// Ensure the required launch digests have been pre-computed.
+		if snpLaunchDigests[platform] == nil || snpLaunchDigests[platform][cpuCount] == nil ||
+			snpLaunchDigests[platform][cpuCount][genoa].LaunchDigest == "" || snpLaunchDigests[platform][cpuCount][milan].LaunchDigest == "" {
+			return meta, spec, fmt.Errorf("missing launch digest configuration for runtime %s with %s CPUs", platform, cpuCount)
+		}
+
+		// Retrieve the guest policies from the manifest (first matching entry per product).
```
burgerdev: Could we implement the logic independent of milan/genoa in a function and then call that twice here?

internal/manifest/referencevalues.go
```diff
@@ -78,6 +79,12 @@ func (r ReferenceValues) Validate() error {
 		} else if err != nil {
 			errs = append(errs, newValidationError(fmt.Sprintf("snp[%d]", i), err))
 		}
+
+		if firstSnpPolicy == nil {
+			firstSnpPolicy = &v.GuestPolicy
+		} else if *firstSnpPolicy != v.GuestPolicy {
+			errs = append(errs, fmt.Errorf("all SNP reference values must have the same GuestPolicy"))
```
burgerdev: Why do we have this requirement?

cli/cmd/generate.go
```diff
@@ -778,6 +832,31 @@ func patchIDBlockAnnotation(res any) error {
...
 }
 
+func snpGuestPolicyForProduct(refVals []manifest.SNPReferenceValues, platform platforms.Platform, product string) (abi.SnpPolicy, bool) {
+	for _, rv := range refVals {
+		if rv.Platform == platform.String() && string(rv.ProductName) == product {
+			return rv.GuestPolicy, true
+		}
+	}
+
+	// Fallback: the manifest may not have SNP ref values yet. Use the embedded reference values
```
burgerdev: When does that happen?

charludo: Never anymore, after implementing [this suggestion](https://github.com/edgelesssys/contrast/pull/2316#discussion_r3077898329). `patchIDBlockAnnotation` previously failed if genoa OR milan were missing.

internal/manifest/referencevalues.go
charludo: I initially added that because you could theoretically have multiple SNP reference value entries with differing guest policies, and it would not be clear which one should be annotated on a resource. 

But yeah, it's a bit too strict of a restriction. Fixup commit instead selects guest policy to annotate based on vCPU count. Just need to be careful when rebasing this on #2301 (or vice versa).

cli/cmd/generate.go
```diff
@@ -711,9 +714,63 @@ func patchRuntimeClassName(defaultRuntimeHandler string) func(*applycorev1.PodSp
 	}
 }
 
-func patchIDBlockAnnotation(res any) error {
-	var snpIDBlocks SNPIDBlocks
-	if err := json.Unmarshal(SNPIDBlockData, &snpIDBlocks); err != nil {
+func computeIDBlockAnnotations(targetPlatform platforms.Platform, cpuCount string, productName string, snpLaunchDigests SNPLaunchDigests, mnf *manifest.Manifest, annotations map[string]string) error {
```
burgerdev: Please document that this function modifies the `annotations`.

docs/docs/architecture/components/manifest.md
```diff
@@ -206,7 +206,8 @@ Always `None` for Milan and Genoa platforms.
 This is the guest policy according to Section 4.3 of the [SEV ABI Spec].
 It's enforced during the launch of the confidential VM.
 
-The guest policy is currently static in Contrast, values can't be changed.
+The guest policy is configurable. `contrast generate` reads the policy from the manifest and uses it
+to compute the SNP ID block that's injected as a pod annotation.
```
burgerdev: Not sure if it's obvious that you need to re-run generate such that the modifications take effect.

burgerdev (approved):
lgtm % nits

docs/docs/architecture/components/manifest.md
charludo: Good point. Added a warning.

cli/cmd/generate.go
charludo: done

charludo:
- [x] [regression test, TDX](https://github.com/edgelesssys/contrast/actions/runs/24494383153)
- [x] [regression test, SNP-GPU](https://github.com/edgelesssys/contrast/actions/runs/24494377876)

(all individual test case failures are unrelated / well known and re-occurring in our nightlies)

PR #2320, opened 2026-04-14
e2e: wait for container running instead of pod running

davidweisse:
We had sporadic failures where the `WaitForCoordinator` method returned true but the container inside the pod could not be found afterwards (e.g., https://github.com/edgelesssys/contrast/actions/runs/24319240177/job/71002411757). I changed the logic to not check for the pod status to be `Running` but check if there is a running *container* matching the desired container name in the pod. My guess is that there is a small window where the pod is considered `Running` but the containers inside might not have startet completely yet. Hopefully this check works a bit better.


burgerdev (approved):
Thanks, I agree that this looks like an improvement, looking forward to the e2e results.

PR #2321, opened 2026-04-14
sets: enable debug output for badaml sets

sespiros:
Build OVMF with debug output to the serial port and enable the kata debug runtime for badaml-vuln and badaml-sandbox so that firmware, agent, and runtime messages are captured in the host journal when a CVM crashes. This gives us the visibility needed to investigate the intermittent crash seen in nightly runs.


burgerdev (approved):
Thanks, excitedly waiting for the failure now!

sespiros:
https://github.com/edgelesssys/contrast/pull/2368 resolves this issue so no need for extra debugging information now.

PR #2337, opened 2026-04-21
Support non-CC runtimes for development

msanft:
This adds support for insecure non-CC runtimes. These can be used both for testing in PoC-like settings where Contrast needs to be integrated into a given platform where CC hardware is not yet available, but also eases benchmarking of CC against non-CC performance for Contrast consumers.

This PR only adds the technical implementation for the insecure runtimes, but doesn't yet expose it to the user via release artifacts or documentation.

Despite the PR size, I opted to combine the changes into a single PR with reviewable commits, since -imo- having the full test coverage of all the code makes it easier to reason about the treewide changes. Let me know if you want anything adjusted.

---

Closes CON-2, CON-3, CON-4

### DIFF
--- a/cli/verifier/image_ref_valid.go
+++ b/cli/verifier/image_ref_valid.go
@@ -25,8 +25,7 @@ func (v *ImageRefValid) Verify(toVerify any) error {
 	kuberesource.MapPodSpec(toVerify, func(
 		spec *applycorev1.PodSpecApplyConfiguration,
 	) *applycorev1.PodSpecApplyConfiguration {
-		if spec == nil || spec.RuntimeClassName == nil ||
-			!(strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") || strings.HasPrefix(*spec.RuntimeClassName, "contrast-insecure")) {
+		if !kuberesource.IsContrastPod(spec) {
 			// Non-Contrast pods are not subject to this verification.
 			return spec
 		}
--- a/cli/verifier/image_ref_valid.go
+++ b/cli/verifier/image_ref_valid.go
@@ -25,7 +25,9 @@ func (v *ImageRefValid) Verify(toVerify any) error {
 	kuberesource.MapPodSpec(toVerify, func(
 		spec *applycorev1.PodSpecApplyConfiguration,
 	) *applycorev1.PodSpecApplyConfiguration {
-		if spec == nil || spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
+		if spec == nil || spec.RuntimeClassName == nil ||
+			!(strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") || strings.HasPrefix(*spec.RuntimeClassName, "contrast-insecure")) {
+			// Non-Contrast pods are not subject to this verification.
 			return spec
 		}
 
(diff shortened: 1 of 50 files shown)


msanft:
[Insecure E2E Test](https://github.com/edgelesssys/contrast/actions/runs/24732165743)

cli/verifier/image_ref_valid.go
```diff
@@ -25,7 +25,9 @@ func (v *ImageRefValid) Verify(toVerify any) error {
 	kuberesource.MapPodSpec(toVerify, func(
 		spec *applycorev1.PodSpecApplyConfiguration,
 	) *applycorev1.PodSpecApplyConfiguration {
-		if spec == nil || spec.RuntimeClassName == nil || !strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") {
+		if spec == nil || spec.RuntimeClassName == nil ||
+			!(strings.HasPrefix(*spec.RuntimeClassName, "contrast-cc") || strings.HasPrefix(*spec.RuntimeClassName, "contrast-insecure")) {
```
sespiros: Since this is a pattern that is frequently used (and also related to your De Morgan lint issues), one thing you could do is something like:

```go
// contrastRuntimeClassPrefixes lists runtime class prefixes that identify Contrast pods.
var contrastRuntimeClassPrefixes = []string{"contrast-cc", "contrast-insecure"}

// isContrastPod reports whether spec belongs to a Contrast-managed pod.
func isContrastPod(spec *corev1.PodSpec) bool {
    if spec == nil || spec.RuntimeClassName == nil {
        return false
    }
    return slices.ContainsFunc(contrastRuntimeClassPrefixes, func(p string) bool {
        return strings.HasPrefix(*spec.RuntimeClassName, p)
    })
}
```

Then call like:
```go
if !isContrastPod(spec) {
    return spec
}
```

msanft:
~TODO: Handling of insecure runtimes in the SDK~
For Privatemode, this is important as all the attestation parts needed for a CC/Non-CC benchmark are based on the SDK.

EDIT: Changed this to match the CLI's opt-in behavior.

burgerdev (changes requested):
Hey @msanft,

Thanks for the PR. I think this could be split up naturally into at least the following intermediate features which don't need to be submitted all at once:

- prefactoring for the `contrast-cc` runtime class name matching.
- Insecure issuer and verifier, not wired into anything but tested against each other.
- initdata processor
- runtime and nodeinstaller changes
- support for insecure manifests at the Coordinator
- support for insecure manifests at the CLI
- e2e test with insecure
- support for insecure manifests in SDK

That should make review (and iteration) much easier - thanks!

msanft:
> That should make [...] iteration much easier - thanks!

@burgerdev: We'll have to find a good working mode for this. For me, as the PR author, rebasing a stack of N PRs is very tedious. That was my rationale for bundling these changes into a single PR and relying on reviewable commits instead. I'll try to split up into multiple PRs as you suggested.

msanft:
I split this with the help of AI:

--- START AGENT OUTPUT

1. #2352 prefactor: centralize Contrast runtime class matching
2. #2353 attestation: add insecure issuer and validator
3. #2354 initdata-processor: support insecure platforms
4. #2355 runtime/nodeinstaller: add insecure platform variants
5. #2356 coordinator: support insecure manifests behind opt-in
6. #2357 cli: support insecure manifests behind opt-in
7. #2358 e2e: add insecure deployment test
8. #2359 sdk: support insecure manifests behind opt-in

Each PR is based on the previous one.

--- END AGENT OUTPUT

PR #2338, opened 2026-04-21
multiple: fix OVMF-TDX debug build with debug set

sespiros:
`set=debug just` on TDX failed for two unrelated reasons:

1. We carry a patch for OVMF-TDX which enables `BUILD_FIRMWARE_UI=FALSE` which strips `BootManagerMenuApp` from the firmware volume. In OVMF debug builds (which we now do as part of the debug set), `PlatformRegisterOptionsAndKeys` asserts on the resulting `EFI_NOT_FOUND` and halts before the guest boots. Fixed by amending `0002-IntelTdxX64-add-toggle-to-disable-firmware-UI.patch` to return early on `EFI_NOT_FOUND`.

2. The debug set now also sets `use_legacy_serial=true` to make it possible to get early OVMF messages. This changes QEMU's PCI/ACPI topology. OVMF then extends different hashes into RTMR[0] than the hardcoded ones in `tdx-measure`, so attestation fails. Fixed by adding a second hash set for the legacy-serial topology, selected via a new `--legacy-serial` flag.


charludo (approved):
LGTM. Could you rebase #2340 after merging this, so we can try it out in CI as well? :D

PR #2340, opened 2026-04-22
ci: add minimal integration of sets with ci

sespiros:
- run `e2e openssl` on non-gpu nodes and `e2e gpu` on gpu nodes using the debug set, as a smoke test during nightly runs
- add a set selector in the e2e manually dispatched workflow

- [x] [manual run with set=debug just e2e openssl Metal-QEMU-SNP](https://github.com/edgelesssys/contrast/actions/runs/24780943153)
- [ ] [manual run with set=debug just e2e gpu Metal-QEMU-SNP-GPU](https://github.com/edgelesssys/contrast/actions/runs/24780757579)
- [ ] [manual run with set=debug just e2e openssl Metal-QEMU-TDX](https://github.com/edgelesssys/contrast/actions/runs/24780686278) This is the one that is broken and will pass after https://github.com/edgelesssys/contrast/pull/2338 is merged
- [ ] [manual run with set=debug just e2e gpu Metal-QEMU-TDX-GPU](https://github.com/edgelesssys/contrast/actions/runs/24780975309)

### DIFF
--- a/.github/workflows/e2e_manual.yml
+++ b/.github/workflows/e2e_manual.yml
@@ -44,6 +44,17 @@ on:
           - Metal-QEMU-SNP-GPU
           - Metal-QEMU-TDX
           - Metal-QEMU-TDX-GPU
+      set:
+        description: "Package set"
+        required: false
+        type: choice
+        options:
+          - base
+          - debug
+          - badaml-sandbox
+          - badaml-vuln
+          - testcase-default
+        default: testcase-default
       skip-undeploy:
         description: "Skip undeploy"
         required: false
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e_manual.yml
```diff
@@ -43,6 +43,17 @@ on:
...
+        description: "Package set"
+        required: false
+        type: choice
+        options:
+          - default
+          - base
+          - debug
+          - badaml-sandbox
+          - badaml-vuln
+        default: default
```
charludo:
```suggestion
        options:
          - base
          - debug
          - badaml-sandbox
          - badaml-vuln
          - testcase-default
        default: testcase-default
```

Nit: this is a bit confusing otherwise, since from our perspective, `base` is the "default", but here default means "what the e2e has configured".

.github/workflows/e2e_nightly.yml
```diff
@@ -83,6 +84,38 @@ jobs:
...
+            test-name: openssl
+            platform:
+              name: Metal-QEMU-TDX
+              runner: TDX
+              self-hosted: true
+          - set: debug
+            debug-shell: false
+            test-name: gpu
+            platform:
+              name: Metal-QEMU-TDX-GPU
```
charludo, line 116: Oof... GPU can often take a while, and block other deployments. Not sure if it's worth it to run these on GPU. (Although a nice side effect of this would be that the debug versions are always built / available from cachix for us.) Not sure.

charludo: "Wer lesen kann ist klar im Vorteil" (German proverb)

charludo (approved):

PR #2341, opened 2026-04-22
rfc: secret sharing

davidweisse:


rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,90 @@
...
+```go
+type Guard struct {
+    state atomic.Pointer[State]
+    hist  *history.History
+
+    seedSharesMux sync.Mutex
+    seedShares    [][]byte // new field to track received seed shares
+    ...
+}
+````
```
burgerdev:
````suggestion
```
````

File: rfc/013-secret-sharing.md [hunk 2]
```diff
@@ -0,0 +1,90 @@
...
+  string PublicKey = 1;
+  bytes EncryptedSeedShare = 2;
+}
+```
+
+### Security Considerations
+
+The `seedShares` list of the guard must be cleared either after successful recovery or when first entering recovery mode.
+After the manifest is updated at a different Coordinator, the Coordinator transitions into recovery mode after realizing it has a stale state.
+In this case, the `seedShares` list must be empty to prevent old seed shares from being used for recovery.
```
burgerdev: It sounds to me like we should track the received seed shares in the state, which will then automatically result in the seed shares being erased when we update. Would also make the synchronous operations easier (no need for the lock).

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Background
+
+During an initial `contrast set` the secret seed is encrypted with each seed share owner's public key and returned to the user.
+This means that each seed share owner has access to the complete seed, and thus the ability to recover the Coordinator in case of a restart.
+The original idea was to use a secret sharing scheme to split the seed into multiple shares, and require a threshold of them to recover the seed.
+This would have the advantage of not giving any single seed share owner access to the complete seed.
+It also comes with the drawback that the seed share owners would need to coordinate to recover the seed, which adds complexity to the recovery process.
+
+## Proposal
```
burgerdev, line 16: Could you please add considerations for distributed Coordinator recovery? As far as I can tell, this proposal only works with a single Coordinator right now. The conclusion may well be that you need to scale the coordinator down to 1 in order to recover.

burgerdev (commented):
Thanks, great proposal!

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,96 @@
...
+}
+```
+
+This threshold defaults to the number of provided seed share owner public keys, so by default every seed share owner would need to be involved in the recovery process.
+The lowest value would be 1, in which case the behavior is the same as the current implementation, allowing every seed share owner to recover the Coordinator on their own.
+Like the seed share owner public keys, this field can't be updated after the initial `contrast set`.
+
+### UserAPI Changes
+
+For the implementation we can use a secret sharing scheme based on OpenBao's [Shamir](https://pkg.go.dev/github.com/openbao/openbao/sdk/v2/helper/shamir) go library.
```
daniel-weisse: Any specific reason for using OpenBao's version over the [original Hashiscorp version](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go)?

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,96 @@
...
+```json
+{
+    ...
+    "SeedShareOwnerThreshold": 3,
+    "SeedShareOwnerPubKeys": [ ... ]
+}
+```
+
+This threshold defaults to the number of provided seed share owner public keys, so by default every seed share owner would need to be involved in the recovery process.
+The lowest value would be 1, in which case the behavior is the same as the current implementation, allowing every seed share owner to recover the Coordinator on their own.
```
daniel-weisse, line 46: Do note that SSS requires at least 2 shares, meaning the old implementation has to be kept for this case

rfc/013-secret-sharing.md
davidweisse: @burgerdev wdyt?

burgerdev: Vault was relicensed to BSL: https://github.com/hashicorp/vault/blob/main/LICENSE. The headers still say MPL sometimes, but it's maybe safer to use the fork.

burgerdev: It also looks like Bao is better maintained here: https://github.com/openbao/openbao/issues/182

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,96 @@
...
+When the `Recover` RPC is called and the state is stale, the `ResetState` method of the guard is invoked, which takes a `SecretSourceAuthorizer` as an argument.
+This interface currently only has the method `AuthorizeByManifest`, which validates the peer given the manifest from the store and returns the new seed engine and mesh CA key.
+This is used both by the user API for `contrast recover` and by the mesh API during peer recovery.
+The `AuthorizeByManifest` function can now additionally take the seed shares as an argument to reconstruct the seed if the threshold is met and return the received seed share otherwise:
+
+```go
+type SecretSourceAuthorizer interface {
+    AuthorizeByManifest(ctx context.Context, manifest *manifest.Manifest, seedShares [][]byte) (seedengine.SeedEngine, crypto.PublicKey, []byte, error)
+}
+```
```
burgerdev: If the seed shares are not relevant for all implementations, wouldn't it make more sense to make this internal to `seedAuthorizer`?

davidweisse: That makes sense.

burgerdev (approved):
lgtm

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,96 @@
+# Secret Sharing
+
+## Background
+
+During an initial `contrast set` the secret seed is encrypted with each seed share owner's public key and returned to the user.
+This means that each seed share owner has access to the complete seed, and thus the ability to recover the Coordinator in case of a restart.
+The original idea was to use a secret sharing scheme to split the seed into multiple shares, and require a threshold of them to recover the seed.
+This would have the advantage of not giving any single seed share owner access to the complete seed.
```
msanft, line 9: What's the concrete advantage?

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+This threshold defaults to the number of provided seed share owner public keys, so by default every seed share owner would need to be involved in the recovery process.
+The lowest value would be 1, in which case the behavior is the same as the current implementation, allowing every seed share owner to recover the Coordinator on their own.
+Like the seed share owner public keys, this field can't be updated after the initial `contrast set`.
+
+### UserAPI Changes
+
+For the implementation we can use a secret sharing scheme based on OpenBao's [Shamir](https://pkg.go.dev/github.com/openbao/openbao/sdk/v2/helper/shamir) go library.
+During the initial `contrast set`, the seed is split into shares using the specified threshold and the provided seed share owner public keys, and each share is encrypted with the corresponding public key and returned to the user.
```
msanft, line 54: Is it a goal of Contrast to allow up-/downscaling of seed share owners on a live deployment? At least the latter wouldn't work anymore with this, right?

msanft (commented):
Generally LGTM, but I fail to see the practical motivation for this.

rfc/013-secret-sharing.md
burgerdev: Having access to the full secret seed allows deriving workload secrets and volume decryption keys, so basically all persistent state. This means that it's currently not possible to set up secure persistency for multi-party collaboration (think, two mutually distrusting data owners). 

More abstract advantage would be multi-party recovery, i.e. requiring to compromise more than one workflow to obtain the seed.

msanft: Thanks, understood.

rfc/013-secret-sharing.md
burgerdev: Scaling down the seed share owners would need to generate a new seed (otherwise, assumptions about prior manifests' security might not hold anymore). At that point, it's probably easier to just start a fresh manifest history.

msanft (approved):

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,96 @@
...
+
+This approach only works if there is only one Coordinator instance, since `contrast recover` only connects to one Coordinator instance.
+If there are multiple instances, the seed shares would need to be distributed to all of them.
+In the current peer recovery process, Coordinators that enter a stale state call the `Recover` RPC of the mesh API of other Coordinators to recover.
+In order to distribute the seed shares to all instances however, we need an RPC in the opposite direction, where the Coordinator in recovery can send the received seed shares to the other Coordinators.
+
+When implementing this, we need to additionally consider concurrent recovery attempts on different Coordinator instances.
+If 2 seed shares remain to meet the threshold and 2 different seed share owners call `contrast recover` at the same time, both Coordinators have to update each other's state without running into an error.
+
+For now, we can say the solution is to scale down the Coordinator to a single instance.
```
thomasten: This is also what we do in MarbleRun btw

rfc/013-secret-sharing.md
thomasten: For MarbleRun, we decided to vendor shamir.go because we didn't want to add a huge depdendency like Vault or OpenBao SDK to the go.mod just for this small, stable part of code.
We decided to take it from this commit: https://github.com/hashicorp/vault/tree/8d07273d14ae7f5a48cc96f66cc86615dea83390/shamir
It comes with the license next to it. These two files may just be copied like here https://github.com/edgelesssys/marblerun/tree/master/3rdparty/hashicorp/shamir
We also looked at the changes in OpenBao to it since then and concluded that none of them are needed.

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,96 @@
...
+    seedShares [][]byte // new field to track the received seed shares
+    ...
+
+    stale atomic.Bool
+}
+```
+
+If the Coordinator is in recovery mode, this list is used to track the received seed shares via the `Recover` RPC of the user API.
+Peer recovery still works as before, since there are no seed shares involved.
+If a `Recover` call from the user is made where the newly received seed share combined with the already received shares meets the threshold, the seed is reconstructed and the Coordinator can be recovered normally.
```
thomasten, line 82: How should the Coordinator behave if the wrong seed is reconstructed because one of the inputs was wrong?

davidweisse: I guess this would imply a malicious seed share owner? The Coordinator has no way to identify a wrong share, so it will create a new seed engine anyway. This should however be immediately noticeable when attesting the Coordinator with `contrast verify` because the root CA changed.

davidweisse: No, the signature of the latest transition is checked during recovery, so directly after the seed engine has been created. If the seed engine is wrong, the user that called `Recover` last will receive an error `verifying latest transition...`. Since the Coordinator doesn't know which seed share is incorrect, maybe we have to transition to a new state with no seed shares and start again?

burgerdev: Yes, if we can't verify the signature with the combined seed we should throw all shares away and start from scratch.

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,99 @@
...
+If the list contains some number of seed shares below the threshold and peer recovery triggers, the Coordinator is recovered via peer recovery as expected.
+
+When the `Recover` RPC is called and the state is stale, the `ResetState` method of the guard is invoked, which takes a `SecretSourceAuthorizer` as an argument.
+This interface currently only has the method `AuthorizeByManifest`, which validates the peer given the manifest from the store and returns the new seed engine and mesh CA key.
+This is used both by the user API for `contrast recover` and by the mesh API during peer recovery.
+The `AuthorizeByManifest` function now additionally returns the reconstructed seed if the threshold is met and the received seed share otherwise:
```
sespiros: s/and/or

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,99 @@
+# Secret Sharing
+
+## Background
+
+During an initial `contrast set` the secret seed is encrypted with each seed share owner's public key and returned to the user.
+This means that each seed share owner has access to the complete seed, and thus the ability to recover the Coordinator in case of a restart.
+The original idea was to use a secret sharing scheme to split the seed into multiple shares, and require a threshold of them to recover the seed.
+This would have the advantage of not giving any single seed share owner access to the complete seed.
```
sespiros, line 9: The background section should mention the motivation behind this change more explicitly (by incorporating Markus' comment about mutually distrusting data owners or the more theoretical benefit of multi-party recovery). 

Another necessity that also emerges then is that we need to extend our threat model to include a new actor, the less-than-fully-trusted seed share owner. This needs to be described either as part of this proposal or at least be documented as follow-up TODO to our threat model (security.md and threat model docs page).

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,99 @@
...
+
+For recovery, the Coordinator needs to track the received seed shares and combine them when the threshold is met.
+For this, we add a new field to the `State` struct:
+
+```go
+type State struct {
+    seedEngine    *seedengine.SeedEngine
+    ...
+
+    seedShares [][]byte // new field to track the received seed shares
```
sespiros: Overloading `State`'s stale state with another type of "staleness", for when we are in the middle of recovery, is confusing. A manifestation of this "confusion" is having to change `AuthorizeByManifest`'s contract to behave differently depending on whether we have shares in flight or not. 

Reading the code there, it seems to me that `Guard` could be a more suitable home for the intermediate shares until a threshold is met. An example sketch of a design that seems more straight-forward to me:

```go
type Guard struct {
    ...
    partialMu sync.Mutex
    // nil when no user recovery is in flight.
    // Replaced as a whole on each share submission..
    // Never mutated in place because callers retain the pointer as a snapshot after AddSeedShare returns.
    partialRecovery *PartialRecovery
}

// PartialRecovery is the immutable accumulator of seed shares submitted
// during a multi-party user recovery. It is bound to a specific recovery
// target (insecure latest transition + manifest) and a single salt.
type PartialRecovery struct {
    TransitionHash [history.HashSize]byte // hist.GetLatestInsecure() at first submission
    Salt           []byte                  // captured from the first submitter
    Shares         map[ShareID][]byte      // keyed by submitter pubkey. Treated as read-only
}

// ShareID identifies a seed share submitter by their hex-encoded RSA
// public key, matching an entry in Manifest.SeedshareOwnerPubKeys.
type ShareID = manifest.HexString

func (g *Guard) AddSeedShare(ctx context.Context, callerID ShareID, share, salt []byte) (*PartialRecovery, error)
```

Then `AddSeedShare` would implement the logic that checks for errors, checks the threshold etc. If the threshold is met, it would reconstruct the seed, build a `seedAuthorizer`, call `Guard.ResetState` etc. It could be made to also return the complete seed, this is just an example.

We would clear `PartialRecovery` in case of either success or error. This is also in line with the other dicsussion about what to do is the wrong seed is reconstructed etc.

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,99 @@
...
+If the list contains some number of seed shares below the threshold and peer recovery triggers, the Coordinator is recovered via peer recovery as expected.
+
+When the `Recover` RPC is called and the state is stale, the `ResetState` method of the guard is invoked, which takes a `SecretSourceAuthorizer` as an argument.
+This interface currently only has the method `AuthorizeByManifest`, which validates the peer given the manifest from the store and returns the new seed engine and mesh CA key.
+This is used both by the user API for `contrast recover` and by the mesh API during peer recovery.
+The `AuthorizeByManifest` function now additionally returns the reconstructed seed if the threshold is met and the received seed share otherwise:
+
+```go
+type SecretSourceAuthorizer interface {
+    AuthorizeByManifest(ctx context.Context, manifest *manifest.Manifest) (seedengine.SeedEngine, crypto.PublicKey, []byte, error)
```
sespiros: I believe this is meant to be `*seedengine.SeedEngine` and `*ecdsa.PrivateKey` but ideally see my other comment that suggests not changing this method's contract at all.

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,99 @@
...
+    ...
+
+    stale atomic.Bool
+}
+```
+
+If the Coordinator is in recovery mode, this list is used to track the received seed shares via the `Recover` RPC of the user API.
+Peer recovery still works as before, since there are no seed shares involved.
+If a `Recover` call from the user is made where the newly received seed share combined with the already received shares meets the threshold, the seed is reconstructed and the Coordinator can be recovered normally.
+If the list contains some number of seed shares below the threshold and peer recovery triggers, the Coordinator is recovered via peer recovery as expected.
```
sespiros, line 83: My understanding from the proposal section was that in the first version we would need to scale down the coordinator to one instance which would be in non-recovered state right? So how is it possible for peer recovery to trigger mid-user-recovery via shares?

Is there a scenario where while we are scaling down to one instance, a non-recovered coordinator with seed shares accumulated, does peer recover from a recovered peer before the recovered one gets torn down?

Should it be specified more like:
- if a recovered coordinator exists, use peer recovery
- if no recovered coordinator exists, scale down to one and perform threshold user recovery
- don't use `--force` for threshold recovery while recovered peers exist

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,99 @@
...
+This means that each seed share owner has access to the complete seed, and thus the ability to recover the Coordinator in case of a restart.
+The original idea was to use a secret sharing scheme to split the seed into multiple shares, and require a threshold of them to recover the seed.
+This would have the advantage of not giving any single seed share owner access to the complete seed.
+It also comes with the drawback that the seed share owners would need to coordinate to recover the seed, which adds complexity to the recovery process.
+
+## Proposal
+
+We implement a secret sharing scheme for the seed, and each seed share owner only has an encrypted seed share.
+To recover the Coordinator, a threshold of seed shares need to be combined to recover the seed.
+This is done by each seed share owner calling `contrast recover` with their encrypted seed share, and the Coordinator combines the shares and recovers the seed if the threshold is met.
```
sespiros, line 20: Technically seed share owners call `contrast recover` with a plain text seed share (but over encrypted transport/aTLS).

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,99 @@
...
+
+If the returned seed engine is `nil` and we get a seed share that isn't already in the list of received seed shares, we return a new state.
+This state has the `stale` boolean still set to `true` but now includes the list of received seed shares from before plus the new received seed share.
+After updating the state, we return an error indicating that the threshold isn't met yet.
+If we get a valid seed engine, we proceed as before.
+
+If the Coordinator has enough seed shares to recover the seed but one of the seed shares is invalid, through, for example, a malicious seed share owner, the following signature verification of the latest transition will fail.
+In that case, we don't know which seed shares are valid or invalid, so the only option is to reset the state and start the recovery process from scratch with no seed shares.
+
+The proto messages for the `Recover` API remain basically unchanged, now including the encrypted seed share instead of the encrypted seed:
```
sespiros: Maybe this is obvious but I think it should be explicitly mentioned here that each share owner is expected to send the same `Salt` as part of `RecoverRequest` and that this should be validated by the coordinator (that's why I included the `Salt` in my suggested design change).

I have also made the comment in the proposal section but, wdym by "encrypted seed share", is this encrypted with something else that I have missed?

sespiros: Looking at https://github.com/edgelesssys/contrast/blob/44a892dcb71d660aa3ba405807bf98768ecb2c80/internal/manifest/crypto.go#L95

I wonder if we should have a more specific label for these new shares (I gather this is like aad in aead), that would also bind them to properties of their intended use like "intended to be used by sss", the threshold, total share count etc.

I see this as a robustness/ux improvement that would help in potential early mismatch detection/cleaner domain separation.

sespiros (commented):
Thanks for this proposal. A couple of comments inline. The 2 most important ones are:
- the consideration around how this should affect our threat model
- the suggestion to put the shares in `Guard` instead of `State`

The rest are really comments on the implementation.

rfc/013-secret-sharing.md
davidweisse: This is more of an edge case: if one recovered Coordinator and one non-recovered Coordinator exist (e.g., 2 Coordinators and I restart one), and I call `recover` with `--force`, the non-recovered Coordinator can accumulate seed shares before peer recovery triggers. I think otherwise the behavior is the same as before, i.e., prefer peer recovery when `--force` is not specified. Scaling down to 1 for user recovery should then be documented in the docs.

rfc/013-secret-sharing.md
davidweisse: The seed share is encrypted on disk, so from a user perspective we call `contrast recover` with the encrypted seed share which is then decrypted by the CLI before sending it to the Coordinator.

rfc/013-secret-sharing.md
davidweisse: Yes, I forgot about that. It only makes sense to also validate the salt.

The proto paragraph needs rewriting: the recover request now includes the seed share (not encrypted), and the `SeedShare` message type belongs to the `SetManifestResponse`, where the encrypted seed share is correct.

davidweisse: I think we can do something like `"seedshare-k/n"`, for a k out of n threshold scheme. Both the user and the Coordinator have access to the manifest, so they both know the threshold and number of seed share owners. I'm not sure how useful this is though. This would only help when a user accidentally runs `contrast recover` with a seed share from a threshold scheme with different parameters. Maybe even `"seedshare-k-{hash(seedshare owner keys)}"`.

But keep in mind that the untrusted seed share owner can simply choose the label himself and forge an incorrect share with the correct label, so this would only be a UX improvement, not adding any robustness in that scenario.

rfc/013-secret-sharing.md
davidweisse: Updated the background section.

I'm not sure how to fit that into our current docs yet, but I added a section to the rfc.

rfc/013-secret-sharing.md
davidweisse: See also https://github.com/edgelesssys/contrast/pull/2341#discussion_r3132189106.

I think semantically, it makes more sense to handle this in the state, since the seed shares are somewhat tied to the state anyway. Apart from handling locks automatically, this would also simplify concurrent recovery requests. If we handle this in the guard, we probably need additional logic to cover possible concurrency problems.

> Then AddSeedShare would implement the logic that checks for errors, checks the threshold etc. If the threshold is met, it would reconstruct the seed, build a seedAuthorizer, call Guard.ResetState etc. It could be made to also return the complete seed, this is just an example.

The first implementation side problem I see here, is that for validating the seed share, we would already need to do basically the same thing we do in `ResetState`, i.e., get the peer public key and validate it via the latest insecure manifest. Which is, why I think the `AuthorizeByManifest` function is a good position to put this in.

To maybe make the code flow a bit clearer, we could change the `SecretSourceAuthorizer` interface in the following way:

```go
type SecretSourceAuthorizer interface {
	AuthorizeByManifest(context.Context, *manifest.Manifest) error
	AddSeedShare() (*PartialRecovery, error)
	GetSeedEngine() (*seedengine.SeedEngine, *ecdsa.PrivateKey, error)
}
```

`AuthorizeByManifest` now only calls `validatePeer`, and the call could look something like this:

```go
_ = authorizer.AuthorizeByManifest(ctx, mnfst)
partialRecovery, _, := authorizer.AddSeedShare()
se, meshCAKey, err := authorizer.GetSeedEninge()
if errors.Is(err, ErrThresholdNotReached) {
    // set new state with partialRecovery
}
```

This is also no problem for peer recovery, where `AddSeedShare` is just a dummy implementation that returns nil and `GetSeedEngine` will never return the specific error.

Other than that, I adopted the `PartialRecovery` struct for now to store the salt. I'm not sure why we need the manifest transition. Is there a scenario where this can change during the recovery process?

burgerdev: The potential for concurrency issues was originally why I wanted to handle everything in `State`. If both `guard.state` and `guard.partialRecovery` can be updated simultaneously, we need to pay close attention not to end up inconsistent.

We could implement the seed share collection logic outside of the stateguard, though. Since the userapi is supplying the `SecretSourceAuthorizer`, we could return a marker error if reconstruction needs more seed shares, and handle that after `ResetState` returns. There's still a chance of inconsistency, but in userapi it is at least not directly security critical. The struct should then probably look like

```
type seedAuthorizer struct {
	seedShares map[manifest.hexString][]byte // public key -> seed share
}
```

rfc/013-secret-sharing.md
sespiros: The idea is that on the Coordinator side you could bail early if you received a seedshare with unexpected parameters. This doesn't add any security, yes.

rfc/013-secret-sharing.md
sespiros: > I'm not sure why we need the manifest transition. Is there a scenario where this can change during the recovery process?

The idea of the transitionHash in my original comment was so that we keep track of the state from which the transition started. I wrote this without taking into account that in v1 we would scale down the Coordinator to 1 instance (as a prerequisite covered in documentation). In that case then yes there isn't nobody else around to write a new transition. Now I don't know if we should be extra defensive against an operator skipping this step so another instance is still alive and can accept a `SetManifest` while shares are being collected. Markus informed me that recover is disabled when there are live Coordinators but there is still `--force`.

> We could implement the seed share collection logic outside of the stateguard, though. Since the userapi is supplying the `SecretSourceAuthorizer`, we could return a marker error if reconstruction needs more seed shares, and handle that after `ResetState` returns.

I think that also makes sense yes; do the seed share accumulation in the `Server`. I didn't think of seed shares as state but as input that contributes towards a future state transition, so this makes as much sense to me as the original idea to put it in guard.

rfc/013-secret-sharing.md
burgerdev: Note that the label is not present at the Coordinator, since the seedshare is decrypted at the client. In any case, I'd like to decouple this UX discussion from this RFC.

rfc/013-secret-sharing.md
burgerdev: @davidweisse: Please add a sentence about immutability of `SeedShareOwnerThreshold` to the proposal.  This is in line with the existing immutability of the owners list: https://github.com/edgelesssys/contrast/blob/90538866932bcd3712e7c98aac586ea6ece1f11b/coordinator/internal/userapi/userapi.go#L110-L113

We need to document that, too.

rfc/013-secret-sharing.md
```diff
@@ -0,0 +1,140 @@
...
+
+```go
+type seedAuthorizer struct {
+    seedShares map[manifest.HexString][]byte
+}
+```
+
+The `AuthorizeByManifest` method verifies the origin of the seed share, checks if the threshold is met, and if so, reconstructs the seed engine.
+Otherwise, return an error indicating that the threshold isn't met yet and propagate this to the user.
+If the seed share can't be verified by the manifest, also return a specific error.
```
burgerdev:
```suggestion
If the seed share owner's public key is not authorized by the manifest, also return a specific error.
```

burgerdev (approved):

rfc/013-secret-sharing.md
davidweisse: Immutability is already document under the "Manifest Changes" section:

```Like the seed share owner public keys, this field can't be updated after the initial `contrast set`.```

burgerdev: I suppose that comment should have targeted https://github.com/edgelesssys/contrast/pull/2341#discussion_r3457734708?

davidweisse: Yes, but that comment does not have a reply box for me..
Ah, it is because it was part of a previous thread.

sespiros:
Thanks gain for the changes. I summarized a couple of more comments in a single comment because it is hard to keep track of all the changes in the Github UI:

- [x] Mentioning that the current path will be maintained and that the sss path will be "activated" when `SeedShareOwnerThreshold >= 2`.

Describe explicitly some edge cases like: 
- [ ] we will validate that `threshold <= len(SeedshareOwnerPubKeys)` and reject with a clear error at manifest set time 
- [x] what happens with `len(SeedshareOwnerPubKeys) == 0`, does it represent a deployment that opts out of recovery? or do we reject it?
- [x] what happens when processing old manifests (unmarshalled as `SeedShareOwnerThreshold==0`)

For that last one, to clarify, I would expect to read something like:
- field absent -> `threshold = len(SeedshareOwnerPubKeys)`
- field present and >= 1 -> use the value
- field present and == 0 -> reject; RFC says minimum is 1

then explain that this value 1 is a user-facing threshold which actually falls back to the existing direct-encrypted-seed etc.

- [x] `--force` semantics for threshold recovery like 2 sentences saying that we prefer peer recovery when `--force` is not specified and that "scale down to 1" for user recovery should be documented
- [x] what happens if the Coordinator dies mid share accumulation? Worth a sentence noting that the ceremony is expected to complete within a single pod lifetime; in-flight shares are lost on Coordinator restart and the ceremony must be restarted.
- [ ] There is a sentence mentioning "If 2 seed shares remain to meet...". This one should go to a future section since we have concluded on the "scale down to 1" design.

burgerdev:
> Thanks gain for the changes. I summarized a couple of more comments in a single comment because it is hard to keep track of all the changes in the Github UI:
> 
> * [ ]  Mentioning that the current path will be maintained and that the sss path will be "activated" when `SeedShareOwnerThreshold >= 2`.

I don't think this needs to be explicit in the RFC, but leaving a note might save the implementor time if we don't get to it right away.

> Describe explicitly some edge cases like:
> 
> * [ ]  we will validate that `threshold <= len(SeedshareOwnerPubKeys)` and reject with a clear error at manifest set time

That's a very minor detail.

> * [ ]  what happens with `len(SeedshareOwnerPubKeys) == 0`, does it represent a deployment that opts out of recovery? or do we reject it?

Yes, and that is existing logic we don't need to touch here: https://docs.edgeless.systems/contrast/howto/immutable-deployments#preventing-decryption-of-state-and-recovery.

> * [ ]  what happens when processing old manifests (unmarshalled as `SeedShareOwnerThreshold==0`)
>
> For that last one, to clarify, I would expect to read something like:
> 
> * field absent -> `threshold = len(SeedshareOwnerPubKeys)`
> * field present and >= 1 -> use the value
> * field present and == 0 -> reject; RFC says minimum is 1

I don't think you can decide between "field absent" and "field present but default" in proto3.

The change would be backwards compatible if we interpreted 0 as 1.

> then explain that this value 1 is a user-facing threshold which actually falls back to the existing direct-encrypted-seed etc.
> 
> * [ ]  `--force` semantics for threshold recovery like 2 sentences saying that we prefer peer recovery when `--force` is not specified and that "scale down to 1" for user recovery should be documented

The RFC does not change `--force` semantics - do you want that to be explicit? Scale down to 1 is in the RFC right now.

> * [ ]  what happens if the Coordinator dies mid share accumulation? Worth a sentence noting that the ceremony is expected to complete within a single pod lifetime; in-flight shares are lost on Coordinator restart and the ceremony must be restarted.

Fair.

> * [ ]  There is a sentence mentioning "If 2 seed shares remain to meet...". This one should go to a future section since we have concluded on the "scale down to 1" design.

That is a motivational edge case for the conclusion "ask the user to scale down to 1".

sespiros:
> I don't think this needs to be explicit in the RFC, but leaving a note might save the implementor time if we don't get to it right away.

ack, it could also be an implementation note or something. Daniel left the same comment so I simply gathered it in the list again because the conversation is unresolved.

> That's a very minor detail.

idd, I just find it helpful though to have these details specified when implementing specs.

> I don't think you can decide between "field absent" and "field present but default" in proto3.
> 
> The change would be backwards compatible if we interpreted 0 as 1.

afaict the `Manifest field is bytes in `SetManifestRequest` (internal/userapi/userapi.proto) and the deserialization is doing:

https://github.com/edgelesssys/contrast/blob/11986ab8b83413ad2e9db2e84552d77214369c7e/coordinator/internal/userapi/userapi.go#L90

That being said your suggestion also works (in case you have something in mind that I missed)

> The RFC does not change `--force` semantics - do you want that to be explicit? Scale down to 1 is in the RFC right now.

Explicit would be nice. This took a second to grok on first read.

> That is a motivational edge case for the conclusion "ask the user to scale down to 1".

ack, it just read ambiguous to me.

davidweisse:
> * [ ]  Mentioning that the current path will be maintained and that the sss path will be "activated" when `SeedShareOwnerThreshold >= 2`.

> * [ ]  what happens when processing old manifests (unmarshalled as `SeedShareOwnerThreshold==0`)

Expanded the manifest section. Since this is JSON, we can make this a pointer in go, so it is nullable. If the value is absent, it is treated as `len(SeedShareOwnerPubKeys)`. This is also in line with the default value, so when generating a new manifest, we don't need to explicitly set this. This isn't backwards compatible, but IMO more what I would expect, since it is falling back to the default.

> * [ ]  we will validate that `threshold <= len(SeedshareOwnerPubKeys)` and reject with a clear error at manifest set time

This should be part of the manifest validation, when the CLI first reads a manifest. Added a sentence.

> * [ ]  `--force` semantics for threshold recovery like 2 sentences saying that we prefer peer recovery when `--force` is not specified and that "scale down to 1" for user recovery should be documented

Added this for visibility.

> * [ ]  what happens if the Coordinator dies mid share accumulation? Worth a sentence noting that the ceremony is expected to complete within a single pod lifetime; in-flight shares are lost on Coordinator restart and the ceremony must be restarted.

Yes.

burgerdev:
As you both probably noticed, I was assuming our manifest is proto, for some reason. 😅

sespiros (approved):
Thanks for the changes!

PR #2344, opened 2026-04-23
cli: extrapolate vCPU launch measurements from single-cpu value

charludo:


charludo:
- [x] [SNP multi-cpus](https://github.com/edgelesssys/contrast/actions/runs/24889435671)
- [x] [SNP-GPU multi-cpus](https://github.com/edgelesssys/contrast/actions/runs/24889446549)
- [x] [SNP regression](https://github.com/edgelesssys/contrast/actions/runs/24889456667)
- [x] [SNP-GPU regression](https://github.com/edgelesssys/contrast/actions/runs/24889464230)

packages/by-name/contrast/snp-ap-eip/package.nix
```diff
@@ -0,0 +1,26 @@
...
+
+  dontUnpack = true;
+
+  buildPhase = ''
+    mkdir $out
+    ${lib.getExe kata.sev-snp-measure} ap-eip \
+      --ovmf ${ovmf-snp} > $out/ap-eip.hex
+    for file in $out/*.hex; do
+      truncate -s -1 "$file"
+    done
```
burgerdev, line 25: A comment like `# strip trailing newlines` might be helpful (at least I had to look up what `-s -1` does).

tools/sev-snp-measure-go/main.go
```diff
@@ -105,5 +105,39 @@ Kata Containers SNP measurement). Output is the hex-encoded SHA-384 digest.`,
...
+func newAPEIPCmd() *cobra.Command {
+	var ovmfPath string
+
+	cmd := &cobra.Command{
+		Use:   "ap-eip",
+		Short: "Print the SEV-ES AP reset EIP from an OVMF firmware image",
+		Long: `ap-eip reads the SEV-ES AP reset EIP from the OVMF footer table and prints
+it as an 8-digit lowercase hex value. This value is specific to the OVMF build
+and is needed to calculate launch measurements for varying vCPU counts at
+verify time without storing per-vCPU measurements.`,
```
burgerdev: Could we link to any documentation about this value, or at least the OVMF source for this?

cli/cmd/generate.go
```diff
@@ -178,6 +179,21 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	// Inject the APEIP (from the OVMF passed at build time) into every SNP
+	// reference value entry. This allows SNPValidateOpts to derive launch
+	// measurements for all vCPU counts at verify time without storing them all.
+	// In unit tests the embedded ap-eip.hex is a placeholder, so we skip
+	// using the value. This means we fall back to assuming the launch digest to be exact.
+	if apEIP, err := parsedAPEIP(); err != nil {
+		log.Warn("AP EIP not available; falling back to pre-computed launch digests", "err", err)
```
burgerdev, line 188: What kind of situation may lead to this?

cli/cmd/generate.go
```diff
@@ -716,32 +732,40 @@ func patchRuntimeClassName(defaultRuntimeHandler string) func(*applycorev1.PodSp
...
-	digestBytes, err := hex.DecodeString(snpLaunchDigests[platform][cpuCount][productName].LaunchDigest)
+	// Derive the exact launch digest for the given vCPU count from the 1-vCPU seed and APEIP.
+	seed, apeip, ok := snpSeedAndAPEIPForProduct(mnf.ReferenceValues.SNP, targetPlatform, productName)
+	if !ok || apeip == "" {
+		return fmt.Errorf("no SNP reference values with APEIP for product %q on platform %q", productName, targetPlatform)
+	}
+	seedBytes, err := seed.Bytes()
 	if err != nil {
-		return fmt.Errorf("decode %s launch digest: %w", productName, err)
+		return fmt.Errorf("decode TrustedMeasurement: %w", err)
```
burgerdev: Don't we want the `productName` in ther error message anymore?

cli/cmd/generate.go
```diff
@@ -766,16 +790,21 @@ func computeAndAnnotateIDBlockAnnotations(targetPlatform platforms.Platform, cpu
...
+		}
+	}
+	return "", "", false
+}
+
 // patchIDBlockAnnotation computes the SNP ID block and ID auth for each CC workload and injects
 // them as Kata annotations. The guest policy is read from the manifest's SNP reference values;
-// the launch digest is looked up from the embedded snp-launch-digests.json. Both are combined at
-// generate time via idblock.IDBlocksFromLaunchDigest.
+// the launch digest is derived at generate time from the 1-vCPU seed and APEIP via idblock.IDBlocksFromLaunchDigest.
```
burgerdev: That does not sound right: `IDBlocksFromLaunchDigest` does not derive a launch digest, does it?

internal/attestation/snp/validator.go
```diff
@@ -155,6 +158,140 @@ func (s snpReport) ClaimsToCertExtension() ([]pkix.Extension, error) {
...
+
+	if err := verify.SnpAttestationContext(ctx, attestationData, v.verifyOpts); err != nil {
+		return fmt.Errorf("verifying report: %w", err)
+	}
+	v.logger.Info("Successfully verified report signature")
+
+	actualMeasurement := attestationData.Report.GetMeasurement()
+
+	for vcpus := 1; vcpus <= 220; vcpus++ {
+		expected, err := snpmeasure.ExtendSNPLaunchDigest(v.seed, vcpus, v.apEIP, v.vcpuSig)
```
burgerdev: Does that result in `O(max_vcpus^2)` sha256 operations? Maybe we could keep the state of the previous iteration...

internal/manifest/manifest.go
```diff
@@ -150,19 +147,20 @@ func (m *Manifest) SNPValidateOpts(kdsGetter *certcache.CachedHTTPSGetter) ([]SN
...
+			allowedChipIDs = append(allowedChipIDs, chipID)
 		}
-		idKeyHash := sha512.Sum384(idKeyBytes)
 
 		validateOpts := &snpvalidate.Options{
-			Measurement:  trustedMeasurement,
+			// Measurement holds the 1-vCPU seed. When APEIP is set the IterativeValidator
+			// reads it as the seed and overrides it per vCPU count; when APEIP is absent it
+			// is treated as the exact expected measurement (backwards compatibility).
+			Measurement:  seed,
```
burgerdev, line 163: According to the docstring of `NewIterativeValidator`, the field `Measurement` should not be set. Am I reading that wrongly?

cli/verifier/cpu_count_valid.go
```diff
@@ -25,7 +25,7 @@ func (v *CPUCountValid) Verify(toVerify any) error {
 		}
 
 		cpuCount := kuberesource.GetPodCPUCount(spec)
-		if cpuCount < 1 || cpuCount > 8 {
+		if cpuCount < 1 || cpuCount > 220 {
 			// TODO(charludo): find way to add pod name to error message
 			findings = errors.Join(findings, fmt.Errorf("pod failed verification: currently only 0-7 additional vCPUs are supported (1-8 vCPUs in total), but %d were requested", cpuCount))
```
burgerdev: The error message needs to be updated, too.

tools/sev-snp-measure-go/main.go
charludo: I think I found the primary source in the edk2 source. Linked it. Not sure how helpful though. The original comment mostly just described what was in the original/ported code.

cli/cmd/generate.go
charludo: The AP EID not being available, you mean? As the comment says, unit tests (or more precisely, unit tests ran with `go test`), since the `ap-eid.hex` file gets populated only during `nix build`. Pretty sure we have/had exceptions like this for the embedded reference values as well, since they have the same issue.

cli/cmd/generate.go
charludo: sorry, re-added, also in the error above it.

cli/cmd/generate.go
charludo: Yeah sorry I should have read that twice.

charludo:
- [x] [Metal-QEMU-SNP-GPU, multiple-cpus](https://github.com/edgelesssys/contrast/actions/runs/25540046289)
- [x] [Metal-QEMU-TDX-GPU, multiple-cpus](https://github.com/edgelesssys/contrast/actions/runs/25540047263)

internal/manifest/manifest.go
charludo: No, sorry. I just missed removing that after fixing backwards compatibility. `Measurement` is used both in the new iterative validator and in the "classical" one. For the iterative one just as a means to get the seed in, for the classical one for actual use. Fixed the docstring.

burgerdev (approved):
Thanks!

PR #2345, opened 2026-04-23
kata: add darwin-compat patch for genpolicy's FileType::from

sespiros:
Lets genpolicy build as an aarch64-darwin Mach-O binary for the contrast CLI.

https://github.com/kata-containers/kata-containers/compare/main...sespiros:kata-containers:darwin-libc-mode

### DIFF
--- a/packages/by-name/kata/genpolicy/package.nix
+++ b/packages/by-name/kata/genpolicy/package.nix
@@ -11,6 +11,7 @@
   libiconv,
   zlib,
   cmake,
+  stdenv,
   stdenvNoCC,
   applyPatches,
 }:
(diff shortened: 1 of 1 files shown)


packages/by-name/kata/runtime/package.nix
```diff
@@ -149,6 +149,11 @@ buildGoModule (finalAttrs: {
       # Drop after upgrading to Kata 3.29.0.
       # https://github.com/kata-containers/kata-containers/security/advisories/GHSA-q49m-57vm-c8cc
       ./0022-genpolicy-restrict-symlinks-in-CopyFile.patch
+
+      # Makes kata-agent-policy's FileType::from compile on Darwin, where libc's mode_t (and the
+      # S_IF* constants) is u16 rather than u32. Required to build genpolicy as an aarch64-darwin
+      # Mach-O binary for the contrast CLI. Not yet submitted upstream.
+      ./0023-genpolicy-make-FileType-from-portable-across-Darwin.patch
```
burgerdev: I just tried building `genpolicy` on main and that worked. Only if I run `cargo test`, the `src/agent/policy` crate is pulled in. Should we just turn off the integration tests in our builds (they are not nix-compatible anyway)?

burgerdev (approved):
Thanks!

PR #2346, opened 2026-04-23
dev-docs: pin git format-patch to myers diff algorithm

sespiros:
Users with a non-default 'diff.algorithm' in their global git config (e.g. 'patience') otherwise produce byte-different patches from the same commit range, which creates spurious churn whenever the patch set is regenerated.


burgerdev (approved):
Nice corner case, thanks!

PR #2348, opened 2026-04-24
Post 1.20.0 release updates to main

edgelessci:
Updating main as part of the 1.20.0 release.


charludo:
@burgerdev this can/should be megered, right? Just noticed the version change hasn't landed in main yet.  I think according to our dev-docs at least, this should also happen before the actual release, usually.

burgerdev:
Yes, I always forget about that step. Maybe we should automate it.

burgerdev (approved):

PR #2353, opened 2026-04-27
[2/8] attestation: add insecure issuer and validator

msanft:
Split from #2337 as part of a stacked review series.

Depends on: https://github.com/edgelesssys/contrast/pull/2352

This PR adds the insecure attestation pieces only:
- insecure issuer
- insecure validator
- tests for issuer/validator interaction

It is intentionally not wired into the rest of the product yet.

PR #2354, opened 2026-04-27
[4/8] initdata-processor: support insecure platforms

msanft:
Split from #2337 as part of a stacked review series.

Depends on: https://github.com/edgelesssys/contrast/pull/2353

This PR teaches the initdata-processor how to operate on insecure platforms:
- skip TEE digest validation when no TEE platform is present
- serve hostdata for insecure attestation
- switch the service unit to notify mode for the long-running insecure case

### DIFF
--- a/initdata-processor/main.go
+++ b/initdata-processor/main.go
@@ -5,16 +5,25 @@ package main
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
 	"log"
+	"net"
+	"net/http"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"slices"
+	"strings"
 	"syscall"
+	"time"
 
 	"github.com/edgelesssys/contrast/initdata-processor/policy"
 	"github.com/edgelesssys/contrast/initdata-processor/validator"
+	"github.com/edgelesssys/contrast/internal/attestation/insecure"
 	"github.com/edgelesssys/contrast/internal/initdata"
 )
 
@@ -23,13 +32,19 @@ const (
 	insecureConfigPath = "/run/insecure-cfg"
 )
 
-var version = "0.0.0-dev"
+var (
+	version           = "0.0.0-dev"
+	kernelCmdlinePath = "/proc/cmdline"
+)
 
 // We always exit with status code 0 so that the Kata agent can start and propagate errors to
 // the runtime.
 func main() {
 	log.Printf("Contrast initdata-processor %s", version)
 	log.Print("Report issues at https://github.com/edgelesssys/contrast/issues")
+
+	var hostdata []byte
+	var insecurePlatform bool
 
 	// Handle initdata.
 	if err := os.MkdirAll(measuredConfigPath, 0o755); err != nil {
@@ -46,7 +61,8 @@ func main() {
 		failf("%s is not an initdata device: %v", device, err)
 		return
 	}
-	if err := handleInitdata(doc); err != nil {
+	hostdata, insecurePlatform, err = handleInitdata(doc)
+	if err != nil {
 		failf("handling initdata: %v", err)
 		return
 	}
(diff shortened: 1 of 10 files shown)


initdata-processor/main.go
```diff
@@ -59,44 +68,89 @@ func main() {
...
-		return fmt.Errorf("creating validator: %w", err)
+		return nil, false, fmt.Errorf("computing initdata digest: %w", err)
 	}
-	if err := validator.ValidateDigest(digest); err != nil {
-		return fmt.Errorf("validating initdata digest: %w", err)
+
+	v, verr := validator.New()
+	if errors.Is(verr, validator.ErrNoPlatform) {
+		log.Print("WARNING: No TEE platform detected, skipping initdata digest validation. This is expected on insecure platforms.")
+		insecurePlatform = true
```
burgerdev: I'm a bit worried that there could be a split brain scenario, where the initdata-processor could not get a working validator but the initializer could, but the initdata would now not be validated. Could we decide whether to enter the insecure path by an initdata field instead?

msanft: Thanks, adjusted!

internal/initdata/initdata.go
```diff
@@ -106,9 +106,14 @@ func (r Raw) EncodeKataAnnotation() (string, error) {
...
-	Version   string            `toml:"version"`
-	Algorithm string            `toml:"algorithm"`
-	Data      map[string]string `toml:"data"`
+	Version   string `toml:"version"`
+	Algorithm string `toml:"algorithm"`
+	// Insecure allows running workloads on non-TEE development platforms.
+	// When set, the initdata-processor serves the initdata digest to the
+	// insecure attestation issuer via HTTP instead of validating it against
+	// TEE hostdata.
+	Insecure bool              `toml:"insecure,omitempty"`
```
burgerdev: Adding a top-level field may break the Kata/CoCo parts of initdata processing. This should go under `Data["contrast-initdata-processor.json"] = "{ insecure: true }"`, rather.

burgerdev (commented):
mostly lgtm

burgerdev (changes requested):
Unfortunately, this does not work with initdata: An attacker can launch the VM with TEE support and set HOSTDATA to their desired value, but supply an initdata file that includes `Insecure`. HOSTDATA is now unvalidated, but still shows in reports.

The previous version was arguably safer, but safe enough? idk.

initdata-processor/main.go
```diff
@@ -59,44 +73,107 @@ func main() {
...
+// serveHostdata starts an HTTP server that serves the hostdata digest.
+func serveHostdata(hostdata []byte) error {
+	mux := http.NewServeMux()
+	mux.HandleFunc("GET /hostdata", func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/octet-stream")
+		if _, err := w.Write(hostdata); err != nil {
+			log.Printf("hostdata write error: %v", err)
+		}
+	})
+	return http.ListenAndServe(insecure.HostdataAddr, mux)
```
charludo: Should probably add a `http.Server.Shutdown` hook, just for cleanliness' sake, since this currently can never exit cleanly, right?

charludo (approved):
I just have one nit, LGTM otherwise.

nodeinstaller/internal/kataconfig/config.go
```diff
@@ -75,6 +76,9 @@ func KataRuntimeConfig(
 
 	// Replace the kernel params entirely (and don't append) since that's
 	// also what we do when calculating the launch measurement.
+	if platforms.IsInsecure(platform) {
+		qemuExtraKernelParams = strings.TrimSpace(qemuExtraKernelParams + " contrast.allow_insecure_attestation=1")
+	}
 	config.Hypervisor["qemu"]["kernel_params"] = qemuExtraKernelParams
```
burgerdev: The comment above now makes a little less sense, consider moving the if block above it.

initdata-processor/main.go
```diff
@@ -59,44 +68,89 @@ func main() {
...
+		return nil, false, fmt.Errorf("computing initdata digest: %w", err)
 	}
-	if err := validator.ValidateDigest(digest); err != nil {
-		return fmt.Errorf("validating initdata digest: %w", err)
+
+	v, verr := validator.New()
+	if errors.Is(verr, validator.ErrNoPlatform) {
+		log.Print("WARNING: No TEE platform detected, skipping initdata digest validation. This is expected on insecure platforms.")
+		insecurePlatform = true
+	} else if verr != nil {
```
burgerdev, line 125: I think 9d7ae288094aab7542a801edf57b4b42417e4901 should be squashed with the next commit, such that we don't have an intermediate state where validation is skipped but the kernel command lie is not checked.

burgerdev (approved):
lgtm, just a nit and a squash

initdata-processor/main.go
```diff
@@ -59,46 +76,128 @@ func main() {
...
+	}
+	go func() {
+		<-ctx.Done()
+		shutdownCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)
+		defer cancel()
+		if err := server.Shutdown(shutdownCtx); err != nil {
+			log.Printf("hostdata server shutdown error: %v", err)
+		}
+	}()
+	if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
sespiros, line 178: This came up during AI review, all incoming connections to the guest are disabled in:

https://github.com/edgelesssys/contrast/blob/25c0a7169a9a562dddced9fff62dcdc5ba50ab25/packages/nixos/kata.nix#L188

so I assume connections to the service will be dropped by default when service mesh is enabled. Service mesh seems relevant for the benchmarking use case (comparing across CC vs non-CC needs the same proxy/iptables stack I assume), but not sure if that combination is actually in scope here.

It seems it is (also afaict nothing in https://github.com/edgelesssys/contrast/blob/main/internal/kuberesource/mutators.go stops you from using service mesh with an insecure pod).

So depending on what we need here we either have to have a rule that allows this connection or refuse enabling the service mesh when an insecure platform is detected during generate (by catching that in the mutators).

sespiros (commented):
Thanks, an inline comment and some others that I had looking at the merged PRs so far, so feel free to incorporate as you see fit here/in one of the upcoming PRs/follow-ups:

* I think the relevant PR is already merged by now but raising this, this line here:
https://github.com/edgelesssys/contrast/blob/25c0a7169a9a562dddced9fff62dcdc5ba50ab25/cli/verifier/cpu_count_valid.go#L23
is not using the `IsContrastPod` helper, so insecure pods bypass that CPU count check.

* The insecure validator https://github.com/edgelesssys/contrast/blob/main/internal/attestation/insecure/validator.go is not logging any message (using the logger) like the other validators. I think we would want to see in the logs at least:
  - when an insecure attestation is accepted
  - log on entry/exit (like the rest)

* I noticed insecure configurations use the `snp` name in https://github.com/edgelesssys/contrast/blob/25c0a7169a9a562dddced9fff62dcdc5ba50ab25/nodeinstaller/internal/targetconfig/targetconfig.go#L40. Shouldn't we rename these to something more fitting like `configuration-qemu-bm.toml`?

* This is a readability nit but it took me a minute to parse https://github.com/edgelesssys/contrast/blob/25c0a7169a9a562dddced9fff62dcdc5ba50ab25/internal/manifest/runtimehandler.go#L46 I would either add a comment explaining the bounds or make the range prefix-aware i.e `minParts, maxParts := 4, 5; if isInsecure { minParts, maxParts = 3, 4 }`

msanft:
> Shouldn't we rename these to something more fitting like `configuration-qemu-bm.toml`?

I explicitly opted not to do so, since we want the insecure runtimes to reflect the CC versions as closely as possible, to allow easy performance benchmarking, for example. We do want the platform-specific kernel, etc., for example.

msanft:
See CON-210, CON-211

PR #2356, opened 2026-04-27
[5/8] coordinator: support insecure manifests behind opt-in

msanft:
Split from #2337 as part of a stacked review series.

Depends on: https://github.com/edgelesssys/contrast/pull/2355

This PR wires insecure manifests into the Coordinator behind explicit opt-in:
- manifest helpers for insecure reference values
- Coordinator opt-in gate
- validator registration for insecure manifests
- user API and tests

Closes CON-212

### DIFF
--- a/coordinator/internal/userapi/userapi.go
+++ b/coordinator/internal/userapi/userapi.go
@@ -60,6 +60,10 @@ type Server struct {
 	guard     guard
 	discovery discovery
 
+	// allowInsecure selects the manifest security level accepted by the Coordinator. It defaults to
+	// secure manifests and can be switched to insecure manifests via MakeInsecure.
+	allowInsecure bool
+
 	userapi.UnimplementedUserAPIServer
 }
 
@@ -70,6 +74,14 @@ func New(logger *slog.Logger, guard guard, discovery discovery) *Server {
 		guard:     guard,
 		discovery: discovery,
 	}
+}
+
+// MakeInsecure configures the Server to accept only manifests that contain insecure platforms.
+//
+// This voids the security guarantees of the deployment and must only be used for testing or
+// benchmarking. It should be called before the Server starts serving requests.
+func (s *Server) MakeInsecure() {
+	s.allowInsecure = true
 }
 
 // SetManifest registers a new manifest at the Coordinator.
@@ -146,6 +158,11 @@ func (s *Server) SetManifest(ctx context.Context, req *userapi.SetManifestReques
 			Salt:       salt,
 			SeedShares: seedShares,
 		}
+	}
+
+	if err := s.checkManifestSecurity(m); err != nil {
+		s.logger.Warn("SetManifest rejected the manifest", "err", err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	state, err := s.guard.UpdateState(ctx, oldState, se, req.GetManifest(), req.GetPolicies())
@@ -229,18 +246,40 @@ func (s *Server) Recover(ctx context.Context, req *userapi.RecoverRequest) (*use
 		s.logger.Info("Skipping sanity checks because user recovery was forced")
 	}
 
-	_, err = s.guard.ResetState(ctx, oldState, &seedAuthorizer{req: req})
+	_, err = s.guard.ResetState(ctx, oldState, &seedAuthorizer{req: req, checkManifestSecurity: s.checkManifestSecurity})
 	if err != nil {
 		return nil, fmt.Errorf("resetting state: %w", err)
 	}
 	return &userapi.RecoverResponse{}, nil
 }
 
+// checkManifestSecurity verifies that the manifest doesn't mix secure and insecure platforms and
+// that its security level matches the Coordinator's configuration.
+func (s *Server) checkManifestSecurity(mnfst *manifest.Manifest) error {
+	hasInsecure := mnfst.HasInsecurePlatforms()
+	hasSecure := mnfst.HasSecurePlatforms()
+
+	if hasInsecure && hasSecure {
+		return ErrMixedManifestNotAllowed
+	}
+	if hasInsecure && !s.allowInsecure {
+		return ErrInsecureNotAllowed
+	}
+	if !hasInsecure && s.allowInsecure {
+		return ErrSecureNotAllowed
+	}
+	return nil
+}
+
 type seedAuthorizer struct {
-	req *userapi.RecoverRequest
+	req                   *userapi.RecoverRequest
+	checkManifestSecurity func(*manifest.Manifest) error
 }
 
 func (a *seedAuthorizer) AuthorizeByManifest(ctx context.Context, mnfst *manifest.Manifest) (*seedengine.SeedEngine, *ecdsa.PrivateKey, error) {
+	if err := a.checkManifestSecurity(mnfst); err != nil {
+		return nil, nil, status.Error(codes.FailedPrecondition, err.Error())
+	}
 	if err := validatePeer(ctx, mnfst.SeedshareOwnerPubKeys); err != nil {
 		return nil, nil, status.Errorf(codes.PermissionDenied, "peer not authorized to recover existing state: %v", err)
 	}
@@ -337,6 +376,18 @@ var (
 	ErrAlreadyRecovered = errors.New("coordinator is already recovered")
 	// ErrNeedsRecovery is returned if state exists, but no secrets are available, e.g. after restart.
 	ErrNeedsRecovery = errors.New("coordinator is in recovery mode")
+
+	// ErrInsecureNotAllowed is returned when a manifest contains insecure platforms but the
+	// Coordinator is not configured to allow them.
+	ErrInsecureNotAllowed = errors.New("manifest contains insecure platforms, but the coordinator is not configured to allow them")
+
+	// ErrMixedManifestNotAllowed is returned when a manifest mixes secure and insecure platforms.
+	// Manifests must be either all-secure or all-insecure.
+	ErrMixedManifestNotAllowed = errors.New("manifest must not mix secure and insecure platforms")
+
+	// ErrSecureNotAllowed is returned when a secure manifest is submitted to a Coordinator that is
+	// configured to accept only insecure manifests.
+	ErrSecureNotAllowed = errors.New("secure manifest is not allowed when the coordinator is configured for insecure manifests")
 
 	errNoSignature = errors.New("manifest signature is empty")
 )
--- a/coordinator/main.go
+++ b/coordinator/main.go
@@ -52,6 +52,7 @@ import (
 
 const (
 	metricsEnvVar       = "CONTRAST_METRICS"
+	allowInsecureEnvVar = "CONTRAST_ALLOW_INSECURE"
 	probeAndMetricsPort = 9102
 	// transitEngineAPIPort specifies the default port to expose the transit engine API.
 	transitEngineAPIPort = "8200"
@@ -126,7 +127,12 @@ func run() (retErr error) {
 
 	userAPICredentials := atlscredentials.New(issuer, nil, atls.NoMetrics, loggerpkg.NewNamed(logger, "atlscredentials"))
 	userAPIServer := newGRPCServer(userAPICredentials, serverMetrics)
-	userapi.RegisterUserAPIServer(userAPIServer, userapiserver.New(logger, meshAuth, discovery))
+	userapiService := userapiserver.New(logger, meshAuth, discovery)
+	if os.Getenv(allowInsecureEnvVar) != "" {
+		logger.Warn("Coordinator is configured to allow insecure manifests")
+		userapiService.MakeInsecure()
+	}
+	userapi.RegisterUserAPIServer(userAPIServer, userapiService)
 	serverMetrics.InitializeMetrics(userAPIServer)
 
 	month := 30 * 24 * time.Hour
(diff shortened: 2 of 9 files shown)


coordinator/internal/stateguard/stateguard.go
```diff
@@ -271,6 +282,9 @@ func (g *Guard) UpdateState(_ context.Context, oldState *State, se *seedengine.S
 	if err := json.Unmarshal(manifestBytes, &mnfst); err != nil {
 		return nil, fmt.Errorf("unmarshaling manifest: %w", err)
 	}
+	if !g.allowInsecure && mnfst.AllowInsecure() {
+		return nil, ErrInsecureNotAllowed
+	}
```
charludo: Sorry if this has been decided already, I might just be OOTL and wanted to make sure: Do we also care about the reverse? I.e. is it acceptable to allow users to have a secure deployment run with an insecure coordinator?

coordinator/internal/stateguard/stateguard.go
```diff
@@ -73,7 +80,10 @@ type metrics struct {
 }
 
 // New creates a new state Guard instance.
-func New(hist *history.History, reg *prometheus.Registry, log *slog.Logger) *Guard {
+//
+// If allowInsecure is true, the Guard will accept manifests that contain insecure platforms.
+// Otherwise, setting such a manifest will be rejected with ErrInsecureNotAllowed.
+func New(hist *history.History, reg *prometheus.Registry, log *slog.Logger, allowInsecure bool) *Guard {
```
charludo: Not super happy with just passing an unnamed `bool` into here. Invocations don't make it super obvious what a serious effect this has. How would you feel about instead leaving `New` as-is, without the option to `allowInsecure`, and instead add something like `func (g *Guard) MakeInsecure()` that needs to be explicitly called?

internal/atls/issuer/issuer_linux.go
```diff
@@ -29,6 +29,7 @@ func New(log *slog.Logger) (atls.Issuer, error) {
 			logger.NewWithAttrs(logger.NewNamed(log, "issuer"), map[string]string{"tee-type": "tdx"}),
 		), nil
 	default:
-		return nil, fmt.Errorf("unsupported platform: %T", cpuid.CPU)
+		log.Warn("No TEE platform detected, using insecure attestation issuer")
+		return insecure.NewIssuer(), nil
 	}
```
charludo, line 34: This should not default to insecure. Any chance to something similar here as with `allowInsecureAttestation()` in the initedata processor?

coordinator/internal/stateguard/credentials.go
```diff
@@ -104,6 +105,12 @@ func (c *Credentials) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.A
 			logger.NewWithAttrs(logger.NewNamed(c.logger, "validator"), map[string]string{"reference-values": name}), &authInfo, name))
 	}
 
+	if state.Manifest().AllowInsecure() {
```
charludo: IMO there need to be two variants of `AllowInsecure`. One to use here, that is only true when *all* platforms allow insecure in the manifest (just as an additional guard), and the other one in the current form for use in 
```go
    if !g.allowInsecure && mnfst.AllowInsecure() {
		return nil, ErrInsecureNotAllowed
	}
```
below, that is true when *any* platform allows insecure.

coordinator/main.go
```diff
@@ -115,7 +116,12 @@ func run() (retErr error) {
 
 	hist := history.NewWithStore(logger.WithGroup("history"), store)
 
-	meshAuth := stateguard.New(hist, promRegistry, logger)
+	_, allowInsecure := os.LookupEnv(allowInsecureEnvVar)
+	if allowInsecure {
+		logger.Warn("Coordinator is configured to allow insecure manifests")
+	}
```
charludo: Check value, not just presence. `CONTRAST_ALLOW_INSECURE=0` should not enable insecure mode. I can easily see users deliberately "disabling" insecure mode like this. `strconv.ParseBool` should do the job.

charludo (changes requested):
Thanks! Sorry review took so long. I have a couple of comments, see below; the one architectural one I want to get your opinion on is if we should allow mixed manifests. Is that something you actually require? Because if not, then I'd strongly urge to allow either all-secure, or all-insecure manifests, no mix.

coordinator/internal/stateguard/stateguard_test.go
```diff
@@ -543,6 +574,28 @@ func newManifest(t *testing.T) (*manifest.Manifest, []byte, [][]byte) {
 	return mnfst, mnfstBytes, [][]byte{policy}
 }
 
+func newInsecureManifest(t *testing.T) (*manifest.Manifest, []byte, [][]byte) {
```
sespiros: There should also be a unit test that exercises mixed manifests:
- Build a manifest with a secure and insecure refvals and two paired policy entries.
- Runs `ServerHandshake` twice, once where the peer presents OID-1 attestation, once with OID-99, both authorizing matching policies.
- Asserts both succeed and the mesh credentials issued are functionally equivalent.

sespiros (commented):
Thanks, I second Charlotte's comment about needing to decide whether we want to allow mixed manifests. My understanding was that we actually do need to do that for the benchmarking use case that triggered this PR series. 

**If** that's the case we should at least close the impersonation gap by adding a per-policy attestation-type binding (which doesn't exist today). An attacker who controls an insecure node can produce an attestation that authenticates as any CC policy in the manifest, because `meshapi.NewMeshCert` does `Policies[hostData]` without checking which validator produced the hostData.

For example like:
```go
type PolicyEntry struct {
    SANs                   []string
    WorkloadSecretID       string
    Role                   Role
    AllowedAttestationOIDs []asn1.ObjectIdentifier `json:",omitempty"`
}
```

internal/manifest/manifest.go
```diff
@@ -120,6 +120,21 @@ func (m *Manifest) CoordinatorPolicyHash() (HexString, error) {
 	return "", errors.New("no coordinator found in manifest")
 }
 
+// AllowInsecure returns true if the manifest contains reference values for insecure platforms.
+func (m *Manifest) AllowInsecure() bool {
```
sespiros: Maybe rename to `ContainsInsecure()` or `HasInsecurePlatforms()`

internal/manifest/manifest.go
```diff
@@ -129,6 +144,13 @@ func (m *Manifest) SNPValidateOpts(kdsGetter *certcache.CachedHTTPSGetter) ([]SN
 
 	var out []SNPValidatorOptions
 	for _, refVal := range m.ReferenceValues.SNP {
+		if p, err := platforms.FromString(refVal.Platform); err == nil && platforms.IsInsecure(p) {
+			continue
+		}
+		if len(refVal.TrustedMeasurement) == 0 {
```
sespiros, line 173: Could we add a unit test for this that exercises this code path directly? The check is non-redundant despite looking like a duplicate of Validate's. Validate's new insecure early-return means the "non-empty TrustedMeasurement" case no longer holds for the whole SNP refval slice, so this is the actual check for anything that bypasses Validate. None of the existing empty-TrustedMeasurement tests actually reach SNPValidateOpts. Something like:

```go
func TestSNPValidateOptsRejectsEmptyMeasurement(t *testing.T) {
    m := newTestManifestSNP()
    m.ReferenceValues.SNP[0].TrustedMeasurement = ""
    _, err := m.SNPValidateOpts(nil)
    require.Error(t, err)
}
```

msanft:
> the one architectural one I want to get your opinion on is if we should allow mixed manifests. Is that something you actually require?

I've thought about this for some time now, and couldn't think of a use-case we'd need a mixed manifest for. I second your opinion that it should not be allowed then.

coordinator/internal/stateguard/stateguard.go
```diff
@@ -92,16 +97,23 @@ func New(hist *history.History, reg *prometheus.Registry, log *slog.Logger, allo
...
 		metrics: metrics{
 			manifestGeneration: manifestGeneration,
 		},
 		clock: clock.RealClock{},
 	}
 }
 
+// MakeInsecure configures the Guard to accept manifests that contain insecure platforms.
+//
+// This significantly lowers the security guarantees of the deployment and must only be used for
```
msanft: s/significantly lowers/voids/

coordinator/internal/stateguard/stateguard.go
msanft: Covered by disallowing mixed manifests

msanft:
ping, this should be ready for another round of review @burgerdev @sespiros @charludo

sespiros (changes requested):
Thanks for the changes. We recently did a significant refactor on our validation logic so this needs to be rebased for a final clean review.

The rest of the changes lgtm.

coordinator/internal/stateguard/stateguard.go
```diff
@@ -271,6 +294,14 @@ func (g *Guard) UpdateState(_ context.Context, oldState *State, se *seedengine.S
 	if err := json.Unmarshal(manifestBytes, &mnfst); err != nil {
 		return nil, fmt.Errorf("unmarshaling manifest: %w", err)
 	}
+	if mnfst.HasInsecurePlatforms() {
```
sespiros: A Coordinator that has switched to an insecure state must not transition back to a secure state. Check `oldState` and return a specific error when an insecure-to-secure transition is attempted. We can also add a test covering this transition or even better a test matrix testing all transitions.

sespiros (changes requested):
Thanks, left a few comments.

I am thinking that recovering an insecure manifest should require `CONTRAST_ALLOW_INSECURE`. We can check the recovered manifest in `ResetState` and return `ErrInsecureNotAllowed` when it is insecure and `allowInsecure` is false. We can also add a test that simulates a restart: create a `Guard`, call `MakeInsecure`, and use `UpdateState` to set an insecure manifest. Then create a new `Guard` using the same `History`, without calling `MakeInsecure`, and verify that `ResetState` returns `ErrInsecureNotAllowed`.

charludo (approved):
Thanks! My original concerns are all covered, and their implementation LGTM. From my side, and not including @sespiros comments of course, this is good to go.

coordinator/internal/stateguard/stateguard.go
```diff
@@ -65,6 +77,10 @@ type Guard struct {
 	logger  *slog.Logger
 	metrics metrics
 
+	// allowInsecure controls whether manifests with insecure platforms are accepted.
+	// It defaults to false and can be enabled via MakeInsecure.
+	allowInsecure bool
```
burgerdev: The stateguard's purpose is to store, retrieve and update state, and I tried to avoid putting unrelated business logic in here as much as possible. For example, completeness of the manifest, permission to update a manifest and recovery security are checked in userapi (we added a callback to `ResetState` for that). Arguably that's also not pretty, but I don't have a good alternative idea right now.

coordinator/main.go
```diff
@@ -117,7 +118,19 @@ func run() (retErr error) {
 
 	hist := history.NewWithStore(logger.WithGroup("history"), store)
 
+	allowInsecure := false
+	if v, ok := os.LookupEnv(allowInsecureEnvVar); ok {
+		allowInsecure, err = strconv.ParseBool(v)
+		if err != nil {
+			return fmt.Errorf("parsing %s: %w", allowInsecureEnvVar, err)
+		}
+	}
```
burgerdev: This could be simplified to only check whether the env var is not empty. I can't think of a scenario where I'd want to set `CONTRAST_ALLOW_INSECURE=false` (and the absent case needs to be secure anyway).

burgerdev (commented):
Overall lgtm, thanks!

coordinator/internal/stateguard/stateguard.go
burgerdev: Maybe I was not clear enough: let's move this logic out of the stateguard and into the userapi for now. Slicing the userapi into subpackages for request processing, authz and such can then be done later.

sespiros (approved):
Thanks for the changes, nothing more to add.

coordinator/internal/userapi/userapi.go
```diff
@@ -229,18 +246,41 @@ func (s *Server) Recover(ctx context.Context, req *userapi.RecoverRequest) (*use
...
+		}
+		if mnfst.HasSecurePlatforms() {
+			return ErrMixedManifestNotAllowed
+		}
+	}
+	if oldState != nil && oldState.Manifest().HasInsecurePlatforms() && !mnfst.HasInsecurePlatforms() {
+		return ErrInsecureToSecureTransition
+	}
+	return nil
+}
```
burgerdev: Is there a valid use case for transitioning from secure to insecure (which is currently allowed)? If not, we could simplify this a lot:

```go
// checkManifestSecurity verifies that
// - The manifest contains either no insecure refvals, or only insecure ones.
// - The manifest security matches the coordinator security.
func (s *Server) checkManifestSecurity(mnfst *manifest.Manifest) error {
	hasInsecure := mnfst.HasInsecurePlatforms()
	hasSecure := mnfst.HasSecurePlatforms()

	if hasInsecure && hasSecure {
		return ErrMixedManifestNotAllowed
	}

	if hasInsecure && !s.allowInsecure {
		return ErrInsecureNotAllowed
	}

	if !hasInsecure && s.allowInsecure {
		return ErrSecureNotAllowed
	}

	// No need to check history, would have been caught by prior checks. 
	return nil
}
```

msanft: Good point, I don't think we'd need to support transitions between the security levels at all.

burgerdev (approved):
Thanks, lgtm!

PR #2351, opened 2026-04-27
e2e: start nightly earlier to avoid PM interference

burgerdev:

PR #2360, opened 2026-04-27
.github: dry-run release pipeline nightly

charludo:
`e2e_nightly` workflow no longer triggers automatically. Instead, `release` workflow dry-runs every night. "dry run" here means no backport PRs, no draft releases. The artifacts are still published to S3, so we/pm can use them.

### DIFF
--- a/.github/actions/parse_version/action.yml
+++ b/.github/actions/parse_version/action.yml
@@ -1,0 +1,73 @@
+name: Parse version
+description: Parses a vX.Y.Z version string into its components
+
+inputs:
+  version:
+    description: Full version string, e.g. v1.2.3 or v1.2.3-2024-01-01
+    required: true
+
+outputs:
+  without_v:
+    description: Base version without v prefix and pre-release suffix, e.g. 1.2.3
+    value: ${{ steps.parse.outputs.without_v }}
+  part_major:
+    description: Major version number
+    value: ${{ steps.parse.outputs.part_major }}
+  part_minor:
+    description: Minor version number
+    value: ${{ steps.parse.outputs.part_minor }}
+  part_patch:
+    description: Patch version number
+    value: ${{ steps.parse.outputs.part_patch }}
+  major_minor:
+    description: Major.Minor, e.g. 1.2
+    value: ${{ steps.parse.outputs.major_minor }}
+  major_minor_patch:
+    description: Major.Minor.Patch, e.g. 1.2.3
+    value: ${{ steps.parse.outputs.major_minor_patch }}
+  release_branch:
+    description: Release branch name, e.g. release/v1.2
+    value: ${{ steps.parse.outputs.release_branch }}
+  next_minor:
+    description: Next minor version, e.g. 1.3.0
+    value: ${{ steps.parse.outputs.next_minor }}
+  next_minor_pre_without_v:
+    description: Next minor pre-release version without v, e.g. 1.3.0-pre
+    value: ${{ steps.parse.outputs.next_minor_pre_without_v }}
+  next_patch_pre_without_v:
+    description: Next patch pre-release version without v, e.g. 1.2.4-pre
+    value: ${{ steps.parse.outputs.next_patch_pre_without_v }}
+  previous_minor_tag:
+    description: Previous minor release tag, e.g. v1.1.0. Empty if minor is 0.
+    value: ${{ steps.parse.outputs.previous_minor_tag }}
+
+runs:
+  using: composite
+  steps:
+    - id: parse
+      shell: bash
+      env:
+        FULL_VERSION: ${{ inputs.version }}
+      run: |
+        WITHOUT_V=$(echo "${FULL_VERSION#v}" | sed 's/-.*$//')
+        PART_MAJOR=${WITHOUT_V%%.*}
+        PART_MINOR=${WITHOUT_V#*.}
+        PART_MINOR=${PART_MINOR%%.*}
+        PART_PATCH=${WITHOUT_V##*.}
+        {
+          echo "without_v=${WITHOUT_V}"
+          echo "part_major=${PART_MAJOR}"
+          echo "part_minor=${PART_MINOR}"
+          echo "part_patch=${PART_PATCH}"
+          echo "major_minor=${PART_MAJOR}.${PART_MINOR}"
+          echo "major_minor_patch=${PART_MAJOR}.${PART_MINOR}.${PART_PATCH}"
+          echo "release_branch=release/v${PART_MAJOR}.${PART_MINOR}"
+          echo "next_minor=${PART_MAJOR}.$((PART_MINOR + 1)).0"
+          echo "next_minor_pre_without_v=${PART_MAJOR}.$((PART_MINOR + 1)).0-pre"
+          echo "next_patch_pre_without_v=${PART_MAJOR}.${PART_MINOR}.$((PART_PATCH + 1))-pre"
+          if (( PART_MINOR > 0 )); then
+            echo "previous_minor_tag=v${PART_MAJOR}.$((PART_MINOR - 1)).0"
+          else
+            echo "previous_minor_tag="
+          fi
+        } | tee -a "$GITHUB_OUTPUT"
--- a/.github/actions/prepare_github_release/action.yml
+++ b/.github/actions/prepare_github_release/action.yml
@@ -1,0 +1,45 @@
+name: Create backport label and milestone
+description: Creates a backport label for the release branch and the next minor milestone
+
+inputs:
+  release_branch:
+    description: Release branch, e.g. release/v1.2
+    required: true
+  next_minor:
+    description: Next minor version, e.g. 1.3.0
+    required: true
+  github_token:
+    description: GitHub token
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Create backport label
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github_token }}
+        GH_REPO: ${{ github.repository }}
+        RELEASE_BRANCH: ${{ inputs.release_branch }}
+      run: |
+        gh label create "backport ${RELEASE_BRANCH}" --color 576F61 --force
+    - name: Create milestone
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github_token }}
+        NEXT_MINOR: ${{ inputs.next_minor }}
+        REPO: ${{ github.repository }}
+      run: |
+        gh api \
+          -H "Accept: application/vnd.github+json" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          /repos/${REPO}/milestones |
+          jq -r '.[] | .title' | \
+          grep -xqF "v${NEXT_MINOR}" && exit 0
+        gh api \
+          --method POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          /repos/${REPO}/milestones \
+          -f title="v${NEXT_MINOR}" \
+          -f state='open'
--- a/.github/actions/prepare_post_release_pr/action.yml
+++ b/.github/actions/prepare_post_release_pr/action.yml
@@ -1,0 +1,152 @@
+name: Update main branch
+description: >
+  Updates the main branch with release artifacts, docs, and a version bump, then opens a PR. Assumes the workspace root is already checked out to the working/release ref.
+inputs:
+  checkout_main_ref:
+    description: Git ref for the main branch
+    required: false
+    default: main
+  version:
+    description: Full release version, e.g. v1.2.3
+    required: true
+  major_minor:
+    description: Major.Minor, e.g. 1.2
+    required: true
+  major_minor_patch:
+    description: Major.Minor.Patch, e.g. 1.2.3
+    required: true
+  without_v:
+    description: Version without v prefix, e.g. 1.2.3
+    required: true
+  next_minor_pre_without_v:
+    description: Next minor pre-release version without v, e.g. 1.3.0-pre
+    required: true
+  pr_branch_suffix:
+    description: Suffix for the automated PR branch (automated/update-main-after-<suffix>)
+    required: true
+  create_docs_release:
+    description: Whether to create a versioned docs snapshot
+    required: false
+    default: 'true'
+  bump_version:
+    description: Whether to bump the flake version post-release
+    required: false
+    default: 'true'
+  github_token:
+    description: GitHub token
+    required: true
+  nunki_ci_token:
+    description: Token with push and PR creation permissions (NUNKI_CI_COMMIT_PUSH_PR)
+    required: true
+  cachix_token:
+    description: Cachix auth token
+    required: true
+  artifacts_run_id:
+    description: Workflow run ID to download artifacts from. Defaults to the current run.
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout main
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        ref: ${{ inputs.checkout_main_ref }}
+        path: contrast-main
+        persist-credentials: false
+    - uses: ./.github/actions/setup_nix
+      with:
+        githubToken: ${{ inputs.github_token }}
+        cachixToken: ${{ inputs.cachix_token }}
+    - name: Configure git
+      shell: bash
+      run: |
+        git config --global user.name "edgelessci"
+        git config --global user.email "edgelessci@users.noreply.github.com"
+    - name: Create docs release
+      if: inputs.create_docs_release == 'true'
+      shell: bash
+      working-directory: contrast-main/docs
+      env:
+        MAJOR_MINOR: ${{ inputs.major_minor }}
+      run: |
+        nix run ".#base.nixpkgs.yarn" install
+        nix run ".#base.nixpkgs.yarn" docusaurus docs:version "${MAJOR_MINOR}"
+        git add .
+        git commit -am "docs: release ${MAJOR_MINOR}"
+        # Clean up auxiliary files, so next steps run on a clean tree
+        git clean -fx :/
+        git reset --hard HEAD
+    - name: Download image replacements
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+      with:
+        name: image-replacements
+        run-id: ${{ inputs.artifacts_run_id || github.run_id }}
+        github-token: ${{ inputs.github_token }}
+    - name: Update release urls in docs with tags
+      shell: bash
+      working-directory: contrast-main
+      run: nix run ".#base.scripts.update-release-urls"
+    - name: Commit updated docs
+      shell: bash
+      working-directory: contrast-main
+      env:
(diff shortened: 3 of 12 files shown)


.github/workflows/release.yml
```diff
@@ -1,6 +1,8 @@
 name: release
 
 on:
+  schedule:
+    - cron: "30 23 * * *" # 23:30 UTC, every day dry-run release pipeline
```
burgerdev: Please consider https://github.com/edgelesssys/contrast/pull/2351 here.

burgerdev (approved):
Thanks, lgtm, but please hold until tomorrow, 2026-04-28 for the actual release.

.github/workflows/release.yml
sespiros: This test calls `fetchRelease` which attempts to download a release based on a tag from Github. Does this nightly release run create a draft release? Because I otherwise I suspect this to fail.

sespiros: Is this version different than the other version? I see this step is not happening when the job is running on schedule but maybe rename the `VERSION` field to something like `RESOLVED_VERSION` or `EFFECTIVE_VERSION`?

charludo: Heh.... Good catch. I think the options are either, live with this step not being available in the nightly, or actually creating / editing a draft release nightly. What do you think?

sespiros: Any counter-arguments against creating/editing a draft release? Didn't we want to do that anyway?

sespiros: A couple of more thoughts, if we do create nightly draft releases, we should name them something like nightly-vX.Y.Z to avoid collision with an actual release triggered manually and also make it clear that this is a nightly release. Then we should consider whether each day we will be doing new nightlies or overriding an existing one. Not a fan of overriding but also daily nightlies would pollute the release view. Same collision for the images that would get uploaded daily. otoh if we don't create nightly draft releases, we would need to fetch the built artifacts elsewhere (refactor `fetchRelease`), which is not that bad.

charludo: Unfortunately that is mutually exclusive with being able to continue the nightly pipeline into a full release as @burgerdev suggested in the channel, since the version is an input to pretty much everything we build, and changing it changes all the output hashes.

We could of course still name the outputted binary something different, is that what you meant?

sespiros:
Another thing to consider, with the refactor, `e2e nightly` run will now depend on `release-x86_64-linux` to succeed. If that fails for any reason we lose the entire night's e2e run.

.github/workflows/release.yml
sespiros: Which part is mutually exclusive, doing separate draft releases? So the idea is to do a draft release with the next version (`fetchRelease` step works), then subsequent runs will keep overwriting that draft release until some point where the release would be ready, we could just simply "promote" it into a full release?

charludo: No, what I mean is: if we build the release artifacts for version `v1.x.0-2026-05-01`, we then cannot release those artifacts as `v1.x.0`.

So it's either naming them in the nicer way, with the date, *or* building this pipeline out to be able to, in the morinng, click a button to go "actually, let's release this". In the former case, we'd have to run the pipeline again, as we currently do.

sespiros: Yes, that's what I am saying so, say we are in `v1.x.0` currently, each night we will be building and overwriting the next version `v1.y.0`. Then we would have a separate `release_promote.yml` workflow that:

1. verifies the draft exists and tests passed last night
2. creates the `Post 1.y.0 release updates to main` PR
3. creates the backport label
4. creates the `v1.z.0` milestone
5. bumps flake on main to `v1.z.0-pre` (I think that is all?)
6. flips the draft to a normal release

charludo:
Sorry @burgerdev @sespiros, with your suggestions + the promote-to-release idea + some refactoring to remove repetition, this might as well be v2 of this PR.

The good thing about this version though, is we can just start the nightly run manually later, before merging any of this, to see if it works out.

.github/actions/release_build_linux/action.yml
```diff
@@ -0,0 +1,92 @@
...
+    - name: Bump flake version to release version
+      uses: ./.github/actions/bump_version
+      with:
+        version: ${{ inputs.without_v }}
+        commit: ${{ inputs.commit_version_bump }}
+    - name: Create release artifacts
+      id: create-artifacts
+      uses: ./.github/actions/release_artifacts
+      with:
+        version: ${{ inputs.effective_version }}
```
sespiros: Say we are on `v1.15.0`, this step here calls `release_artifacts` with `effective_version` which (through a few steps) comes from: 
```yml
      - name: Determine nightly version
        id: nightly-version
        run: |
          base=$(sed 's/-.*$//' version.txt)
          echo "value=v${base}-$(date -u +%Y-%m-%d)" >> "$GITHUB_OUTPUT"
```
iiuc this will create an `image-replacements.txt` file that has images which use the nightly version i.e `v1.16.0-2026-01-01`. And this then gets embedded in the contrast cli (which has the "without_v" version `v1.16.0` since the previous step is doing `bump_version`). The images are also uploaded with this "effective version".

If we then hit promote, the cli binary gets uploaded under the `v1.16.0` tag but the image-replacements inside it still point to `v1.16.0-2026-01-01`. The images themselves don't get re-uploaded on promote so I think this works at the end but it is just a bit inconsistent.

A way to solve this I guess is to build and override nightly the same next tag i.e `v.1.16.0` in this example. Rebuilding the cli is not an option because then we lose the claim that release e2e has run with the exact published binary.

.github/actions/prepare_github_release/action.yml
sespiros, line 1: I would pick a more explicit name like `prepare_next_github_release` or `setup_next_release_gh_metadata`?

.github/actions/prepare_post_release_pr/action.yml
sespiros, line 1: How about renaming it more explicitly like `prepare_post_release_pr`?

sespiros (changes requested):
Thanks for your effort, I find this easier to follow than the current workflow. Did a first pass (still reading) and left a few comments.

Also wdyt about having a folder under actions to group all the release relevant actions?

.github/actions/release_build_linux/action.yml
charludo: Thanks for catching this. Was indeed wrong. I have changed it in the line you commented on to `inputs.without_v`, and AFAIK, that should be enough. Darwin is unaffected anyways.

Hmm, on second though, is that enough? I'd honestly like to try running this and just see where exactly it fails now -.-

charludo:
Thanks for the review! And sorry, I know it's a lot o yaml 😄 

> Also wdyt about having a folder under actions to group all the release relevant actions?

I think more organization is never a bad idea. Unfortunately, after moving every action that is *only* required by the release workflows into an `.github/actions/release/` dir, most actions are now in that dir, and some of those that aren't are also used by the release (just not exclusively).

I applied your renaming suggestions, and I think with those clearer names the actions dir is more readable already. We *can* move the actions into the mentioned new dir, but I'm not completely sure it would help with organization as much as I'd like.

charludo:
(force-push because of rebase on main after the upload/download-artifacts PR merge)

charludo:
(and again for the flake update)

charludo:
@sespiros could you take a final look at the 3 new commits? They're relatively small. Also already includes the docs additions.

.github/actions/parse_version/action.yml
```diff
@@ -62,4 +65,9 @@ runs:
           echo "next_minor=${PART_MAJOR}.$((PART_MINOR + 1)).0"
           echo "next_minor_pre_without_v=${PART_MAJOR}.$((PART_MINOR + 1)).0-pre"
           echo "next_patch_pre_without_v=${PART_MAJOR}.${PART_MINOR}.$((PART_PATCH + 1))-pre"
+          if (( PART_MINOR > 0 )); then
+            echo "previous_minor_tag=v${PART_MAJOR}.$((PART_MINOR - 1)).0"
```
sespiros, line 69: Will this work if there were patch releases? So let's say we have published `1.20.1`, shouldn't the nightly's `1.21.0-XXXX-YY-ZZ` compare vs that patch release instead of `1.20.0`? I think we should solve this in code, maybe grab list of releases and figure out the latest one.

.github/workflows/release_promote.yml
```diff
@@ -0,0 +1,222 @@
...
+            !./release-artifacts/LICENSE
+            ./darwin-artifacts/contrast-aarch64-darwin
+      - name: Delete nightly draft release
+        env:
+          NIGHTLY_TAG: ${{ needs.prepare.outputs.nightly_tag }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          gh release delete "${NIGHTLY_TAG}" --repo "${GH_REPO}" --yes --cleanup-tag
+
+  create-github-stuff:
```
sespiros: Shouldn't we also rename this to `prepare-github-release`?

.github/workflows/release_promote.yml
```diff
@@ -0,0 +1,222 @@
...
+        with:
+          ref: ${{ needs.prepare.outputs.nightly_ref }}
+          persist-credentials: false
+      - uses: ./.github/actions/prepare_github_release
+        with:
+          release_branch: ${{ needs.prepare.outputs.RELEASE_BRANCH }}
+          next_minor: ${{ needs.prepare.outputs.NEXT_MINOR }}
+          github_token: ${{ github.token }}
+
+  update-main:
```
sespiros: and this to `prepare-post-release-pr`?

.github/workflows/release_promote.yml
```diff
@@ -0,0 +1,222 @@
...
+          next_minor_pre_without_v: ${{ needs.prepare.outputs.NEXT_MINOR_PRE_WITHOUT_V }}
+          pr_branch_suffix: ${{ needs.prepare.outputs.EFFECTIVE_VERSION }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          nunki_ci_token: ${{ secrets.NUNKI_CI_COMMIT_PUSH_PR }}
+          cachix_token: ${{ secrets.CACHIX_AUTH_TOKEN }}
+          artifacts_run_id: ${{ inputs.nightly_run_id }}
+
+  publish:
+    name: Publish release
+    needs: [prepare, promote, create-github-stuff, update-main]
```
sespiros: I think the workflow would be easier to read like:

```suggestion
    needs: [prepare, promote, prepare-github-release, prepare-post-release-pr]
```

`prepare` still feels a bit generic but ok.

.github/workflows/release_promote.yml
```diff
@@ -0,0 +1,222 @@
...
+    permissions:
+      contents: write
+    steps:
+      - name: Publish release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          VERSION: ${{ needs.prepare.outputs.EFFECTIVE_VERSION }}
+        run: |
+          gh release edit "${VERSION}" --draft=false --repo "${{ github.repository }}"
+      - name: Regenerate release notes against the now-existing tag
```
sespiros: Shouldn't we add an equivalent step in `release.yml` as well?

.github/workflows/release_promote.yml
```diff
@@ -0,0 +1,222 @@
...
+          nightly_tag="${FULL_VERSION}-${nightly_date}"
+          echo "nightly_ref=${nightly_ref}" | tee -a "$GITHUB_OUTPUT"
+          echo "nightly_tag=${nightly_tag}" | tee -a "$GITHUB_OUTPUT"
+      - name: Verify nightly tests passed
+        env:
+          NIGHTLY_RUN_ID: ${{ inputs.nightly_run_id }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          failed=$(gh run view "${NIGHTLY_RUN_ID}" --repo "${GH_REPO}" --json jobs \
+            --jq '[.jobs[] | select(.name | startswith("e2e release")) | select(.conclusion != "success")] | map("\(.name): \(.conclusion)") | .[]')
```
sespiros: Are the `e2e release` tests the only ones we care about here? Shouldn't we check that all tests have passed? That would also make the manual check step 1 of the playbook optional. If any test has failed, running the promotion workflow would fail here too.

dev-docs/release.md
```diff
@@ -1,6 +1,44 @@
 # How to release
 
-## Minor
+## Minor, by promoting nightly
+
+The CI should create a nightly minor draft release, called `v1.x.0-yyyy-mm-dd`, and runs the full test suite against it.
+This draft release can be promoted to actual release.
+
+1. Check that in the latest nightly, all checks and tests passed. A draft release is created even if that's not the case, but the promotion won't work if a job failed.
```
sespiros: How about:

```
1. (Optional) Check that in the latest nightly, all checks and tests passed. Promotion won't be able to run if job has failed during the nightly draft release. 
```

To signify that this is not a load-bearing check, but see also my other comment in `release_promote.yml` about having these checks be checked automatically.

dev-docs/release.md
```diff
@@ -1,6 +1,44 @@
...
+
+2. Find and export the workflow run ID of the release you want to promote:
+
+    ```sh
+    gh run list --workflow=release_nightly.yml --limit 10
+    export NIGHTLY_RUN_ID=<run-id>
+    ```
+
+3. Sanity-check the corresponding draft release on GitHub. The tag should match `vX.Y.Z-yyyy-mm-dd` where
```
sespiros: Isn't this step also enforced by code? Why have a separate step for it? Personally I find it confusing but if you think the is some possibility for the code to misbehave then maybe either fix the code or clarify this here or mark it also as `(Optional)`?

dev-docs/release.md
```diff
@@ -1,6 +1,44 @@
...
+  `X.Y.Z` is `version.txt` on main with the `-pre` suffix stripped, and the draft must have artifacts attached.
+
+4. Trigger the promote workflow:
+
+    ```sh
+    gh workflow run release_promote.yml -f nightly_run_id="$NIGHTLY_RUN_ID"
+    ```
+
+5. Review the release notes. If label/title/description changes are necessary, change them on the PR itself. Test the binary artifact.
+   The release notes will be re-generated by the CI once the new version tag has been created.
```
sespiros: So after step 4, the promotion step is blocked at human review (awaiting for someone to click the button). Then if I follow step 5, review the notes and I make edits, then click the button, iiuc the regeneration step that happens after publish will eat my edits no? Do we want this?

A couple of ways to resolve this:

- push the tag -> generates notes -> then publish. Any edits the user will do before clicking submit are the final ones.
- Make it clear that user's shouldn't be doing manual edits to the release text at this point because changes are overwritten. 

What adds to the confusion is "change them on the PR itself" (and I saw we have this in the other section too). It doesn't mean the auto-generated PR, it means the contributor's PR in the list of PRs. So changing an individual contribution's PR would allow the regeneration to pick the changes but changing anything in the release's text at this step would cause it to be overwritten.

dev-docs/release.md
```diff
@@ -1,6 +1,44 @@
...
+This draft release can be promoted to actual release.
+
+1. Check that in the latest nightly, all checks and tests passed. A draft release is created even if that's not the case, but the promotion won't work if a job failed.
+
+2. Find and export the workflow run ID of the release you want to promote:
+
+    ```sh
+    gh run list --workflow=release_nightly.yml --limit 10
+    export NIGHTLY_RUN_ID=<run-id>
+    ```
```
sespiros: Process-wise, should we allow selecting arbitrary nightly release ids here? There might be subtle bugs that can happen since now nightly releases override the same tag (I am looking at https://github.com/edgelesssys/contrast/blob/3eb1a586d1d15e703e0ccd3112166ae4d3076a95/.github/actions/release_build_linux/action.yml#L63), say if someone selects yesterday's nightly for promotion while registries have today's nightly images under the latest tag.

I think in this scenario, the cli would continue working because it picks up images by hash but if someone pulls an image by the release's tag and compare against the one the cli is using will get a hash mismatch because yesterday's nightly cli (which was promoted) had an old hash while the tag points to today's nightly.

So I think this should be solved:

- either by re-tagging the images after promotion if we want to allow promoting older nightlies or actually better,
- make this impossible in code. Only allow the latest nightly to be promoted (and have that be reflected somehow in the playbook here).

sespiros (commented):
Thanks for changes. Left some inline comments. And also some general comments:

- I think the first and the second commit could be squashed since the first one references `steps.version-info.outputs.previous_minor_tag` which doesn't exist yet.
- In general I prefer playbook-style pages to not contain steps that leave the reader wondering whether a step is enforced by the code or not. Steps that can be enforced in code should be enforced in code. If there is a possibility for the code to be wrong, either fix the code or try to make this known as part of the step. As an example, going through the playbook and reading "Check that the latest nighty has passed tests" makes me think "if I skip this I could promote a release with failing tests?", "why isn't the code checking that already?" etc.

.github/actions/parse_version/action.yml
charludo: No, but that's intentional. Our minor releases are always based on the previous minor, not patch. (See the dev-docs release document)

.github/workflows/release_promote.yml
charludo: right. extracted into action

.github/workflows/release_promote.yml
charludo: done

dev-docs/release.md
charludo: Not really. If yesterday's run completely failed, the latest nightly is from the day before and might not be what you wanted to release. If it failed midway, there might be a draft release with (some) artifacts missing.

So I don't really consider this step optional: before releasing, check that what you're about to release exists, and is the correct thing that you want to release.

.github/workflows/release_promote.yml
charludo: Hm... The `e2e release` requires success of both the release and the nightly, so implicitly this is already correct. But I can see the argument for checking everything here.

As for the docs, that's mostly a "don't bother running this if nightly failed", I'll amend the instructions to clarify that.

dev-docs/release.md
charludo: Oh. That's a  good point. I don't think we'd ever want to release an outdated run, and if so, we can always use the manuyal release flow. Changed it so the most recent run is auto-selected.

dev-docs/release.md
charludo: Yes, that was a bit confusing. Notes now get re-generated right after tag creation, and docs mention reviewing them at a later, more fitting point.

charludo:
@sespiros thanks for the thorough review. Addressed the comments in a fixup commit. If things look good to you, I'll squash that into the first commit as well tomorrow.

dev-docs/release.md
```diff
@@ -1,6 +1,37 @@
 # How to release
 
-## Minor
+## Minor, by promoting nightly
+
+The CI should create a nightly minor draft release, called `v1.x.0-yyyy-mm-dd`, and runs the full test suite against it.
+This draft release can be promoted to actual release.
+
+1. Check that in the latest nightly, all checks and tests passed. A draft release is created even if that's not the case, but the promotion won't work if a job failed,
+   but trying to promote a nightly release where linux/darwin releases, nightly e2es or the release e2e failed will automatically fail.
```
sespiros:
```suggestion
1. Check that in the latest nightly, all checks and tests passed. A draft release is created even if that's not the case, but trying to promote a nightly release where linux/darwin releases, nightly e2es or the release e2e failed, will automatically fail anyway.
```

sespiros (approved):
Thanks for the changes, left one minor comment inline.

PR #2361, opened 2026-04-28
justfile: `just e2e-ci` to start and post additional e2e runs on a PR

charludo:
Recently had to quite often start or re-start additional e2e tests on PRs and got increasingly annoyed with having to manually start tests using `gh workflow run`, then write a schema-f comment.

New just recipe automates this. The comment below will have been generated and posted by:
```sh
just e2e-ci imagestore Metal-QEMU-SNP,Metal-QEMU-TDX auto
```

(failures are due to a bug fixed in #2279, not related to this PR)

### DIFF
--- a/justfile
+++ b/justfile
@@ -126,6 +126,61 @@ _e2e target=default_deploy_target platform=default_platform set=default_set: sof
             --sync-ticket-file ./{{ workspace_dir }}/just.sync-ticket \
             --insecure-enable-debug-shell-access=${debug:-false}
 
+# Run additional tests for a PR. Set `post` to `true` to auto-post the comment.
+e2e-ci target platforms="all" post="false" set="testcase-default" debug_shell="false" skip_undeploy="false":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    platforms="{{ platforms }}"
+    case "$platforms" in
+        "all")
+            platforms="Metal-QEMU-SNP,Metal-QEMU-TDX,Metal-QEMU-SNP-GPU,Metal-QEMU-TDX-GPU"
+            ;;
+        "snp")
+            platforms="Metal-QEMU-SNP,Metal-QEMU-SNP-GPU"
+            ;;
+        "tdx")
+            platforms="Metal-QEMU-TDX,Metal-QEMU-TDX-GPU"
+            ;;
+        "gpu")
+            platforms="Metal-QEMU-SNP-GPU,Metal-QEMU-TDX-GPU"
+            ;;
+        *)
+            ;;
+    esac
+
+    pr_number=""
+    branch=$(git branch --show-current)
+    if [[ "{{ post }}" == "true" ]]; then
+        pr_number=$(gh pr view --json number --jq '.number' 2>/dev/null || true)
+        if [[ -z "$pr_number" ]]; then
+            echo "error: could not auto-detect PR for branch '$branch'" >&2
+            exit 1
+        fi
+    fi
+
+    results=()
+    IFS=',' read -ra platform_list <<< "$platforms"
+    for platform in "${platform_list[@]}"; do
+        run_url=$(gh workflow run e2e_manual.yml \
+            --ref "$branch" \
+            -f "test-name={{ target }}" \
+            -f "platform=$platform" \
+            -f "set={{ set }}" \
+            -f "skip-undeploy={{ skip_undeploy }}" \
+            -f "debug-shell={{ debug_shell }}" \
+            2>&1 | grep -oE 'https://[^ ]+/actions/runs/[0-9]+' || echo "")
+        line="- [ ] [$platform, {{ target }}]($run_url)"
+        results+=("$line")
+        echo "$line"
+    done
+
+    if [[ -n "$pr_number" ]]; then
+        echo ""
+        comment=$(printf '%s\n' "${results[@]}")
+        gh pr comment "$pr_number" --body "$comment"
+        echo "Posted comment to PR #$pr_number"
+    fi
+
 e2e-release version platform=default_platform set=default_set: soft-clean
     #!/usr/bin/env bash
     set -euo pipefail


charludo:
- [ ] [Metal-QEMU-SNP, imagestore](https://github.com/edgelesssys/contrast/actions/runs/25039514733)
- [ ] [Metal-QEMU-TDX, imagestore](https://github.com/edgelesssys/contrast/actions/runs/25039515921)

sespiros (changes requested):
Nice! The invocation seems a bit confusing to me though, some suggestions:

- Let's say I am working in a branch already, `just e2e-ci <test-name>` should detect whether there is a PR for it and launch in all platforms for it.
- then if you need a specific platform keep the comma list as an optional second argument
- this is a nit/personal preference but the `auto` argument feels weird. I *think* I've noticed recently that manually triggered e2e tests show in the default github "Checks" list of running tests alongside the other tests so that should be enough no? If the author wants to specifically point out a specific test for review in their PR message or comment, maybe it's fine for them to do this manually?

charludo:
>     * Let's say I am working in a branch already, `just e2e-ci <test-name>` should detect whether there is a PR for it and launch in all platforms for it.

Quite often you want to test on specific platforms though, e.g. because a feature or bug is only on that CPU type / with or without GPU. I mostly tried to force us to be explicit to not swamp the GPU hosts.

About auto-detecting the PR... probably personal preference, but I didn't want the tool to comment in my name without explicit approval. I also think there's quite often the case where you re-run the command (e.g. because tests failed and you made changes), and then probably you should replace the comment, not add another new one.

>     * then if you need a specific platform keep the comma list as an optional second argument

I could see that though.

>     * this is a nit/personal preference but the `auto` argument feels weird. I _think_ I've noticed recently that manually triggered e2e tests show in the default github "Checks" list of running tests alongside the other tests so that should be enough no? If the author wants to specifically point out a specific test for review in their PR message or comment, maybe it's fine for them to do this manually?

Ah, didn't know that. Yes, this is basically intended for making sure others are aware that certain tests passed. We can remove the `pr=` arg from the just recipe, but then again, does it hurt to have it? ^^

sespiros:
> Quite often you want to test on specific platforms though, e.g. because a feature or bug is only on that CPU type / with or without GPU. I mostly tried to force us to be explicit to not swamp the GPU hosts.

My comment was mostly about the default behavior. If you think the most sane default is running on a specific builder then I guess the current approach is fine. What about adding an "all" helper? So when you run the command you have a) a way to run it for all easily while at the same time b) are forced to think about the implications :)

> Ah, didn't know that. Yes, this is basically intended for making sure others are aware that certain tests passed. We can remove the `pr=` arg from the just recipe, but then again, does it hurt to have it? ^^

If it makes your life easier, I'm fine with it. My confusion was mostly around seeing "auto" there as a flag, not knowing that "auto" really translate to "auto-discover the PR number and also leave a comment on it" so it's like a single flag for doing 2 things. What about renaming it to "post"? Is there a use-case for wanting to set an explicit PR number and not work on the auto-detected one? 

Making the comment text more explicit like "Manual runs triggered: ..." would be another suggestion. About detecting and changing vs creating a new one, I don't have a preference. By default github Checks would simply list the last one.

justfile
```diff
@@ -126,6 +126,43 @@ _e2e target=default_deploy_target platform=default_platform set=default_set: sof
...
+    IFS=',' read -ra platform_list <<< "{{ platforms }}"
+    for platform in "${platform_list[@]}"; do
+        run_url=$(gh workflow run e2e_manual.yml \
+            --ref "$branch" \
+            -f "test-name={{ test }}" \
+            -f "platform=$platform" \
+            -f "set={{ set }}" \
+            -f "skip-undeploy={{ skip_undeploy }}" \
+            -f "debug-shell={{ debug_shell }}" \
+            2>&1 | grep -oE 'https://[^ ]+/actions/runs/[0-9]+' || echo "(url unavailable)")
```
burgerdev: The fallback likely breaks the markdown in the next line.

justfile
```diff
@@ -126,6 +126,43 @@ _e2e target=default_deploy_target platform=default_platform set=default_set: sof
             --sync-ticket-file ./{{ workspace_dir }}/just.sync-ticket \
             --insecure-enable-debug-shell-access=${debug:-false}
 
+# Run additional tests for a PR. Set `pr` to `auto` or a PR number to auto-post the comment.
+e2e-ci test platforms pr="" set="testcase-default" skip_undeploy="false" debug_shell="false":
```
burgerdev: `pr` is a mandatory input, isn't it?

burgerdev (commented):
Neat idea!

justfile
charludo: Not really, if absent, you just get the markdown output and no post is made. But I think I'll go with @sespiros suggestion and make this a true/false toggle that always autodetects the pr number.

charludo:
- [ ] [Metal-QEMU-SNP, imagestore](https://github.com/edgelesssys/contrast/actions/runs/25098619406)
- [ ] [Metal-QEMU-SNP-GPU, imagestore](https://github.com/edgelesssys/contrast/actions/runs/25098620393)

Edit: ignore that.

charludo:
@sespiros added multiple shorthands for common sets of runners, and pr => post, must be true to auto-detect PR number and then creates comment.

justfile
```diff
@@ -126,6 +126,61 @@ _e2e target=default_deploy_target platform=default_platform set=default_set: sof
             --sync-ticket-file ./{{ workspace_dir }}/just.sync-ticket \
             --insecure-enable-debug-shell-access=${debug:-false}
 
+# Run additional tests for a PR. Set `pr` to `auto` or a PR number to auto-post the comment.
```
sespiros: Thanks! The comment needs updating. I guess if this fails often for some reason (like pr number not being auto-detected) we can revisit.

sespiros (approved):
Approving with a minor comment fix suggestion.

justfile
charludo: Oops, thanks for catching that. Done.

PR #2362, opened 2026-04-28
dev-shell: configure pre-push hooks

charludo:
Note: you need to exit and re-enter the nix shell for the hooks to get installed and become active.


burgerdev (approved):
Thanks!

PR #2363, opened 2026-04-29
release: push k8s-log-collector

sespiros:
As it can be seen in https://github.com/edgelesssys/contrast/actions/runs/25005789022/job/73530829709, release workflow jobs fail to find the `collect-host-logs` binary because they are using an older `k8s-log-collector` image. That is because the `k8s-log-collector` image used to be hard-coded with a hash before the log host collection was added.

Also surface an error if the host log collection steps fail.

Fix this by making the release workflow also push a fresh k8s-log-collector image.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -279,6 +279,8 @@ jobs:
       contents: write
       packages: write
       id-token: write
+    outputs:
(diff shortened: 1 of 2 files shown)


burgerdev (approved):
Thanks!

PR #2366, opened 2026-05-04
tools: use k8s lease for GPU sync FIFO

burgerdev:
Adds a library and a binary for mutual exclusion using only Kubernetes constructs.

The tools/fifo/internal/lease package describes the ideas behind this.

---

Switching the actual FIFO callers to the new binary requires coordination and is therefore not included in this PR.


tools/fifo/internal/lease/lease.go
```diff
@@ -0,0 +1,326 @@
...
+		if item.Spec.AcquireTime.After(earliest) {
+			l.log.Debug("skipping older entry", "name", item.Name)
+			continue
+		}
+		earliest = item.Spec.AcquireTime.Time
+		holder = *item.Spec.HolderIdentity
+	}
+	if holder == "" {
+		return false, fmt.Errorf("no valid queue entries found, can't continue")
+	}
```
charludo: Why do we want to error out when the holder is unknown? (Or rather, does it maybe make sense to retry this?)

tools/fifo/release.go
```diff
@@ -0,0 +1,54 @@
...
+	cmd := &cobra.Command{
+		Use:   "release <holder-id>",
+		Short: "Release the named lock",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runRelease(cmd, namespace, lockName, args[0], verbose)
+		},
+	}
+	cmd.Flags().StringVar(&namespace, "namespace", defaultNamespace, "Kubernetes namespace for the Lease objects")
+	cmd.Flags().StringVar(&lockName, "lock", defaultLeaseName, "Name of the lock Lease")
```
charludo: Seems to be called `--lease` everywhere else

charludo (approved):
I really, really hate to do this... but... LGTM 😅

tools/fifo/internal/lease/lease.go
burgerdev: Ah, that's a safety mechanism I should have documented.

The lease mechanism is designed so that you can simply clean up all leases in the namespace and nothing bad will happen (if the holding process is gone, at least). If we can't find a lease with any name, that implies that our own lease expired. That should not happen without external interventiuon,  because we're renewing it in the background. So, it's most likely that either the cluster is broken or the queue was deleted, and thus we just stop.

Now that I'm thinking of it, we could try to re-queue, but I'm not sure whether it's desirable.

tools/fifo/release.go
burgerdev: Good catch, I started with lock but then renamed to lease at some point, let me fix that.

tools/fifo/internal/lease/lease.go
burgerdev: I found a better way to cover this condition and a single queue entry disappearing: we're now looking for our own queue entry: if that doesn't exist, we need to error out, if it does, it means that we are either in front or we still need to wait (no need to check for the current holder anymore).

charludo (approved):
The changed behavior with looking for our own ticket SGTM.

PR #2368, opened 2026-05-04
badaml-payload: start coarse scan at 1GB to make badaml-vuln pass reliably on SEV-SNP

sespiros:
This fixes the flakiness we have observed in nightly ci tests on the SEV-SNP runners.

Supersedes https://github.com/edgelesssys/contrast/pull/2321

### DIFF
--- a/tools/badaml-payload/deadbeef-file.asl
+++ b/tools/badaml-payload/deadbeef-file.asl
@@ -12,9 +12,9 @@
     To mount the attack, we use a shell wrapper around QEMU that adds the malicious table via -acpitable flag.
 
     The AML code does the following:
-    - Scans ~4GB of guest physical memory from 0x100000 (1MB, skipping legacy area) in
-      8MB steps (510 iterations), looking for the 0xDEADBEEF pattern. Since the target
-      file is 16MB, an 8MB step guarantees a hit.
+    - Scans guest physical memory from 0x40000000 (1GB) in 8MB steps,
+      looking for the 0xDEADBEEF pattern. Since the target file is 16MB,
+      an 8MB step guarantees a hit.
     - Once found, scans backward to find the start of the pattern block.
     - Overwrites the first 4 bytes with 0xCAFEBABE.
 */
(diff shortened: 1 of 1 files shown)


tools/badaml-payload/deadbeef-file.asl
```diff
@@ -12,9 +12,9 @@
     To mount the attack, we use a shell wrapper around QEMU that adds the malicious table via -acpitable flag.
 
     The AML code does the following:
-    - Scans ~4GB of guest physical memory from 0x100000 (1MB, skipping legacy area) in
-      8MB steps (510 iterations), looking for the 0xDEADBEEF pattern. Since the target
-      file is 16MB, an 8MB step guarantees a hit.
+    - Scans guest physical memory from 0x40000000 (1GB) in 8MB steps,
+      looking for the 0xDEADBEEF pattern. Since the target fils is 16MB,
+      an 8BM step guarantees a hit.
```
burgerdev: typo: 8MB

tools/badaml-payload/deadbeef-file.asl
```diff
@@ -107,16 +107,27 @@ DefinitionBlock ("", "SSDT", 6, "BADAML", "BADAML", 0x20240306)
             {
                 Debug = "BADAML: _INI started"
 
-                // Scan base: 0x100000 (1MB) to skip legacy BIOS area.
-                Local0 = 0x100000
+                // Scan base: 0x40000000 (1GB). The initrd is loaded by Linux
+                // toward the upper end of guest RAM (empirically observed
```
burgerdev: The initrd is loaded into memory by qemu, isn't it? Could we try to find out how that placement is decided?

sespiros: Yes it is but when we originally wrote this we concluded that it wasn't possible to compute the initrd's exact end memory location, hence the scan. An alternative we considered at the time was to boot the vm once, get that address then reboot and pass that information in but that was deemed impractical.

burgerdev: What I meant to say:

1. There's a mistake in this docstring ("loaded by Linux").
2. There is probably logic in qemu for placing the initrd, which we could mention here.

sespiros: "loaded into memory" is a bit generic, and the actual process is quite more complex:

- qemu passes the initrd file to the guest through fw_cfg (that's host userspace QEMU memory)
- OVMF installs a `LoadFile2` protocol for the initrd then loads the EFI kernel image. 
- Linux's EFI stub will then call to that protocol, map memory in the guest for the initrd and then
- asks the firmware to load it there iirc 

so the actual loading in guest memory is done by OVMF but the where is dictated by linux (well its EFI stub) so "loaded by OVMF" would be the most accurate term here.

Edit: Interestingly enough I compared this with stage0's mechanism which decides itself where to load the initrd in guest memory so I might have to revisit this in the future and make it even more generic.

burgerdev: > qemu passes the initrd file to the guest through fw_cfg (that's host userspace QEMU memory)

That does not seem to be true, or why do we get a measurement for the initrd?

https://github.com/edgelesssys/contrast/blob/d0573da1490f853bcb2436ab6a898cda3ac7cc7d/internal/snp/sev_hashes.go#L69

burgerdev: Nevermind, I was mistaken here. sgtm

burgerdev (approved):

PR #2369, opened 2026-05-04
go: migrate crypto/elliptic -> filippo.io/nistec

charludo:
No idea why `generate` keeps failing, checks out locally.

Disclaimer: migration to `nistec` package was done with help from an LLM.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ replace github.com/google/go-tdx-guest => github.com/edgelesssys/go-tdx-guest v0
 
 require (
 	filippo.io/keygen v0.0.0-20260114151900-8e2790ea4c5b
+	filippo.io/nistec v0.0.4
 	github.com/containerd/ttrpc v1.2.8
 	github.com/coreos/go-iptables v0.8.0
 	github.com/coreos/go-systemd/v22 v22.7.0
--- a/internal/idblock/idblock.go
+++ b/internal/idblock/idblock.go
@@ -4,6 +4,7 @@
 package idblock
 
 import (
+	"bytes"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/sha512"
@@ -12,8 +13,11 @@ import (
 	"math/big"
 	"slices"
 
+	"filippo.io/nistec"
 	"github.com/google/go-sev-guest/abi"
 )
+
+const p384CoordSize = 48
 
 // IDBlock is the ID block.
 // https://github.com/microsoft/igvm-tooling/blob/main/src/igvm/structure/igvmfileformat.py#L453
@@ -264,8 +268,14 @@ func recoverPublicKey(curve elliptic.Curve, r, s, z *big.Int) ([]ecdsa.PublicKey
 			pointY = new(big.Int).Sub(params.P, pointY)
 		}
 
-		// Verify if the point is on the curve
-		if !curve.IsOnCurve(pointX, pointY) { //nolint:staticcheck // SA1019: low-level arithmetic is needed for public key recovery
+		// Construct the candidate as an uncompressed, encoded point (0x04 || X || Y) as per https://www.secg.org/sec1-v2.pdf section 2.3.3,
+		// and rely on the documented behaviour of SetBytes to reject the point if it is not on the curve.
+		qBytes := make([]byte, 1+2*p384CoordSize)
+		qBytes[0] = 0x04
+		pointX.FillBytes(qBytes[1 : 1+p384CoordSize])
+		pointY.FillBytes(qBytes[1+p384CoordSize:])
+		q, err := nistec.NewP384Point().SetBytes(qBytes)
+		if err != nil {
 			continue
 		}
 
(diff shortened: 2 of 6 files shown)


internal/idblock/idblock.go
thomasten, line 1: What error do you actually get before the change? On my machine, the current idblock package compiles just fine with go 1.26.2.

charludo: As part of the last flake update, I had to add a bunch of `//nolint:staticcheck // SA1019` linter exceptions. They are sufficient in go >1.26 as well. I suppose we can keep them for now, if we're ok with those exceptions / anticipate that go 2.0 is far off 😄

internal/idblock/idblock.go
```diff
@@ -4,17 +4,35 @@
...
 	"fmt"
 	"math/big"
 	"slices"
 
+	"filippo.io/nistec"
 	"github.com/google/go-sev-guest/abi"
 )
 
+// p384 holds the curve parameters for NIST P-384. P, N, B are not exposed in nistec, but are constant:
+// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf#%5B%7B%22num%22%3A232%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C70%2C264%2C0%5D
```
burgerdev:
```suggestion
// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf, section 3.2.1.4.
```

internal/idblock/idblock.go
```diff
@@ -4,17 +4,35 @@
...
+var p384 = struct {
+	P, N, B *big.Int
+}{
+	P: hexBig("fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffeffffffff0000000000000000ffffffff"),
+	N: hexBig("ffffffffffffffffffffffffffffffffffffffffffffffffc7634d81f4372ddf581a0db248b0a77aecec196accc52973"),
+	B: hexBig("b3312fa7e23ee7e4988e056be3f82d19181d9c6efe8141120314088f5013875ac656398d8a2ed19d2a85c8edd3ec2aef"),
+}
```
burgerdev: These should still be exposed and not deprecated under

```go
elliptic.P384().Params()
```

internal/idblock/idblock.go
```diff
@@ -217,91 +235,85 @@ func (pubKey *Ecdsa384PublicKey) UnmarshalBinary(data []byte) error {
...
-
-		threeX := new(big.Int).Lsh(x, 1)
-		threeX.Add(threeX, x)
-
-		x3.Sub(x3, threeX)
-		x3.Add(x3, curve.B)
-		x3.Mod(x3, curve.P)
-
-		return x3
-	}
```
burgerdev, line 239: Imho this was clearer than the inlined version.

internal/idblock/idblock.go
```diff
@@ -345,19 +357,20 @@ func IDBlocksFromLaunchDigest(launchDigest [48]byte, guestPolicy abi.SnpPolicy)
...
-	slices.Reverse(xLittleEndian)
-
-	yLittleEndian := pubKey.Y.Bytes()
-	slices.Reverse(yLittleEndian)
+	xBigEndian := slices.Clone(pubKey[1 : 1+p384CoordSize])
+	yBigEndian := slices.Clone(pubKey[1+p384CoordSize:])
+	slices.Reverse(xBigEndian)
+	slices.Reverse(yBigEndian)
+	xLittleEndian := xBigEndian
+	yLittleEndian := yBigEndian
```
burgerdev: I actually find both of these variants confusing. Maybe a function could help:

```go
func reverse(x []byte) []byte {
  x = slices.Clone(x)
  slices.Reverse(x)
  return x
}

xBigEndian := reverse(pubKey[1 : 1+p384CoordSize])
```

(note the remark about diff, though - this should go in a separate commit)

internal/idblock/idblock.go
```diff
@@ -217,91 +235,85 @@ func (pubKey *Ecdsa384PublicKey) UnmarshalBinary(data []byte) error {
...
+	u1 := new(big.Int).Mul(z, rInv)
+	u1.Mod(u1, p384.N)
+	u1.Neg(u1)
+	u1.Mod(u1, p384.N)
+	u2 := new(big.Int).Mul(s, rInv)
+	u2.Mod(u2, p384.N)
+	u1Bytes := make([]byte, p384CoordSize)
+	u1.FillBytes(u1Bytes)
+	u2Bytes := make([]byte, p384CoordSize)
+	u2.FillBytes(u2Bytes)
```
burgerdev: This is an optimization (pulling it out of the loop), right? I'd prefer if we sticked to the current implementation as close as we can for the migration to `nistec` (i.e., keep the diff minimal), and do the rest in a separate commit.

internal/idblock/idblock.go
```diff
@@ -265,7 +268,12 @@ func recoverPublicKey(curve elliptic.Curve, r, s, z *big.Int) ([]ecdsa.PublicKey
 		}
 
 		// Verify if the point is on the curve
-		if !curve.IsOnCurve(pointX, pointY) { //nolint:staticcheck // SA1019: low-level arithmetic is needed for public key recovery
+		qBytes := make([]byte, 1+2*p384CoordSize)
+		qBytes[0] = 0x04
+		pointX.FillBytes(qBytes[1 : 1+p384CoordSize])
+		pointY.FillBytes(qBytes[1+p384CoordSize:])
+		q, err := nistec.NewP384Point().SetBytes(qBytes)
```
burgerdev, line 299: We should mention that the strategy is to construct an uncompressed, encoded point as per https://www.secg.org/sec1-v2.pdf 2.3.3, and then rely on the specified behaviour of `SetBytes` to reject the point if it's not on the curve.

internal/idblock/idblock.go
```diff
@@ -279,16 +287,24 @@ func recoverPublicKey(curve elliptic.Curve, r, s, z *big.Int) ([]ecdsa.PublicKey
...
+		if err != nil {
+			return nil, fmt.Errorf("scalar base mult: %w", err)
+		}
+		p2, err := nistec.NewP384Point().ScalarMult(q, u2Bytes)
+		if err != nil {
+			return nil, fmt.Errorf("scalar mult: %w", err)
+		}
+		sumBytes := nistec.NewP384Point().Add(p1, p2).Bytes()
 
 		publicKeys = append(publicKeys, ecdsa.PublicKey{
```
burgerdev: `ecdsa.PublicKey.X` (and Y) are deprecated, too. Maybe we should use [`ecdsa#ParseUncompressedPublicKey`](https://pkg.go.dev/crypto/ecdsa#ParseUncompressedPublicKey).

go.mod
```diff
@@ -46,6 +46,7 @@ require (
 
 require (
 	filippo.io/bigmod v0.1.1-0.20260103110540-f8a47775ebe5 // indirect
+	filippo.io/nistec v0.0.4 // indirect
```
burgerdev: Why is this indirect?

internal/idblock/idblock.go
```diff
@@ -261,6 +267,20 @@ func recoverPublicKey(curve elliptic.Curve, r, s, z *big.Int) ([]ecdsa.PublicKey
 		return nil, fmt.Errorf("no square root")
 	}
 
+	u1 := new(big.Int).Mul(z, rInv)
+	u1.Mod(u1, order)
+	u1.Neg(u1)
+	u1.Mod(u1, order)
+
+	// Compute u1 and u2
```
burgerdev, line 276: That comment is misplaced, but was so before. However, now that I'm trying to understand what u1 and u2 actually are, I notice that the nomenclature in this function seems to be far off from SEC-1. We should address that eventually, but if you don't already know how just leave a TODO.

go.mod
charludo: No idea how that happened. `go mod tidy` turned it direct.

internal/idblock/idblock.go
charludo: done

internal/idblock/idblock.go
charludo: Indeed, other accesses of X and Y as well. Replaced in fixup commit in the middle.

burgerdev (approved):

charludo:
- [x] [Metal-QEMU-SNP, regression](https://github.com/edgelesssys/contrast/actions/runs/27538025009)
- [x] [Metal-QEMU-TDX, regression](https://github.com/edgelesssys/contrast/actions/runs/27538026657)
- [x] [Metal-QEMU-SNP-GPU, regression](https://github.com/edgelesssys/contrast/actions/runs/27538028260)
- [x] [Metal-QEMU-TDX-GPU, regression](https://github.com/edgelesssys/contrast/actions/runs/27538030033)

PR #2371, opened 2026-05-05
scripts: pass tags to modernize

burgerdev:


charludo (approved):
I assume the cli/coordinator changes are the result of the flags being passed? Reason I'm asking is that these same changes are curiously also present in #2369

burgerdev:
> I assume the cli/coordinator changes are the result of the flags being passed? Reason I'm asking is that these same changes are curiously also present in #2369

That's where I found the error messages to begin with. I assumed that the `gofix` step was simply being skipped for all packages that consume `sdk`.

burgerdev:
@charludo fixed the other issue in the `gofix` flow, too.

PR #2372, opened 2026-05-05
[3/8] runtime/nodeinstaller: add insecure platform variants

msanft:
Reopened after reordering the stack: this PR now comes before initdata-processor.

### DIFF
--- a/internal/platforms/platforms.go
+++ b/internal/platforms/platforms.go
@@ -24,11 +24,18 @@ const (
 	MetalQEMUSNPGPU
 	// MetalQEMUTDXGPU is the generic platform for bare-metal TDX deployments with GPU passthrough.
 	MetalQEMUTDXGPU
+	// MetalQEMUInsecure is the platform for bare-metal deployments with a non-CC runtime class.
+	MetalQEMUInsecure
+	// MetalQEMUInsecureGPU is the platform for bare-metal deployments with GPU passthrough and a non-CC runtime class.
+	MetalQEMUInsecureGPU
 )
 
 // All returns a list of all available platforms.
 func All() []Platform {
-	return []Platform{MetalQEMUSNP, MetalQEMUTDX, MetalQEMUSNPGPU, MetalQEMUTDXGPU}
+	return []Platform{
+		MetalQEMUSNP, MetalQEMUTDX, MetalQEMUSNPGPU, MetalQEMUTDXGPU,
+		MetalQEMUInsecure, MetalQEMUInsecureGPU,
+	}
 }
 
 // AllStrings returns a list of all available platforms as strings.
@@ -51,8 +58,25 @@ func (p Platform) String() string {
 		return "Metal-QEMU-TDX"
 	case MetalQEMUTDXGPU:
 		return "Metal-QEMU-TDX-GPU"
+	case MetalQEMUInsecure:
+		return "Metal-QEMU-Insecure"
+	case MetalQEMUInsecureGPU:
+		return "Metal-QEMU-Insecure-GPU"
 	default:
 		return "Unknown"
+	}
+}
+
+// InsecureVariant returns the insecure (non-CC) variant of the
+// platform, or Unknown if there is no such variant.
+func (p Platform) InsecureVariant() Platform {
+	switch p {
+	case MetalQEMUSNP, MetalQEMUTDX:
+		return MetalQEMUInsecure
+	case MetalQEMUSNPGPU, MetalQEMUTDXGPU:
+		return MetalQEMUInsecureGPU
+	default:
+		return Unknown
 	}
 }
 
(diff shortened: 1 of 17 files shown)


internal/platforms/platforms.go
```diff
@@ -24,11 +24,22 @@ const (
...
 	// MetalQEMUTDXGPU is the generic platform for bare-metal TDX deployments with GPU passthrough.
 	MetalQEMUTDXGPU
+	// MetalQEMUSNPInsecure is the platform for bare-metal SNP deployments with a non-CC runtime class.
+	MetalQEMUSNPInsecure
+	// MetalQEMUTDXInsecure is the platform for bare-metal TDX deployments with a non-CC runtime class.
+	MetalQEMUTDXInsecure
+	// MetalQEMUSNPGPUInsecure is the platform for bare-metal SNP deployments with GPU passthrough and a non-CC runtime class.
+	MetalQEMUSNPGPUInsecure
+	// MetalQEMUTDXGPUInsecure is the platform for bare-metal TDX deployments with GPU passthrough and a non-CC runtime class.
+	MetalQEMUTDXGPUInsecure
```
burgerdev: Shouldn't `insecure` be an alternative to `TDX` and `SNP` here?

justfile
```diff
@@ -224,7 +169,7 @@ runtime target=default_deploy_target platform=default_platform set=default_set:
         --namespace {{ target }}${namespace_suffix-} \
         --node-installer-target-conf-type ${node_installer_target_conf_type} \
         --platform "$platforms" \
-        runtime >> "./{{ workspace_dir }}/runtime/runtime.yml"
+        runtime > "./{{ workspace_dir }}/runtime/runtime.yml"
```
burgerdev, line 226: That's a bug fix, right? Would be good to separate that out.

internal/platforms/platforms.go
msanft: No, since we'll need to use different kernels depending on what's desired.

justfile
msanft: I wasn't sure about this being a "bug" or just a papercut in development. Can still split this out.

internal/platforms/platforms.go
burgerdev: The kernel does not depend on TDX vs. SNP, does it?

msanft: Sorry. Bad wording: For GPU vs. no-GPU, the kernel (i.e. the modules) differ. For SNP vs. TDX it's just the command line.

burgerdev: I also don't see a command line dependence on the TEE flavour.

msanft: Confused this with the host command line, sorry. For TDX, it needs `nohibernate kvm_intel.tdx=1`, which it doesn't for SNP.

I still think there's more value in considering the insecure runtimes as another matrix dimension to keep things simple over trying to compact here, and then rather implementing [RFC 8](https://github.com/edgelesssys/contrast/blob/main/rfc/008-platform-support.md) at some point. WDYT?

burgerdev: The runtime `metal-QEMU-XXX(-GPU)?` names currently map 1:1 to the XXX attestation method. I'd really like to avoid adding dimensions to the matrix until we have RFC 8, if only to not run large batteries of tests that are identical except for the name.

msanft: Ok. Just to confirm before making such a large change over the whole stack: You'd want `Metal-QEMU-Insecure` and `Metal-QEMU-Insecure-GPU`?

burgerdev: That's what I had in mind, yes.

internal/atls/issuer/issuer_linux.go
```diff
@@ -29,6 +29,7 @@ func New(log *slog.Logger) (atls.Issuer, error) {
 			logger.NewWithAttrs(logger.NewNamed(log, "issuer"), map[string]string{"tee-type": "tdx"}),
 		), nil
 	default:
-		return nil, fmt.Errorf("unsupported platform: %T", cpuid.CPU)
+		log.Warn("No TEE platform detected, using insecure attestation issuer")
+		return insecure.NewIssuer(), nil
```
burgerdev: Is this intentionally part of this PR? I don't see where it's used by the nodeinstaller.

burgerdev (commented):
lgtm % issuer question

internal/atls/issuer/issuer_linux.go
msanft: It's not. Needs to go into the coordinator PR. Solved.

burgerdev (approved):
Thanks!

burgerdev:
The generated code check failure is due to the Kata update, I assume.

PR #2374, opened 2026-05-06
kata-agent: configure debug console and log level via initdata

davidweisse:
Instead of rebuilding the images with the debug set enabled, we can configure the agent log level and debug console via initdata when running `contrast generate` with `--insecure-enable-debug-shell-access`. We can write an agent configuration file directly into the `Data` field and pass the resulting path `/run/measured-cfg/agent.toml` to the agent via a command line argument. The default behavior of the agent is to use only the provided config if such a config exists and no options passed via kernel cmdline. The patch is there to use the provided config as an override after the kernel cmdline has been parsed. (Consider a scenario were the debug set is enabled but we want to manipulate the agent config via initdata. The kernel parameters added from the debug set would be ignored.)

Not sure if ignoring the error when reading `/run/measured-cfg/agent.toml` is the best. If the initdata is not manipulated, the initdata processor won't create a `agent.toml` file, so the agent would crash. In that case we could also write an empty config file in, e.g., the initdata processor.

For now, I added 3 lines of TOML config manually, but if we want to do more with this, we could also bring the `AgentConfig` struct from the agent into go code for better configurability.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -220,6 +220,7 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 		fmt.Fprintln(cmd.OutOrStdout(), "⚠️ Insecure debug shell access enabled!")
 		initdataManipulators = append(initdataManipulators, func(id *initdata.Initdata) error {
 			id.Data["contrast.insecure-debug"] = "true"
+			id.Data["agent.toml"] = "log_level = \"debug\"\ndebug_console = true\ndebug_console_vport = 1026"
 			return nil
 		})
 	}
--- a/packages/by-name/kata/runtime/0025-agent-use-config-file-as-override.patch
+++ b/packages/by-name/kata/runtime/0025-agent-use-config-file-as-override.patch
@@ -1,0 +1,280 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: davidweisse <98460960+davidweisse@users.noreply.github.com>
+Date: Wed, 13 May 2026 16:19:54 +0200
+Subject: [PATCH] agent: use config file as override
+
+---
+ src/agent/src/config.rs | 239 +++++++++++++++++++++-------------------
+ 1 file changed, 125 insertions(+), 114 deletions(-)
+
+diff --git a/src/agent/src/config.rs b/src/agent/src/config.rs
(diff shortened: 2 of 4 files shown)


cli/cmd/generate.go
```diff
@@ -204,6 +204,7 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 		fmt.Fprintln(cmd.OutOrStdout(), "⚠️ Insecure debug shell access enabled!")
 		initdataManipulators = append(initdataManipulators, func(id *initdata.Initdata) error {
 			id.Data["contrast.insecure-debug"] = "true"
+			id.Data["agent.toml"] = "log_level = \"debug\"\ndebug_console = true\ndebug_console_vport = 1026"
```
burgerdev, line 223: Regarding the absent agent config file question: I think it might make sense to write an empty config by default, as there's a chance we'll want to use this in the future for non-debug things.

davidweisse: Added this, and changed the patch to throw an error if the config doesn't exist (behavior as before).

burgerdev: Ok, I just realized while debugging an unrelated issue that we should not make the agent fail on an absent config file: if a user tries to launch a Contrast pod without initdata, or with some handcrafted initdata that does not contain a config file, they will not get a proper error message but rather just see their VM collapse. So, at the end of the day, we'll still need to either support an absent config file, or always write one in the initdata processor. In the interest of keeping separate things separate, I'd pick the former. Sorry for the distraction!

packages/by-name/kata/runtime/0025-agent-use-config-file-as-override.patch
```diff
@@ -0,0 +1,277 @@
...
+ 
++        // If config file specified in the args, use it as an override
++        let config_position = args.iter().position(|a| a == "--config" || a == "-c");
++        if let Some(config_position) = config_position {
++            if let Some(config_file) = args.get(config_position + 1) {
++                if let Ok(s) = fs::read_to_string(config_file) {
++                    let agent_config_builder: AgentConfigBuilder =
++                        toml::from_str(&s).map_err(anyhow::Error::new)?;
++                    config.override_config_from_builder(agent_config_builder)?;
++                }
```
burgerdev: Please log an error in the `else` case.

burgerdev (approved):
lgtm, just a nit

PR #2375, opened 2026-05-06
e2e: collect logs synchronously to fix namespace teardown race

sespiros:
This was prompted by the failure in https://github.com/edgelesssys/contrast/actions/runs/25005789022/job/73530829709 which was mitigated by https://github.com/edgelesssys/contrast/pull/2363 but the root cause is actually a race condition that this PR fixes.

The release test (`TestRelease` in `e2e/release/release_test.go`) deletes the `contrast-system` namespace in a `t.Cleanup` callback at test exit. Previously, the workflow's separate `Download logs` step ran *after* the test process exited, racing the cluster's namespace teardown.

When the deployed log-collector image was newer than the SHA pinned in `packages/log-collector.yaml`, the `kubectl exec collect-host-logs` retry loop ate ~25 seconds of the 30-second teardown grace period. The subsequent `kubectl exec ... tar` step then hit "task not found" because the log-collector pod and its container were already gone, and the workflow failed despite the test itself passing.

# Fix

Move log download inside the test process and remove the extra "Download logs" action step.

I've also fixed 2 adjacent issues and another drive-by fix:

- The `default` namespace's `creationTimestamp` in the testing cluster is months old, which made `journalctl --since` pull a year of host logs. The new `LOG_COLLECT_SINCE_FLOOR` env var allows configuring that.
- also the `e2e-release` target had the same issue like the original mitigation PR, it needs to push a fresh `k8s-log-collector` image. I plan to push a follow-up PR that will refactor k8s-log-collector a bit so this is not needed.
- clarified the comment for the `GH_TOKEN` use

# Test

- [x] Tested with `just e2e-release v1.20.0` in the dev cluster.
- [ ] CI release workflow

### DIFF
--- a/e2e/internal/logcollect/logcollect.go
+++ b/e2e/internal/logcollect/logcollect.go
@@ -1,0 +1,33 @@
+// Copyright 2026 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+// Package logcollect provides a thin Go wrapper around the get-logs script
+// for e2e tests that want to collect cluster + host logs synchronously,
+// before any test cleanup tears down the namespaces being collected from.
+package logcollect
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"time"
+)
+
+// Download invokes `get-logs download` against namespaceFile, blocking until
+// log collection completes or ctx is canceled.
+//
+// If sinceFloor is non-zero, it is passed to get-logs as a lower bound for
+// the host-log --since timestamp via the LOG_COLLECT_SINCE_FLOOR env var.
+func Download(ctx context.Context, namespaceFile string, sinceFloor time.Time) error {
+	cmd := exec.CommandContext(ctx, "get-logs", "download", namespaceFile)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if !sinceFloor.IsZero() {
+		cmd.Env = append(os.Environ(), "LOG_COLLECT_SINCE_FLOOR="+sinceFloor.UTC().Format(time.RFC3339))
+	}
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("running get-logs download %s: %w", namespaceFile, err)
+	}
+	return nil
+}
--- a/justfile
+++ b/justfile
@@ -194,7 +194,6 @@ e2e-release version platform=default_platform set=default_set: soft-clean
             --node-installer-target-conf ${node_installer_target_conf_type} \
             --namespace-file ./{{ workspace_dir }}/just.namespace \
             --use-loadbalancer=true
-    nix run .#{{ set }}.scripts.get-logs download ./{{ workspace_dir }}/just.namespace
 
 # Generate policies, apply Kubernetes manifests.
 deploy target=default_deploy_target cli=default_cli platform=default_platform: (populate target platform) (runtime target platform) (write-namespace target) (apply "runtime" platform) (generate cli platform) (apply target platform)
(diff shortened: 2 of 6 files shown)


e2e/internal/logcollect/logcollect.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2026 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+// Package logcollect provides a thin Go wrapper around the get-logs script
+// for e2e tests that want to collect cluster + host logs synchronously,
+// before any test cleanup tears down the namespaces being collected from.
+package logcollect
```
burgerdev, line 7: Is this going to stay, or are we moving the backing implementation from shell script to `internal/` eventually?

justfile
```diff
@@ -188,13 +188,14 @@ e2e-release version platform=default_platform set=default_set: soft-clean
     mkdir -p ./{{ workspace_dir }}
     nix run .#{{ set }}.scripts.get-logs start ./{{ workspace_dir }}/just.namespace &
     trap "kubectl delete -f ./{{ workspace_dir }}/log-collector.yaml; rm -f ./{{ workspace_dir }}/just.namespace" EXIT
-    nix shell .#{{ set }}.contrast.e2e --command release.test -test.v \
+    # get-logs must be on PATH so the test (via e2e/internal/logcollect)
+    # can call `get-logs download` synchronously before its cleanup.
+    nix shell .#{{ set }}.contrast.e2e .#{{ set }}.scripts.get-logs --command release.test -test.v \
```
burgerdev: Alternatively, we could add the script as input to the e2e derivation.

justfile
```diff
@@ -551,8 +551,13 @@ CONTRAST_LOG_LEVEL=""
...
 # Should be set by running just get-ghcr-read-token.
 contrast_ghcr_read=""
-# A Github token with contents:write permissions for accessing draft releases.
-# Set this manually if you want to use the e2e-release target on a draft release.
+# A GitHub token, used by:
+#   - e2e-release: download release tarballs from the contrast repo
+#     (when testing drafts, the token's user must be able to see them)
+#   - just fmt / treefmt: zizmor calls the GitHub API to resolve action
+#     refs and check for known issues; without the token it errors out
+# A fine-grained PAT with "Contents: Read-only" on edgelesssys/contrast
```
burgerdev: This assumption came from the experience writing

https://github.com/edgelesssys/contrast/blob/3ee35fd309f5ffeca2d91874c6355681534ec81d/.github/workflows/release.yml#L378-L383

where read permissions proved not to be enough. Cool that it works with a fine-grained token!

However, I don't quite understand the commit message:

> The "draft release" caveat is real but separate: the token's user has to be able to see the draft, which is an access-control question, not a permission scope.

How would I set up access control for draft releases?

e2e/internal/logcollect/logcollect.go
sespiros: I have already started a Go rewrite but it is more complex than I expected because if we want to make this more generic, so it can be wired up to a separate subcommand that users could also use, there are other considerations to keep in mind. But yes when that's done I expect it to live in `/internal` not under `e2e`. I thought to fix this in the meantime until I get the rewrite to a shape I like.

justfile
sespiros: Thanks, pushed a fix.

justfile
sespiros: I must have mixed up the tokens I had generated. I re-ran it now and idd I can't see draft releases with a ro pat. I amended the commit and the message. I keep recommending a rw pat scoped to the Contrast repo.

burgerdev (approved):
Thanks, lgtm!

PR #2376, opened 2026-05-06
imagepuller: upgrade go.podman.io/storage to v1.62.0

burgerdev:
This removes a temporary override we put in to resolve a security scanner finding, https://pkg.go.dev/vuln/GO-2025-4098.


charludo (approved):
Nice, thank you!

PR #2377, opened 2026-05-07
contrast-releases: rename contrast key to contrast-x86_64-linux

sespiros:
and also fix the `nix develop .#demo-v1-20-0` workflow for darwin dev machines.

Tested with `nix develop .#demo-v1-20-0 --command contrast --version`


charludo (approved):
Maybe being more explicit with the override name (`linux_x86-64-filename` or something) could be helpful, but that's a cosmetic change, so, approved.

PR #2382, opened 2026-05-11
dev-shells.kata: let rust-toolchain follow kata.runtime version

charludo:
Noticed a new version was available during kata update.

### DIFF
--- a/dev-shells/by-name/kata.nix
+++ b/dev-shells/by-name/kata.nix
@@ -5,6 +5,7 @@
   # keep-sorted start
   clang,
   cmake,
+  contrastPkgs,
   fenix,
   lib,
   lld,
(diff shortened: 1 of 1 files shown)


dev-shells/by-name/kata.nix
```diff
@@ -18,15 +18,15 @@
 }:
 
 let
-  pinnedCommit = "b6c60d9229b210a1925f43a1a20af1882cee255b";
+  pinnedCommit = "2ffd1538a296cff93a357bfba0dfca747480a1f8";
```
burgerdev: Why are we using a pinned commit here instead of the version from `kata/runtime`?

charludo: Ha, fair question. *Originally* because we didn't have a good/any mechanism to pass the source/version in yet, now because I was used to updating it this way.

Changed it to consume the actual runtime source now. Thanks for the suggestion ^^

burgerdev (approved):

PR #2383, opened 2026-05-11
kata: build incrementally using crane

charludo:
- patches are applied in a new `kata.source` package and consumed from there
- changes to root `Cargo.lock`/`Cargo.toml` files require full rebuild (doesn't matter of upstream or through our patches)
- building e.g. the `agent` still takes ~1:30min, because the entire agent still needs to be rebuilt and tested, but that's still significantly faster than before


burgerdev:
Let's fix this in #2387.

burgerdev:
Oh no, that was not what I meant to close!

charludo:
Nice, this seems to fully wok now.

charludo:
Added a second commit, making things a little more incremental. On my machine, reduces `agent` build times from the (already reduced) 1:22min to 0:38min (of which 22s are tests), `genpolicy` to 2:18min, and `runtime-rs` to 2:09min.

One caveat of this is that changing e.g. a `runtime-rs` patch still causes the next `agent` build to pay the full 1:22min build cost. That's because the intermediate source is different for each workspace member.

Yeah tbh I am not super happy with this step. It's a bit faster, but also less clean. I still have one more idea I'll add in a third commit, and then we'll have three choices to pick from.

charludo:
With crane replaced by crate2nix instead:
- `agent`: 1:06min
- `genpolicy`: 1:51min
- `runtime-rs`: 2:45min

Which means that this solution is not *only* much uglier and harder to maintain, it's also not straightforward to compare performance 🙂

The upside of this solution is that changing e.g. agent code, then rebuilding genpolicy immediately uses the cached output, which is nice.

I have a personal preference but am curious to see what you think. Each commit is a slightly different approach that warrants separate testing and consideration.

charludo:
The third option (crate2nix) doesn't play nice at all with static package builds, it seems. Left the commit in for now, just in case one of you had a smart idea about this. But most likely, option 1 or 2 is the only way to go.

packages/by-name/kata/agent/package.nix
```diff
@@ -17,29 +19,36 @@
...
-      "api_client-0.1.0" = "sha256-RdwQg6/EI+oGkyNXnu5t1q87oTXev25XpIaE+PWDTx4=";
-      "micro_http-0.1.0" = "sha256-XemdzwS25yKWEXJcRX2l6QzD7lrtroMeJNOUEWGR7WQ=";
-      "regorus-0.9.1" = "sha256-+TCq9r8kTNM0URbcDP4D9/lKA6Bni7+KgrGRTJFbQPM=";
-      "s390_pv_core-0.11.0" = "sha256-P275gUoF4JtaKvKPvzhCsBuo882kKCYebtNpCDEmTP0=";
-    };
-  };
+  cargoExtraArgs = lib.concatStringsSep " " (
+    [
+      "--target"
+      stdenv.hostPlatform.rust.rustcTarget
```
burgerdev: Will that be the correct target when building on Darwin?

packages/by-name/kata/runtime-rs/package.nix
```diff
@@ -3,40 +3,34 @@
...
-      --replace-fail @RUNTIME_NAME@ containerd-shim-kata-v2 \
-      --replace-fail @CONTAINERD_RUNTIME_NAME@ io.containerd.kata.v2
-  '';
+  inherit (source) version cargoVendorDir src;
+  strictDeps = true;
+
+  cargoExtraArgs = lib.concatStringsSep " " (
+    [
+      "--target"
+      stdenv.hostPlatform.rust.rustcTarget
```
burgerdev: Same here.

packages/by-name/kata/runtime/package.nix
```diff
@@ -4,184 +4,18 @@
...
-      # We need the layer information to calculate the memory overhead for the VM during generate.
-      ./0027-genpolicy-write-processed-layer-information-to-file.patch
-    ];
-  };
+  inherit (source) version src;
 
   sourceRoot = "${finalAttrs.src.name}/src/runtime";
 
-  vendorHash = "sha256-TVIWv+b94z3QuhAxRbRpCuMxQtE8Nq2fuQ4R9I6UqMs=";
+  vendorHash = null;
```
burgerdev: Where do we pin that now?

packages/by-name/kata/source/package.nix
```diff
@@ -0,0 +1,218 @@
...
+
+  srcRaw = fetchFromGitHub {
+    owner = "kata-containers";
+    repo = "kata-containers";
+    rev = version;
+    hash = "sha256-LSwmeNO5mEWZ2NCrC1o3JXPeZJed0eljIUak/J2fkv0=";
+  };
+
+  src = applyPatches {
+    src = srcRaw;
```
burgerdev: Is there a reason why this is not inlined? I think we're nesting `applyPatches -> fetchFromGithub` in most other packages.

packages/by-name/kata/source/package.nix
```diff
@@ -0,0 +1,218 @@
...
+
+    patches = [
+      ./0001-emulate-CPU-model-that-most-closely-matches-the-host.patch
+
+      #
```
burgerdev, line 32: It might be helpful to split the commit such that `runtime/package.nix` is recognized as the origin of `source/package.nix`, to preserve history, especially for the patches.

burgerdev (commented):
Review of patch 1.

packages/by-name/kata/source/package.nix
charludo: Won't we have to remove the runtime part of this anyways with the upcoming kata release?

charludo:
Sorry about the force push, just rebasing to solve the merge conflict.

PR #2384, opened 2026-05-11
docs: PIID extraction using PoE Generator

burgerdev:
The currently documented method is cumbersome and has some subtly dangerous components. Let's recommend users to switch to the PoE tool, which was specifically designed for this use case.


sespiros (approved):
LGTM, thanks

PR #2387, opened 2026-05-12
just: don't attempt to kill log-collector twice

burgerdev:
The `get-logs start` command itself spawns background processes and exits early. 

> You can not kill that which is already dead.

### DIFF
--- a/justfile
+++ b/justfile
@@ -113,9 +113,8 @@ _e2e target=default_deploy_target platform=default_platform set=default_set: sof
         just containerd-reproducer
     fi
     get_logs=$(nix build .#{{ set }}.scripts.get-logs --no-link --print-out-paths)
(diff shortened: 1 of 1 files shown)


sespiros (approved):
Thanks for this fix, I had it already bundled in an upcoming PR. This was a refactor artifact after changing the way `start` works too many times.

burgerdev:
Turns out that the command is short-lived, but does not return immediately.

sespiros:
@burgerdev ptal https://github.com/edgelesssys/contrast/pull/2388, I opened it as draft because I haven't tested it yet thoroughly.

burgerdev:
Let's fix this in https://github.com/edgelesssys/contrast/pull/2387.

PR #2392, opened 2026-05-12
contrast.reference-values: enable multi-CPU for non-GPU TDX

charludo:


charludo:
Closed in favor of working out a proper solution.

PR #2393, opened 2026-05-12
e2e: dont remove labeled resource during release e2e test

daniel-weisse:
The release e2e test kept removing RuntimeClasses that were still in use by other deployments.
Update the test to respect the `ci.contrast.edgeless.systems/keep` label


burgerdev (approved):
Thanks!

PR #2394, opened 2026-05-13
.github: do not attempt to delete nonexistent tag

charludo:
https://github.com/edgelesssys/contrast/actions/runs/25761363482/job/75663107538

The tag deletion was a mistake, we only create tags after/during release promotion, so the deletion fails.

(The actual draft deletion worked like a charm, the error comes from then trying to clean up the tag.)

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -62,7 +62,7 @@ jobs:
             | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+-[0-9]{4}-[0-9]{2}-[0-9]{2}$' \
             | while IFS= read -r tag; do
                 echo "Deleting nightly draft release: ${tag}"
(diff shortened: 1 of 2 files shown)


sespiros (approved):
I think maybe we shouldn't name "tag" the `{FULL_VERSION}-${nightly_date}` because it is confusing. I read "delete the release with tag foo but don't cleanup the tag". 

Maybe something simpler like `release_name` and `NIGHTLY_RELEASE_NAME` would be better in `release_{nightly,promote}.yml` respectively.

charludo:
> I think maybe we shouldn't name "tag" the `{FULL_VERSION}-${nightly_date}` because it is confusing. I read "delete the release with tag foo but don't cleanup the tag".
> 
> Maybe something simpler like `release_name` and `NIGHTLY_RELEASE_NAME` would be better in `release_{nightly,promote}.yml` respectively.

Yeah fair. pushed a commit renaming these.

PR #2395, opened 2026-05-13
kata: use modern syntax for error message policy

burgerdev:
The `regorus` crate was bumped in the last Kata update, and the new version does not support the older Rego syntax anymore. Add `if` keywords to the synthetic rules we create for propagating error messages from initdata-processor to runtime.

Included are two commits without which debugging the situation was harder than necessary.


charludo (approved):
Good luck for upstreaming the patch 😛

PR #2397, opened 2026-05-18
ci/release_e2e_test: scope fifo ticket to GPU platforms only

sespiros:
## Summary

Non-GPU darwin entries of the nightly release matrix were failing at `just request-fifo-ticket` with `services "sync" not found`.

The condition gating the ticket step matched every darwin job because `inputs.self_hosted` is `'false'` for darwin (it is derived from `cli-arch == 'x86_64-linux'` in `release_nightly.yml`), so the first clause of the OR short-circuited the rest.
The sync fifo only exists on GPU clusters (see `bm_maintenance.yml`), so the platform name is the only signal that matters.

## Test plan

- [ ] Re-run the failing nightly release job and confirm non-GPU darwin entries skip the ticket step.
- [ ] Confirm GPU entries (both darwin and linux) still acquire and release a ticket.

### DIFF
--- a/.github/actions/release_e2e_test/action.yml
+++ b/.github/actions/release_e2e_test/action.yml
@@ -123,7 +123,7 @@ runs:
 
     # --- common steps ---
     - name: Request fifo ticket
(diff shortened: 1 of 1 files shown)


charludo (approved):
Thanks!

PR #2398, opened 2026-05-18
multiple: use resourcegen to dynamically generate k8s-log-collector image and use it in the nightly release workflow

sespiros:
2 changes:
- Replace the static `log-collector.yaml` file for a dynamically generated image built with resourcegen. 
- Address https://github.com/edgelesssys/contrast/actions/runs/26001951657/job/76441097386 by pushing the log-collector image


internal/kuberesource/sets.go
```diff
@@ -57,6 +58,79 @@ func Runtime(platform platforms.Platform) ([]any, error) {
 	}, nil
 }
 
+// LogCollector returns the DaemonSet and PriorityClass that deploy the
+// k8s-log-collector image. This is dev/test-only infrastructure consumed by
+// the get-logs script; the image is the lookup placeholder, substituted at
+// deploy time via just.containerlookup.
```
charludo: Please edit LLM comments down, esp. since things like the container-lookup are in no way unique to this set.

charludo (approved):
LGTM as a fix, and I don't want to stand in the way of this.

However, I've been wondering: maybe we should (not in this PR) actually use the `codegen` script for this as well, both for the `log-collector` and all other images. We could keep a file in-tree with all the digests of our own images (a commited `just.containerlookup`, so to speak), and auto-update that on container push. Any CI jobs requiring specific images could then just use that lookup file, including e.g. things like the `release_artifacts` action.

That *should* also fix this issue, right?

LMK what you think. In any case, that's not for this PR. Also I'm not sure if I'm maybe not seeing some complications here.

sespiros:
> LGTM as a fix, and I don't want to stand in the way of this.
> 
> However, I've been wondering: maybe we should (not in this PR) actually use the `codegen` script for this as well, both for the `log-collector` and all other images. We could keep a file in-tree with all the digests of our own images (a commited `just.containerlookup`, so to speak), and auto-update that on container push. Any CI jobs requiring specific images could then just use that lookup file, including e.g. things like the `release_artifacts` action.
> 
> That _should_ also fix this issue, right?
> 
> LMK what you think. In any case, that's not for this PR. Also I'm not sure if I'm maybe not seeing some complications here.

Yes that would also fix it. The fix really is pushing the image in the release test. 

The refactor to use resourcegen was an idea I had because I didn't like that I had to do the following (because there is a hardcoded hash in there): https://github.com/edgelesssys/contrast/blob/43ef0c180d34fa9f41bbc53437f6012bbebb83c6/packages/by-name/scripts/get-logs/get-logs.sh#L53-L58
but isn't related to the fix strictly speaking.

PR #2400, opened 2026-05-19
ci: embed image-replacements in darwin CLI assets

sespiros:
The x86_64-linux build commits cli/cmd/assets/image-replacements.txt locally before building the CLI. The darwin build runs on a fresh macos-latest checkout that never sees that commit and embeds the placeholder comment instead, so `contrast generate` passes `initializer:latest` to genpolicy and trips ManifestUnknown.

Pull the image-replacements artifact (already uploaded by the linux build, gated via needs:) and drop it into cli/cmd/assets/ before the Nix build, mirroring what the linux job does in-tree.

### DIFF
--- a/.github/actions/release_build_darwin/action.yml
+++ b/.github/actions/release_build_darwin/action.yml
@@ -27,6 +27,14 @@ runs:
       with:
         version: ${{ inputs.without_v }}
         commit: false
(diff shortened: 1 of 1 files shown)


charludo (approved):
Ahh, nice catch.

PR #2401, opened 2026-05-19
justfile: auto-enable debug shell for badaml e2e tests

sespiros:
BadAMLTest's first action is `require.True(t, Flags.InsecureEnableDebugShell, ...)` (`e2e/badaml-vuln/badaml.go:27`), so `just e2e badaml-vuln` on a fresh checkout aborts immediately with the default `debug=false`. There is no discoverability today: users find the requirement by hitting the assert.

Mirror the existing `required_sets` pattern with a `required_debug` map that unconditionally overrides `debug` for the two affected targets. Same semantics, same silent-override trade-off, no caller-side changes. CI workflows already pass `debug-shell=true` for badaml runs.

### DIFF
--- a/justfile
+++ b/justfile
@@ -80,7 +80,7 @@ collect-platforms platform=default_platform set=default_set:
         --deployment "$deployment" \
         collect-platforms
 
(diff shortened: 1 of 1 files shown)


charludo (approved):
LGTM, thanks!

PR #2402, opened 2026-05-19
collateral-proxy: init and use in e2e tests

charludo:
Left in this repo for now/review/iteration speed, but I have created a task to move the actual proxy out of the repo after reviews are completed and this is merge-ready.

BTW: we should probably come up with a new name, since this is not JUST a kds proxy.


charludo:
@thomasten we could probably replace this with something like [Squid](https://github.com/squid-cache/squid) + its ssl_bump plugin, which can also self-mint leaf certs and do the MITM. Not sure if that would be easier than the bespoke solution though. Esp. since I don't think we can get CRL-aware freshness checks there.

thomasten:
Thanks for checking!

> Esp. since I don't think we can get CRL-aware freshness checks there.

Yes, that's what I thought too.

Maybe the other reviewers have thoughts on this.

internal/kuberesource/parts.go
```diff
@@ -573,3 +573,58 @@ func GetPodCPUCount(spec *applycorev1.PodSpecApplyConfiguration) uint64 {
...
+			WithRequests(corev1.ResourceList{corev1.ResourceStorage: resource.MustParse("1Gi")}))
+	if storageClassName != "" {
+		pvcSpec = pvcSpec.WithStorageClassName(storageClassName)
+	}
+	pvc := applycorev1.PersistentVolumeClaim(name+"-state", namespace).WithSpec(pvcSpec)
+
+	mem := corev1.ResourceList{corev1.ResourceMemory: resource.MustParse("256Mi")}
+	deployment := Deployment(name, namespace).
+		WithSpec(DeploymentSpec().
+			WithReplicas(1).
```
burgerdev: Should we treat this as a `StatefulSet` with a volume claim template, since the PVC is fixed anyway?

internal/kuberesource/parts.go
```diff
@@ -573,3 +573,58 @@ func GetPodCPUCount(spec *applycorev1.PodSpecApplyConfiguration) uint64 {
...
+						WithName(stateVol).
+						WithPersistentVolumeClaim(applycorev1.PersistentVolumeClaimVolumeSource().
+							WithClaimName(name + "-state"))).
+					WithContainers(applycorev1.Container().
+						WithName(name).
+						WithImage("ghcr.io/edgelesssys/contrast/kds-proxy:latest").
+						WithArgs(fmt.Sprintf("-addr=:%d", port), "-state-dir="+stateDir).
+						WithPorts(applycorev1.ContainerPort().
+							WithName("proxy").
+							WithContainerPort(port)).
```
burgerdev, line 641: If we used a standard HTTPS port here, `ServiceForDeployment` would use it too, and we could just configure the endpoint as `https://kds-proxy.default` without remembering an obscure port.

kds-proxy/internal/allowlist/allowlist.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2026 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package allowlist
+
+// Default lists the hosts that kds-proxy will proxy to. Any other host is rejected.
+var Default = map[string]struct{}{
```
burgerdev: Does this need to be exported?

kds-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,208 @@
...
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+	}
+}
+
+func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
+	host := r.URL.Host
+	if host == "" {
+		host = r.Host
+	}
+	sniHost := stripPort(host)
```
burgerdev: Does this have anything to do with SNI?

kds-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,208 @@
...
+	if allows == nil {
+		allows = allowlist.Allows
+	}
+	return &Server{log: log, ca: ca, cache: c, upstream: u, allows: allows}
+}
+
+func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case http.MethodConnect:
+		s.handleConnect(w, r)
```
burgerdev: Is there a reason why we build a `CONNECT` proxy instead of an application layer proxy, other than that the endpoints are currently not configurable in libraries?

kds-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,208 @@
...
+		if entry, fresh := s.cache.Get(fullURL); fresh {
+			s.hits.Add(1)
+			s.log.Info("cache hit", "url", fullURL)
+			writeRaw(conn, entry.Status, entry.Header, entry.Body)
+			continue
+		}
+
+		res, err := s.upstream.Get(ctx, fullURL)
+		if err != nil {
+			if entry, _ := s.cache.Get(fullURL); entry != nil {
```
burgerdev: We could already have `entry` from a couple of lines above.

kds-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,208 @@
...
+				s.stale.Add(1)
+				s.log.Warn("serving stale on upstream error", "url", fullURL, "err", err)
+				writeRaw(conn, entry.Status, entry.Header, entry.Body)
+				continue
+			}
+			s.errors.Add(1)
```
burgerdev: A metric field for HTTP error or so might be interesting. Maybe also type of document (CRL, attestation key certificate, collateral, ?).

kds-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+}
+
+func (s *Server) writeMetrics(w http.ResponseWriter) {
+	fmt.Fprintf(w, "kds_proxy_cache_hits %d\n", s.hits.Load())
+	fmt.Fprintf(w, "kds_proxy_cache_misses %d\n", s.misses.Load())
+	fmt.Fprintf(w, "kds_proxy_cache_stale %d\n", s.stale.Load())
+	fmt.Fprintf(w, "kds_proxy_rejected %d\n", s.rejected.Load())
+	fmt.Fprintf(w, "kds_proxy_upstream_errors %d\n", s.errors.Load())
+}
```
burgerdev: idk ... can't we just use `promauto` or something?

kds-proxy/internal/cache/cache.go
```diff
@@ -0,0 +1,166 @@
...
+		Header:     header.Clone(),
+		Body:       body,
+		FreshUntil: time.Now().Add(freshness(header, body)),
+	}
+	if err := c.writeDisk(e); err != nil {
+		return nil, err
+	}
+	c.mu.Lock()
+	c.entries[url] = e
+	c.mu.Unlock()
```
burgerdev: It looks like this can cause inconsistencies between memory cache and disk cache:

1. T1 writes the file.
2. T2 writes the file
3. T2 updates `entries`.
4. T1 updates `entries`.

kds-proxy/internal/cache/cache.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func freshness(header http.Header, body []byte) time.Duration {
+	if d, ok := crlFreshness(body); ok {
+		return d
+	}
+	if d, ok := cacheControlMaxAge(header); ok {
+		return d
+	}
+	return minTTL
```
burgerdev: Isn't this rather `maxTTL`, or `defaultTTL`?

kds-proxy/internal/cache/cache.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func cacheControlMaxAge(header http.Header) (time.Duration, bool) {
+	cc := header.Get("Cache-Control")
+	if cc == "" {
+		return 0, false
+	}
+	if strings.Contains(cc, "no-store") || strings.Contains(cc, "no-cache") {
+		return 0, false
+	}
```
burgerdev: We should also understand the `must-understand` directive.

kds-proxy/internal/cache/cache.go
```diff
@@ -0,0 +1,166 @@
...
+		return time.Duration(secs) * time.Second, true
+	}
+	return 0, false
+}
+
+func crlFreshness(body []byte) (time.Duration, bool) {
+	if len(body) == 0 {
+		return 0, false
+	}
+	crl, err := x509.ParseRevocationList(body)
```
burgerdev: Do we know that CRLs come as DER, and not PEM-encoded?

e2e/internal/contrasttest/contrasttest.go
```diff
@@ -85,6 +86,10 @@ type ContrastTest struct {
 	GHCRToken                      string
 	Kubeclient                     *kubeclient.Kubeclient
 
+	// SkipKDSProxy disables injection of kds-proxy env vars + CA bundle into.
+	// Used by kds-pcs-downtime test.
+	SkipKDSProxy bool
```
burgerdev: Why do we deploy the proxy per-run, not per-cluster?

internal/kuberesource/parts.go
charludo: done

internal/kuberesource/parts.go
charludo: done, as talked about yesterday, port 80/plain http.

kds-proxy/internal/proxy/proxy.go
charludo: as decided yesterday, remains a connect proxy for now to not have to patch any libraries, unless we see issues cropping up.

kds-proxy/internal/proxy/proxy.go
charludo: done

kds-proxy/internal/proxy/proxy.go
charludo: Yeah good idea. Done. I just had never done much with prometheus/promauto, so this wasn't on my mind for a debugging tool :D

kds-proxy/internal/cache/cache.go
charludo: oh yeah, good catch. Lock held across disk write and map update now.

kds-proxy/internal/cache/cache.go
charludo: done. Also, TIL.

kds-proxy/internal/cache/cache.go
charludo: Looks like pcs is indeed PEM encoded, hadn't noticed since I was only testing on the dev server. Proxy understands both now.

e2e/internal/contrasttest/contrasttest.go
charludo: n/a

charludo:
@burgerdev sorry about the force-pushes / history rewrites. Thought it prudent to also rename the proxy.

Edit: ah, the proxy is deployed on all nodes; let's see if the tests take to it as they should 👀

charludo:
- [x] [Metal-QEMU-SNP, regression](https://github.com/edgelesssys/contrast/actions/runs/27752219475)
- [x] [Metal-QEMU-TDX, regression](https://github.com/edgelesssys/contrast/actions/runs/27752221701)
- [x] [Metal-QEMU-SNP-GPU, regression](https://github.com/edgelesssys/contrast/actions/runs/27752224014)
- [x] [Metal-QEMU-TDX-GPU, regression](https://github.com/edgelesssys/contrast/actions/runs/27752225836)

charludo:
@thomasten NV RIM is now (again) implemented, and the new version of the proxy is deployed everywhere I have access to (esp. palutena/dgx-007). As far as I'm aware, you should be able to simply change `https://rim-cache/` to `http://collateral-proxy.default.svc/` and use a default `httpClient` in the `rimClient`. Is this an acceptable solution for your needs?

@burgerdev from my side, ready to review again 🙂 We could probably squash some of the commits, I just left them as-is for reviewability.

thomasten:
> simply change `https://rim-cache/` to `http://collateral-proxy.default.svc/` and use a default `httpClient` in the `rimClient`. Is this an acceptable solution for your needs?

Thanks, works great!

We may need optional TLS certificate configuration for the server later for compliance reasons in production, but (1) I guess this should be easily addable and (2) doesn't need to be part of this PR.

charludo:
Fantastic!

> (1) I guess this should be easily addable and (2) doesn't need to be part of this PR

Yes, should be easy to add later on. Sounds good!

collateral-proxy/internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	tmp := path + ".tmp"
+	if err := os.WriteFile(tmp, raw, 0o600); err != nil {
+		return err
+	}
+	return os.Rename(tmp, path)
+}
+
+func keyHash(url string) string {
+	sum := sha256.Sum256([]byte(url))
+	return hex.EncodeToString(sum[:])
```
burgerdev: It would be easier to debug a cache store if the path was reversible to the URL, for example with `base64url` encoding.

collateral-proxy/internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+
+func crlFreshness(body []byte) (time.Duration, bool) {
+	if len(body) == 0 {
+		return 0, false
+	}
+	der := body
+	if block, _ := pem.Decode(body); block != nil {
+		der = block.Bytes
+	}
+	crl, err := x509.ParseRevocationList(der)
```
burgerdev, line 184: We're feeding responses to `crlFreshness` that are not CRLs, which seems to only work because `x509.ParseRevocationList` is robust. Could we determine the CRL-ness more accurately somehow?

collateral-proxy/internal/cache/cache_test.go
burgerdev, line 1: The tests should use `require` and `assert`, by convention.

collateral-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,266 @@
...
+
+func documentType(path string) string {
+	p := strings.ToLower(path)
+	switch {
+	case strings.Contains(p, "crl"):
+		return "crl"
+	case strings.Contains(p, "cert_chain"), strings.Contains(p, "certchain"),
+		strings.Contains(p, "tcb"), strings.Contains(p, "identity"),
+		strings.Contains(p, "/rim"):
+		return "collateral"
```
burgerdev: These matchers look overly permissive, can we make them more specific?

collateral-proxy/internal/proxy/proxy_test.go
burgerdev, line 1: Same comment about `require`.

collateral-proxy/internal/upstream/upstream.go
```diff
@@ -0,0 +1,63 @@
...
+	Body   []byte
+}
+
+// Fetcher pulls upstream responses.
+type Fetcher struct {
+	client *http.Client
+	group  singleflight.Group
+}
+
+// New returns a Fetcher backed by client. The client must not be pre-configured with a proxy.
```
burgerdev: Why not? I could imagine the proxy running in an environment that needs a proxy to connect to the public internet.

collateral-proxy/main.go
```diff
@@ -0,0 +1,82 @@
...
+	go func() { errCh <- httpSrv.ListenAndServe() }()
+
+	select {
+	case err := <-errCh:
+		if !errors.Is(err, http.ErrServerClosed) {
+			return err
+		}
+	case <-ctx.Done():
+		log.Info("shutdown signal received")
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
burgerdev: This should have been flagged by golangci-lint (probably a missing entry for this package?). What you usually do to work around this is `context.WithTimeout(context.WithoutCancel(ctx), ...)`.

internal/kuberesource/mutators_test.go
burgerdev, line 1: Why are there a bunch of format changes in this commit (c2e42b9fd85b54b7adf2af247eb4ad1a4c51b29d)?

packages/containers.nix
```diff
@@ -73,6 +73,16 @@
     ];
   };
 
+  collateral-proxy = contrastPkgs.buildOciImage {
+    name = "collateral-proxy";
+    tag = "v${contrastPkgs.collateral-proxy.version}";
+    copyToRoot = (with pkgs; [ busybox ]) ++ (with dockerTools; [ caCertificates ]);
+    config = {
+      Entrypoint = [ "${contrastPkgs.collateral-proxy}/bin/collateral-proxy" ];
```
burgerdev: That could use `lib.getExe`, right?

cli/cmd/generate.go
```diff
@@ -97,6 +97,7 @@ subcommands.`,
 	must(cmd.Flags().MarkHidden("image-replacements"))
 	cmd.Flags().Bool("skip-initializer", false, "skip injection of Contrast Initializer")
 	cmd.Flags().Bool("skip-service-mesh", false, "skip injection of Contrast service mesh sidecar")
+	cmd.Flags().String("collateral-proxy", "", "route attestation-collateral fetches through the caching proxy at this base URL")
```
burgerdev, line 100: I wonder if set/verify/recover should get the flag, too, for local use by the CLI. But that can be a separate PR.

internal/kuberesource/parts.go
```diff
@@ -605,7 +605,7 @@ func GetPodCPUCount(spec *applycorev1.PodSpecApplyConfiguration) uint64 {
 	return uint64(totalCPUs)
 }
 
-// CollateralProxy returns the resources for an in-cluster HTTPS forward proxy caching responses from AMD KDS, Intel PCS, and NVIDIA RIM endpoints.
+// CollateralProxy returns the resources for an in-cluster read-through caching reverse proxy for responses from AMD KDS, Intel PCS, and NVIDIA RIM endpoints.
```
burgerdev: It's still a forward proxy, right?

burgerdev (commented):
Please squash the commits such that the CA and CONNECT bits are gone, the PR is a bit hard to follow right now.

internal/kuberesource/mutators_test.go
charludo: Graaaah because somewhere in the last 2 weeks some of my editor tooling must have gone out of sync with what we're using, but in a way that our formatters still accept the changes.

I don't like that this is happening. Any objections to adding `gofumpt` as a write-formatter to treefmt and running it on the codebase in a separate PR? The diff is not huge.

collateral-proxy/internal/cache/cache.go
charludo: done

collateral-proxy/internal/cache/cache.go
charludo: The idea was that a body that parses as a `RevocationList  with a `NextUpdate` is a CRL for our caching purposes. What alternative did you have in mind?

collateral-proxy/internal/cache/cache_test.go
charludo: sorry, yes. done. The other one too.

collateral-proxy/internal/proxy/proxy.go
charludo: This is just for the prometheus labels, but yeah. Unnecessary anyways to duplicate the logic from the routing decision. Put them together into a single function.

collateral-proxy/internal/upstream/upstream.go
charludo: Missed this in the rewrite, sorry. No longer true. At one point I was worried about building a loop from the proxy to itself 😄

collateral-proxy/main.go
charludo: I don't think so, `just lint` includes the proxy, since it's in the `go.work`. Anyways. Fixed.

packages/containers.nix
charludo: indeed

cli/cmd/generate.go
charludo: Yeah, good idea. Created the followup ticket.

internal/kuberesource/parts.go
charludo: Mhhh... Not really, right? Previously clients tunneled arbitrary TLS traffic through the proxy, which we then intercepted and broke open. Now we offer specific endpoints, which we map to to some fixed upstream URLs.

Just from NGINX docs (was just the first thing that came up) I also don't think we any longer match (their) defintion of a forward proxy.

burgerdev: In my mind, a forward proxy sits next to a client ([corporate firewalls](https://www.checkpoint.com/cyber-hub/network-security/what-is-a-forward-proxy/), SOCKS5 proxies) and a reverse proxy sits next to a server (load balancer, ingress, [API gateway](https://www.cloudflare.com/learning/cdn/glossary/reverse-proxy/)).

internal/kuberesource/mutators_test.go
burgerdev: If it's too big of a hassle to revert the format changes, that's ok.

collateral-proxy/internal/cache/cache.go
burgerdev: As long as we need to support DER, too, I don't see a clean way.

internal/kuberesource/mutators_test.go
charludo: Alright, thanks. Still though: argh.

internal/kuberesource/parts.go
charludo: OK I'm just going to accept your judgement and adjust the comment, I do not care sufficiently about the nomenclature here 😅

sespiros:
I removed myself from this PR as it has received plenty of review comments already. Please add me back if my input is needed.

collateral-proxy/internal/cache/cache.go
```diff
@@ -0,0 +1,193 @@
...
+func (c *Cache) writeDisk(e *Entry) error {
+	raw, err := json.Marshal(e)
+	if err != nil {
+		return err
+	}
+	path := filepath.Join(c.dir, base64.RawURLEncoding.EncodeToString([]byte(e.URL))+".json")
+	tmp := path + ".tmp"
+	if err := os.WriteFile(tmp, raw, 0o600); err != nil {
+		return err
+	}
```
burgerdev, line 114: This could be constructing something like a `json.Writer(base64.Encoder(os.File))`.

collateral-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,221 @@
...
+			Help:      `Upstream fetch outcomes by HTTP status code (or "error") and document type.`,
+		}, []string{"code", "document"}),
+	}
+	for _, t := range documentTypes {
+		for _, r := range requestResults {
+			m.requests.WithLabelValues(r, t)
+		}
+		m.upstream.WithLabelValues("error", t)
+	}
+	m.requests.WithLabelValues("rejected", "unknown")
```
burgerdev, line 52: I don't think this is necessary: the field values don't need to be preallocated, and filtering by a specific value that's not yet been used (while aggregating metrics somewhere else) will just produce a 0 naturally. If the point is to make manual interpretation of the metrics more intuitive from the start of the proxy, we can keep it, but should maybe leave a comment for future readers.

collateral-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,221 @@
...
+		return "certificates.trustedservices.intel.com", intelCertsDocType(path), true
+	case strings.HasPrefix(path, "/v1/rim/"):
+		return "rim.attestation.nvidia.com", "collateral", true
+	default:
+		return "", "", false
+	}
+}
+
+// amdHardwareID matches the hex hardware ID that addresses a VCEK/VLEK certificate.
+var amdHardwareID = regexp.MustCompile(`^[0-9a-fA-F]+$`)
```
burgerdev, line 96: Should that be a specific number of hex bytes?

collateral-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,221 @@
...
+func intelCertsDocType(path string) string {
+	if lastSegment(path) == "IntelSGXRootCA.der" {
+		return "crl"
+	}
+	return "unknown"
+}
+
+func lastSegment(path string) string {
+	return path[strings.LastIndex(path, "/")+1:]
+}
```
burgerdev, line 136: Could this just be [`path.Base`](https://pkg.go.dev/path#Base)?

collateral-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,221 @@
...
+		return
+	}
+	s.serveCollateral(w, r, upstreamHost, docType)
+}
+
+func (s *Server) serveCollateral(w http.ResponseWriter, r *http.Request, upstreamHost, docType string) {
+	upstreamURL := "https://" + upstreamHost + r.URL.Path
+	if r.URL.RawQuery != "" {
+		upstreamURL += "?" + r.URL.RawQuery
+	}
```
burgerdev, line 166: This would be a bit safer with the `url.URL` type, instead of string concatenation.

packages/by-name/collateral-proxy/package.nix
```diff
@@ -0,0 +1,44 @@
...
+  env.CGO_ENABLED = 0;
+
+  ldflags = [
+    "-s"
+    "-X main.version=v${finalAttrs.version}"
+  ];
+
+  checkPhase = ''
+    runHook preCheck
+    go test ./...
```
burgerdev, line 39: Should we add `-race`?

packages/containers.nix
```diff
@@ -73,6 +74,16 @@
     ];
   };
 
+  collateral-proxy = contrastPkgs.buildOciImage {
+    name = "collateral-proxy";
+    tag = "v${contrastPkgs.collateral-proxy.version}";
+    copyToRoot = (with pkgs; [ busybox ]) ++ (with dockerTools; [ caCertificates ]);
+    config = {
+      Entrypoint = [ (lib.getExe contrastPkgs.collateral-proxy) ];
+      Env = [ "PATH=/bin" ];
```
burgerdev, line 83: Not necessary for non-confidential images.

burgerdev (approved):
This lgtm, just update the fork refs when we're ready rebasing them, merge this PR and fix the remaining nits in a new PR.

collateral-proxy/internal/proxy/proxy.go
charludo: That was indeed the intention. Getting back a completely empty response body on the metrics was unexpected, at least to me. IMO it's also useful to see all available categories from the start.

collateral-proxy/internal/proxy/proxy.go
charludo: indeed

collateral-proxy/internal/proxy/proxy.go
charludo: indeed

collateral-proxy/internal/proxy/proxy.go
charludo: done

packages/by-name/collateral-proxy/package.nix
charludo: done

packages/containers.nix
charludo: done

collateral-proxy/internal/cache/cache.go
charludo: done

PR #2403, opened 2026-05-20
tools/badaml-payload: avoid fine scan after coarse hit

sespiros:
This is another attempt at making the reproducer less prone to flaky failures because of #VC 404 terminations.

The new badaml-vuln payload proves the attack once it finds and patches **any** word inside the initrd marker file. The backward 4-byte fine scan only finds byte zero of the marker, but can add millions of extra SystemMemory reads after a coarse hit.

Those reads are another potential surface that can trigger the SEV-SNP #VC 0x404 termination seen in CI. Patch the coarse hit directly and make the e2e check look for cafebabe anywhere in the copied file.

### DIFF
--- a/e2e/badaml-vuln/badaml.go
+++ b/e2e/badaml-vuln/badaml.go
@@ -52,7 +52,7 @@ func BadAMLTest(t *testing.T, expectSuccessfulAttack bool) {
 	require.True(t, t.Run("get content of /run/deadbeef.bin", func(t *testing.T) {
 		ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(1*time.Minute))
 		defer cancel()
-		cmd := []string{"debugshell", `hexdump -e '1/1 "%02x"' -n4 /run/deadbeef.bin`}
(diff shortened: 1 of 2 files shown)


charludo:
Hmm. What was the original intention with the course-then-fine scanning? I assume to prevent false-positive `0xCAFEBABE` matches by adding the position as an additional pointer to this being intentional?

sespiros:
> Hmm. What was the original intention with the course-then-fine scanning? I assume to prevent false-positive `0xCAFEBABE` matches by adding the position as an additional pointer to this being intentional?

I don't think this makes the check more intentional. I think `deadbeef` instead of `cafebabe` anywhere in the file is plenty enough proof that this was intentional :D iirc we were just trying to modify the first word of the file initially and we were fixated in making that particular goal happen. My intention with this is to try and cut down on the amount of memory accesses we do from asl since they seem to be triggering the VM exits.

charludo (approved):
> iirc we were just trying to modify the first word of the file initially

Ah, makes sense. In taht case, LGTM.

PR #2404, opened 2026-05-20
.github: run nightly tests from per-platform sub-workflow for better grouping

charludo:
This moves per-platform tests into a reusable workflow. *Calling* that workflow should group the jobs similar to how they were before.

TBH I'd like to hold off on merging this until we've had a successful release, since it's just cosmetic for us and I don't want to risk new unforeseen failures tonight.

### DIFF
--- a/.github/workflows/e2e_nightly_platform.yml
+++ b/.github/workflows/e2e_nightly_platform.yml
@@ -1,0 +1,99 @@
+name: e2e test nightly (per platform)
+
+on:
+  workflow_call:
+    inputs:
+      platform-name:
+        description: "Platform name (e.g. Metal-QEMU-SNP)"
+        type: string
+        required: true
+      runner:
+        description: "Self-hosted runner label"
+        type: string
+        required: true
+      self-hosted:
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e_nightly_platform.yml
```diff
@@ -0,0 +1,99 @@
...
+          - workloadsecret
+          # keep-sorted end
+        set: [""]
+        exclude:
+          - debug-shell: false
+            test-name: badaml-vuln
+          - debug-shell: false
+            test-name: badaml-sandbox
+            # If the platform is GPU, exclude the (non-existent) sentinel test, otherwise, exclude the (existing) "gpu" test.
+            # matrix context is not available here, so instead, this workaround.
```
sespiros:
```suggestion
          # If the platform is GPU, exclude the (non-existent) sentinel test, otherwise, exclude the (existing) "gpu" test.
          # matrix context is not available here, so instead, this workaround.
```

sespiros (approved):
Thanks.

PR #2406, opened 2026-05-20
genpolicy: save layer sizes to extra file

davidweisse:
To configure the pod memory requests automatically during `contrast generate`, we need access to the compressed and uncompressed layer sizes used within the pod. The sensible place to do this is during `genpolicy`, since the layers are pulled here anyway. This also means we should use the `layers-cache.json` file to cache the sizes.

The layers cache in `genpolicy` is indexed by the `diffID`, i.e., the hash over the uncompressed layer. If we want to store the compressed layer sizes as well, it only makes sense to change this to the `layer_digest` instead (hash over the compressed layer). Consider the same layer that has been compressed with different compressing options: compressed layer size would differ, but `diffID` would be the same. We have to adjust our precomputation of the layers cache file during the OCI image build accordingly.

Migrating from an old `layers-cache.json` is done automatically by `genpolicy` by discarding an invalid file.

The second step is to get the information of which layer is needed for which pod to Contrast. The layers cache file is a bad place to store this, because
1. we would need some form of unique identifier per pod resource and
2. if the deployment changes, the layers cache file contains stale data (since this is a global cache file, putting any deployment specific data in here wouldn't make much sense anyway).

Since we only ever invoke `genpolicy` on one pod-generating resource at a time, I decided to add another file output, that basically outputs a subset of the layers cache file, containing only the layers used in the processed resource. On Contrast side, we can then parse this file after running `genpolicy` and add up the memory requirements.

### DIFF
--- a/packages/by-name/kata/runtime/0026-genpolicy-cache-un-compressed-layer-sizes.patch
+++ b/packages/by-name/kata/runtime/0026-genpolicy-cache-un-compressed-layer-sizes.patch
@@ -1,0 +1,224 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: davidweisse <98460960+davidweisse@users.noreply.github.com>
+Date: Mon, 18 May 2026 14:42:01 +0200
+Subject: [PATCH] genpolicy: cache (un-)compressed layer sizes
+
+---
+ src/tools/genpolicy/src/layers_cache.rs       |  4 +-
+ src/tools/genpolicy/src/registry.rs           | 29 +++++++------
+ .../genpolicy/src/registry_containerd.rs      | 42 +++++++++++--------
+ 3 files changed, 44 insertions(+), 31 deletions(-)
+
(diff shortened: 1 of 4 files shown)


packages/by-name/kata/runtime/0029-genpolicy-write-processed-layer-information-to-file.patch
```diff
@@ -0,0 +1,173 @@
...
++                .config
++                .layers_cache
++                .get_layers_by_digests(&used_layer_digests);
++            let file = std::fs::OpenOptions::new()
++                .write(true)
++                .truncate(true)
++                .create(true)
++                .open(path)
++                .unwrap();
++            serde_json::to_writer(file, &layers).unwrap();
```
charludo: Nit: This should happen atomically IMO. E.g. serialize first, write to temp path, and then `rename` atomically.
Probably not going to matter too much in this specific case, as long as we always pass a fresh `pod_layers_file_path`, but still good convention.

charludo (approved):
Very nice! I just have one nit.

charludo (approved):
Thanks! You can ignore the govulncheck error, it just wants to taunt me 😆

PR #2407, opened 2026-05-21
.github: various small fixes for release_promote

charludo:
- release/tmp branch logic was faulty
- rename release_publish.yml to release_cleanup.yml since that's the only thing it does now, and include deleting old nightly assets
- read version.txt in release_promote.yml from nightly_ref instead of main


.github/workflows/release_cleanup.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Semantic version tag of the release (vX.Y.Z)."
+        required: true
+      latest:
```
sespiros: Seems like this input is not used anywhere.

.github/workflows/release_promote.yml
```diff
@@ -152,7 +155,7 @@ jobs:
 
   prepare-github-release:
     name: Create backport label and milestone
-    needs: prepare
+    needs: [prepare, promote]
```
sespiros, line 159: Is this so that we don't create milestones for a failed release? Maybe add a comment here to make it clear.

.github/workflows/release_promote.yml
sespiros: This has the same name as the job above it.

```suggestion
    name: Publish release
```

.github/workflows/release_promote.yml
```diff
@@ -213,6 +216,11 @@ jobs:
         run: |
           git tag "${VERSION}" "${REF}"
           git push origin "${VERSION}"
+      - name: Update release branch to nightly ref
+        env:
+          REF: ${{ needs.prepare.outputs.nightly_ref }}
+          RELEASE_BRANCH: ${{ needs.prepare.outputs.RELEASE_BRANCH }}
+        run: git push origin "${REF}:refs/heads/${RELEASE_BRANCH}"
```
sespiros, line 227: `release/*` branches have branch protection rules that only allow edgelessci to push without review (you can find that in repo settings) and I think this job here is missing the `NUNKI_CI_COMMIT_PUSH_PR` (the edgelessci's token)

This will break for patch releases most probably where the branch already exists but haven't verified.

.github/workflows/release.yml
```diff
@@ -268,6 +268,15 @@ jobs:
           git fetch origin "${target}"
           git tag "${VERSION}" "${target}"
           git push origin "${VERSION}"
+      - name: Update release branch to release tag
+        env:
+          VERSION: ${{ needs.process-inputs.outputs.EFFECTIVE_VERSION }}
+          RELEASE_BRANCH: ${{ needs.process-inputs.outputs.RELEASE_BRANCH }}
+        run: git push origin "refs/tags/${VERSION}^{commit}:refs/heads/${RELEASE_BRANCH}"
```
sespiros, line 278: `release/*` branches have branch protection rules that only allow edgelessci to push without review (you can find that in repo settings) and I think this job here is missing the `NUNKI_CI_COMMIT_PUSH_PR` (the edgelessci's token).

This will break for patch releases most probably where the branch already exists but haven't verified.

.github/workflows/release_promote.yml
```diff
@@ -62,10 +54,22 @@ jobs:
...
-          echo "nightly_version=${nightly_version}" | tee -a "$GITHUB_OUTPUT"
+          echo "nightly_date=${nightly_date}" | tee -a "$GITHUB_OUTPUT"
+      - name: Determine version
+        id: effective-version
+        env:
+          GH_REPO: ${{ github.repository }}
+          NIGHTLY_REF: ${{ steps.nightly-info.outputs.nightly_ref }}
+          NIGHTLY_DATE: ${{ steps.nightly-info.outputs.nightly_date }}
+        run: |
+          version_txt=$(gh api "repos/${GH_REPO}/contents/version.txt?ref=${NIGHTLY_REF}" --jq '.content' | base64 -d)
```
sespiros: Suggestion by Claude:

```suggestion
          version_txt=$(gh api -H "Accept: application/vnd.github.raw" "repos/${GH_REPO}/contents/version.txt?ref=${NIGHTLY_REF}")
```

.github/workflows/release.yml
```diff
@@ -268,6 +268,15 @@ jobs:
           git fetch origin "${target}"
           git tag "${VERSION}" "${target}"
           git push origin "${VERSION}"
+      - name: Update release branch to release tag
```
sespiros, line 273: Add a comment:
```yaml
# release/* requires linear history so this push must be a fast-forward.
```

sespiros (changes requested):
Thanks, a few inline comments.

.github/workflows/release_promote.yml
charludo: Ah, good catch, thank you.

charludo:
Thanks @sespiros, should all be addressed.

sespiros (approved):
Thanks.

PR #2409, opened 2026-05-21
ci: minor release workflow fixes

sespiros:


.github/workflows/release_promote.yml
```diff
@@ -75,14 +75,21 @@ jobs:
...
+            echo "::error::No 'release-requirement: ' jobs found in nightly run ${NIGHTLY_RUN_ID}; job naming may have changed."
+            exit 1
+          fi
+          failed=$(jq -rs '.[] | select(.conclusion != "success") | "\(.name): \(.conclusion)"' <<<"${required_jobs}")
           if [[ -n "${failed}" ]]; then
             echo "::error::The following required jobs did not succeed in nightly run ${NIGHTLY_RUN_ID}:"
             echo "${failed}"
             exit 1
           fi
+          echo "All ${required} release-requirement job(s) passed in nightly run ${NIGHTLY_RUN_ID}."
```
charludo, line 108:
```suggestion
          echo "${required} release-requirement job(s) passed in nightly run ${NIGHTLY_RUN_ID}."
```

The nittiest nit I have ever nitted; but technically, the check above doesn't tell us if only *some* jobs were accidentally renamed... 😁

charludo (approved):
Thanks. Feel free to ignore the nit 😆

.github/workflows/release_promote.yml
sespiros: Fair point, I pushed a fixup with some code to check explicitly the list of tests that need to pass.

charludo: Hm... At that point, we could probably drop the "release requirement: " prefix. I wanted to move them to a "(release requirement)" suffix anyways since currently the step-labels in the graph are a bit hard to read. What do you think?

sespiros: Yes I agree. Should we a) merge this one b) run promote to release (because this promote will require the old names) c) do your rename/update after?

charludo: Ahhhh thanks for thinking of that. Yeah let's merge now.

PR #2410, opened 2026-05-22
kata.runtime: 3.30.0 -> 3.31.0

charludo:
Pretty straightforward this time. Only two things of note:
- `runtime-rs` received `emptydir_mode` support, whihc we disabled (set to `shared-fs`) in `runtime` a while back, so I took the liberty to do the same in `runtime-rs`, and
- kata's bundled vendor for `runtime` seems to be stale, hence the need to specify a `vendorHash` in the nix package where we didn't need one previously.


sespiros (approved):
Thanks.

PR #2415, opened 2026-05-22
ci: keep release_promote actions on workflow ref

sespiros:
Promotion should use the nightly commit as release data, but the promotion machinery must come from the dispatched workflow ref so fixes on main are not bypassed. Keep the remaining local composite actions on that ref and fetch the nightly commit explicitly before tagging it.

Also guard the post-release PR flow by checking that checkout_main_ref still carries the release version being promoted.


charludo (approved):
Thanks, this seems like a good solution.

PR #2417, opened 2026-05-26
docs: update debugshell docs

sespiros:
The container is named contrast-debug-shell, not debug-shell as the howto claimed in four places, so following the docs verbatim was not working.

Also moving the settings.json prerequisites (ExecProcessRequest.regex, ReadStreamRequest, WriteStreamRequest) before the generate step and explain what each one controls, so the policy is permissive before the user tries to exec.

I also added a `just docs` helper.


docs/docs/howto/debugshell.md
```diff
@@ -68,17 +36,53 @@ Ensure the `settings.json` used for generate includes the following settings to
         "ReadStreamRequest": true,
         ...
         "WriteStreamRequest": true
+    }
+}
+```
+
+What each knob controls:
```
charludo:
```suggestion
```

That's just too casual imo 😄

charludo (approved):

docs/docs/howto/debugshell.md
sespiros: Thanks for catching that, pushing a fixup.

PR #2418, opened 2026-05-26
ci: cap bm-maintenance jobs at 15m

sespiros:
Jobs had no timeout, so they inherited GHA's 360m default. If a step hangs (e.g. the cleanup step's kubectl delete on a stuck Terminating namespace), the job wastes 6h before being reaped.


charludo (approved):
oof yeah good idea

sespiros:
15 minutes might still be too much but oh well better than 6 hours.

PR #2419, opened 2026-05-26
grpc: retry aTLS handshake failures caused by connection resets

sespiros:
https://github.com/edgelesssys/contrast/actions/runs/26252518697/job/77319263422#step:3:1634 points to a transient failure that triggered a CI failure instead of being autoretried.

The root cause was traced back to our grpc/retry logic which only auto-retries on "EOF" or "context deadline exceeded" but not on "connection reset by peer".

### DIFF
--- a/internal/grpc/retry/retry.go
+++ b/internal/grpc/retry/retry.go
@@ -6,14 +6,15 @@ package retry
 
 import (
 	"regexp"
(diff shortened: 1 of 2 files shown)


charludo (approved):
Nice, thank you!

PR #2421, opened 2026-05-28
fix `codegen` cycle, fix `govulncheck` findings

charludo:
Instead of just bumping the dependency, this PR includes a second commit that breaks a cycle we had when running `just codegen`.

Previously,
- `contrast.node-installer` inherits `vendorHash` from `contrast.contrast` (as do `cli`, `e2e`, `resourcegen`)
- to update this one, shared `vendorHash`, building `nodeinstaller` is required, because the `reference-values` depend on it, and e.g. the `cli` requires those to build
- this requires the `vendorHash` already be correct

Fixed it by moving from `pre/postPatch` to `pre/postConfigure` hooks. This causes them to no longer propagate into the `goModules` `vendorHash`. They still affect our package's output hashes, of course.


sespiros (approved):
Nice! I am curious how we did the Go updates with that dep cycle in place...

charludo:
> Nice! I am curious how we did the Go updates with that dep cycle in place...

Might just be that we recently had a bunch of changes to the `reference-values`, *and* this specific update now touched a lot of modules' `go.sum` files. But I do remember running into either the same thing or something very similar not long ago, and just fixed it by hand since I didn't have the time to figure this out.

Plus as long as a single element in the cycle is invariant in its output hash, I guess this is not technically an issue 🤔

PR #2422, opened 2026-05-28
.github/e2e_nightly: manually declare platforms instead of via matrix

charludo:
readability in the gh graph view > less code repetition 🤷🏼‍♀️


charludo:
Oh yeah that looks better: https://github.com/edgelesssys/contrast/actions/runs/26564110415

sespiros (approved):

PR #2425, opened 2026-06-01
.github: group cleanup jobs by runner, notify after e2e

charludo:
The TDX-GPU runner was intentionally shut off over the weekend. This caused no e2es to be run, because the cleanup jobs were grouped by job type, i.e. each job required the TDX-GPU runner. 

This PR re-groups these steps by runner type. Once cleanup for a runner is complete, e2e tests for this runner can immediately begin, regardless of the failure state of the other cleanup jobs.

Also moved the notification job to after the e2e tests.

### DIFF
--- a/.github/workflows/bm_build_image.yml
+++ b/.github/workflows/bm_build_image.yml
@@ -1,0 +1,57 @@
+name: build bare metal maintenance image
+
+on:
+  workflow_call:
+    secrets:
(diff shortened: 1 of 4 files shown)


charludo:
(Example, cancelled so as to not clog the pipeline: https://github.com/edgelesssys/contrast/actions/runs/26742269440)

burgerdev (approved):

PR #2426, opened 2026-06-01
cli/generate: add option to automatically calculate pod memory

davidweisse:


cli/cmd/generate.go
```diff
@@ -96,6 +96,7 @@ subcommands.`,
 	cmd.Flags().Bool("skip-service-mesh", false, "skip injection of Contrast service mesh sidecar")
 	cmd.Flags().Bool("inject-image-store", false, "inject an ephemeral storage device to pull images onto instead of into memory")
 	cmd.Flags().Bool("insecure-enable-debug-shell-access", false, "enable the debug shell service in the pod CVM to get access from container to guest VM")
+	cmd.Flags().Bool("calculate-pod-memory", false, "calculate pod memory based on image layer sizes and container resource limits")
```
burgerdev, line 99: Regarding our private discussion about the default here: right now, I think this feature should be off by default, as it's unlikely to play nice with existing pod resources. That means we'll probably want a new e2e test, or add a test to `imagestore`.

cli/cmd/generate.go
```diff
@@ -485,12 +486,37 @@ func generatePolicies(ctx context.Context, flags *generateFlags, fileMap map[str
 				meta.Annotations = make(map[string]string)
 			}
 			meta.Annotations[initdata.InitdataAnnotationKey] = initdataAnno
+			if flags.calculatePodMemory && !flags.injectImageStore {
```
burgerdev: This would probably need to respect the `image-store-size=0` annotation, too. https://docs.edgeless.systems/contrast/architecture/k8s-yaml-elements#overview-of-contrast-annotations

cli/genpolicy/genpolicy.go
```diff
@@ -52,12 +53,18 @@ func New(rulesPath, settingsPath, cachePath string, bin []byte) (*Runner, error)
...
+	if err != nil {
+		return "", nil, fmt.Errorf("creating pod layers file: %w", err)
+	}
+	defer os.Remove(podLayersFile.Name())
 	args := []string{
 		"--runtime-class-names=contrast-cc",
 		"--rego-rules-path=" + r.rulesPath,
 		"--json-settings-path=" + r.settingsPath,
 		"--layers-cache-file-path=" + r.cachePath,
+		"--pod-layers-file-path=" + podLayersFile.Name(),
```
burgerdev: Just wanted to mention that we need to count layers multiple times if they're used by different images, due to the coarse caching we apply in the imagepuller: https://github.com/edgelesssys/contrast/blob/f99be92aa64f2143285b07bb1b1e396cc9bf0720/imagepuller/internal/service/service.go#L75

burgerdev (commented):
Yep, looks like the right direction to me.

cli/cmd/generate.go
davidweisse: While using the flag, injected resources like initializer/debugshell/coordinator still have the predefined memory limits which is now way too high. Should these be changed dependent on the flag?

burgerdev: Yes, and I think that may be a good thing to extract into its own PR because it should also apply to `inject-image-store`.

cli/cmd/generate.go
```diff
@@ -485,12 +487,67 @@ func generatePolicies(ctx context.Context, flags *generateFlags, fileMap map[str
...
 			}
 			meta.Annotations[initdata.InitdataAnnotationKey] = initdataAnno
+
+			podMemory, err := calculatePodMemory(spec, layersCache)
+			if err != nil {
+				retError = fmt.Errorf("calculating pod memory: %w", err)
+				return meta, spec
+			}
+			imageStoreSize := meta.Annotations[imageStoreSizeAnnotationKey]
+			if flags.calculatePodMemory && (!flags.injectImageStore || imageStoreSize == "0") {
```
charludo: Since it returns error when layer cache info is not available, `calculatePodMemory` should probably only run conditionally on the if check below.

cli/genpolicy/genpolicy.go
```diff
@@ -88,12 +89,25 @@ func (r *Runner) Run(ctx context.Context, res any, extraPath string, logger *slo
...
 
-	return out.String(), err
+	layersCacheData, err := os.ReadFile(r.cachePath)
+	if err != nil {
+		return "", nil, fmt.Errorf("reading layers cache file: %w", err)
+	}
+	var layersCacheEntries []ImageLayersCacheEntry
+	if err := json.Unmarshal(layersCacheData, &layersCacheEntries); err != nil {
+		return "", nil, fmt.Errorf("decoding layers cache file: %w", err)
+	}
```
charludo: Similar here, this shouldn't error for users who don't use the feature.

cli/cmd/generate.go
```diff
@@ -43,6 +43,7 @@ const (
 	kataAnnotationPrefix          = "io.katacontainers.config.hypervisor."
 	contrastRoleAnnotationKey     = annotationPrefix + "pod-role"
 	workloadSecretIDAnnotationKey = annotationPrefix + "workload-secret-id"
+	imageStoreSizeAnnotationKey   = annotationPrefix + "image-store-size"
```
charludo: Please export and re-use the const from kuberesources' mutators.go

cli/cmd/generate.go
```diff
@@ -485,12 +487,67 @@ func generatePolicies(ctx context.Context, flags *generateFlags, fileMap map[str
 				meta.Annotations = make(map[string]string)
 			}
 			meta.Annotations[initdata.InitdataAnnotationKey] = initdataAnno
+
+			podMemory, err := calculatePodMemory(spec, layersCache)
+			if err != nil {
+				retError = fmt.Errorf("calculating pod memory: %w", err)
+				return meta, spec
```
charludo: Only the last err survives here. Maybe the entire func should be migrated from `MapPodSpecWithMeta` to `MapPodSpecWithMetaAndErrors`. If you don't want to this in this PR that's fine IMO, but please leave a TODO

cli/cmd/generate.go
```diff
@@ -485,12 +487,67 @@ func generatePolicies(ctx context.Context, flags *generateFlags, fileMap map[str
...
+						WithMemoryLimitAndRequest(2 * podMemory / 1024 / 1024),
+				)
+			}
 			return meta, spec
 		})
 		return res, retError
 	})
 }
 
+func calculatePodMemory(spec *applycorev1.PodSpecApplyConfiguration, podLayers *genpolicy.LayersCache) (int64, error) {
```
charludo, line 509: A unittest for this would be nice, just for pods with/without limits, sidebars, restartpolicy=always,... and the expected memory.

charludo (commented):
Nice, thanks! I have a couple of comments, but most are straightforward.

cli/cmd/generate.go
davidweisse: I think we should move these to a uniform location at some point. We have multiple duplicates in kuberesource/cli.

charludo: Yeah, that sounds sensible. Do you want to create a ticket?

charludo (approved):
Great, thank you! Marked all of my comments as resolved; I think @burgerdev's have also all been resolved?

davidweisse:
I messed up the condition in the last rebase. Should be correct now.

charludo (approved):

PR #2427, opened 2026-06-01
ci: cleanup old containerd backups

davidweisse:
Cleans up all `config.toml.*.bak` or `config.toml.tmpl.*.bak` in the containerd directory that are older than 5 days.

Workflow run:  https://github.com/edgelesssys/contrast/actions/runs/26757480960


burgerdev (approved):
Thanks!

PR #2429, opened 2026-06-03
flake: introduce special `+` syntax to combine sets at invocation time

charludo:
This is a suggestion on how to combine arbitrary sets. Any permutation of set names joined by `+` accesses the package set with all specified overlays applied. ~~To keep eval time down and to prevent accidental overrides, set names are internally sorted after splitting on `+`.~~


flake.nix
```diff
@@ -78,7 +67,47 @@
           (import ./overlays/runtimepkgs.nix)
         ];
 
-        sets = setsFromDirectory ./overlays/sets;
+        setsDir = ./overlays/sets;
```
sespiros: AI comment suggestion, feel free to keep or drop.

```suggestion
        # Sets are built from the overlays in ./overlays/sets. Beyond the single-file
        # sets, any combination of up to maxCombineDepth overlays is exposed under a
        # name joining the set names with "+". Each combination is evaluated once and
        # stored under its sorted (canonical) name; every permutation of the names is
        # registered as an alias of that same evaluation, so the order the user types
        # does not matter and does not cost an extra eval.
        setsDir = ./overlays/sets;
```

flake.nix
```diff
@@ -78,7 +67,47 @@
...
+        setNames = map (nixpkgs.lib.removeSuffix ".nix") (builtins.attrNames (builtins.readDir setsDir));
+
+        nonEmptySubsets =
```
sespiros: Add a comment on what this function does in a similar spirit to the one `setsFromDirectory` had. Same for the rest of the functions that build towards the final attribute set (`permutations`, `canonicalName`, `canonicalSets`, `sets`).

flake.nix
```diff
@@ -78,7 +67,47 @@
           (import ./overlays/runtimepkgs.nix)
         ];
 
-        sets = setsFromDirectory ./overlays/sets;
+        setsDir = ./overlays/sets;
+        setNames = map (nixpkgs.lib.removeSuffix ".nix") (builtins.attrNames (builtins.readDir setsDir));
```
sespiros: See my high level comment too but what do you think about excluding here sets starting with "_" to allow for parts of sets?

flake.nix
```diff
@@ -78,7 +67,47 @@
...
+            nixpkgs.lib.concatLists (
+              nixpkgs.lib.imap0 (
+                i: x: map (p: [ x ] ++ p) (permutations (nixpkgs.lib.take i list ++ nixpkgs.lib.drop (i + 1) list))
+              ) list
+            );
+
+        canonicalName =
+          subset: nixpkgs.lib.concatStringsSep "+" (nixpkgs.lib.sort builtins.lessThan subset);
+
+        subsets = builtins.filter (s: builtins.length s <= 3) (nonEmptySubsets setNames);
```
sespiros: Extract literal 3 as `maxCombineDepth` or equivalent.

flake.nix
```diff
@@ -78,7 +67,47 @@
...
+
+        canonicalSets = builtins.listToAttrs (
+          map (s: {
+            name = canonicalName s;
+            value = mkSet ((defaultOverlays (canonicalName s)) ++ map (n: import (setsDir + "/${n}.nix")) s);
```
sespiros: This is nitpicking/not sure if intentional but the overlays here are applied in the order of s but s comes from `nonEmptySets` which is in reverse alphabetical order. So applying `a+b` applies the overlays in reverse order.

sespiros (changes requested):
Thanks, this indeed looks technically neat and I am happy to merge since this is dev-only and also easily reversible, but I do have some concerns (that I tried to express in https://github.com/edgelesssys/contrast/pull/2405#issuecomment-4518862761) on how useful this added complexity will be. Could you tell me some concrete use cases you have in mind before we start building on this? Other than this high level comment/question, just some inline comments.

To share my thoughts a bit, in the concrete use cases I had so far (badaml tests), I am not sure if this composability would bring that much benefit other than simply taking the debug overrides out of the badaml sets. Most of the other overrides couldn't be split out in separate composable sets that would be useful to any other combination of sets. Could I split out `qemu-wrapped` too as a reusable set for instance? I could create a `qemu-acpi` set (that overrides `qemu-wrapped` with the `withACPITable` flag) and a `qemu-serial` set (for the `withSerialLog` flag). But then, does running `... .#badaml-vuln+debug+qemu-serial+qemu-acpi.foo` vs `... .#badaml-vuln.foo` make my life any easier than simply having a set which is monolithic and per use-case? That actually wouldn't even be possible due to the limit of 3 which I believe is there so that the possibilities don't explode. Moreover there are cross-dependencies, i.e supplying `withACPITable` to qemu without specifying `withACPIDebug` to kernel-uvm is not useful so now I need to be aware of that vs having a per-use case set in that folder. That's why in https://github.com/edgelesssys/contrast/pull/2405 I opted for this simple approach of prepending sets with `_` to indicate a part of a set. otoh we could have both (parts of sets and composable sets and only allow composing sets).

Other concrete upcoming use cases I see is the oak set and the runtime-rs set but will we want to compose these/is it meaningful to be able to compose all those different combinations? If it is, it seems much more intentional to me to see a set in that folder named `runtime-rs_oak` that indicates that we do care about that particular combination.

Then if it's only some flavor of debug that we want to attach in these per-use-case sets, I think that could be done in some simpler and more straightforward way.

charludo:
> Could you tell me some concrete use cases you have in mind before we start building on this? Other than this high level comment/question, just some inline comments.

Sure! The main goal is not to have to introduce inidividual set files for every new usecase. The `_`-prefixed set parts you proposed to allow you to reuse bits, but you still need to explicitly define every set.

We currently only really use the `base` and `debug` sets (apart from the CI also using the `badaml-` ones). Then, the `runtime-rs` set was recently introduced, and I assume we will get an `oak` set from your current project. 

With this PR, it's immediately and trivially possible to run deployments:
- in debug mode, with runtime-rs;
- in debug mode, using oak;
- with runtim-rs, using oak;
- in debug mode, with runtime-rs, using oak.

I assume we will want to have all of these combinations testable. 

(Also, just as an aside, while `runtime-rs`/`oak` might need special e.g. `debug` options, we can just add them to the shared `debug` set and they will have no effect, until used in the combined sets.)

To do the same with set parts, we'd have to extract the reusable parts, and then create 4 new files on top of it. IMO that becomes more of a headache to manage than the combination approach, but I can see how that's also individual preference.

> To share my thoughts a bit, in the concrete use cases I had so far (badaml tests), I am not sure if this composability would bring that much benefit other than simply taking the debug overrides out of the badaml sets. Most of the other overrides couldn't be split out in separate composable sets that would be useful to any other combination of sets. 

> Could I split out `qemu-wrapped` too as a reusable set for instance? I could create a `qemu-acpi` set (that overrides `qemu-wrapped` with the `withACPITable` flag) and a `qemu-serial` set (for the `withSerialLog` flag). But then, does running `... .#badaml-vuln+debug+qemu-serial+qemu-acpi.foo` vs `... .#badaml-vuln.foo` make my life any easier than simply having a set which is monolithic and per use-case? 

Why would you though? Seems to me there's not really a usecase for splitting out just `qemu-wrapped`. I'm not trying to advocate for decomposing all existing sets into their smallest components, but rahter to split sets along the line of usecase, and then be able to combine usecases.

For the `badaml-` case in particular, I'd say let's take the current `badaml-sandbox.nix`as-is, and change `badaml-vuln.nix` to the only content being:

```nix
  contrastPkgs = prev.contrastPkgs.overrideScope (badamlScope {
    withAMLSandbox = false;
  });
```

(copied from #2405).

(Admittedly... Here the alphabetic sorting thing rears its head. I actually think we should maybe drop it, and for set order to equal overlay apply order, than we can just keep the name `badaml-sandbox.nix` and gain more flexibility.)

> That actually wouldn't even be possible due to the limit of 3 which I believe is there so that the possibilities don't explode. 

It's there because this can not be an infinite set in nix, and the limit of 3 was chosen arbitrarily. We could easily bump it to 5, or (actually, that's probably the right choice: ) the number of sets in the `overlays/sets/` dir). 

> Moreover there are cross-dependencies, i.e supplying `withACPITable` to qemu without specifying `withACPIDebug` to kernel-uvm is not useful so now I need to be aware of that vs having a per-use case set in that folder. That's why in #2405 I opted for this simple approach of prepending sets with `_` to indicate a part of a set. otoh we could have both (parts of sets and composable sets and only allow composing sets).

Hope the first part is addressed by my comment above. The second part... IDK, I'm struggling to see what additional usefulness the parts bring. Do you have a concrete usecase for this, that can *not* be expressed through set combination?

> Other concrete upcoming use cases I see is the oak set and the runtime-rs set but will we want to compose these/is it meaningful to be able to compose all those different combinations? If it is, it seems much more intentional to me to see a set in that folder named `runtime-rs_oak` that indicates that we do care about that particular combination.

See above, I think this gets out of hand fast.

flake.nix
charludo: oops, that was not intentional. Fixed.

flake.nix
charludo: done

flake.nix
charludo: see discussion below

flake.nix
charludo: done

flake.nix
charludo: Don't know if `setsDir` is the right place for this. I rather commented the inidividual critical bits, so IMO no longer required.

charludo:
@sespiros Refactored, and added two new commits, one for removing the limitations discussed above, and one refactoring the `badaml-` sets to use the mechanism.

lib/sets.nix
```diff
@@ -0,0 +1,97 @@
...
+      # maxCombineDepth caps how many sets may be combined at once. A combination uses each set at most once, so the limit is the number of available sets.
+      maxCombineDepth = builtins.length setNames;
+      # mkOrderedSet builds the package set for a single ordered combination of set names by layering the corresponding set overlays on top of the default overlays.
+      mkOrderedSet =
+        p:
+        mkSet { inherit nixpkgs system; } (
+          (defaultOverlays { inherit self; } (lib.concatStringsSep "+" p))
+          ++ map (n: import (setsDir + "/${n}.nix")) p
+        );
+      combinations = builtins.filter (p: builtins.length p <= maxCombineDepth) (
```
sespiros: iiuc this filter never removes anything since an ordered combination can't be longer than the number of sets so

```suggestion
      combinations = orderedCombinations setNames
```

sespiros (approved):
Thanks (and thanks in particular for adding all those clarifying comments in the nix heavy parts :))

lib/sets.nix
charludo: Ah, thanks for catching that, that's indeed superfluous. Removed.

PR #2432, opened 2026-06-08
deps: update nix lock file

charludo:
Most pressingly, this addresses the `403`s we get from `crates.io` when vendoring cargo lock files.

Notable changes:
- Qemu 11
  - now forwards the host's CET capability to the guest TD, resulting in a changed XFAM vallue only on those hosts. We previously hardcoded this. I opted for now to mask CET out when validating XFAM, but I'm open to other suggestions... AFAICT we should also be able to disable this feature in QEMU, but since it's a hardening feature, it seemed best to allow it where possible
- Systemd 260


charludo:
- [x] [Metal-QEMU-SNP, regression](https://github.com/edgelesssys/contrast/actions/runs/27343481379)
- [x] [Metal-QEMU-TDX, regression](https://github.com/edgelesssys/contrast/actions/runs/27343482690)
- [x] [Metal-QEMU-SNP-GPU, regression](https://github.com/edgelesssys/contrast/actions/runs/27343484214)
- [x] [Metal-QEMU-TDX-GPU, regression](https://github.com/edgelesssys/contrast/actions/runs/27343485824)

e2e/openssl/openssl_test.go
```diff
@@ -364,6 +364,10 @@ func TestOpenSSL(t *testing.T) {
 			// (e.g. default kata commandline), MCE can't enable its logger.
 			// See https://www.firstfloor.org/~andi/mce.pdf for details on MCE.
 			"mce: Unable to init MCE device (rc: -5)",
+			// We actively mask systemd-networkd.service in the cmdline.
+			// Warning showing since systemd 260, from whereon statistics from services are reported via Varlink.
+			"systemd-networkd-varlink-metrics.socket: Socket service systemd-networkd.service not loaded, refusing.",
+			"Failed to listen on Network Management Metrics Varlink Socket.",
```
burgerdev: Shouldn't we just mask the `systemd-networkd-varlink-metrics.socket`, too?

internal/attestation/tdx/validator.go
```diff
@@ -166,6 +177,27 @@ func (t Report) ClaimsToCertExtension() ([]pkix.Extension, error) {
 	return claimsToCertExtension(t.Quote)
 }
 
+// xfamCETMask covers the CET XFAM bits CET_U (bit 11) and CET_S (bit 12).
+// Whether they are set depends on host CPU support.
+const xfamCETMask = uint64(1)<<11 | uint64(1)<<12
```
burgerdev, line 183:
```suggestion
// xfamCETMask covers the CET XFAM bits CET_U (bit 11) and CET_S (bit 12).
// Whether they are set depends on host CPU support.
// Source: Intel SDM Vol. 1 Sec. 13.1, https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html 
const xfamCETMask = uint64(1)<<11 | uint64(1)<<12
```

internal/attestation/tdx/validator.go
```diff
@@ -166,6 +177,27 @@ func (t Report) ClaimsToCertExtension() ([]pkix.Extension, error) {
...
+	if len(quoteXfam) != xfamSize {
+		return fmt.Errorf("quote XFAM has length %d, want %d", len(quoteXfam), xfamSize)
+	}
+	if len(refXfam) != xfamSize {
+		return fmt.Errorf("reference XFAM has length %d, want %d", len(refXfam), xfamSize)
+	}
+	got := binary.LittleEndian.Uint64(quoteXfam) &^ xfamCETMask
+	want := binary.LittleEndian.Uint64(refXfam) &^ xfamCETMask
+	if got != want {
+		return fmt.Errorf("XFAM mismatch: got %#016x, want %#016x (ignoring CET bits)", got, want)
```
burgerdev:
```suggestion
		return fmt.Errorf("XFAM mismatch: got %x, want %x (ignoring CET bits)", quoteXfam, refXfam)
```

e2e/openssl/openssl_test.go
charludo: That.... makes almost too much sense :D

charludo: Argh. This requires changing the kernel command line. Is this really worth a patch?
Edit: wait I think I might be a bit dense here.

charludo: Yeah indeed I was being dense 😆

burgerdev (approved):
Thanks!

PR #2434, opened 2026-06-08
docs: use Ubuntu 26.04 in bare-metal setup

thomasten:
Ubuntu 25.10 is nearing EOL. dgx-007 already runs with 26.04.


charludo (approved):
Thanks!

PR #2435, opened 2026-06-08
kata: stop shim cleanup from hanging on a dead kata-agent

sespiros:
For a cleaner diff:
https://github.com/kata-containers/kata-containers/compare/main...sespiros:kata-containers:shim-bound-cleanup

## To reproduce

On palutena, running the below on main leaks 4-6 sandboxes. With current branch it leaks none:

```bash
# Pre-test baseline
sudo bash << 'EOF'
for d in /run/containerd/io.containerd.runtime.v2.task/k8s.io/*/; do
  id=$(basename "$d")
  if pgrep -f "id $id" >/dev/null 2>&1; then echo LIVE; else echo LEAK; fi
done | sort | uniq -c
EOF

# Deploy + tear down
just e2e openssl
NS=$(kubectl get ns | grep "testopenssl-[a-f0-9]*-sse" | awk '{print $1}')
echo "test namespace: $NS"
just undeploy
# wait for $NS to disappear before measuring

# Count leaks attributed to the test namespace (filters out concurrent nightlies)
sudo bash << EOF
for d in /run/containerd/io.containerd.runtime.v2.task/k8s.io/*/; do
  id=\$(basename "\$d")
  if pgrep -f "id \$id" >/dev/null 2>&1; then continue; fi
  ns=\$(jq -r '.annotations."io.kubernetes.cri.sandbox-namespace"' "\$d/config.json" 2>/dev/null)
  mtime=\$(stat -c "%Y" "\$d")
  if [ "\$mtime" -gt "\$(date -d "1 hour ago" +%s)" ]; then
    echo "\$ns"
  fi
done | sort | uniq -c
EOF
```

### DIFF
--- a/packages/by-name/kata/runtime/0028-runtime-stop-shim-cleanup-from-hanging-on-a-dead-kat.patch
+++ b/packages/by-name/kata/runtime/0028-runtime-stop-shim-cleanup-from-hanging-on-a-dead-kat.patch
@@ -1,0 +1,163 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Spyros Seimenis <sse@edgeless.systems>
+Date: Tue, 9 Jun 2026 13:36:29 +0300
+Subject: [PATCH] runtime: stop shim cleanup from hanging on a dead kata-agent
+
+When the kata-agent becomes unreachable mid-teardown, agent-client
+calls hang in commonDialer for the full 30s dial_timeout, far beyond
+containerd's 5s per-shim cleanup deadline. Containerd SIGKILLs the
+cleanup binary, leaks the task dir, and the next containerd restart
+trips TimeoutStartUSec=90s and cascades through kubelet's
+Requires=containerd.service.
+
+The five fixes sit on one teardown chain. The shim's wait() goroutine
+must first finish sandbox.Stop/Delete against the dead guest:
+
+* agent/client commonDialer closes the cancel channel instead of
+  unbuffered send, so cancellation unblocks the caller.
+* agent.connect honors the caller's context deadline and forwards
+  it to the agent client.
+
+That lets sandbox.Delete tear down the sandbox's on-disk state.
+Containerd then drives task.Delete and the cleanup binary, which
+must fit in its 5s window:
+
+* shim.Delete honors caller's context and defaults to 30s.
+* shim.cleanupContainer defaults to 3s.
+
+Finally the cleanup binary calls CleanupContainer on the bundle and
+must not treat the already-removed sandbox state as an error:
+
+* virtcontainers.CleanupContainer with force=true treats an
+  already-gone sandbox as success.
+
+Signed-off-by: Spyros Seimenis <sse@edgeless.systems>
+---
+ src/runtime/pkg/containerd-shim-v2/service.go |  7 ++++++-
+ src/runtime/pkg/containerd-shim-v2/utils.go   | 17 +++++++++++++++--
+ src/runtime/virtcontainers/api.go             |  8 ++++++++
+ src/runtime/virtcontainers/kata_agent.go      | 19 ++++++++++++++++++-
+ .../pkg/agent/protocols/client/client.go      |  4 +++-
+ 5 files changed, 50 insertions(+), 5 deletions(-)
+
+diff --git a/src/runtime/pkg/containerd-shim-v2/service.go b/src/runtime/pkg/containerd-shim-v2/service.go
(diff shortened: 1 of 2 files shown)


packages/by-name/kata/runtime/0028-runtime-stop-shim-cleanup-from-hanging-on-a-dead-kat.patch
```diff
@@ -0,0 +1,135 @@
...
++			s := uint32((remaining + time.Second - 1) / time.Second)
++			if dialTimout == 0 || s < dialTimout {
++				dialTimout = s
++			}
++		}
++	}
++
+ 	k.Logger().WithField("url", k.state.URL).Info("New client")
+-	client, err := kataclient.NewAgentClient(k.ctx, k.state.URL, k.dialTimout)
++	client, err := kataclient.NewAgentClient(k.ctx, k.state.URL, dialTimout)
```
charludo: Should this maybe be `ctx` instead of `k.ctx`? If `ctx` is cancelled, this won't be if I understand correctly. But tbh I'm also not sure if it matters, since you're checking `ctx.Err` just before as well.

packages/by-name/kata/runtime/0028-runtime-stop-shim-cleanup-from-hanging-on-a-dead-kat.patch
```diff
@@ -0,0 +1,135 @@
...
+-		return err
++	if _, hasDeadline := ctx.Deadline(); !hasDeadline {
++		var cancel context.CancelFunc
++		ctx, cancel = context.WithTimeout(ctx, cleanupBinaryTimeout)
++		defer cancel()
++	}
++
++	if err := vci.CleanupContainer(ctx, sandboxID, cid, true); err != nil {
++		shimLog.WithError(err).WithField("container", cid).Warn("failed to cleanup container; proceeding with rootfs unmount")
+ 	}
```
charludo, line 91: I know that not erroring here is the point, but can we find a way to return the error *eventually*? Like with a defer or something. Doesn't feel right that this is completely swallowed.

charludo (approved):
LGTM. Thanks for figuring this out. I have two small things, but they are in no way critical. If you want to, just go ahead and merge today.

packages/by-name/kata/runtime/0028-runtime-stop-shim-cleanup-from-hanging-on-a-dead-kat.patch
sespiros: The context here is actually completely irrelevant (it is never used inside the function because `commonDialer` uses a new timeout for its deadline) but pushed a fixup so that this at least reads more nicely.

packages/by-name/kata/runtime/0028-runtime-stop-shim-cleanup-from-hanging-on-a-dead-kat.patch
sespiros: Thanks for pointing this out. The intention is indeed to swallow that error here, that's the point but I took a closer look inside the implementation of `CleanupContainer` and it seems the implementation is already doing that for some errors when force is passed. So I pushed a fix where I don't modify the code here at all but instead modify the implementation side so that when force is supplied, more errors are ignored (the ones relevant to the sandbox folder not being present any more) so that we allow containerd to proceed with cleanup.

packages/by-name/kata/runtime/0028-runtime-stop-shim-cleanup-from-hanging-on-a-dead-kat.patch
charludo: Ah, OK. Thanks for the explanation!

packages/by-name/kata/runtime/0028-runtime-stop-shim-cleanup-from-hanging-on-a-dead-kat.patch
charludo: OK interesting. Yeah that seems like a good way forward - I assume you've already ran this and the fix still works? 😅

sespiros: Yes the reproducer works the same. I think if this works for us, I could propose it upstream but I would want to more closely look at these errors on the implementation side before doing so (and why they originally opted for ignoring them). But then also maybe it's not worth it after the switch to runtime-rs. I will have to test this again when we make the switch.

charludo (approved):
Thanks (again! :D) for taking the time!

PR #2436, opened 2026-06-09
ci: remove fifo server

burgerdev:
This replaces all usage of the server-based fifo and replaces it with Kubernetes leases.


charludo:
@burgerdev the cleanup error on SNP-GPU seems genuine and related

packages/scripts.nix
```diff
@@ -623,42 +612,26 @@ lib.makeScope pkgs.newScope (scripts: {
...
     text = ''
       echo "Requesting fifo ticket from sync server" >&2
-      sync_ip=$(kubectl get svc sync -n default -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
-      sync_uuid=$(kubectl get configmap sync-server-fifo -n default -o jsonpath='{.data.uuid}')
-      path="ticket"
-      if [[ "$#" -ge 1 ]]; then
-        path="ticket?done_timeout=$1"
+      timeout_flag=()
+      if [[ "$#" -ge 1 && -n "$1" ]]; then
+        timeout_flag=(--timeout "$1")
```
charludo: `acquire` has both a timeout and a duration arg, would be nice to be able to set both here (if we're already able to customize one of them)

packages/scripts.nix
```diff
@@ -623,42 +612,26 @@ lib.makeScope pkgs.newScope (scripts: {
...
       fi
-      sync_ticket=$(curl -fsSL "$sync_ip:8080/fifo/$sync_uuid/$path" | jq -r '.ticket')
-      echo "Waiting for lock on fifo $sync_uuid with ticket $sync_ticket" >&2
-      curl -fsSL "$sync_ip:8080/fifo/$sync_uuid/wait/$sync_ticket"
-      echo "Acquired lock on fifo $sync_uuid with ticket $sync_ticket" >&2
-      echo "$sync_ticket"
+      fifo acquire "''${timeout_flag[@]}"
```
charludo: Also, the `request-fifo-ticket` just recipe is called with 90m in the justfile. The `get-sync-ticket` script turns that into `fifo acquire --timeout 90m`, leaving the duration of the claim at 60m. I assume that's intentional? But it does mean that an actual long-running GPU task will have its ticket stolen by the next job that comes along.

Yeah the more I think about this the less of an issue this is.

packages/scripts.nix
```diff
@@ -623,42 +612,26 @@ lib.makeScope pkgs.newScope (scripts: {
...
   get-sync-ticket = writeShellApplication {
     name = "get-sync-ticket";
-    runtimeInputs = with pkgs; [
-      kubectl
-      jq
-      curl
-    ];
+    runtimeInputs = [ contrastPkgs.fifo ];
     text = ''
       echo "Requesting fifo ticket from sync server" >&2
```
charludo:
```suggestion
      echo "Requesting fifo ticket" >&2
```

charludo (approved):
Thanks! To a never-again locked GPU :D

packages/scripts.nix
burgerdev: Good catch, I believe we're passing the wrong argument here: that should be the waiting time, not the lease holding time. Let me just pass the arguments through unmodified and adjust the callsites.

packages/scripts.nix
burgerdev: The overall idea is to keep the waiting time a bit longer than the holding time, as evident in the default

```
--timeout 65m --duration 60m
```

We're marking the expected holding time as 90m such that the cleanup job will remove the namespace and release the lease before it expires.

I believe that, with the old sync server, waiting time was not capped. However, I'm not sure whether this is actually a good idea.

PR #2437, opened 2026-06-09
kata: propagate (some) early VM errors to k8s

burgerdev:
This allows error messages to propagate to the RPC caller (i.e., containerd). 

As an example, I removed the `initdata` annotation from a Contrast pod. Before this change, we saw messages like this in `kubectl describe`:

```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to start sandbox "f0e9f9deec7732c28b8f69f9ba6e09fdf7a59da04701dff0f12f74dbfe8aa76a": failed to create containerd task: failed to create shim task: create container timeout: f0e9f9deec7732c28b8f69f9ba6e09fdf7a59da04701dff0f12f74dbfe8aa76a
```

After the change, the error messages show correctly:

```
Failed to create pod sandbox: rpc error: code = PermissionDenied desc = failed to start sandbox "29130d147c4784857969ea65cc2f294a2386367dbdf55b5674a88ddda9d20833": failed to create containerd task: failed to create shim task: "UpdateInterfaceRequest is blocked by policy: /run/measured-cfg/policy.rego:84: Internal error: no initdata device found"
```

---

See the patch commit message for details on what we're doing in Kata, and why it's ok. This PR is probably insufficient to deal with all sorts of VM creation failures. For example, it's unlikely to help if the VM suddenly crashed, or if the agent never started.

Fixes CON-27.

### DIFF
--- a/e2e/internal/kubeclient/deploy.go
+++ b/e2e/internal/kubeclient/deploy.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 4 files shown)


e2e/internal/kubeclient/deploy.go
sespiros: Suggestion to rename to `IsBlockedByMissingInitdata` or equivalent as it is now seems to be checking for this specific error.

sespiros (approved):
Are we going to upstream any of this or are we going to drop it entirely after the switch to runtime-rs? If yes maybe we should split the patch in 2 to have the general improvements as a separate patch.

burgerdev:
> Are we going to upstream any of this or are we going to drop it entirely after the switch to runtime-rs? If yes maybe we should split the patch in 2 to have the general improvements as a separate patch.

I'd rather not. I haven't tested this in other configurations (dynamic resource management, cgroupv1, no systemd) and there's an argument for not touching this in the soon-deprecated runtime-go. That even goes for the error checking mistakes, which may have become load-bearing by accident.

PR #2438, opened 2026-06-09
kubeclient: simplify waiting logic

burgerdev:
The notification-based waiting logic was implemented when we did not know about the wait package yet. Since we saw a couple of CI flakes that can be explained by missed notifications (pod was ready, but test logic said it wasn't), moving to a polling model might be more robust, albeit a bit wasteful. In any case, the polling logic is much easier to reason about.

Fixes CON-33.

### DIFF
--- a/e2e/internal/kubeclient/wait.go
+++ b/e2e/internal/kubeclient/wait.go
@@ -13,10 +13,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/informers"
-	listerscorev1 "k8s.io/client-go/listers/core/v1"
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/klog/v2"
 )
 
 // WaitForDeployment waits until the Deployment is ready.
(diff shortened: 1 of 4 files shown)


e2e/internal/kubeclient/wait.go
```diff
@@ -151,68 +147,55 @@ type PodCondition interface {
...
-		select {
-		case <-notifications:
-			done, err := podCondition.Check(podLister)
+	poll := func(ctx context.Context) (done bool, err error) {
+		return podCondition.Check(ListerFunc(func(selector labels.Selector) (ret []*corev1.Pod, err error) {
+			podList, err := c.Client.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: selector.String()})
 			if err != nil {
-				logger.Warn("error checking pods", "error", err)
-				continue
+				// TODO(burgerdev): should we just log and return false here?
```
davidweisse: That sounds correct to me. Context errors will be caught by the calling `PollUntilContextCancel` function anyway.

e2e/internal/kubeclient/wait.go
```diff
@@ -151,68 +147,55 @@ type PodCondition interface {
...
+				out = append(out, &podList.Items[i])
 			}
-		case <-ctx.Done():
-			break loop
-		}
+			return out, nil
+		}))
+	}
+
+	err := wait.PollUntilContextCancel(ctx, 2*time.Second /*immediate*/, true, poll)
```
davidweisse:
```suggestion
	err := wait.PollUntilContextCancel(ctx, 2*time.Second, /*immediate*/ true, poll)
```
Is the inline comment really necessary though? We have many instances of this and you could just use inlay hints.

e2e/internal/kubeclient/wait.go
burgerdev: Changed accordingly.

e2e/internal/kubeclient/wait.go
burgerdev: Removed

e2e/internal/kubeclient/wait.go
```diff
@@ -151,68 +147,55 @@ type PodCondition interface {
...
+			return out, nil
+		}))
+	}
+
+	err := wait.PollUntilContextCancel(ctx, 2*time.Second, true, poll)
+	if err == nil {
+		return nil
 	}
 
 	logger.Debug("context expired while waiting", "condition", podCondition)
```
davidweisse:
```suggestion
	logger.Debug("context expired while waiting")
```

davidweisse (approved):

davidweisse (approved):

PR #2439, opened 2026-06-10
just: rename set target to avoid keyword clash

burgerdev:
Not sure if I'm using something else than others, but my just LSP recognizes `set` as a reserved keyword and renders the entire second half of the justfile as a syntax error. Seems to not be an issue for just itself, though.

### DIFF
--- a/justfile
+++ b/justfile
@@ -5,7 +5,7 @@ default_set := "${set}"
 workspace_dir := "workspace"
 
 # Undeploy, rebuild, deploy.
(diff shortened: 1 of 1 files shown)


charludo (approved):
WAIT THAT'S THE REASON?!

I don't have an LSP for just, (treesitter)-highlighting just always... stopped halfway through the file. I also remember Paul at one point looking at my monitor and going "haha, justfile highlighting is broken for you too, huh".

Very cool, thanks for fixing this :D

PR #2440, opened 2026-06-10
kata: build incrementally using crate2nix

charludo:
Should probably merge this instead of #2383.


charludo:
Build times after touching a patch relevant to the given bin:
```
     agent:    1m 08s
 genpolicy:    1m 46s
runtime-rs:    1m 59s
```
(on my machine, of course, and including compiling and running tests)

Touching e.g. an agent patch does not require rebuilds of the other two packages.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
```diff
@@ -0,0 +1,47 @@
...
+set -euo pipefail
+
+repo=$(git rev-parse --show-toplevel)
+out="$repo/packages/by-name/kata/source/Cargo.nix"
+
+echo "Building patched kata source" >&2
+src=$(cd "$repo" && nix build --no-link --print-out-paths .#base.kata.source.src)
+
+workdir=$(mktemp -d)
+trap 'rm -rf "$workdir"' EXIT
```
burgerdev, line 15: I believe the build already sets `TMPDIR` to something cleaned up by the daemon.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
```diff
@@ -0,0 +1,47 @@
...
+# Copyright 2026 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+set -euo pipefail
+
+repo=$(git rev-parse --show-toplevel)
+out="$repo/packages/by-name/kata/source/Cargo.nix"
+
+echo "Building patched kata source" >&2
+src=$(cd "$repo" && nix build --no-link --print-out-paths .#base.kata.source.src)
```
burgerdev: Shouldn't `nix build` work from anywhere in the repo?

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
```diff
@@ -0,0 +1,47 @@
...
+
+workdir=$(mktemp -d)
+trap 'rm -rf "$workdir"' EXIT
+cp -r --no-preserve=mode,ownership "$src"/. "$workdir/"
+cd "$workdir"
+
+echo "Running crate2nix generate" >&2
+crate2nix generate -f Cargo.toml -o Cargo.nix
```
burgerdev, line 24: `crate2nix` produces a `crate-hashes.json` file. Don't we need this later on?

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# crate2nix hardcodes crate sources as relative paths, rewrite them to resolve against workspaceSrc
+sed -i \
+  -e 's|^{ nixpkgs ? <nixpkgs>$|{ workspaceSrc\n, nixpkgs ? <nixpkgs>|' \
+  -e 's|src = \./|src = workspaceSrc + "/|' \
+  -e 's|src = workspaceSrc + "/\([^;]*\);|src = workspaceSrc + "/\1";|' \
+  Cargo.nix
```
burgerdev: Why can't we use relative paths? Sounds like the more portable thing to do.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
```diff
@@ -0,0 +1,47 @@
...
+  /^      "registry\+https:\/\/github.com\/rust-lang\/crates.io-index#safe-path@0.1.0" = rec \{$/ { skip = 1 }
+  skip { if ($0 == "      };") skip = 0; next }
+  { print }
+' Cargo.nix >Cargo.nix.tmp && mv Cargo.nix.tmp Cargo.nix
+sed -i \
+  's|registry+https://github.com/rust-lang/crates.io-index#safe-path@0.1.0|path+file:///tmp/c2n-gen/src/libs/safe-path#0.1.0|g' \
+  Cargo.nix
+
+# crate2nix' test-runner derivation writes its log directly to $out via tee, so the default installPhase which mkdirs $out collides
+sed -i '/buildInputs = testInputs;/a\            dontInstall = true;' Cargo.nix
```
burgerdev, line 36: What does the `a\` do here?

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
```diff
@@ -0,0 +1,47 @@
...
+# crate2nix hardcodes crate sources as relative paths, rewrite them to resolve against workspaceSrc
+sed -i \
+  -e 's|^{ nixpkgs ? <nixpkgs>$|{ workspaceSrc\n, nixpkgs ? <nixpkgs>|' \
+  -e 's|src = \./|src = workspaceSrc + "/|' \
+  -e 's|src = workspaceSrc + "/\([^;]*\);|src = workspaceSrc + "/\1";|' \
+  Cargo.nix
+
+# crate2nix includes the absolute generation directory into safe-path's package id.
+# Normalize it to a stable placeholder so regenerations don't produce diff.
+sed -i "s|$workdir|/tmp/c2n-gen|g" Cargo.nix
```
burgerdev: Any chance this can go when we fixed the double `safe-path`? At least, it seems to be only happening for that crate, not the other local ones.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
charludo: This doesn't run in a build sandbox unfortunately, but directly on the host. I don't think the nix daemon will now about this dir.

(The build of the kata packages itself it of course sandboxed, but we need network connectivity to generate the Cargo.nix)

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
charludo: Yeah that makes no sense

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
charludo: No, just needed at generation time. We can still commit it though, this saves us the 5 lookups of the git dependencies we previously hand-maintained in nix code.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
charludo: Unfortunately not. Amended the comment. The relative paths are relative to the generation dir, i.e. some nix store path. At build time, crate2nix tries to look these up in the build dir which fails.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
charludo: Appends a line after match. commented.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
charludo: Yep indeed, leftover. Removed it, thanks for noticing.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
burgerdev: Thanks, forgot about that.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
burgerdev: Ah, this would only work if the `Cargo.nix` file was in the Kata source? Would it be less hacky to add it there?

charludo: Yeah, the idea behind the file is to have it committed into the actual project, not some downstream consumer.
 
> less hacky

I mean... If you think we can get kata upstream to commit this, maybe together with a flake building it all... Yeah, sure.

But instead adding it as a patch seems like a nightmare to maintain/update.

packages/by-name/scripts/update-kata-cargo-nix/update-kata-cargo-nix.sh
charludo: Committed it, script completes slightly faster and we have an easy-to-find reference for the hashes 🤷🏼‍♀️

burgerdev (approved):

PR #2443, opened 2026-06-11
.github: bump sev-snp-measure-consistency timeout

charludo:
I don't think this completed in-time even once 😄


charludo:
Alternatively, we could reduce this from 1..220 CPUs checked to something like 1..16.... not sure if going to the max is thaaaaat beneficial.

burgerdev (approved):

PR #2444, opened 2026-06-11
coordinator: include HTTP API in shutdown process

burgerdev:
We forgot to add the HTTP API server to the list of things to shut down, which prevented the Coordinator from terminating after receiving a signal. Add a shutdown command and rename the HTTP servers so that there are no ambiguities.

Fixes CON-56

### DIFF
--- a/coordinator/main.go
+++ b/coordinator/main.go
@@ -137,7 +137,8 @@ func run() (retErr error) {
 	meshapi.RegisterMeshAPIServer(meshAPIServer, meshapiserver.New(logger))
 	serverMetrics.InitializeMetrics(meshAPIServer)
 
-	httpServer := &http.Server{}
+	metricsServer := &http.Server{}
+	httpAPIServer := &http.Server{}
 
 	var userapiStarted, meshapiStarted, recoveryStarted atomic.Bool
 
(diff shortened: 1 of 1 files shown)


coordinator/main.go
```diff
@@ -280,7 +280,8 @@ func run() (retErr error) {
 		gracefulStopGRPC(ctx, wg, meshAPIServer) //nolint:contextcheck
 		wg.Wait()
 		_ = transitAPIServer.Shutdown(ctx) //nolint:contextcheck
-		return httpServer.Shutdown(ctx)    //nolint:contextcheck
+		_ = httpAPIServer.Shutdown(ctx)    //nolint:contextcheck
+		return metricsServer.Shutdown(ctx) //nolint:contextcheck
```
davidweisse: Not sure, but this looks a bit weird. Should this be
```go
return errors.Join(
	transitAPIServer.Shutdown(ctx), //nolint:contextcheck
	httpAPIServer.Shutdown(ctx),    //nolint:contextcheck
	metricsServer.Shutdown(ctx),    //nolint:contextcheck
)
```

davidweisse (approved):
Otherwise lgtm

coordinator/main.go
burgerdev: Thanks, makes perfect sense. I also modified the cleanup context so that we don't need `nolint` anymore.

davidweisse (approved):

PR #2445, opened 2026-06-11
package: build qemu without GUI backends

charludo:
So maybe `boot-microvm` does not *absolutely need* gtk, alsa, pulse, pipewire, jack, vnc, ... 🙃


sespiros (commented):
~~Are the packages mentioned the only ones that get removed in qemu's minimal build? How did you test that nothing else breaks?~~ I just understood the spirit of the PR, we don't use this `boot-microvm` anywhere it just costs us time when we evaluate everything.

sespiros (approved):

sespiros (commented):
I don't use these but we may need to add under `-nographic` in both invocations otherwise qemu will fail to boot:

```bash
      -vga none \
      -nic none \
```

or do
```bash
  runtimeInputs = [ (qemu.override { minimal = true; enableBlobs = true; }) ];
```

charludo:
So... should we just remove both instead?

burgerdev (approved):

PR #2446, opened 2026-06-11
ci: wait for contrast-system deletion in release e2e cleanup

sespiros:
Same-cluster matrix entries are serialized, but the namespace delete finalizes asynchronously, so the next entry could race a still-Terminating contrast-system. Block until it's gone.

Fix for  https://github.com/edgelesssys/contrast/actions/runs/27235297604/job/80462312665

### DIFF
--- a/.github/actions/release_e2e_test/action.yml
+++ b/.github/actions/release_e2e_test/action.yml
@@ -167,6 +167,11 @@ runs:
       shell: bash
       run: |
         kubectl delete -f workspace/log-collector.yaml
+        # Wait for contrast-system to be fully gone so the next same-cluster
+        # matrix entry doesn't race this job's async namespace teardown.
+        if kubectl get namespace contrast-system >/dev/null 2>&1; then
+          kubectl wait --for=delete namespace/contrast-system --timeout=3m
+        fi
         rm -f workspace/just.namespace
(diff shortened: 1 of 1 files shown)


.github/actions/release_e2e_test/action.yml
```diff
@@ -167,6 +167,11 @@ runs:
       shell: bash
       run: |
         kubectl delete -f workspace/log-collector.yaml
+        # Wait for contrast-system to be fully gone so the next same-cluster
+        # matrix entry doesn't race this job's async namespace teardown.
+        if kubectl get namespace contrast-system >/dev/null 2>&1; then
+          kubectl wait --for=delete namespace/contrast-system --timeout=3m || true
```
burgerdev: What's the idea behind the `|| true`? If we failed here, we'd have a good indicator of what job botched the namespace so that couldn't be cleaned up.

sespiros: Oversight, AI seems to like adding these `|| true` for commands in actions where the failure is incidental to a step (and benign). In this case we do want to fail if that steps fails even after its 3m timeout. Pushed a fixup.

burgerdev (approved):
👍

PR #2447, opened 2026-06-11
e2e: retry demo request until the LB endpoint serves

sespiros:
WaitForService returns once the Service has an ingress IP, not once the Scaleway LoadBalancer backend marks the just-started web pod healthy, so the single demo GET could race the readiness window and fail with EOF. Retry the request until it serves a response.

Fix for https://github.com/edgelesssys/contrast/actions/runs/27235297604/job/80462312700


burgerdev (approved):
Thanks!

PR #2451, opened 2026-06-15
e2e: fix initContainerRunningCondition

burgerdev:
The condition should obviously check the iteration's current pod, not the first one in the list.


charludo (approved):
:D

PR #2452, opened 2026-06-15
e2e/release: restrict node-installers to their intended nodes

charludo:
Mirrors what the "normal" e2es are doing with `PatchNodeInstallers`. Not reused sadly, because it requires typed args, and converting the yaml into something typed and then back goes against the intention of using the "raw" released yaml.

### DIFF
--- a/internal/kuberesource/mutators.go
+++ b/internal/kuberesource/mutators.go
@@ -32,6 +32,12 @@ const (
 	securePVAnnotationKey         = "contrast.edgeless.systems/secure-pv"
 	workloadSecretIDAnnotationKey = "contrast.edgeless.systems/workload-secret-id"
 	imageStoreSizeAnnotationKey   = "contrast.edgeless.systems/image-store-size"
+
+	// TDXEnabledNodeLabel is the node-feature-discovery label that marks a node as TDX-capable.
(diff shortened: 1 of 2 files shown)


internal/kuberesource/mutators.go
```diff
@@ -32,6 +32,12 @@ const (
 	securePVAnnotationKey         = "contrast.edgeless.systems/secure-pv"
 	workloadSecretIDAnnotationKey = "contrast.edgeless.systems/workload-secret-id"
 	imageStoreSizeAnnotationKey   = "contrast.edgeless.systems/image-store-size"
+
+	// TDXEnabledNodeLabel is the node-feature-discovery label that marks a node as TDX-capable.
+	TDXEnabledNodeLabel = "feature.node.kubernetes.io/tdx.enabled"
```
burgerdev, line 37: There's a similar label for SNP nodes, not sure why we're not using that one, too. We don't need to add it in this PR, though.

burgerdev (approved):

internal/kuberesource/mutators.go
charludo: Hm, probably just using this to differentiate which node to run on in the dgx cluster?

PR #2453, opened 2026-06-15
runtime-rs: add missing kernel parameters

burgerdev:
... and add a derivation for manually building the expected command line.
... and add a basic test to run on Kata changes.


charludo (approved):
LGTM 🎉

PR #2454, opened 2026-06-16
imagepuller: reclaim storage from unreferenced containers

charludo:
Multiple restarts of a container could lead to the pod VM storage filling up, esp. if the container wrote to disk during its lifetime.
While the imagepuller does reuse cached images, the image store did not clean up the store container's unreferenced layers.
This is now fixed, and the imagepuller cleans the store of unreferenced containers before creating a new container.

Reproducer:
1. run `just` on main
2. run 
    ```
    n=0
    while [ "$n" -lt 8 ]; do
      id=$(crictl ps -q --name openssl-backend | head -n1)
      [ -z "$id" ] && { sleep 2; continue; }
      echo "round $n: $id"
      crictl exec "$id" sh -c 'dd if=/dev/zero of=/leak.bin bs=1M count=200' || echo "  write failed (store full?)"
      crictl stop "$id" >/dev/null 2>&1 || true
      n=$((n+1))
      # wait for the replacement container before the next round
      for _ in $(seq 1 30); do
        new=$(crictl ps -q --name openssl-backend | head -n1)
        [ -n "$new" ] && [ "$new" != "$id" ] && break
        sleep 1
      done
    done
    ```
    from the node
3. see that eventually, new containers fail to come up due to lack of storage
4. do the same on this branch, containers continue to come up and the journalctl log for the imagepuller contains cleanup messages

### DIFF
--- a/imagepuller/internal/service/mount_linux.go
+++ b/imagepuller/internal/service/mount_linux.go
@@ -15,7 +15,10 @@ import (
 )
 
 func (s *ImagePullerService) createAndMountContainer(log *slog.Logger, imageID, bundlePath string) (string, error) {
-	container, err := s.Store.CreateContainer("", nil, imageID, "", "", nil)
+	rootfs := filepath.Join(bundlePath, "rootfs")
+
+	// Store the rootfs path as the container's metadata so that orphaned containers can be reclaimed later (see cleanupOrphanedContainers)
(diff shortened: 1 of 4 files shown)


burgerdev (approved):
Thanks, good find!

daniel-weisse (approved):

PR #2457, opened 2026-06-16
initdata-processor: don't rely on magic udev links

burgerdev:
Try detecting initdata and imagepuller devices by looking at the `/dev/disk/by-id/virtio-*` path AND by scanning all other block devices for the expected magic number.

This allows us to stay closer to upstream, which implements a similar scanning method.

Fixes CON-159.


initdata-processor/main.go
```diff
@@ -104,18 +106,66 @@ func handleImagepullerAuthConfig(doc initdata.Raw) error {
...
+		case major == 8:
+			// SCSI / SATA
+		case major == 3:
+			// IDE / PATA
+		case 240 <= major && major <= 254:
+			// Dynamic, probably virtio.
+		default:
+			// Probably not a disk.
+			continue
+		}
```
charludo, line 158: Does this mirror what kata does for discovery, or is it specific to us?

In the former case, a link would be nice, in the latter, we should at least log the seen and discarded devices to be able to check why a device was not found, if a device is ever not found.

initdata-processor/main.go
```diff
@@ -35,7 +37,7 @@ func main() {
 		failf("Could not create directory %q: %v", measuredConfigPath, err)
 		return
 	}
-	device, err := checkDeviceAvailability("initdata")
+	device, err := checkDeviceAvailability("initdata", []byte("initdata"))
 	if err != nil {
 		failf("no initdata device found")
 		return
```
charludo, line 43: The new, more descriptive errors from `checkDeviceAvailability`  are being swallowed

charludo (commented):
Thanks. Two small things, plus: do we even need the by-id path anymore?

initdata-processor/main.go
burgerdev: The implementation is in progress and happening here: https://github.com/kata-containers/kata-containers/pull/11655/changes#diff-e35f61faa38539c49d18a25597e2ddc3e61b031540e46486aa1be3920b18db65. I diverged from what is currently there, but intend to update upstream to this mode of operation. Let me link the PR in lieu of something stable.

Regarding the logging: I was afraid to pollute the logs with a bunch of clearly unrelated devices, but this case should be rare enough to justify it.

initdata-processor/main.go
burgerdev: Fixed.

initdata-processor/main.go
charludo: Great, thanks ,that's the context I was looking for :D 

If you want to, you could only log the list of found and discarded devices when we found no suitable device. But IMO it's fine as-is 🤷🏼‍♀️

charludo (approved):
Thanks!

PR #2459, opened 2026-06-17
e2e: test pod memory flag in imagestore test

davidweisse:
Tests, if the accounted memory is enough to pull a large image.

I couldn't run the imagestore test on the dev machine, but this should hopefully work: https://github.com/edgelesssys/contrast/actions/runs/28096059141/job/83185751294

Fixes CON-142

### DIFF
--- a/e2e/imagestore/imagestore_test.go
+++ b/e2e/imagestore/imagestore_test.go
@@ -59,13 +59,17 @@ func TestImageStore(t *testing.T) {
 		resources = append(resources, testPod(tc.name, tc.annotation))
 	}
 
+	resources = append(resources, tensorflowPod())
+
 	resources = kuberesource.PatchRuntimeHandlers(resources, runtimeHandler)
 	resources = kuberesource.AddPortForwarders(resources)
 	resources = kuberesource.AddImageStore(resources)
 
 	ct.Init(t, resources)
 
-	require.True(t.Run("generate", ct.Generate), "contrast generate needs to succeed for subsequent tests")
+	require.True(t.Run("generate", func(t *testing.T) {
+		require.NoError(ct.RunGenerate(t.Context(), "--inject-image-store", "--calculate-pod-memory"))
+	}), "contrast generate needs to succeed for subsequent tests")
 	require.True(t.Run("apply", ct.Apply), "Kubernetes resources need to be applied for subsequent tests")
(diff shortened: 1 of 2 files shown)


e2e/imagestore/imagestore_test.go
```diff
@@ -143,6 +157,19 @@ func TestImageStore(t *testing.T) {
...
+		ctx, cancel := context.WithTimeout(t.Context(), ct.FactorPlatformTimeout(5*time.Minute))
+		t.Cleanup(cancel)
+
+		require = req.New(t)
+
+		// The container is equipped with a very low memory limit, not nearly enough to pull the ~7Gi image.
+		// This test verifies that the added pod memory limit during generate is enough to pull the image
+		// when the imagestore is disbabled.
+		require.NoError(ct.Kubeclient.WaitForPod(ctx, ct.Namespace, "tensorflow"))
+	})
```
burgerdev, line 174: Looks like something is missing here.

e2e/imagestore/imagestore_test.go
```diff
@@ -84,6 +88,11 @@ func TestImageStore(t *testing.T) {
 			var expectedKibiBytes int
 			switch tc.annotation {
 			case "0":
+				require.NotNil(pod.Spec.Resources)
+				require.NotNil(pod.Spec.Resources.Limits)
+				require.NotNil(pod.Spec.Resources.Limits.Memory())
```
burgerdev, line 93: Should we test in the other cases that pod resources are not set?

e2e/imagestore/imagestore_test.go
davidweisse: This only checks if the pod comes up. If the pod resource limits are not correctly set, the pod shouldn't have enough memory to pull the image, so the test would fail.

e2e/imagestore/imagestore_test.go
davidweisse: I thought this didn't work because we are manually adding the image stores via `AddImageStore` since we have extra logic for overriding the storage class in `contrasttest`. But adding the `--inject-image-store` flag seems to work fine, since already present image stores are not touched. With the flag we can test it.

burgerdev (approved):
Thanks, lgtm!

burgerdev:
* [x] [imagestore @ palutena](https://github.com/edgelesssys/contrast/actions/runs/28178262416)

PR #2460, opened 2026-06-18
ci: retry release lookup, surface HTTP status on failure

sespiros:
ListReleases occasionally returns an empty page even when the release exists. A single empty response failed the test immediately. Retry the lookup a few times so a transient empty page recovers, and on failure log the HTTP status, request ID, and response body, which %#v hid behind the embedded *http.Response pointer.

Closes CON-173

### DIFF
--- a/e2e/release/release_test.go
+++ b/e2e/release/release_test.go
@@ -339,17 +339,34 @@ func fetchRelease(ctx context.Context, t *testing.T) string {
 	}
 
 	// Find our target release. There is GetReleaseByTag, but we may be looking for a draft release.
-	rels, resp, err := gh.Repositories.ListReleases(ctx, *owner, *repo, nil)
-	require.NoError(err)
 	var release *github.RepositoryRelease
-	for _, rel := range rels {
-		t.Logf("Checking release %q", *rel.TagName)
-		if *rel.TagName == *tag {
-			release = rel
-			break
+	var lastErr error
+	var lastInfo string
+	var lastCount int
+	for attempt := range 5 {
+		if attempt > 0 {
+			time.Sleep(time.Duration(attempt) * 2 * time.Second)
 		}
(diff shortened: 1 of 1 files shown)


e2e/release/release_test.go
```diff
@@ -339,17 +339,32 @@ func fetchRelease(ctx context.Context, t *testing.T) string {
...
 	var release *github.RepositoryRelease
-	for _, rel := range rels {
-		t.Logf("Checking release %q", *rel.TagName)
-		if *rel.TagName == *tag {
-			release = rel
-			break
+	var lastErr error
+	var lastInfo string
+	var lastCount int
+	for attempt := 0; attempt < 5 && release == nil; attempt++ {
```
burgerdev:
```suggestion
	for attempt := range 5 {
```

You're already `break`ing after `release = rel`.

e2e/release/release_test.go
```diff
@@ -339,17 +339,32 @@ func fetchRelease(ctx context.Context, t *testing.T) string {
...
+		}
+		for _, rel := range rels {
+			if rel.TagName != nil && *rel.TagName == *tag {
+				release = rel
+				break
+			}
+		}
+		if release == nil {
+			t.Logf("attempt %d: release %q not found among %d releases (%s)", attempt, *tag, len(rels), info)
 		}
```
burgerdev: I don't think we want to retry if we found some releases, but not the requested ones.

burgerdev (commented):
Thanks!

e2e/release/release_test.go
sespiros: Pushed a fixup, thanks.

e2e/release/release_test.go
sespiros: Pushed a fix for this too

burgerdev (approved):

PR #2467, opened 2026-06-22
kata: do not build standalone crate2nix packages 

charludo:
- [x] [build and cache](https://github.com/edgelesssys/contrast/actions/runs/27954090262)

This removes the `cargoNixPackage.*` standalone rust dependency packages generated by `crate2nix` from the matrix. Those actually required by the subset of crates we build and the features we enable are still built as part of those crates, of course.

As far as I can tell this was a perfect storm of runtime-rs and the crate2nix and matrix PRs all contributing their individual bits of the puzzle for this error to occur 😂

### DIFF
--- a/lib/matrix.nix
+++ b/lib/matrix.nix
@@ -61,6 +61,7 @@ rec {
           "contrastPkgsStatic"
           "contrast-releases"
           "matrix"
+          "cargoNixPackage" # crate2nix standalone outputs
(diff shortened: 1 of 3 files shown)


charludo:
...also, add missing dep to qemu-cc, and use our patched qemu-cc in boot-microvm.

burgerdev (approved):
Thanks!

PR #2470, opened 2026-06-23
manifest: allow more than one Coordinator policy

burgerdev:
The intention here was to avoid incompatible coordinators. However, we now have the `VersionsMatch` check that should give stronger guarantees, and allowing a slightly different Coordinator (for example: different name) unblocks some use cases.

Fixes CON-174.

### DIFF
--- a/coordinator/internal/peerdiscovery/peerdiscovery.go
+++ b/coordinator/internal/peerdiscovery/peerdiscovery.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/edgelesssys/contrast/internal/kuberesource"
+	"github.com/edgelesssys/contrast/internal/manifest"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -39,7 +41,7 @@ func (d *Discovery) GetPeers(ctx context.Context) ([]string, error) {
 	}
 	var peers []string
 	for _, pod := range pods.Items {
-		if pod.Annotations["contrast.edgeless.systems/pod-role"] != "coordinator" ||
+		if pod.Annotations[kuberesource.ContrastRoleAnnotationKey] != string(manifest.RoleCoordinator) ||
 			pod.Name == os.Getenv("HOSTNAME") {
 			continue
 		}
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -88,8 +88,8 @@ func (m *Manifest) Validate() error {
 			coordinatorCount++
 		}
 	}
-	if coordinatorCount != 1 {
-		return &CoordinatorCountError{Count: coordinatorCount}
+	if coordinatorCount == 0 {
+		return ErrMissingCoordinator
 	}
 
 	if err := m.ReferenceValues.Validate(); err != nil {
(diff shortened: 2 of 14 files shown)


coordinator/internal/peerdiscovery/peerdiscovery.go
```diff
@@ -39,7 +40,7 @@ func (d *Discovery) GetPeers(ctx context.Context) ([]string, error) {
 	}
 	var peers []string
 	for _, pod := range pods.Items {
-		if pod.Annotations["contrast.edgeless.systems/pod-role"] != "coordinator" ||
+		if pod.Annotations[kuberesource.ContrastRoleAnnotationKey] != "coordinator" ||
```
sespiros: Nit/improvement/follow-up: maybe we could extract "coordinator" as a constant in `internal/kuberesource` too.

internal/manifest/manifest.go
sespiros, line 117: I think there is an assumption here that there is a single Coordinator while after this PR we could have an snp coordinator and a tdx coordinator with different policy hashes so this check here would pick the first one found.

sespiros (commented):
Thanks! An inline comment.

internal/manifest/manifest.go
burgerdev: Oof, good catch! That needs some more thinking.

burgerdev: Fixed in the latest HEAD.

sdk/common.go
```diff
@@ -24,46 +24,51 @@ import (
 // Originally an unexported function in the contrast CLI.
 // Can be made unexported again, if we decide to move all userapi calls from the CLI to the SDK.
 // Validators MUST NOT be used concurrently.
+// The returned validators only allow COordinators.
```
sespiros, line 27: Coordinators

sdk/common.go
```diff
@@ -24,46 +24,51 @@ import (
...
-		return nil, fmt.Errorf("generating TDX validation options: %w", err)
-	}
-	var mrConfigID [48]byte
-	copy(mrConfigID[:], coordPolicyHashBytes)
-	for i, opt := range tdxOpts {
-		name := fmt.Sprintf("tdx-%d", i)
-		opt.ValidateOpts.TdQuoteBodyOptions.MrConfigID = mrConfigID[:]
-		validators = append(validators, tdx.NewValidator(opt.VerifyOpts, &tdx.StaticValidateOptsGenerator{Opts: opt.ValidateOpts}, opt.AllowedPIIDs, logger.NewWithAttrs(logger.NewNamed(log, "validator"), map[string]string{"reference-values": name}), name))
+		for i, opt := range opts {
+			opt.ValidateOpts.HostData = coordPolicyHashBytes
```
sespiros, line 47: `snp.NewValidator` stores `opt.ValidateOpts` as a pointer, so mutating `opt.ValidateOpts.HostData` here also changes every SNP validator already built in a previous iteration. Since opts is computed once outside the loop, building validators for a second coordinator hash overwrites the `HostData` of the first one, and all SNP validators end up pinned to whichever hash was iterated last (random again). TDX avoids this only because `tdxOpts` is regenerated inside the loop.

We should move the line `opts, err := m.SNPValidateOpts(kdsGetter)` and its error handling inside the loop as we do with TDX.

sespiros (changes requested):
Thanks, did a quick pass.

burgerdev:
This needs to be implemented on top of a safer API. See #2483

PR #2472, opened 2026-06-23
just: only acquire lease for gpu target

burgerdev:
We want to avoid collisions of workloads using the GPU, not workloads running at all. This changes `just default` and `just e2e` to only acquire the lease if we're using the `gpu` set.

Note that this allows `just default custom` to bypass leases although there might be GPUs used by the custom resources. Alternatively, we could opportunistically acquire the lease on `custom`, or do a `grep NVIDIA` on the resources.


sespiros (approved):
Thanks! I don't use custom that much but I guess either (always acquiring the lock or grep for NVIDIA) would be better than the possibility of contention. Maybe?
```bash
if [[ {{ target }} == "gpu" ]] || \
   { [[ {{ target }} == "custom" ]] && [[ {{ platform }} == *-GPU ]]; }; then
    just request-fifo-ticket 90m
    trap 'just release-fifo-ticket' ERR
    kubectl label ns ... contrast.edgeless.systems/sync-ticket=... --overwrite
fi
```

PR #2473, opened 2026-06-23
initdata-processor: cast Rdev to uint64 to satisfy linter running on darwin

sespiros:
The pre-commit hook (the linter specifically) was failing today locally for me because of https://github.com/edgelesssys/contrast/commit/346c35add6a8a1f0e62dc592a71441c80bc2f065 with

```
Running golangci-lint with tags contrast_unstable_api on /Users/sespiros/git-pulls/edgelesssys/contrast/.worktree/release-e2e-genpolicy-cache/initdata-processor
initdata-processor/main.go:1  typecheck  : # github.com/edgelesssys/contrast/initdata-processor
initdata-processor/main.go:147:23: cannot use stat.Rdev (variable of type int32) as uint64 value in argument to unix.Major
1 issues:
* typecheck: 1
```

I also added a test to run on new PRs to try and catch these in the future.

### DIFF
--- a/initdata-processor/main.go
+++ b/initdata-processor/main.go
@@ -16,7 +16,6 @@ import (
 	"github.com/edgelesssys/contrast/initdata-processor/policy"
 	"github.com/edgelesssys/contrast/initdata-processor/validator"
 	"github.com/edgelesssys/contrast/internal/initdata"
-	"golang.org/x/sys/unix"
 )
 
 const (
@@ -144,7 +143,7 @@ func checkDeviceAvailability(id string, magic []byte) (string, error) {
 
 		// Check whether the device major number indicates a disk type.
 		// https://www.kernel.org/doc/html/latest/admin-guide/devices.html
-		major := unix.Major(stat.Rdev)
+		major := devMajor(stat)
 		switch {
 		case major == 8:
 			// SCSI / SATA
--- a/initdata-processor/rdev_fallback.go
+++ b/initdata-processor/rdev_fallback.go
@@ -1,0 +1,20 @@
+// Copyright 2026 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build !linux
+
+package main
+
+import (
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
+// devMajor returns the major device number from stat.Rdev.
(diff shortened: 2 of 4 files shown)


initdata-processor/main.go
```diff
@@ -144,7 +144,7 @@ func checkDeviceAvailability(id string, magic []byte) (string, error) {
 
 		// Check whether the device major number indicates a disk type.
 		// https://www.kernel.org/doc/html/latest/admin-guide/devices.html
-		major := unix.Major(stat.Rdev)
+		major := unix.Major(uint64(stat.Rdev))
```
burgerdev: The linter complains on some platform, no matter what. Wouldn't the solution be to `GOOS=linux just lint` instead? Otherwise, we'd need alternative implementations (`main_darwin.go`) that stub out the incompatible parts.

sespiros: If we do `GOOS=linux` we would lose linting for the files that are meant to be linted for darwin (because they build code meant for darwin i.e the cli). Some of them are in the cli's dependency tree:

```bash
❯ rg "\!linux"
internal/atls/issuer/issuer_universal.go
4://go:build !linux

imagestore/internal/service/mount_fallback.go
4://go:build !linux

internal/cryptsetup/ext4_fallback.go
4://go:build !linux

internal/embedbin/install_fallback.go
4://go:build !linux

nodeinstaller/internal/fileop/copy_general.go
4://go:build !linux

imagepuller/internal/service/mount_fallback.go
4://go:build !linux
```

Force pushed an alternative, ptal.

initdata-processor/rdev_fallback.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
+// devMajor returns the major device number for the given stat. Off Linux,
+// Stat_t.Rdev is not uint64 (e.g. int32 on darwin), so convert it. This file
+// exists only to keep the package building and linting on non-Linux hosts;
+// the initdata-processor itself only runs on Linux.
```
burgerdev: I'd put the platform difference explanation somewhere other than the function docstring (maybe the body), and try to keep the docstring the same in both files.

burgerdev (approved):

burgerdev (approved):

PR #2474, opened 2026-06-23
e2e: persist genpolicy layer cache across nightly release runs

sespiros:
The release e2e test runs `contrast generate` against published images pulled straight from ghcr.io with no warm layer cache: the default cache lands in the throwaway release dir, so every nightly re-pulls and re-decompresses every layer from cold. When ghcr is slow this overruns the 4-minute generate deadline and the whole TestRelease fails (e.g. the 2026-06-23 nightly, where a single service-mesh-proxy layer took ~100s).

https://github.com/edgelesssys/contrast/actions/runs/27984552112

Warm the genpolicy layer cache once in the release build job, which runs on a github-hosted runner with fast ghcr access, and hand it to the e2e jobs as a workflow artifact. The build copies the workspace, runs generate over it to populate layers-cache.json, and uploads it; the bare-metal e2e jobs download it and point --genpolicy-cache-path at it, so their generate reads the cache instead of pulling layers over the slow lab link. The cache is rebuilt fresh from each run's own images, so there's no cross-run or per-runner state to manage, and warm-up is non-fatal: if it fails the e2e jobs just pull cold like today.

Only the generate pull path is covered here; the node-installer pull during apply-runtime is a separate slow path and unaffected.

### DIFF
--- a/.github/actions/release_build_linux/action.yml
+++ b/.github/actions/release_build_linux/action.yml
@@ -79,6 +79,30 @@ runs:
         path: |
           LICENSE
           ${{ steps.create-artifacts.outputs.paths }}
+    # Warm the genpolicy layer cache on this fast github-hosted runner so the
+    # self-hosted e2e jobs read it instead of pulling image layers over their
+    # slow ghcr link, which can exceed generate's timeout.
+    - name: Warm genpolicy layer cache for e2e
+      shell: bash
+      env:
+        PATHS: ${{ steps.create-artifacts.outputs.paths }}
+      run: |
+        set -u
+        mkdir -p genpolicy-warmup
+        cp $PATHS genpolicy-warmup/
+        touch genpolicy-warmup/layers-cache.json
+        nix build -L ".#base.contrast.cli-release"
+        ./result/bin/contrast generate \
+          --reference-values Metal-QEMU-SNP \
+          --genpolicy-cache-path genpolicy-warmup/layers-cache.json \
+          genpolicy-warmup/ \
+          || echo "genpolicy warm-up failed; e2e will pull cold"
+    - name: Upload genpolicy layer cache
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: genpolicy-layer-cache
+        path: genpolicy-warmup/layers-cache.json
+        if-no-files-found: warn
     # Dev/test-only image, but pushed here so it can share this job's
     # packages:write permission. The e2e release matrix consumes the digest
     # to substitute the :latest placeholder when get-logs generates the
--- a/.github/actions/release_e2e_test/action.yml
+++ b/.github/actions/release_e2e_test/action.yml
@@ -138,6 +138,11 @@ runs:
       run: |
         mkdir -p workspace
         echo "ghcr.io/edgelesssys/contrast/k8s-log-collector:latest=${LOG_COLLECTOR_IMAGE}" >> workspace/just.containerlookup
+    - name: Download warmed genpolicy layer cache
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+      with:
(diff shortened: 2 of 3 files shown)


.github/actions/release_e2e_test/action.yml
```diff
@@ -154,6 +155,11 @@ runs:
         if [[ "${CLI_ARCH}" == "aarch64-darwin" ]]; then
           use_loadbalancer="true"
         fi
+        if [[ "${SELF_HOSTED}" == "true" ]]; then
+          cache_dir="${HOME}/.cache/contrast-release-e2e/${PLATFORM_NAME}"
+          mkdir -p "${cache_dir}"
+          export GENPOLICY_CACHE_PATH="${cache_dir}/layers-cache.json"
+        fi
```
burgerdev: That's a bit heavy on the side effects, isn't it? How about we included a layer cache with the release, or used action artifacts instead?

sespiros: The side effects are not ideal idd. I picked it as the simplest one and maybe we could add a bm-maintenance job that purges `~/.cache/contrast-release-e2e` weekly.

Shipping as a release artifact seemed too much, at the end of the day this started just as a workaround for ghcr occasionally being slow and our CI becoming red as a result.

I force-pushed an approach with download/upload actions.

.github/actions/release_build_linux/action.yml
```diff
@@ -79,6 +79,27 @@ runs:
...
           ${{ steps.create-artifacts.outputs.paths }}
+    # Warm the genpolicy layer cache on this fast github-hosted runner so the
+    # self-hosted e2e jobs read it instead of pulling image layers over their
+    # slow ghcr link, which can exceed generate's timeout.
+    - name: Warm genpolicy layer cache for e2e
+      shell: bash
+      run: |
+        set -u
+        mkdir -p genpolicy-warmup/deployment
+        cp workspace/coordinator.yml workspace/emojivoto-demo.yml genpolicy-warmup/deployment/
```
burgerdev: Should we copy the entire `${{ steps.create-artifacts.outputs.paths }}`? That would also avoid tying the paths here to the unexposed paths in `.github/actions/release_artifacts`.

burgerdev (commented):
Thanks, I think it looks much better with artifacts.

.github/actions/release_build_linux/action.yml
sespiros: The release test needs only:

https://github.com/edgelesssys/contrast/blob/d0bb5441fdec77872ca5cb8fa4735db4f2d251db/e2e/release/release_test.go#L163-L165

If we pushed everything we would be running generate for `mysql-demo.yml` and `vault-demo.yml` too. I guess that's not that big of a deal in favor of a cleaner action. Pushed a fix.

burgerdev: What I meant to say: the step currently relies upon the `workspace` directory to be populated, but that is an implementation detail of the `.github/actions/release_artifacts` action.

sespiros: ptal I pushed another fixup for this.

.github/actions/release_build_linux/action.yml
```diff
@@ -79,6 +79,30 @@ runs:
...
+    # self-hosted e2e jobs read it instead of pulling image layers over their
+    # slow ghcr link, which can exceed generate's timeout.
+    - name: Warm genpolicy layer cache for e2e
+      shell: bash
+      env:
+        PATHS: ${{ steps.create-artifacts.outputs.paths }}
+      run: |
+        set -u
+        mkdir -p genpolicy-warmup
+        cp $PATHS genpolicy-warmup/
```
burgerdev, line 92: I suppose word splitting is intentional here?

burgerdev (approved):
Thanks!

.github/actions/release_build_linux/action.yml
sespiros: Yes, `$PATHS` contains multiple entries and globs.

PR #2475, opened 2026-06-23
just: use gnused for get-credentials

burgerdev:
This improves life on Mac, where `sed -i` requires an argument.


sespiros (approved):
Thanks!

PR #2476, opened 2026-06-24
kata: 3.31.0 -> 3.32.0

charludo:
Two things of note:
- second commit fixes an issue that otherwise prevents pods to come up where some container is supposed to get a GPU
- the `tdx_quote_generation_service_socket_port` already falls back to 4050, but I thought I'd rather make this explicit. We can think about migrating to unix sockets instead, but for now, let everything be as-is.


nodeinstaller/internal/kataconfig/config.go
```diff
@@ -30,6 +30,8 @@ func KataRuntimeConfig(
 			return nil, fmt.Errorf("failed to unmarshal kata runtime configuration: %w", err)
 		}
 		config.Hypervisor["qemu"]["firmware"] = filepath.Join(baseDir, "tdx", "share", "OVMF.fd")
+		// TODO: Migrate our custom calls from vsock to unix socket
+		config.Hypervisor["qemu"]["tdx_quote_generation_service_socket_port"] = 4050
```
burgerdev, line 34: Upstream implemented the fallback mechanism because the QGS is not configurable (one version does only VSOCK, the newer version only UDS). We can leave this as is for this PR, but eventually need to switch to 0. I created an issue for this.

packages/by-name/kata/source/0020-runtime-do-not-add-nr_vcpus-to-kernel-command-line.patch
```diff
@@ -4,19 +4,21 @@ Date: Fri, 27 Mar 2026 13:28:10 +0100
...
+@@ -195,11 +195,6 @@ func (q *qemu) kernelParameters() string {
  	// use default parameters
  	params = append(params, defaultKernelParameters...)
  
--	// set the maximum number of vCPUs
--	params = append(params, Param{"nr_cpus", fmt.Sprintf("%d", q.config.DefaultMaxVCPUs)})
+-	// set the maximum number of vCPUs (not applicable for confidential guests)
+-	if !q.config.ConfidentialGuest {
+-		params = append(params, Param{"nr_cpus", fmt.Sprintf("%d", q.config.DefaultMaxVCPUs)})
+-	}
```
burgerdev: I believe we can drop this patch, has been fixed upstream in https://github.com/kata-containers/kata-containers/pull/12539.

packages/by-name/kata/source/package.nix
```diff
@@ -179,6 +179,12 @@ rec {
       # to the corresponding DiffID. Entries with a DiffID key now also store the uncompressed size of the layer.
       # This allows us to calculate the pod memory requirements based on the image sizes.
       ./0026-genpolicy-cache-image-sizes-and-layer-info.patch
+
+      # Cold-plugged VFIO devices (GPUs) land in the pause container's CreateContainerRequest, but the cdi.k8s.io/vfio<n> annotation
+      # is only added to workload containers. The agent policy requires the device and its annotation to both exist, so the sandbox request can
+      # never be authorized and GPU pods fail with PermissionDenied.
+      # TODO(charludo): open upstream issue and PR.
+      ./0027-runtime-don-t-expose-VFIO-devices-to-the-sandbox-con.patch
```
burgerdev: We might need to turn the logic around: add the missing annotations to the sandbox container. Something needed the devices to be in the sandbox container (maybe https://github.com/kata-containers/kata-containers/pull/13103, but hard to tell), but the policy was not updated accordingly?

charludo: Hm, also unsure. Added a fixup commit which should do as suggested.

charludo: Tests all passed, I'd be happy to squash and keep the new patch rather than the old approach. Your call

burgerdev (approved):
I'd take the `packages/by-name/kata/source/0026-runtime-annotate-the-sandbox-container-with-VFIO-met.patch`.

PR #2479, opened 2026-06-25
docs: document `collateral-proxy`

charludo:


docs/docs/howto/collateral-proxy.md
```diff
@@ -0,0 +1,124 @@
...
+With this flag, coordinators and initializers route their AMD KDS and Intel PCS collateral fetches through the proxy instead of contacting the vendor endpoints directly.
+Without the flag, components fetch collateral directly and the proxy isn't required.
+
+## Observability
+
+The proxy exposes Prometheus metrics on its `/metrics` endpoint and a liveness check on `/healthz`.
+To inspect the metrics, query the endpoint from inside the cluster, for example:
+
+```bash
+kubectl exec statefulset/collateral-proxy -- wget -qO- http://127.0.0.1/metrics
```
burgerdev, line 80: Could also be done with port forwarding (since this is not a CC pod). The script would be more complicated in that case, but we could at least mention the possibility.

docs/docs/howto/collateral-proxy.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Prerequisites
+
+A running Contrast deployment.
+
+## How-To
+
+Using the proxy requires deploying the proxy once per cluster, then pointing your Contrast components at it when generating your deployment.
+
+### Deploy the proxy
```
burgerdev, line 23: I'm not sure whether we should even document the server at this point.

1. It's not going to stay in this repo for too long.
2. k8s resource defs should be released.

Maybe shelve this section for now and focus on the client side and how Contrast handles this.

docs/docs/howto/collateral-proxy.md
```diff
@@ -0,0 +1,124 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: collateral-proxy
+  namespace: default
+spec:
+  ports:
+    - name: proxy
+      port: 80
+      targetPort: 80
```
burgerdev: We need a security considerations section that explains why port 80 is ok (answer: the libraries _we_ use check all certificates against an embedded root cert).

charludo: done

charludo:
Sorry, it appears I forgot to push -.-

docs/docs/howto/collateral-proxy.md
```diff
@@ -0,0 +1,106 @@
...
+
+This section describes how to deploy and use the Contrast collateral proxy, an in-cluster cache for the attestation collateral that Contrast fetches from hardware vendors.
+
+## Applicability
+
+To verify an attestation report, Contrast retrieves *collateral* from external vendor endpoints:
+VCEK/VLEK certificates and CRLs from the [AMD Key Distribution Service (KDS)](../architecture/attestation/amd-details.md#attestation-report),
+PCK certificates, TCB info, and QE identities from the Intel Provisioning Certification Service (PCS), and RIM files from NVIDIA.
+Both the [Coordinator](../architecture/components/coordinator.md) and the [Initializer](../architecture/components/initializer.md) fetch this collateral
+before attesting a peer.
```
burgerdev:
```suggestion
Both the [Coordinator](../architecture/components/coordinator.md) and the [Initializer](../architecture/components/initializer.md) need to fetch this collateral to verify a peer.
```

docs/docs/howto/collateral-proxy.md
```diff
@@ -0,0 +1,106 @@
...
+
+```bash
+contrast generate --collateral-proxy http://collateral-proxy.default.svc resources/
+kubectl apply -f resources/
+```
+
+With this flag, coordinators and initializers route their AMD KDS and Intel PCS collateral fetches through the proxy instead of contacting the vendor endpoints directly.
+Without the flag, components fetch collateral directly and the proxy isn't required.
+
+The proxy is a soft dependency: if a component can't reach it, the component logs a warning and falls back to fetching directly from the vendor endpoint, then retries the proxy after a short cooldown.
```
burgerdev, line 55: Is that correct for the verifier libraries, or only for the proxy itself?

docs/docs/howto/collateral-proxy.md
```diff
@@ -0,0 +1,106 @@
...
+
+The proxy exposes Prometheus metrics on its `/metrics` endpoint and a liveness check on `/healthz`.
+To inspect the metrics, query the endpoint from inside the cluster, for example:
+
+```bash
+kubectl exec statefulset/collateral-proxy -- wget -qO- http://127.0.0.1/metrics
+```
+
+Alternatively, configure port forwarding as you usually would.
+No special considerations are required, since the proxy doesn't run in a Confidential Containers pod.
```
burgerdev:
```suggestion
No special considerations are required, since the proxy doesn't run in a confidential VM.
```

docs/docs/howto/collateral-proxy.md
```diff
@@ -0,0 +1,106 @@
...
+
+Instead, every piece of collateral is authenticated cryptographically after it's fetched:
+
+- Each Contrast component ships with the hardware vendors' root certificates embedded in its binary, namely the AMD ARK and ASK certificates for SEV-SNP, and the Intel SGX Provisioning Certification root CA for TDX.
+- The underlying libraries used in Contrast verify the full certificate chain of the fetched collateral against these embedded roots, and check the attestation report signature against the leaf key.
+
+Because this verification is anchored in keys baked into the component at build time rather than in the connection to the proxy, transport encryption adds no additional security, and conversely, lack of transport encryption doesn't detract from security.
+
+:::warning
+**A compromised proxy, or an attacker on the network path to it, can't forge collateral that passes verification.**
```
burgerdev: That's an odd warning. Maybe convert to `tip`, or just omit the admonition and keep the bold.

docs/docs/howto/collateral-proxy.md
```diff
@@ -0,0 +1,106 @@
...
+
+The worst it can do is withhold collateral or serve a stale copy, which either fails verification or, if the proxy becomes unreachable, triggers the fallback to direct vendor fetching described above.
```
burgerdev, line 101: At least the go-sev-guest library does not check the CRL validity if it was fetched remotely: 

https://github.com/google/go-sev-guest/blob/c930ed67bebfe7245c0309888ec185bd9ad35899/verify/verify.go#L283-L297

and then

https://github.com/google/go-sev-guest/blob/c930ed67bebfe7245c0309888ec185bd9ad35899/verify/verify.go#L302-L325

That should be addressed upstream, I guess.

docs/docs/howto/collateral-proxy.md
charludo: Changed to `info`, IMO these are the cliffnotes that should stand out.

docs/docs/howto/collateral-proxy.md
charludo: Adressed in https://github.com/google/go-sev-guest/pull/191 and https://github.com/edgelesssys/go-sev-guest/pull/5

docs/docs/howto/collateral-proxy.md
charludo: No, we actually do both. If we can't reach the proxy from the verifier, we fall back to upstream and retry using the proxy only after a timeout.

burgerdev (approved):
Thanks, lgtm!

PR #2480, opened 2026-06-26
collateral-proxy: various small fixups

charludo:
Follow-up to #2402.

### DIFF
--- a/collateral-proxy/internal/cache/cache.go
+++ b/collateral-proxy/internal/cache/cache.go
@@ -8,6 +8,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"encoding/pem"
+	"errors"
 	"io/fs"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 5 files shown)


collateral-proxy/internal/cache/cache.go
```diff
@@ -103,16 +105,23 @@ func (c *Cache) loadAll() error {
...
-	if err := os.WriteFile(tmp, raw, 0o600); err != nil {
+
+	enc := base64.NewEncoder(base64.StdEncoding, f)
+	if err := json.NewEncoder(enc).Encode(e); err != nil {
+		return err
+	}
+	if err := enc.Close(); err != nil {
+		return err
+	}
+	if err := f.Close(); err != nil {
```
burgerdev: Sorry, I believe I screwed this up in the last review: what we wanted to base64-encode is the file name, not the data, so we just need a `json.Encoder(os.File)`.

charludo: Ahhhh, sorry, I was already wondering :D Makes sense. Done.

burgerdev (approved):

PR #2481, opened 2026-06-26
flake: enable gofumpt as a formatter for more consistent go style

charludo:
Another followup to a conversation on #2402 😄 

Disabled on `.pb.go` since we autogenerate those.


burgerdev (approved):
Thanks a lot!

PR #2482, opened 2026-06-26
collateral-proxy: use in set/verify/recover/SDK, use per-getter config instead of global

charludo:
This is a couple of related things thrown together in this PR. The commits themselves are rather stand-alone.

- the remaining ticket was to make the proxy be used in set/verify/recover, that is done here
- noticed it should also be usable through the SDK, leading to a small refactor
- noticed when using the proxy in set/verify/recover in e2es I had to clean up the proxy setting since it was global, so refactored to use per-getter config. TBH should probably have been like that from the start.
- and finally, use in set/verify/recover in e2es

### DIFF
--- a/internal/attestation/certcache/cached_client.go
+++ b/internal/attestation/certcache/cached_client.go
@@ -6,18 +6,25 @@ package certcache
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"log/slog"
 	"net/http"
 	neturl "net/url"
 	"regexp"
 	"strings"
+	"sync/atomic"
 	"time"
 
-	"github.com/google/go-sev-guest/kds"
-	"github.com/google/go-tdx-guest/pcs"
 	"github.com/google/go-tdx-guest/verify/trust"
 	"k8s.io/utils/clock"
 	testingclock "k8s.io/utils/clock/testing"
+)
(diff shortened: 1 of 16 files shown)


internal/attestation/certcache/cached_client.go
```diff
@@ -34,55 +41,66 @@ func alwaysRevalidate(rawURL string) bool {
...
+		return header, body, nil
 	}
-	return c
+	var httpErr *httpError
+	if errors.As(err, &httpErr) {
+		return nil, nil, err
+	}
+	if c.proxyUnreachable.CompareAndSwap(false, true) {
+		c.logger.Warn("collateral proxy not reachable, falling back to direct upstream fetching", "url", url, "error", err)
+	}
```
burgerdev: The `proxyUnreachable` state is never reset - could we somehow retry after some time?

burgerdev (commented):
Otherwise lgtm

internal/attestation/certcache/cached_client.go
charludo: Added a new commit which retries the proxy every 2mins and reverts to using it permanently if successful.

burgerdev (approved):

PR #2483, opened 2026-06-26
atls: move validators into dedicated package 

burgerdev:
My overall plan is to consolidate the attestation validation logic into one package. We currently construct validators (at least) here:

https://github.com/edgelesssys/contrast/blob/41ae00c1bfa42b4073f317b5f14658d027222703/sdk/common.go#L27-L29

https://github.com/edgelesssys/contrast/blob/41ae00c1bfa42b4073f317b5f14658d027222703/coordinator/internal/stateguard/credentials.go#L74

https://github.com/edgelesssys/contrast/blob/41ae00c1bfa42b4073f317b5f14658d027222703/e2e/atls/atls_test.go#L153-L156

Since the manifest is the authoritative source for admission policy, I would like to move all of this into the manifest package.

Such a refactoring is currently hard to do, because what were supposed to be validation internals leak throughout the code base.

- The `reportSetter` is a special purpose implementation for aTLS only, but leaves its traces in validator constructors and duplicated implementations. If validators returned a report struct it would not be necessary, as the Coordinator could just wrap the validator.
- The OID `Getter` design only makes sense in this particular loop over validators: https://github.com/edgelesssys/contrast/blob/41ae00c1bfa42b4073f317b5f14658d027222703/internal/atls/atls.go#L249-L254 Now we're forced to present an exact list of validators that can each only validate one type of document. But the Coordinator can't wrap a list.
- Something that's been bothering me for quite some time: an empty list of validators means "skip validation", while a populated list means "at least one validator must pass". The logic should not surprise people assuming [vacuous truth ](https://en.wikipedia.org/wiki/Vacuous_truth). I want my sums of empty lists to be 0. https://github.com/edgelesssys/contrast/blob/41ae00c1bfa42b4073f317b5f14658d027222703/internal/atls/atls.go#L399-L404 If we did not deal in lists, this would be trivial to fix in a `validators` package.

---

That being said, this PR takes an initial step into this direction.

* Clean up around the existing validator definition (first two commits).
* Move the `Validator` into its own package.

No functional change, but this paves the way for above plan.

### DIFF
--- a/coordinator/internal/peerrecovery/recovery.go
+++ b/coordinator/internal/peerrecovery/recovery.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/edgelesssys/contrast/coordinator/internal/stateguard"
 	"github.com/edgelesssys/contrast/internal/atls"
+	"github.com/edgelesssys/contrast/internal/atls/validators"
 	"github.com/edgelesssys/contrast/internal/attestation/certcache"
 	"github.com/edgelesssys/contrast/internal/grpc/dialer"
 	"github.com/edgelesssys/contrast/internal/manifest"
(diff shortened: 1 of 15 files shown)


sespiros (approved):
Thanks.

PR #2484, opened 2026-06-26
atls: invert the validator OID lookup logic 

burgerdev:
This is step 2  teasered in #2483. I'd like to change the APIs in the`atls` package to only work with a single validator and leave the logic of walking over the individual validators in the `validators` package. For this to work, I need the interface to allow for the implementation to decide whether a given report can be processed, not the `atls` package. This PR changes the interface accordingly.


burgerdev:
Ok, I learned now that the implementation with two functions does not work how I thought it would. I think I need a 

```go
func Validate(context.Context, asn1.ObjectIdentifier, []byte, []byte) error
```

and dedicated errors for OID mismatch.

PR #2487, opened 2026-06-29
coordinator: include report OID in HTTP response

burgerdev:
The Coordinator's HTTP API has a new field in the attestation response: The `attestation_type` field now contains the OID Contrast uses internally to identify different attestation flavours, telling the validator how to interpret `raw_attestation_doc`. This change is backwards-compatible for now: old validators just ignore the field populated by new Coordinators, while new validators fall back to the old code if the field was left unset by an old Coordinator. 

*In a future Contrast version, the fallback path will be removed.*

---

Original PR description:

This PR changes the `validators.Validator` interface to only consist of a single function, `Validate`. The idea is that the validator itself should decide whether it supports a given OID, allowing for future validators to accept more than one OID (such as for a single combined validator for all reference values).  More background in the description of #2483.

This change surfaced a bug in the SDK, where we are not checking for OID support before invoking `Validate`. This is not problematic because reports are signed anyway, but I needed to add a bit of helper logic to make the old HTTP response work with the new API. A test is added to ensure we stay backwards compatible.

### DIFF
--- a/e2e/coordinator/coordinator_test.go
+++ b/e2e/coordinator/coordinator_test.go
@@ -6,10 +6,12 @@
 package coordinator
 
 import (
+	"bytes"
 	"context"
 	"crypto/ecdsa"
 	"crypto/rand"
 	"crypto/sha256"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"log/slog"
@@ -212,6 +214,17 @@ func TestCoordinator(t *testing.T) {
 		}))
 
 		state, err := client.ValidateAttestation(ctx, nonce[:], report)
+		require.NoError(err)
+		require.Equal(manifestBytesExpected, state.Manifests[0])
+
+		// Test backwards compatibility of AttestationType.
+		var resp httpapi.AttestationResponse
+		require.NoError(json.NewDecoder(bytes.NewReader(report)).Decode(&resp))
+		resp.AttestationType = nil
+		reportWithoutType, err := json.Marshal(resp)
+		require.NoError(err)
+
+		state, err = client.ValidateAttestation(ctx, nonce[:], reportWithoutType)
 		require.NoError(err)
 		require.Equal(manifestBytesExpected, state.Manifests[0])
 
--- a/e2e/coordinator/coordinator_test.go
+++ b/e2e/coordinator/coordinator_test.go
@@ -11,6 +11,8 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"flag"
+	"fmt"
+	"log/slog"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -18,9 +20,11 @@ import (
 	"time"
 
 	"github.com/edgelesssys/contrast/e2e/internal/contrasttest"
+	"github.com/edgelesssys/contrast/internal/httpapi"
 	"github.com/edgelesssys/contrast/internal/kuberesource"
 	"github.com/edgelesssys/contrast/internal/manifest"
 	"github.com/edgelesssys/contrast/internal/platforms"
+	"github.com/edgelesssys/contrast/sdk"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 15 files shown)


internal/httpapi/attest.go
```diff
@@ -33,6 +34,8 @@ type AttestationResponse struct {
 	Version string `json:"version"`
 	// RawAttestationDoc is a raw attestation report.
 	RawAttestationDoc []byte `json:"raw_attestation_doc"`
+	// RawAttestationDoc is a raw attestation report.
```
sespiros: This seems like wrong copy paste

sdk/verify.go
```diff
@@ -211,3 +224,26 @@ type CoordinatorState struct {
 	// Signature of the latest transition hash by the Coordinator.
 	LatestTransitionSignature []byte
 }
+
+// withoutOID determines the right OID for the given validator and overrides the input OID with it.
+//
+// This is needed because the OID was not part of the HTTP API response in prior versions.
+//
+// TODO(burgerdev): remove once all clients receive the OID from the HTTP API.
+func withoutOID(v validators.Validator) (validators.Validator, error) {
```
sespiros: maybe rename to `validatorForLegacyResponse` or `withInferredOID`

internal/attestation/snp/validator_test.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2024 Edgeless Systems GmbH
```
sespiros: 2026

internal/attestation/insecure/insecure_test.go
```diff
@@ -58,3 +59,6 @@ type stubReportSetter struct {
 func (s *stubReportSetter) SetReport(report attestation.Report) {
 	s.report = report
 }
+
+// Ensure that Validator implements the intended interface.
+var _ = validators.Validator(&Validator{})
```
sespiros: You can do that with zero allocation:

```go
var _ validators.Validator = (*Validator)(nil)
```

same for the rest of the places where we do this.

sespiros (changes requested):
Thanks. 

Is the coordinator-side that sets the OID in the response coming in a follow-up PR because commit https://github.com/edgelesssys/contrast/pull/2487/changes/e1164b2144d02988d84c2a7c5f73350a6115c507 says include OID field but doesn't include it.

sdk/verify.go
```diff
@@ -176,7 +181,15 @@ func (c Client) ValidateAttestation(ctx context.Context, nonce []byte, attestati
...
-		if err := v.Validate(ctx, resp.RawAttestationDoc, reportData[:]); err != nil {
+		if resp.AttestationType == nil {
+			wrapped, err := withoutOID(v)
+			if err != nil {
+				c.log.Warn("using this validator requires a server update", "error", err)
+				continue
+			}
+			v = wrapped
+		}
+		if err := v.Validate(ctx, resp.AttestationType, resp.RawAttestationDoc, reportData[:]); err != nil {
```
sespiros, line 192:
```suggestion
		if err := v.Validate(ctx, resp.AttestationType, resp.RawAttestationDoc, reportData[:]); err != nil {
		      if errors.Is(err, validators.ErrOIDNotSupported) {
              continue   // not this validator's document type; don't pollute errs
      }
```

internal/attestation/insecure/insecure_test.go
```diff
@@ -58,3 +60,6 @@ type stubReportSetter struct {
...
```
sespiros: There should also be an assertion for the `IterativeValidator`.

internal/atls/atls.go
```diff
@@ -237,22 +237,18 @@ func verifyEmbeddedReport(ctx context.Context, validators []validators.Validator
...
 				// The validator has successfully verified the document. We can exit.
 				return nil
+			} else if errors.Is(validationErr, validators.ErrOIDNotSupported) {
+				// This is the wrong validator, don't include the generic error in the response.
+				continue
 			}
 			// Otherwise, we'll keep track of the error and continue with the next validator.
-			retErr = errors.Join(retErr, fmt.Errorf(" validator %s failed: %w", validator.String(), validationErr))
+			foundMatchingValidator = true
+			retErr = errors.Join(retErr, fmt.Errorf(" validator %v failed: %w", validator, validationErr))
```
sespiros: %v resolves to `String()` only if the value implements `Stringer` and there might be some implementer that doesn't. AI raised this but I am not very confident it's an issue, just flagging it for you to see it. It suggested
```go
func validatorName(v validators.Validator) string {
      if s, ok := v.(fmt.Stringer); ok {
              return s.String()
      }
      return fmt.Sprintf("%T", v) // e.g. "*snp.Validator", "validators.ValidatorFunc"
}
```
and then
```go
retErr = errors.Join(retErr, fmt.Errorf(" validator %s failed: %w", validatorName(validator), validationErr))
```

e2e/coordinator/coordinator_test.go
```diff
@@ -172,6 +178,61 @@ func TestCoordinator(t *testing.T) {
...
+		}))
+
+		state, err := client.ValidateAttestation(ctx, nonce[:], report)
+		require.NoError(err)
+		require.Equal(manifestBytesExpected, state.Manifests[0])
+
+		// Test backwards compatibility of AttestationType.
+		var resp httpapi.AttestationResponse
+		require.NoError(json.NewDecoder(bytes.NewReader(report)).Decode(&resp))
+		resp.AttestationType = nil
```
sespiros, line 223: AI caught that this is a noop today (because the current coordinator is not setting the attestation type).

sespiros (commented):
Some more comments from the AI reviewer

internal/attestation/insecure/insecure_test.go
burgerdev: Fixed for this commit, but the rest of the codebase has some more of these.

internal/attestation/insecure/insecure_test.go
burgerdev: Added

internal/attestation/snp/validator_test.go
burgerdev: Fixed

e2e/coordinator/coordinator_test.go
burgerdev: Yes, I forgot to actually set it. Fixed now.

internal/httpapi/attest.go
burgerdev: Yup, fixed.

internal/atls/atls.go
burgerdev: This change was intentional, as I wanted to be able to lift pure functions (in particular, closures) into validators. The suggestion sounds good, though, adopted.

sdk/verify.go
burgerdev: Makes sense, but this was the case before and I'm going to remove this block in the next PR. This is  a good example for why I wanted to clean this up - subtle differences between the validation for loops.

sdk/verify.go
burgerdev: Changed. This is only necessary until I moved the validator creation to the manifest package, where I know the expected OID, so be prepared for this function to go away again.

burgerdev (commented):
Thanks, I believe I addressed all comments.

sespiros (changes requested):
Thanks. 2 more things:

This test fails now:
```bash
go test ./coordinator/internal/httpapi/ -run 'TestAttestationHandler/success' -v
```

The `stubIssuer` in `coordinator/internal/httpapi/attest_test.go` needs:
```go
func (s *stubIssuer) OID() asn1.ObjectIdentifier {
      return oid.RawSNPReport
}
```

And also it's worth noting this wasn't surfaced in the CI so I opened https://github.com/edgelesssys/contrast/pull/2490

sespiros (approved):

PR #2488, opened 2026-06-29
cli: adjust container memory limit when using pod memory limits

davidweisse:
When using the `--calculate-pod-memory` flag during `generate`, the pods get additional memory assigned through pod level resource limits, based on the image sizes. The Initializer, Debugshell and Service Mesh containers have a predefined container memory limit that takes the image sizes already into account. When automatically adding the necessary memory with the `--calculate-pod-memory` flag, we can thus lower the container limits for our injected containers.

I looked at the `memory.peak` cgroup entry to approximately gauge how much memory is actually needed for the containers. Especially for the service mesh container, we should be generous with the necessary memory,  as this is dependent on the deployed workload.

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -19,6 +19,36 @@ import (
 	applyrbacv1 "k8s.io/client-go/applyconfigurations/rbac/v1"
 )
 
+// MemoryProfile represents the memory profile of a container.
+type MemoryProfile bool
+
+const (
+	// MemoryProfileFull includes the image requirements in the container memory limits.
+	MemoryProfileFull MemoryProfile = false
+	// MemoryProfileRuntimeOnly only includes the runtime requirements in the container memory limits.
+	MemoryProfileRuntimeOnly MemoryProfile = true
+)
+
+var containerMemoryLimits = map[string]map[MemoryProfile]int64{
+	"contrast-initializer": {
+		// In v1.18.0, the initializer image was 50MiB compressed, 160MiB uncompressed.
+		// Double that because only 50% of the VM memory are available for /run.
+		MemoryProfileFull: 420,
+		// The container's cgroup memory.peak is at about 5MiB.
+		MemoryProfileRuntimeOnly: 20,
+	},
+	"contrast-service-mesh": {
+		MemoryProfileFull: 400,
+		// When running the emojivoto deployment, the container's cgroup memory.peak is at about 20MiB.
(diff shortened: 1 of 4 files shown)


internal/kuberesource/parts.go
```diff
@@ -547,19 +557,24 @@ func ServiceMeshProxy() *applycorev1.ContainerApplyConfiguration {
...
 		)
 }
 
 // DebugShell creates a new debug shell container.
-func DebugShell() *applycorev1.ContainerApplyConfiguration {
+func DebugShell(adjustMemoryLimit bool) *applycorev1.ContainerApplyConfiguration {
+	memoryLimit := int64(1000)
+	if adjustMemoryLimit {
+		// The container's cgroup memory.peak is at about 4MiB.
+		memoryLimit = int64(20)
```
charludo: But that's idle usage, right? I imagine if we actually use the debug shell for its intended purpose, then we'll easily exceed the 20MiB. I'd set this a good deal higher, just to be safe (cat-ing/cp-ing large file, tar, curl ... can all easily exceed this)

charludo (changes requested):
Thanks! 🎉 

- Passing around contextless bools is a bit hard to read unfortunately. How about:
    ```go
      type MemoryProfile bool

      const (
          FullMemory MemoryProfile = false
          RuntimeOnlyMemory MemoryProfile = true
      )
    ```
    and then you just call e.g. 
    ```go
    ServiceMeshProxy(kuberesource.FullMemory)
    ```

internal/kuberesource/parts.go
davidweisse: True, do you think 200MiB is fine?

charludo: Yeah sounds good

internal/kuberesource/parts.go
```diff
@@ -19,6 +19,36 @@ import (
...
+type MemoryProfile bool
+
+const (
+	// MemoryProfileFull includes the image requirements in the container memory limits.
+	MemoryProfileFull MemoryProfile = false
+	// MemoryProfileRuntimeOnly only includes the runtime requirements in the container memory limits.
+	MemoryProfileRuntimeOnly MemoryProfile = true
+)
+
+var containerMemoryLimits = map[string]map[MemoryProfile]int64{
```
charludo, line 32: Ah, nice!

charludo (approved):
Thanks, looks good!

PR #2489, opened 2026-06-29
manifest: authoritative source for validators

burgerdev:
This PR continues the plan described in #2483, moving all validation logic from the various places we implemented it in to the `manifest` package, which is now the authoritative source for how to construct validators for a given manifest. Two helpers are added:

* This move needs to retain the backwards compat logic from the SDK, so we're adding a wrapper for that to the `validators` package, too.
* The `Any` combinator separates the for loop over eligible validators from their construction and individual invocation, simplifying the respective functions.

### DIFF
--- a/sdk/common.go
+++ b/sdk/common.go
@@ -6,18 +6,12 @@
 package sdk
 
 import (
-	"encoding/binary"
 	"fmt"
 	"log/slog"
-	"strings"
 
 	"github.com/edgelesssys/contrast/internal/atls/validators"
 	"github.com/edgelesssys/contrast/internal/attestation/certcache"
-	"github.com/edgelesssys/contrast/internal/attestation/snp"
-	"github.com/edgelesssys/contrast/internal/attestation/tdx"
-	"github.com/edgelesssys/contrast/internal/logger"
 	"github.com/edgelesssys/contrast/internal/manifest"
-	snpmeasure "github.com/edgelesssys/contrast/internal/snp"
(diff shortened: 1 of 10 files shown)


sdk/common.go
```diff
@@ -6,65 +6,22 @@
...
-	"github.com/edgelesssys/contrast/internal/attestation/tdx"
-	"github.com/edgelesssys/contrast/internal/logger"
 	"github.com/edgelesssys/contrast/internal/manifest"
-	snpmeasure "github.com/edgelesssys/contrast/internal/snp"
 )
 
 // ValidatorsFromManifest returns a list of validators corresponding to the reference values in the given manifest.
 // Originally an unexported function in the contrast CLI.
 // Can be made unexported again, if we decide to move all userapi calls from the CLI to the SDK.
 // Validators MUST NOT be used concurrently.
```
sespiros, line 20: is this still true?

sespiros (approved):
Thanks. just a comment nit

sdk/common.go
burgerdev: That's an excellent question! Turns out to still be the case due to the CRL/collateral caching modifying the validator:

https://github.com/edgelesssys/contrast/blob/1a2157e90fee6fe3c9718d4b55b6809230bdf900/internal/attestation/snp/validator.go#L102-L106

Let me add that warning to the new functions and log an issue to ... not do that.

PR #2490, opened 2026-06-30
justfile: add just unit target and pre-push hook

sespiros:
During review of https://github.com/edgelesssys/contrast/pull/2487/ I noticed that a unit test failure was not being surfaced in the CI during PR review. Adding a new just target and a pre-push hook.

### DIFF
--- a/justfile
+++ b/justfile
@@ -546,6 +546,10 @@ fmt:
 lint:
     nix run -L .#base.scripts.golangci-lint -- run
 
+# Run Go unit tests.
+unit:
+    CGO_ENABLED=1 go test -tags=contrast_unstable_api -v -race ./...
+
 # Check links.
 check-links config="external":
     nix run .#base.nixpkgs.lychee -- --config tools/lychee/config-{{ config }}.toml .
--- a/dev-shells/git-hooks.nix
+++ b/dev-shells/git-hooks.nix
@@ -28,7 +28,7 @@
   };
   go-test = {
     enable = true;
-    entry = "go test -tags contrast_unstable_api -v ./...";
+    entry = "just unit";
     pass_filenames = false;
     stages = [ "pre-push" ];
   };


justfile
```diff
@@ -546,6 +546,10 @@ fmt:
 lint:
     nix run -L .#base.scripts.golangci-lint -- run
 
+# Run Go unit tests.
+unit:
+    CGO_ENABLED=1 go test -tags=contrast_unstable_api -race ./...
+
```
charludo, line 552: I actually like the idea of having this. Should also be used in the `git-hooks.nix`.

```suggestion
# Run Go unit tests.
unit:
    CGO_ENABLED=1 go test -tags=contrast_unstable_api -v -race ./...

```

charludo (commented):
I am a bit confused. We should not need this. Building our packages already runs the tests in the check phase, and fails if any unit test fails.

From your comment on the linked PR:

> This test fails now:
> 
> ```shell
> go test ./coordinator/internal/httpapi/ -run 'TestAttestationHandler/success' -v
> ```

I'm sorry to say but I can't reproduce that. Succeeds for me on 79d5ce4a2814dd0aa26bb7494cd1eacd5a214e1f.

charludo:
> Building our packages already runs the tests in the check phase

...and additionally, I should have said, some go tests *require* being run through nix, since they rely on nix build artifacts to be present at test time. Meaning the CI step introduced in this PR also doesn't capture all unit tests.

sespiros:
@charludo thanks for looking. So the reason you can't reproduce is that Markus had already pushed a fix for this 30 minutes before my review. Also I saw green tests today most probably after the fix so I assumed the unit tests don't run during PR. I am gonna drop the ci commit but leave the just target.

charludo:
> @charludo thanks for looking. So the reason you can't reproduce is that Markus had already pushed a fix for this 30 minutes before my review. Also I saw green tests today most probably after the fix so I assumed the unit tests don't run during PR. I am gonna drop the ci commit but leave the just target.

Ah, did not know that. Can indeed reproduce on 7dbe61abfe3197b39612adaeddf8387c85bdc624. However, `nix build .#base.contrast.coordinator` also fails on that commit for the same reason. I can't speak to why/if/how CI didn't fail.

burgerdev:
> > @charludo thanks for looking. So the reason you can't reproduce is that Markus had already pushed a fix for this 30 minutes before my review. Also I saw green tests today most probably after the fix so I assumed the unit tests don't run during PR. I am gonna drop the ci commit but leave the just target.
> 
> Ah, did not know that. Can indeed reproduce on [7dbe61a](https://github.com/edgelesssys/contrast/commit/7dbe61abfe3197b39612adaeddf8387c85bdc624). However, `nix build .#base.contrast.coordinator` also fails on that commit for the same reason. I can't speak to why/if/how CI didn't fail.

CI did fail - that's how I learned about the problem and why I pushed a fix this morning.

charludo (approved):

PR #2491, opened 2026-06-30
just/e2e: add flag to disable usage of very large images in tests

charludo:
#2459 added the tensorflow image to the imagestore test. Unofrtunately, 7GB is way too large for `generate` to now finish in the 5min timeout window on my dev machine / with my internet connection. I assume I'm not alone. 

This PR adds a flag to just skip that container in the imagestore test. For CI, nothing changes; for devs, you can set `imagestore_skip_large_image="true"` in your `justfile.env` and have the tensorflow image be skipped.

### DIFF
--- a/e2e/imagestore/imagestore_test.go
+++ b/e2e/imagestore/imagestore_test.go
@@ -59,7 +59,9 @@ func TestImageStore(t *testing.T) {
 		resources = append(resources, testPod(tc.name, tc.annotation))
 	}
 
-	resources = append(resources, tensorflowPod())
+	if !*skipLargeImage {
+		resources = append(resources, tensorflowPod())
+	}
 
 	resources = kuberesource.PatchRuntimeHandlers(resources, runtimeHandler)
 	resources = kuberesource.AddPortForwarders(resources)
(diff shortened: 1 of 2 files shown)


burgerdev:
Maybe it would be better to use a smaller image?

charludo:
> Maybe it would be better to use a smaller image?

I assumed using a large image was the point. This isn't urgent, maybe we can wait with a decision for @davidweisse to weigh in.

burgerdev:
> > Maybe it would be better to use a smaller image?
> 
> I assumed using a large image was the point. This isn't urgent, maybe we can wait with a decision for @davidweisse to weigh in.

Yes, but the image only needs to be large enough to explode the VM without the automatic limit. Agreed to wait for @davidweisse.

davidweisse:
With https://github.com/edgelesssys/contrast/pull/2488, I think we only need to be over about 700-800MiB in compressed+uncompressed size. On a GPU platform, the default memory is 1GiB. Then add various small container limits for the initializer, debugshell and tensorflow containers, and divide by 2 to account for the available space in `/run`.

If this is acceptable, I would go with a smaller image (if my calculations are correct). If so, we should at least check manually once that when using no auto-memory limits, the vm still runs out of space.

charludo:
> With #2488, I think we only need to be over about 700-800MiB in compressed+uncompressed size. On a GPU platform, the default memory is 1GiB. Then add various small container limits for the initializer, debugshell and tensorflow containers, and divide by 2 to account for the available space in `/run`.
> 
> If this is acceptable, I would go with a smaller image (if my calculations are correct). If so, we should at least check manually once that when using no auto-memory limits, the vm still runs out of space.

That sounds good. I guess this can then be closed. Do you want to decide on a new image? Otherwise I can also do a followup.

davidweisse:
> That sounds good. I guess this can then be closed. Do you want to decide on a new image? Otherwise I can also do a followup.

Yeah, I can do that.

PR #2493, opened 2026-06-30
manifest: allow more than one Coordinator policy

burgerdev:
Contrast now allows more than one Coordinator in the resources passed to `contrast generate`. This is useful in a couple of scenarios:

- Coordinator instances running with different names or in different namespaces.
- Coordinator instances with different configuration (e.g. logging).
- Blue-green deployments.
- Different runtime class (SNP and TDX produce different policies).

---

Fixes CON-174

### DIFF
--- a/internal/manifest/validators.go
+++ b/internal/manifest/validators.go
@@ -71,13 +71,25 @@ var ErrWrongCoordinatorPolicyHash = errors.New("wrong policy hash for Coordinato
 //
 // This is a more restrictive version of Validator, see the warning there.
 func (m *Manifest) CoordinatorValidator(log *slog.Logger, kdsGetter *certcache.CachedHTTPSGetter) (validators.Validator, error) {
-	coordPolicyHash, err := m.CoordinatorPolicyHash()
+	coordPolicyHashes, err := m.CoordinatorPolicyHashes()
 	if err != nil {
 		return nil, fmt.Errorf("getting coordinator policy hash: %w", err)
 	}
-	coordPolicyHashBytes, err := coordPolicyHash.Bytes()
-	if err != nil {
-		return nil, fmt.Errorf("converting coordinator policy hash to bytes: %w", err)
+	return coordinatorValidator(m.Validator, coordPolicyHashes, log, kdsGetter)
+}
+
+type validatorFactory func(log *slog.Logger, kdsGetter *certcache.CachedHTTPSGetter, reportSetter attestation.ReportSetter) (validators.Validator, error)
+
+// coordinatorValidator is the implementation of the post-validation step for CoordinatorValidator.
+// It can be parametrized for testing.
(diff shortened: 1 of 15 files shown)


internal/manifest/validators.go
```diff
@@ -88,8 +92,16 @@ func (m *Manifest) CoordinatorValidator(log *slog.Logger, kdsGetter *certcache.C
...
+		if !slices.ContainsFunc(allHashes, func(b []byte) bool {
+			return slices.Equal(report.HostData(), b)
+		}) {
+			s := &strings.Builder{}
+			s.WriteString("[ ")
+			for _, h := range coordPolicyHashes {
+				fmt.Fprintf(s, "%s ", h.String())
+			}
+			s.WriteString(" ]")
+			return fmt.Errorf("wrong policy hash for Coordinator: got %x, want %x", report.HostData(), s.String())
```
sespiros:
```suggestion
			return fmt.Errorf("wrong policy hash for Coordinator: got %x, want one of %v", report.HostData(), coordPolicyHashes)
```

internal/manifest/validators.go
```diff
@@ -88,8 +92,16 @@ func (m *Manifest) CoordinatorValidator(log *slog.Logger, kdsGetter *certcache.C
 		if err := validator.Validate(ctx, oid, attDoc, reportData); err != nil {
 			return err
 		}
-		if !slices.Equal(coordPolicyHashBytes, report.HostData()) {
-			return fmt.Errorf("wrong policy hash for Coordinator: got %x, want %x", report.HostData(), coordPolicyHashBytes)
+		if !slices.ContainsFunc(allHashes, func(b []byte) bool {
```
sespiros, line 115: I think we should have a `validators_test.go` that tests a few cases for a multi-coordinator path because currently this is not exercised anywhere. I am thinking a unit test on `CoordinatorValidator` with two coordinator hashes that asserts a) a report whose `HostData` matches the second hash is accepted and b) a report matching neither is rejected.

sdk/verify.go
sespiros:
```suggestion
		return m.CoordinatorValidator(log, kdsGetter)
```

sespiros (changes requested):
Thanks, did a pass.

internal/manifest/validators.go
burgerdev: Yeah, no idea why I did not realize this 😅

sdk/verify.go
burgerdev: That's from the other PR, let me fix it there.

internal/manifest/validators.go
burgerdev: Added a unit test. We should eventually add an e2e test, too, but not in this PR.

sespiros (approved):
Thanks for the changes!

sespiros (approved):

PR #2494, opened 2026-06-30
treewide: pass a single validator to aTLS

burgerdev:
Another continuation of #2483: now that we can combine `Validator`s, use this to simplify the aTLS dialers' and HTTP verifier's APIs to just expect a single validator (which is now a `validators.Any`). Consistent with the `Issuer` argument, skip validation if the validator is `nil`. 

The last commit moves all existing callers of the `sdk.ValidatorsFromManifest` to the `manifest.CoordinatorValidator` function. `sdk.ValidatorsFromManifest` should not have been exported in the first place, as the docstring suggests.

Fixes CON-129.

### DIFF
--- a/internal/atls/atls.go
+++ b/internal/atls/atls.go
@@ -42,10 +42,10 @@ var (
 )
 
 // CreateAttestationServerTLSConfig creates a tls.Config object with a self-signed certificate and an embedded attestation document.
-// Pass a list of validators to enable mutual aTLS.
 // If issuer is nil, no attestation will be embedded.
-func CreateAttestationServerTLSConfig(issuer Issuer, validators []validators.Validator, attestationFailures prometheus.Counter) (*tls.Config, error) {
-	getConfigForClient, err := getATLSConfigForClientFunc(issuer, validators, attestationFailures)
+// If validator is nil, no attestation will be requested from the peer. Otherwise, use mutual TLS.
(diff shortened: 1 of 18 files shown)


internal/atls/atls.go
```diff
@@ -44,8 +44,9 @@ var (
 // CreateAttestationServerTLSConfig creates a tls.Config object with a self-signed certificate and an embedded attestation document.
 // Pass a list of validators to enable mutual aTLS.
```
sespiros:
```suggestion
// Pass a validator to enable mutual aTLS.
```

internal/atls/atls.go
```diff
@@ -219,7 +229,7 @@ func processCertificate(rawCerts [][]byte, _ [][]*x509.Certificate) (*x509.Certi
 // verifyEmbeddedReport verifies an aTLS certificate by validating the attestation document embedded in the TLS certificate.
 //
 // It will check against all applicable validator for the type of attestation document, and return success on the first match.
```
sespiros: Remove?

sespiros (approved):
Thanks

internal/atls/atls.go
burgerdev: Fixed.

internal/atls/atls.go
burgerdev: Fixed.

PR #2498, opened 2026-07-02
.github/release_promote: skipped teams notification is not a release blocker

charludo:
- The teams notification action skips on reruns. IMO that should remain as-is, to not pollute the channel (not just from the release workflow). 
- github prepends the name of the calling step to the action, so because the e2e step is `release requirement: `, so is the teams notification
- hence, this fix :/

### DIFF
--- a/.github/workflows/release_promote.yml
+++ b/.github/workflows/release_promote.yml
@@ -75,7 +75,7 @@ jobs:
           NIGHTLY_RUN_ID: ${{ steps.nightly-info.outputs.nightly_run_id }}
           GH_REPO: ${{ github.repository }}
         run: |
-          required_jobs=$(gh api --paginate \
+          release_requirement_jobs=$(gh api --paginate \
             "repos/${GH_REPO}/actions/runs/${NIGHTLY_RUN_ID}/jobs?per_page=30" \
             --jq '.jobs[] | select(.name | startswith("release-requirement: ")) | {name, conclusion}')
           # Every release-requirement job must be present, so a rename can't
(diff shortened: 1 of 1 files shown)


.github/workflows/release_promote.yml
```diff
@@ -75,9 +75,10 @@ jobs:
           NIGHTLY_RUN_ID: ${{ steps.nightly-info.outputs.nightly_run_id }}
           GH_REPO: ${{ github.repository }}
         run: |
+          # The "Notify teams channel of failure" job in e2e_nightly.yml carries the "release-requirement: " prefix via the reusable-workflow job name, but is not a release gate.
           required_jobs=$(gh api --paginate \
             "repos/${GH_REPO}/actions/runs/${NIGHTLY_RUN_ID}/jobs?per_page=30" \
-            --jq '.jobs[] | select(.name | startswith("release-requirement: ")) | {name, conclusion}')
+            --jq '.jobs[] | select(.name | startswith("release-requirement: ")) | select(.name | endswith("Notify teams channel of failure") | not) | {name, conclusion}')
```
burgerdev: We already have a hard-coded list of expected jobs, could we just check the status of those instead? Or, maybe even better: could we just not care for the nightly status and trust in the action chaining and/or human discretion?

charludo: Fair. Using the list now. 

IDK about not gating at all.... I think the tests passing is a sensible hard requirement for the release. For sporadic failures, we can rerun. For consistent failures, something is broken and should not be released.

sespiros (approved):
Thanks.

PR #2499, opened 2026-07-02
kuberesource: export all annotation vars from kuberesource package

davidweisse:
We define constants for contrast specific K8s annotation keys in multiple locations, mainly the `cli` and `kuberesource` packages. Since there are also multiple duplicate definitions, it makes sense to put them into one location. I put every annotation key variable we define into the `kuberesource` package and exported them. Any future definitions can then go there as well.

Fixes CON-192

### DIFF
--- a/internal/kuberesource/constants.go
+++ b/internal/kuberesource/constants.go
@@ -1,0 +1,62 @@
+// Copyright 2026 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package kuberesource
+
+const (
+	annotationPrefix     = "contrast.edgeless.systems/"
+	kataAnnotationPrefix = "io.katacontainers.config."
+)
+
+const (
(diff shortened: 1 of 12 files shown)


internal/kuberesource/constants.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2026 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+package kuberesource
+
+const (
+	annotationPrefix     = "contrast.edgeless.systems/"
+	kataAnnotationPrefix = "io.katacontainers.config.hypervisor."
```
burgerdev: Can we make the prefix just `io.katacontainers.config.`? We might need to use other subgroups in the future, notably `agent.*`.

burgerdev (approved):
Thanks!

burgerdev (approved):

PR #2500, opened 2026-07-02
coordinator: domain-separate transit engine keys from workload secrets

sespiros:
The transit engine derived its AES keys with DeriveWorkloadSecret("<version>_<name>"), the same HKDF namespace that hands each pod its workload secret. A pod deployed with WorkloadSecretID "0_vault_unsealing" thus received, in its own workload secret, the exact key the transit engine uses to wrap OpenBao's master key. That defeats the transit API's per-workload authorization on both directions: the pod can unwrap the auto-unseal blob offline, and (since it holds the key) forge a blob the Coordinator will decrypt, injecting an attacker-chosen master key on OpenBao's untrusted storage.

Derive transit keys from a dedicated seed and HKDF label so their keyspace can never alias a workload secret. No decrypt-side fallback to the old derivation is provided on purpose: honoring the legacy key would keep the forgery path open, since the server cannot tell an original legacy blob from an attacker-forged one. This is a breaking change for existing OpenBao auto-unseal deployments, which must migrate their master key (seal-migration or re-init) onto the new key on upgrade.

Fixes CON-201

TODO
- [x] ~~Decide on rollout/backwards compatibility as this is a breaking change!~~ This is now tagged as a breaking change for the release notes.

# Migrating an existing Vault after this upgrade

This release changes the transit engine key derivation, so a Vault that uses transit auto-unsealing won't auto-unseal after you upgrade across this version. Migrate by bracketing the upgrade with a Shamir seal. This only re-wraps Vault's root key; the stored data is never touched. You need a threshold of the recovery keys you saved when you initialized the Vault (3 of 5 by default). Run the `bao` commands from your workspace against the Vault endpoint (`VAULT_ADDR`/`VAULT_CACERT` as in the initial setup), the same way you initialized it. The Vault config is policy-pinned, so config changes go through `contrast generate` + `contrast set`, not `kubectl edit`.

**Before upgrading (transit -> Shamir).** In the Vault config (the `VAULT_CONFIG` value of the `vault-config` ConfigMap), add the line `disabled = "true"` inside the `seal "transit"` block (the stanza with `key_name = "vault_unsealing"` and `address = "https://coordinator:8200"`), leaving its other fields unchanged. Then:

```sh
contrast generate --reference-values <platform> resources/
contrast set -c "${coordinator}:1313" resources/
kubectl rollout restart statefulset/vault
bao operator unseal -migrate <recovery-key>   # repeat for the threshold (3)
```

`bao status` should now show `Seal Type: shamir`, `Sealed: false`.

**Deploy the new Coordinator** (and the rest of the release). Because the Vault is Shamir-sealed at this point, the transit key change doesn't affect it.

**After upgrading (Shamir -> transit).** Remove the `disabled = "true"` line, then run the same four commands again (same recovery keys). Vault re-wraps its root key under the new transit key and auto-unseals; `bao status` should show `Seal Type: transit`, `Sealed: false`.

### DIFF
--- a/coordinator/internal/transitengineapi/transitengineapi.go
+++ b/coordinator/internal/transitengineapi/transitengineapi.go
@@ -137,7 +137,7 @@ func getEncryptHandler(guard stateGuard, logger *slog.Logger) http.HandlerFunc {
 			}, logger)
 			return
 		}
-		key, err := deriveEncryptionKey(r.Context(), guard, fmt.Sprintf("%d_%s", encReq.KeyVersion, workloadSecretID))
+		key, err := deriveEncryptionKey(r.Context(), guard, encReq.KeyVersion, workloadSecretID)
 		if err != nil {
 			writeHTTPError(w, httpError{
 				code:          http.StatusInternalServerError,
@@ -209,7 +209,7 @@ func getDecryptHandler(guard stateGuard, logger *slog.Logger) http.HandlerFunc {
 			}, logger)
 			return
 		}
-		key, err := deriveEncryptionKey(r.Context(), guard, fmt.Sprintf("%d_%s", decReq.CiphertextContainer.keyVersion, workloadSecretID))
+		key, err := deriveEncryptionKey(r.Context(), guard, decReq.CiphertextContainer.keyVersion, workloadSecretID)
 		if err != nil {
 			writeHTTPError(w, httpError{
 				code:          http.StatusInternalServerError,
@@ -263,21 +263,21 @@ func authorizeWorkloadSecret(workloadSecretID string, r *http.Request, logger *s
 	return fmt.Errorf("mismatching workloadSecretIDs: name:%s, extension:%s", workloadSecretID, extensionWSID)
 }
 
-// deriveEncryptionKey derives the workload secret used as the encryption key by receiving the seedengine of the current state.
-func deriveEncryptionKey(ctx context.Context, guard stateGuard, workloadSecretID string) ([]byte, error) {
+// deriveEncryptionKey derives the transit engine encryption key from the current state's seed engine.
+func deriveEncryptionKey(ctx context.Context, guard stateGuard, keyVersion uint32, name string) ([]byte, error) {
 	// TODO(burgerdev): we should be using the state from the request context here!
 	state, err := guard.GetState(ctx)
 	if err != nil {
 		return nil, err
 	}
-	derivedWorkloadSecret, err := state.SeedEngine().DeriveWorkloadSecret(workloadSecretID)
+	key, err := state.SeedEngine().DeriveTransitEngineKey(keyVersion, name)
 	if err != nil {
 		return nil, err
 	}
-	if len(derivedWorkloadSecret) < aesGCMKeySize {
+	if len(key) < aesGCMKeySize {
 		return nil, fmt.Errorf("derived key too small, expected key length: %d", aesGCMKeySize)
 	}
-	return derivedWorkloadSecret[:aesGCMKeySize], nil
+	return key[:aesGCMKeySize], nil
 }
 
 // writeJSONResponse wraps any payload inside a "data" object and sends it as an HTTP response.
--- a/coordinator/internal/transitengineapi/transitengineapi_test.go
+++ b/coordinator/internal/transitengineapi/transitengineapi_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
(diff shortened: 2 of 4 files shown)


coordinator/internal/transitengineapi/transitengineapi.go
```diff
@@ -137,7 +137,7 @@ func getEncryptHandler(guard stateGuard, logger *slog.Logger) http.HandlerFunc {
 			}, logger)
 			return
 		}
-		key, err := deriveEncryptionKey(r.Context(), guard, fmt.Sprintf("%d_%s", encReq.KeyVersion, workloadSecretID))
```
burgerdev, line 140: Now that I think of it, I remember that we picked `%d_%s` because generated workloadSecretIDs don't contain an underscore: https://github.com/edgelesssys/contrast/pull/1199#discussion_r1958021042. We could enforce that in `manifest.Validate`, but that would be even more breaking.

coordinator/internal/transitengineapi/transitengineapi_test.go
```diff
@@ -212,6 +213,46 @@ func TestAdversarialDecryptInputs(t *testing.T) {
...
+	decryptReqBody, err := json.Marshal(map[string]string{"ciphertext": string(bytes.Trim(ciphertext, `"`))})
+	require.NoError(err)
+
+	req := httptest.NewRequestWithContext(t.Context(), http.MethodPut, "/v1/transit/decrypt/"+name, bytes.NewReader(decryptReqBody))
+	req.Header.Set("Content-Type", "application/json")
+
+	rec := httptest.NewRecorder()
+	mux.ServeHTTP(rec, req)
+	res := rec.Result()
+	t.Cleanup(func() { _ = res.Body.Close() })
```
burgerdev, line 249: Why did you choose HTTP instead of calling the function directly?

burgerdev (approved):
Thanks for fixing this, lgtm!

As migration path, we could recommend starting a pod with the problematic `0_$ID` to obtain the old sealing secret, then reseal using the transit secret engine. I don't think we need to keep a backwards-compatible implementation around.

coordinator/internal/transitengineapi/transitengineapi_test.go
sespiros: Which function should I be calling? The rest of the test suite exercises the API this way exercising the full end-to-end path. I initially wrote the test and it proved meaningful because in some initial version I experimented with adding various types of fallbacks which obviously all failed this test (as the comment describes).

coordinator/internal/transitengineapi/transitengineapi.go
sespiros: Right! I didn't like that `_` was essentially that load-bearing. I initially solved this by changing the HKDF info label but then I thought that a dedicated seed would split the two keyspaces even more cleanly.

coordinator/internal/transitengineapi/transitengineapi_test.go
burgerdev: I now see that calling `deriveEncryptionKey` does not really cover what this test wants, and I was under the false impression that we had better separation between the HTTP layer and the business logic. This is fine.

burgerdev (approved):

sespiros:
# Migrating an existing Vault after this upgrade

This release changes the transit engine key derivation, so a Vault that uses transit auto-unsealing won't auto-unseal after you upgrade across this version. Migrate by bracketing the upgrade with a Shamir seal. This only re-wraps Vault's root key; the stored data is never touched. You need a threshold of the recovery keys you saved when you initialized the Vault (3 of 5 by default). Run the `bao` commands from your workspace against the Vault endpoint (`VAULT_ADDR`/`VAULT_CACERT` as in the initial setup), the same way you initialized it. The Vault config is policy-pinned, so config changes go through `contrast generate` + `contrast set`, not `kubectl edit`.

**Before upgrading (transit -> Shamir).** In the Vault config (the `VAULT_CONFIG` value of the `vault-config` ConfigMap), add the line `disabled = "true"` inside the `seal "transit"` block (the stanza with `key_name = "vault_unsealing"` and `address = "https://coordinator:8200"`), leaving its other fields unchanged. Then:

```sh
contrast generate --reference-values <platform> resources/
contrast set -c "${coordinator}:1313" resources/
kubectl rollout restart statefulset/vault
bao operator unseal -migrate <recovery-key>   # repeat for the threshold (3)
```

`bao status` should now show `Seal Type: shamir`, `Sealed: false`.

**Deploy the new Coordinator** (and the rest of the release). Because the Vault is Shamir-sealed at this point, the transit key change doesn't affect it.

**After upgrading (Shamir -> transit).** Remove the `disabled = "true"` line, then run the same four commands again (same recovery keys). Vault re-wraps its root key under the new transit key and auto-unseals; `bao status` should show `Seal Type: transit`, `Sealed: false`.

PR #2501, opened 2026-07-02
lib, packages: add and use package meta helper

charludo:
Precursor to the SBOM PR. We don't really *need* this on its own, but it's necessary for filling the SBOMs.

### DIFF
--- a/packages/by-name/contrast/resourcegen/package.nix
+++ b/packages/by-name/contrast/resourcegen/package.nix
@@ -71,8 +71,5 @@ buildGoModule (_finalAttrs: {
   # need any other fixup, saving some seconds.
   dontFixup = true;
 
-  meta = {
-    description = "Resource generator for Contrast";
-    mainProgram = "resourcegen";
-  };
+  meta = lib.contrast.ourMeta { mainProgram = "resourcegen"; };
 })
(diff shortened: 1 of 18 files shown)


packages/by-name/contrast/resourcegen/package.nix
```diff
@@ -71,8 +71,5 @@ buildGoModule (_finalAttrs: {
   # need any other fixup, saving some seconds.
   dontFixup = true;
 
-  meta = {
-    description = "Resource generator for Contrast";
-    mainProgram = "resourcegen";
-  };
+  meta = lib.contrast.ourMeta { mainProgram = "resourcegen"; };
```
sespiros, line 74: Is the description dropped on purpose?

burgerdev (approved):

sespiros (approved):

packages/by-name/contrast/resourcegen/package.nix
charludo: Ah, no. But, should I add it back? It's not super useful... Or descriptive... 😂

sespiros: As you see fit, afaiu it's an optional field and most other packages don't have it anyway.

charludo: Thanks. Gonna leave it as-is for now then.

PR #2504, opened 2026-07-03
coordinator: fix anti-affinity by moving pod-role to label

burgerdev:
The Contrast Coordinator used to specify  an anti-affinity based on pod role label, which was ineffective because the pod role was, in fact, an annotation. With this change, we're moving the pod role to a label.

---

Labels are a bit more restrictive than annotations, but we can easily fit the pod role into a label. Doing so allows using the pod role in selectors, such as for anti-affinity (where we're already selecting by the label, which did not have an effect) or in the peerdiscovery watcher. Having a label that is separate from the pod selector labels also allows exposing Coordinator's from different controllers with services.

Fixes CON-213.
Fixes CON-206.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -336,8 +336,7 @@ func isCoordinator(resource any) bool {
 		r.Spec != nil &&
 		r.Spec.Template != nil &&
 		r.Spec.Template.ObjectMetaApplyConfiguration != nil &&
-		r.Spec.Template.Annotations != nil &&
-		r.Spec.Template.Annotations[kuberesource.ContrastRoleAnnotationKey] == string(manifest.RoleCoordinator) {
+		r.Spec.Template.Labels[kuberesource.ContrastRoleLabelKey] == string(manifest.RoleCoordinator) {
 		return true
 	}
 	return false
--- a/cli/verifier/versions_match.go
+++ b/cli/verifier/versions_match.go
@@ -40,8 +40,8 @@ func (v *VersionsMatch) Verify(toVerify any) error {
 		}
 
 		var podRole string
-		if meta != nil && meta.Annotations != nil {
-			podRole = meta.Annotations[kuberesource.ContrastRoleAnnotationKey]
+		if meta != nil {
+			podRole = meta.Labels[kuberesource.ContrastRoleLabelKey]
 		}
 		for _, container := range spec.Containers {
 			if !needsImageVersionCheck(*container.Name, podRole) {
--- a/internal/kuberesource/constants.go
+++ b/internal/kuberesource/constants.go
@@ -9,10 +9,10 @@ const (
 )
 
 const (
-	// ContrastRoleAnnotationKey assigns a role to the annotated pod.
+	// ContrastRoleLabelKey assigns a role to the labeled pod.
 	//
 	// This is used to determine whether a given pod may act as Coordinator.
-	ContrastRoleAnnotationKey = annotationPrefix + "pod-role"
+	ContrastRoleLabelKey = annotationPrefix + "pod-role"
 
 	// ExposeServiceAnnotationKey is the annotation key used to specify whether a Service should be exposed via a LoadBalancer.
 	ExposeServiceAnnotationKey = annotationPrefix + "expose-service"
--- a/internal/kuberesource/mutators.go
+++ b/internal/kuberesource/mutators.go
@@ -300,7 +300,7 @@ func SetCollateralProxyEnv(resources []any, proxyURL string) []any {
 	out := make([]any, 0, len(resources))
 	for _, resource := range resources {
 		out = append(out, MapPodSpecWithMeta(resource, func(meta *applymetav1.ObjectMetaApplyConfiguration, spec *applycorev1.PodSpecApplyConfiguration) (*applymetav1.ObjectMetaApplyConfiguration, *applycorev1.PodSpecApplyConfiguration) {
-			if meta == nil || meta.Annotations[ContrastRoleAnnotationKey] != string(manifest.RoleCoordinator) {
+			if meta == nil || meta.Labels[ContrastRoleLabelKey] != string(manifest.RoleCoordinator) {
 				return meta, spec
 			}
 			for i := range spec.Containers {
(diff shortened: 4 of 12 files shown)


internal/kuberesource/constants.go
```diff
@@ -9,10 +9,10 @@ const (
 )
 
 const (
-	// ContrastRoleAnnotationKey assigns a role to the annotated pod.
+	// ContrastRoleLabelKey assigns a role to the labeled pod.
 	//
 	// This is used to determine whether a given pod may act as Coordinator.
-	ContrastRoleAnnotationKey = annotationPrefix + "pod-role"
+	ContrastRoleLabelKey = annotationPrefix + "pod-role"
```
davidweisse, line 15: This should also be used here:

https://github.com/edgelesssys/contrast/blob/3f6cc2fadc6e61ad806ffd008cdf6a9202107a44/internal/kuberesource/parts.go#L389

internal/kuberesource/parts.go
```diff
@@ -120,7 +120,7 @@ func NodeInstaller(namespace string, platform platforms.Platform) (*applyappsv1.
 					PodTemplateSpec().
 						WithLabels(map[string]string{"app.kubernetes.io/name": name}).
 						WithAnnotations(map[string]string{
-							ContrastRoleAnnotationKey:            string(manifest.RoleNodeInstaller),
+							ContrastRoleLabelKey:                 string(manifest.RoleNodeInstaller),
```
sespiros: Flagged by AI review: This is still under `WithAnnotations`. Nothing in Go reads it and `packages/cleanup-bare-metal.sh:48` still selects it via `.metadata.annotations[...]`.

internal/kuberesource/parts.go
sespiros:
```suggestion
																		WithMatchLabels(map[string]string{ContrastRoleLabelKey: string(manifest.RoleCoordinator)}),
```

sespiros (approved):
Thanks.

AI also flagged a needed change in https://github.com/edgelesssys/contrast/blob/de312912326847e47012c37805a4f65a01c60c04/rfc/010-distributed-coordinator.md?plain=1#L156

internal/kuberesource/mutators.go
```diff
@@ -300,7 +300,7 @@ func SetCollateralProxyEnv(resources []any, proxyURL string) []any {
 	out := make([]any, 0, len(resources))
 	for _, resource := range resources {
 		out = append(out, MapPodSpecWithMeta(resource, func(meta *applymetav1.ObjectMetaApplyConfiguration, spec *applycorev1.PodSpecApplyConfiguration) (*applymetav1.ObjectMetaApplyConfiguration, *applycorev1.PodSpecApplyConfiguration) {
-			if meta == nil || meta.Annotations[ContrastRoleAnnotationKey] != string(manifest.RoleCoordinator) {
+			if meta == nil || meta.Labels == nil || meta.Labels[ContrastRoleLabelKey] != string(manifest.RoleCoordinator) {
```
davidweisse:
```suggestion
			if meta == nil || meta.Labels[ContrastRoleLabelKey] != string(manifest.RoleCoordinator) {
```
Accesses to a nil map return the default value: https://go.dev/ref/spec#Map_types.

internal/kuberesource/mutators.go
```diff
@@ -724,7 +724,8 @@ func PatchCoordinatorMetrics(resources []any) []any {
 		case *applyappsv1.StatefulSetApplyConfiguration:
 			if r.Spec != nil && r.Spec.Template != nil && r.Spec.Template.Spec != nil &&
 				len(r.Spec.Template.Spec.Containers) > 0 &&
-				r.Spec.Template.Annotations[ContrastRoleAnnotationKey] == string(manifest.RoleCoordinator) {
+				r.Spec.Template.Labels != nil &&
+				r.Spec.Template.Labels[ContrastRoleLabelKey] == string(manifest.RoleCoordinator) {
```
davidweisse:
```suggestion
				r.Spec.Template.ObjetcMetaApplyConfiguration != nil &&
				r.Spec.Template.Labels[ContrastRoleLabelKey] == string(manifest.RoleCoordinator) {
```

Same here, but `Labels` is part of an embedded pointer struct, so we should check that.

internal/kuberesource/mutators_test.go
```diff
@@ -809,7 +809,7 @@ func TestSetCollateralProxyEnv(t *testing.T) {
 	assert := assert.New(t)
 
 	coordinatorPod := applycorev1.Pod("coordinator", "default").
-		WithAnnotations(map[string]string{ContrastRoleAnnotationKey: "coordinator"}).
+		WithLabels(map[string]string{ContrastRoleLabelKey: "coordinator"}).
```
davidweisse:
```suggestion
		WithLabels(map[string]string{ContrastRoleLabelKey: string(manifest.RoleCoordinator)}).
```

cli/cmd/generate.go
```diff
@@ -336,8 +336,8 @@ func isCoordinator(resource any) bool {
 		r.Spec != nil &&
 		r.Spec.Template != nil &&
 		r.Spec.Template.ObjectMetaApplyConfiguration != nil &&
-		r.Spec.Template.Annotations != nil &&
-		r.Spec.Template.Annotations[kuberesource.ContrastRoleAnnotationKey] == string(manifest.RoleCoordinator) {
+		r.Spec.Template.Labels != nil &&
+		r.Spec.Template.Labels[kuberesource.ContrastRoleLabelKey] == string(manifest.RoleCoordinator) {
```
davidweisse:
```suggestion
		r.Spec.Template.Labels[kuberesource.ContrastRoleLabelKey] == string(manifest.RoleCoordinator) {
```

cli/verifier/versions_match.go
```diff
@@ -40,8 +40,8 @@ func (v *VersionsMatch) Verify(toVerify any) error {
 		}
 
 		var podRole string
-		if meta != nil && meta.Annotations != nil {
-			podRole = meta.Annotations[kuberesource.ContrastRoleAnnotationKey]
+		if meta != nil && meta.Labels != nil {
```
davidweisse:
```suggestion
		if meta != nil {
```

internal/kuberesource/mutators.go
```diff
@@ -494,8 +494,8 @@ func AddLogging(resources []any, level, subsystem string) []any {
 	for _, resource := range resources {
 		switch r := resource.(type) {
 		case *applyappsv1.StatefulSetApplyConfiguration:
-			if r.Spec != nil && r.Spec.Template != nil &&
-				r.Spec.Template.Annotations[ContrastRoleAnnotationKey] == string(manifest.RoleCoordinator) {
+			if r.Spec != nil && r.Spec.Template != nil && r.Spec.Template.Labels != nil &&
```
davidweisse:
```suggestion
			if r.Spec != nil && r.Spec.Template != nil && r.Spec.Template.ObjectMetaApplyConfiguration != nil &&
```

internal/kuberesource/constants.go
burgerdev: That was resolved by @sespiros suggestion.

burgerdev (commented):
I addressed the comments, updated the cleanup script and removed the unnecessary platform annotation (ref #2510).

davidweisse (approved):

PR #2507, opened 2026-07-06
e2e: use smaller image for imagestore test

davidweisse:
Changes the image used to test the auto-memory-limit flag. The compressed postgres image is ~150Mi, so it should be relatively quick to pull locally during generate. The compressed+uncompressed size is ~815Mi, which is enough that the pod does not come up on a GPU platform (which has increased default memory) when auto-memory-limits are disabled.

- [x] https://github.com/edgelesssys/contrast/actions/runs/28788831601


burgerdev (approved):
lgtm, thanks!

Note https://linear.app/edgelesssys/issue/CON-218, though.

PR #2508, opened 2026-07-06
treewide: implement fully CRA-compliant, largely BSI-TR-03183-2 compliant SBOMs

charludo:
Supersedes #2428.

General overview:
- use crate2nix fork to add additional metadata to Cargo.nix
- use Cargo.nix to generate rust package SBOMs
- use cyclonedx-gomod + slight post-processing to generate go package SBOMs
- use patched bombon to generate SBOMs based on our nix pkgs, including ingesting those package-level SBOMs
- use an unholy amount of jq to post-process into something resembling BSI-compliance
- since BSI wants a timestamp, add a finalize script that (non-reproducibly, of course) adds that
- build the SBOMs (split by category as in the original tickets) in CI and publish as part of the release flow

Useful for testing:
- `nix run .#base.scripts.finalize-sbom -- cli > contrast-cli.cdx.json`
- https://cyclonedx.github.io/Sunshine/ (nice visualization, component search, but limited metadata view)
- https://sbomviewer.com/ (better metadata view, scan for vulns)

(Edit: sorry for the diff, but >2k lines come from autogenerated Cargo.nix 😄)

### DIFF
--- a/packages/by-name/buildGoModuleSbom/package.nix
+++ b/packages/by-name/buildGoModuleSbom/package.nix
@@ -1,0 +1,93 @@
+# Copyright 2026 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+{
+  lib,
+  runCommand,
+  cyclonedx-gomod,
+  jq,
+  go,
+  git,
+  cacert,
+}:
+
+{
+  package,
+  mains ? package.subPackages or [ "." ],
+  preAnalyze ? package.postConfigure or "",
+  pname ? package.pname,
+}:
+
+let
+  moduleRoot =
+    if package ? sourceRoot then lib.removePrefix "${package.src.name}/" package.sourceRoot else ".";
+  tags = package.tags or [ ];
+  tagsFlag = lib.optionalString (tags != [ ]) "-tags=${lib.concatStringsSep "," tags}";
+in
+
+lib.throwIfNot (package.proxyVendor or false)
+  "buildGoModuleSbom: ${pname} must set proxyVendor = true; cyclonedx-gomod needs a module proxy / go mod graph, a vendor directory is insufficient"
+
+  runCommand
+  "${pname}-sbom"
+  {
+    nativeBuildInputs = [
+      cyclonedx-gomod
+      jq
+      go
+      git
+      cacert
+    ];
+  }
+  # bash
+  ''
+    export HOME=$TMPDIR
+    export XDG_CACHE_HOME=$TMPDIR/xdg-cache
+
+    cp -r --no-preserve=mode,ownership ${package.src} src
+    pushd src >/dev/null
+    git init -q -b main
+    git -c user.email=sbom@contrast -c user.name=sbom add -A
+    git -c user.email=sbom@contrast -c user.name=sbom \
+      -c commit.gpgsign=false commit -q -m sbom
+    popd >/dev/null
+
+    pushd src/${moduleRoot} >/dev/null
+    ${preAnalyze}
+
+    export GOPROXY="file://${package.goModules}"
+    export GOSUMDB=off
+    export GOMODCACHE=$TMPDIR/gomodcache
+    export GOCACHE=$TMPDIR/gocache
+    ${lib.optionalString (tagsFlag != "") ''export GOFLAGS="${tagsFlag}"''}
+    export CGO_ENABLED=0
+
+    mkdir -p "$out"
+    for main in ${lib.escapeShellArgs mains}; do
+      outName=$(echo "$main" | tr '/' '-')
+      [ "$outName" = "." ] && outName=${pname}
+
+      cyclonedx-gomod app \
+        -json \
+        -licenses \
+        -output-version 1.5 \
+        -main "$main" \
+        -output "$out/$outName.cdx.json" \
+        .
+
+      # cyclonedx-gomod's quirks:
+      #  - promote `evidence.licenses` to the concluded `licenses`
+      #  - derive source-code `vcs` reference from the module path
+      outFile="$out/$outName.cdx.json"
+      jq '
+        def promote:
+          if (.evidence.licenses // [] | length) > 0 then .licenses = .evidence.licenses else . end;
+        def with_vcs:
+          if (.externalReferences // []) | any(.type == "vcs") then .
+          else .externalReferences = ((.externalReferences // [])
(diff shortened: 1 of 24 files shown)


packages/by-name/buildGoModuleSbom/package.nix
```diff
@@ -0,0 +1,93 @@
...
+    export GOPROXY="file://${package.goModules}"
+    export GOSUMDB=off
+    export GOMODCACHE=$TMPDIR/gomodcache
+    export GOCACHE=$TMPDIR/gocache
+    ${lib.optionalString (tagsFlag != "") ''export GOFLAGS="${tagsFlag}"''}
+    export CGO_ENABLED=0
+
+    mkdir -p "$out"
+    for main in ${lib.escapeShellArgs mains}; do
+      out_name=$(echo "$main" | tr '/' '-')
```
burgerdev: Nit: I believe we usually do camelCase to silence shell checks that insist on curly braces.

packages/by-name/sbom/finish.jq
burgerdev, line 1: We can keep it for this PR, but I do think this is too complex for a jq script.

packages/by-name/sbom/package.nix
```diff
@@ -0,0 +1,106 @@
...
+    };
+    runtimeConfidential = {
+      inherit (contrastPkgs.contrast) coordinator initializer;
+      inherit (contrastPkgs.kata) image agent;
+      guestSystem = contrastPkgs.kata.image.passthru.toplevel;
+      inherit (contrastPkgs) OVMF-SNP OVMF-TDX;
+    };
+    runtimeNonConfidential = {
+      inherit (contrastPkgs.contrast) nodeinstaller;
+      inherit (contrastPkgs) imagepuller imagestore service-mesh;
```
burgerdev: These three are all confidential, no?

packages/by-name/sbom/package.nix
```diff
@@ -0,0 +1,106 @@
...
+
+  categories = {
+    cli = {
+      inherit (contrastPkgs.contrast) cli;
+      # The CLI go:embeds the (static) genpolicy binary. Embedded bytes carry no store-path reference, so it never appears in the CLI's runtime closure.
+      inherit (contrastPkgsStatic.kata) genpolicy;
+    };
+    runtimeConfidential = {
+      inherit (contrastPkgs.contrast) coordinator initializer;
+      inherit (contrastPkgs.kata) image agent;
```
burgerdev, line 88: We're missing the initdata-processor.

.gitignore
```diff
@@ -27,3 +27,4 @@ uplosi.conf*
 .tmp*
 .lycheecache
 .pre-commit-config.yaml
+*.sqlite
```
burgerdev: Where does this come from? I'd rather not have random DB files in my code tree.

packages/by-name/sbom/package.nix
```diff
@@ -0,0 +1,106 @@
...
+      target = linkFarm "contrast-${name}-matrix" entries;
+    in
+    withMetadata categoryId.${name} target (
+      buildBom target {
+        extraPaths = vendoredSbomsOf entries;
+      }
+    );
+
+  categories = {
+    cli = {
```
burgerdev, line 80: We could also consider splitting into "part of TCB" and its complement. CLI would be in the former, then.

packages/by-name/sbom/package.nix
```diff
@@ -0,0 +1,106 @@
...
+      inherit (contrastPkgs.contrast) cli;
+      # The CLI go:embeds the (static) genpolicy binary. Embedded bytes carry no store-path reference, so it never appears in the CLI's runtime closure.
+      inherit (contrastPkgsStatic.kata) genpolicy;
```
burgerdev, line 84: I'm confused, aren't you fixing this in d018a33b3919fe3802e4f88931fd6cbf46c9f91b?

packages/by-name/sbom/finish.jq
charludo: So rather a go tool?

packages/by-name/sbom/package.nix
charludo: Eh yup. Sorry.

packages/by-name/sbom/package.nix
charludo: I think this is my biggest issue with this whole thing, having to manually keep a list of components. `.#base.sbom` contains *everything*, but everything also includes things like our dev scripts....

.gitignore
charludo: I'm not sure. Might have come from grype, I just remember for a while it came popping back up in my tree, but I can't reproduce that anymore. Alternatively from some of the other tools I had initially eyed, but IDK.Removed.

packages/by-name/sbom/package.nix
charludo: I don't have a strong opinion, the current split is simply what CON-139 suggested.

packages/by-name/sbom/package.nix
charludo: That's about two different things. The commit you mentioned just makes `bombonVendoredSbom` run after the embed, but what's being embedded is still just a blob. Adding this here as an explicit root pulls in genpolicy's actual SBOM component and vendored SBOM.

(In other words, the linked commit is required to make building the SBOM work at all, but then it's still missing genpolicy's tree.)

Amended the comment.

packages/by-name/sbom/finish.jq
burgerdev: Yes, I'd say so.

packages/by-name/sbom/package.nix
```diff
@@ -80,17 +80,24 @@ let
     cli = {
       inherit (contrastPkgs.contrast) cli;
       # The CLI go:embeds the (static) genpolicy binary. Embedded bytes carry no store-path reference, so it never appears in the CLI's runtime closure.
+      # Adding genpolicy as an explicit root is pulls in its Rust component and vendored SBOM.
```
burgerdev: There's a stray "is" here.

packages/by-name/sbom/package.nix
burgerdev: Ok, let's keep it as is for now.

burgerdev (approved):
lgtm, thanks!

thomasten (commented):
Love it! I'm not skilled to review the nix stuff though :)

PR #2509, opened 2026-07-06
rfc: client-side proxy

burgerdev:
[rendered version](https://github.com/edgelesssys/contrast/blob/rfc/014/rfc/014-http-proxies.md)

### DIFF
--- a/rfc/014-http-proxies.md
+++ b/rfc/014-http-proxies.md
@@ -1,0 +1,158 @@
+# RFC 014: HTTP Proxies
+
+## Background
+
+Use of HTTP is pervasive in modern applications, and we expect many Contrast workloads exposing an HTTP service, especially at the cluster edge.
+Pretty much all HTTP clients integrate with the so-called [web PKI], which tightly controls who is authorized to sign certificates for the web.
+Contrast, on the other hand, has its own PKI with signing policies and semantics that differ from the web PKI.
+This prevents regular HTTP clients from interacting with Contrast in the intended fashion.
+
+While it's generally possible to configure clients with additional CA certificates, it's sometimes not easy or may lead to undesirable side effects.
+Browsers, in particular, are often configured centrally and don't allow individuals to add CA certificates for compliance reasons.
+But even if a central authority were to bless the CA certificate, care must be taken to not weaken other connections.
+
+Integrating Contrast logic into standard HTTP clients is infeasible, so we're looking for an alternative solution to make standard HTTP clients compatible with workloads using Contrast certificates.
+
+[web PKI]: https://icannwiki.org/Web_PKI
+
+## Requirements
+
+- Allow standard HTTP clients to connect to servers running behind a service mesh ingress or using the mesh certificate directly.
+- While keeping the traffic confidential up to the client's environment (local machine or internal network).
+- Without increasing risk for applications in the cluster.
+
+## Proposal
+
+This proposal relies on the HTTP `CONNECT` verb and the associated `HTTP_PROXY` settings, which are [widely supported across browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/CONNECT#browser_compatibility) and libraries.
+The overall idea is to deploy a local proxy that terminates TLS traffic using Contrast certificates and re-encrypts them with a web PKI certificate.
+
+### Setup example
+
+In order to make the proposal more concrete, we're going to explain it alongside an example setup.
+The Kubernetes cluster has a confidential `web service`, for example the Emojivoto one.
+It's exposed on a public IP using the Gateway API, possibly together with other services.
+Service mesh ingress is configured to not require a client certificate.
+DNS is configured to resolve the domain `web.cc.cloud` to the gateway IP.
+On premises, there's a proxy running at `proxy.internal` that can connect to the public gateway IP.
+
+```mermaid
+graph TB
+    subgraph OnPrem["On-Premises Environment"]
+        Client[Client]
+        Bot[Bot]
+        Proxy[Proxy]
+        Client -->|CONNECT web.cc.cloud:443| Proxy
+        Proxy -->|Re-encrypt| Proxy
+        Bot --> Proxy
+    end
+
+    subgraph Cloud["Cloud Environment"]
+        LB[Gateway]
+        subgraph K8s["Kubernetes Cluster"]
+            App1[Web service]
+            App2[Voting service]
+        end
+            LB -->|web.cc.cloud| App1
+            LB -->|voting.cc.cloud| App2
+    end
+
+    Proxy -->|ClientHello<br>SNI: web.cc.cloud<br>CAs: mesh-ca.pem| LB
+```
+
+### Client configuration
+
+The client browser is configured to use `proxy.internal` as its `HTTP_PROXY`.
+
+### Forward proxy
+
+The forward proxy runs in a trusted environment and listens for `CONNECT` requests.
+It's configured with the following:
+
+- A CA certificate returned by `contrast verify` (mesh CA or root CA).
+  See the discussion in the appendix about integrating Contrast verification logic directly.
+- A list of hosts and ports that are allowed to be tunnelled.
+  We can introduce a special "everything" syntax, if needed, to accelerate development.
+- The certificate for the domain names used by backend services (`web.cc.cloud` for the example).
+
+The life of a proxied connection then looks like this:
+
+0. The client browses to `https://web.cc.cloud`.
+1. The browser issues a `CONNECT web.cc.cloud:443` request to the proxy.
+2. The proxy establishes a TCP connection to the target service.
+   If that succeeds, the proxy returns `200 OK` to the client to indicate tunnel establishment, otherwise a `503 Service Unavailable`.
+3. The browser initiates a TLS connection over the tunnel, expecting a web PKI certificate for `web.cc.cloud`.
+4. The proxy receives the `ClientHello` and initiates a similar TLS connection (ALPN) to the backend, verified with the Contrast CA certificate.
+5. After the backend session is established, the proxy responds to the client with similar protocol options and uses its web PKI certificate for signing.
+6. When both TLS sessions are established, the proxy routes traffic back and forth until the sessions close.
+
+```mermaid
+sequenceDiagram
+    participant Client as Client
+    participant FwdProxy as Forward Proxy
+    participant Backend as Backend Service
+
+    Note over FwdProxy,Backend: On-Prem <=> Kubernetes Cluster
+
+    Client->>FwdProxy: HTTPS CONNECT request
+    FwdProxy->>Backend: TCP SYN
+    Backend->>FwdProxy: TCP SYN/ACK
+    FwdProxy->>Client: 200 Connection Established
+
+
+    Client->>FwdProxy: TLS handshake
+
+    FwdProxy->>Backend: TLS handshake
+    Backend->>FwdProxy: TLS handshake done<br>(Contrast cert)
+
+    FwdProxy->>Client: TLS handshake done<br>(web PKI cert)
+
+
+    Client->>FwdProxy: HTTP request<br>(over TLS)
+    FwdProxy->>Backend: HTTP request<br>(reencrypted)
+    Backend->>FwdProxy: HTTP response
+    FwdProxy->>Client: HTTP response<br>(reencrypted)
+
+# https://mermaid.ai/d/7acceba4-9e72-4cbd-ab06-54f125d19546
+```
+
+## Alternatives considered
+
+### Application gateway
+
+Instead of a simple reverse proxy, we could come up with a full-fledged application gateway.
+Such a gateway could route requests to backends and take care of user authentication/authorization.
+
+- There's no standard, and thus it's very likely that our gateway would be unusable for most customers.
+- If the gateway terminates TLS traffic, it would need to run inside a TEE itself, making it again more complicated to expose.
+
+### Proxy at the transport layer
+
+- No standard for such proxy (SOCKS5, but not widely usable).
+- Thus, would need a listener for each remote service (like service mesh outbound)
+
+### Hijack local traffic (`TPROXY`)
+
+- Needs elevated privileges and is hard to get right.
+
+### Embed Coordinator certificate into web PKI
+
+- Fails security goals (any web CA can mint certificates), unless attestation is included.
+- Will require attested TLS availability in client software (far out).
+
+### Configure trust store with Contrast CA certs
+
+This is the current status quo, which works but may be inconvenient, especially in large orgs.
+Those large orgs are usually running their own CAs, so they could sign the mesh CA key with another intermediate, possibly adding name constraints.
+
+## Appendix
+
+### Contrast integration
+
+This proposal deliberately focuses on the practical TLS-side of things without talking about Contrast verification.
+However, the proxy could trivially be extended with Contrast functionality, or a verifier could run next to the proxy, supplying it with certificate updates.
+
+The main question here is how such a verifier would learn about a Contrast manifest update, and how it would verify the new manifest.
+It would need to be configured with a source of trusted manifest histories, a transparency log for example.
+Furthermore, it would need direct access to the Coordinator, which may not be a desirable configuration (DoS concerns).
+
+Since these issues are orthogonal to the goals of this RFC, they aren't discussed here.
--- a/tools/vale/styles/config/vocabularies/edgeless/accept.txt
+++ b/tools/vale/styles/config/vocabularies/edgeless/accept.txt
@@ -2,6 +2,7 @@
 Affero
 Agent
 agent
+allowlist
 Ansys
 APIs?
 ASG


rfc/014-http-proxies.md
```diff
@@ -0,0 +1,120 @@
+# HTTP Proxies
```
charludo:
```suggestion
# RFC 014: HTTP Proxies
```
nit nit nit

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Background
+
+Use of HTTP is pervasive in modern applications, and we expect many Contrast workloads exposing an HTTP service, especially at the cluster edge.
+Pretty much all HTTP clients integrate with the so-called [web PKI], which tightly controls who is authorized to sign certificates for the web.
+Contrast, on the other hand, has its own PKI with signing policies and semantics that differ from the web PKI.
+This prevents regular HTTP clients from interacting with Contrast in the intended fashion.
+
+While it's generally possible to configure clients with additional CA certificates, it's sometimes not easy or may lead to undesirable side effects.
+For example, a Contrast CA certificate added to a browser trust store would be accepted for connections that aren't Contrast related.
```
charludo: Is this strictly true though. AFAIK, you can add `nameConstraints` to CA certificates. It seems to me that, if an organization were to run all contrast deployments, and only contrast deployments, under e.g. `.contrast.internal`, and installed the corresponding constraint cert in browsers, then that would be the better / actually e2e solution.

(Not saying we should recommend this *instead* of the proxy, but IMO it should be advised as an alternative.)

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,120 @@
...
+The forward proxy listens on a local port for `CONNECT` requests.
+It's configured with the following:
+
+1. A CA certificate returned by `contrast verify` (mesh CA or root CA).
+2. A list of hosts and ports that are allowed to be tunnelled.
+   We can introduce a special "everything" syntax, if needed, to accelerate development.
+
+When the proxy receives a `CONNECT` request, it establishes a TLS connection to the target service.
+This connection could itself be proxied, see the [reverse proxy section](#reverse-proxy-optional).
+After successful validation of the target service with the CA certificate, the proxy forwards HTTP requests from the client to the backend and vice versa.
```
charludo: I have some trouble seeing how this fulfills the goal of

> Without sacrificing e2e security of the application traffic.

I *could* see it if we explicitly assume here that the forward proxy runs on each client device, and listens only on loopback or something like that.

But from yesterday's conversation I gathered that this will most likely be deployed somewhere in e.g. a company's secure subnet / intranet / .... Even if the proxy would then be configured to use a web PKI cert for HTTPS between client and proxy, that's not e2e.

(The idea that it's e2e between the entire secure subnet and the deployment is not satisfactory IMO, since e2e usually does imply that the client device is one of those ends.)

charludo (commented):
Some preliminary thoughts

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Requirements
+
+- Allow standard HTTP clients to connect to servers running behind a service mesh ingress or using the mesh certificate directly.
+- Without sacrificing e2e security of the application traffic.
+- Without increasing risk for applications in the cluster.
+
+## Proposal
+
+This proposal relies on the HTTP `CONNECT` verb and the associated `HTTP_PROXY` settings, which are [widely supported across browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/CONNECT#browser_compatibility) and libraries.
```
sespiros, line 26: As specified, the forward proxy won't work for browsers, only for CLI tools. Browsers use CONNECT exclusively for https:// URLs, and then perform the TLS handshake themselves inside the tunnel against the browser trust store, so they hit the mesh certificate directly, which is the problem this RFC is solving. For http:// URLs, browsers don't send CONNECT at all; they send the request to the proxy in absolute form (GET http://host/path HTTP/1.1, RFC 9112 section 3.2.2). I verified both behaviors with Firefox against a logging proxy; the only client producing the diagrammed CONNECT-then-plain-HTTP flow was `curl --proxytunnel`.

Suggestion: keep the design, but specify the client leg as two modes. For browsers, absolute-form proxying of http:// URLs, where the proxy upgrades the outbound leg to Contrast TLS. For configurable clients (curl, SDKs), CONNECT tunneling as currently described. 

Worth a note that in the browser mode pages are insecure origins (no Secure cookies, WebCrypto, service workers) and redirects to https:// or HSTS-preloaded domains will fail.

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,120 @@
...
+    Backend->>FwdProxy: HTTPS response (tunnelled by proxy)
+    FwdProxy->>Client: HTTP  response
+
+# https://mermaid.ai/d/7acceba4-9e72-4cbd-ab06-54f125d19546
+```
+
+### Forward proxy (required)
+
+The forward proxy listens on a local port for `CONNECT` requests.
+It's configured with the following:
```
sespiros: Also list somewhere as a prerequisite that applications accessible by that proxy need to have their ingress disable client authentication.

sespiros: What name does the proxy verify the backend cert against? The dialed target (LB hostname, port-forwarded localhost) isn't in the default SANs, so this only works if users add it to the manifest SANs first, worth stating as a prerequisite. Skipping name verification isn't an option: it'd accept any workload of the deployment, so the infra could route the proxy to a different one.

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,120 @@
...
+
+1. A CA certificate returned by `contrast verify` (mesh CA or root CA).
```
sespiros: Suggest pinning a manifest here instead of a CA cert, with the proxy verifying via the SDK and deriving the mesh CA itself: a static CA file goes stale on every manifest update, and the manifest is the actual trust anchor. privatemode-proxy already does exactly this (GetAttestedMeshCA, with static and dynamic manifest modes), so this could reuse that machinery instead of introducing a weaker parallel path.

rfc/014-http-proxies.md
burgerdev: As far as I can tell this is not true. If you look at RFC 9112 3.2.2 and 3.2.3, it explicitly talks about `CONNECT` and that it uses the authority-form. How did you configure Firefox with the proxy? I used FoxyProxy, and I get the expected connect requests.

rfc/014-http-proxies.md
burgerdev: The target of the  incoming `CONNECT` request. But then the server side proxy is required, after all, for resolving the in-cluster addresses.

rfc/014-http-proxies.md
burgerdev: If anything, we'd need to pin a manifest _history_, not only a manifest. PM gets away with this because the deployment is immutable (iirc). The pinned history would be equally stale on manifest updates, though, and cause more connectivity headache (Coordinator needs to be exposed when the proxy comes up).

rfc/014-http-proxies.md
burgerdev: Yes, maybe the goal is too ambitious. What I was actually aiming for is exchanging full e2e security with Contrast certs to "secure until on data owners premises", which would be an individual running the proxy on localhost, or an organization running the proxy internally and applying protections as they see fit.

rfc/014-http-proxies.md
burgerdev: We don't have that feature right now (adding constraints to CA certificates), but maybe we should.

rfc/014-http-proxies.md
burgerdev: If the in-cluster component is undesired, you'd need a SAN entry and a load balancer anyway.

rfc/014-http-proxies.md
sespiros: Tested on Firefox 152.0.5 against a logging proxy (manual proxy config, which is the same machinery FoxyProxy drives): navigating to `http://host/path` produces `GET http://host/path HTTP/1.1` in absolute-form, no CONNECT; a schemeless navigation to `host/path` does the same. Every CONNECT the proxy logged was for an `https://` destination (page loads and Firefox's own telemetry), and in every single case the first in-tunnel bytes after `200 Connection Established` were Firefox's own TLS ClientHello (`0x16 0x03 ...`), never plaintext HTTP. That second part is the load-bearing claim: when a browser does send CONNECT, it runs TLS itself against its own trust store, so the forward proxy can't terminate the mesh TLS for it, and the user gets the exact cert error this RFC wants to remove. The only client that produced the diagrammed CONNECT-then-plain-HTTP was `curl --proxytunnel`.

Can you share your repro? Specifically the URL scheme you navigated to, the FoxyProxy proxy type (HTTP vs HTTPS vs SOCKS), and what your proxy saw inside the tunnel after answering the CONNECT. Two guesses at the reconciliation: either the test URL was `https://`, or HTTPS-First upgraded the navigation to `https://` before the proxy saw it; in both cases the CONNECT is expected and is followed by browser-run TLS, which is the problem, not the mechanism.

<details>
<summary>Logging proxy + repro steps</summary>

Sample output from a run: every CONNECT is an `https://` destination followed by a ClientHello; the `http://` navigation arrives as `GET http://proxytest.invalid/from-firefox HTTP/1.1` with no CONNECT (as does Firefox's own captive-portal probe, `GET http://detectportal.firefox.com/canonical.html`).

```python
#!/usr/bin/env python3
"""Fake HTTP proxy that logs what the client actually sends.

For a CONNECT request it replies "200 Connection Established" and then
logs the first bytes the client sends *inside* the tunnel, so we can see
whether the client starts TLS (0x16 0x03 ...) or speaks plain HTTP.
For any other request it logs the request line and returns a tiny page.
"""
import socket
import sys
import threading

HOST, PORT = "127.0.0.1", 3128
lock = threading.Lock()


def log(msg):
    with lock:
        print(msg, flush=True)


def handle(conn, addr):
    conn.settimeout(10)
    try:
        data = b""
        while b"\r\n\r\n" not in data:
            chunk = conn.recv(4096)
            if not chunk:
                break
            data += chunk
        if not data:
            return
        reqline = data.split(b"\r\n", 1)[0].decode("latin1")
        log(f"REQUEST LINE: {reqline}")
        if reqline.startswith("CONNECT"):
            conn.sendall(b"HTTP/1.1 200 C")
            try:
                inner = conn.recv(16)
            except socket.timeout:
                log("  IN-TUNNEL: (client
                return
            if not inner:
                log("  IN-TUNNEL: (client closed connection)")
                return
            if inner[:1] == b"\x16":
                ver = f"{inner[1]}.{inner
                log(f"  IN-TUNNEL: TLS handshake record (ClientHello), first bytes {inner[:6].hex()} (record
version {ver})")
            else:
                log(f"  IN-TUNNEL: plaint
        else:
            body = b"<html><body>proxyspy
            conn.sendall(
                b"HTTP/1.1 200 OK\r\nContt-Length: "
                + str(len(body)).encode()
                + b"\r\nConnection: close
                + body
            )
    except Exception as e:
        log(f"  error: {e}")
    finally:
        conn.close()


def main():
    srv = socket.socket(socket.AF_INET, s
    srv.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    srv.bind((HOST, PORT))
    srv.listen(16)
    log(f"proxyspy listening on {HOST}:{P
    while True:
        conn, addr = srv.accept()
        threading.Thread(target=handle, args=(conn, addr), daemon=True).start()


if __name__ == "__main__":
    sys.exit(main())
```

Run `python3 proxyspy.py`, point the cliethe log; the interesting field is IN-TUNNEL after any CONNECT.

```sh
# curl, three modes
curl -x http://127.0.0.1:3128 http://proxytest.invalid/hello    # absolute-form GET, no CONNECT
curl -p -x http://127.0.0.1:3128 http://pECT + in-tunnel plaintext (the diagrammedflow)
curl -x http://127.0.0.1:3128 https://proECT + in-tunnel ClientHello

# Firefox, throwaway profile with the pro
mkdir -p ffprofile
cat > ffprofile/user.js <<'EOF'
user_pref("network.proxy.type", 1);
user_pref("network.proxy.http", "127.0.0.
user_pref("network.proxy.http_port", 3128);
user_pref("network.proxy.ssl", "127.0.0.1
user_pref("network.proxy.ssl_port", 3128);
EOF
# macOS: firefox is not on PATH, use the app bundle binary
FIREFOX="/Applications/Firefox.app/ContenFIREFOX=firefox
"$FIREFOX" --headless --no-remote --profile ffprofile "http://proxytest.invalid/from-firefox"
```

</details>

rfc/014-http-proxies.md
sespiros: On the PM point: yes, its attest getter hard-fails unless the history has exactly one manifest (which we can't assume for a generic Contrast deployment). So history is actually the object to pin.

On the staleness point: a stale CA fails with a more generic TLS error while a stale manifest pin fails as "Coordinator enforces manifest X, expected Y". Also it's a question of what you distribute, a CA (which is a derived artifact) or a new manifest. The manifest (or a manifest history) could also be distributed via a remote source that we publish similar to what the pm proxy does when you don't pass a manifest to it.

On the connectivity point: the CA has the same dependency no? Coordinator needs to be exposed also to whoever runs verify (in order to then distribute the CA to the proxy). If a reverse proxy exists, it could relay to the Coordinator too.

rfc/014-http-proxies.md
charludo: > secure until on data owners premises

That I can get behind with the proposed architecture.

burgerdev:
Folks, I updated the proposal, ptal.

rfc/014-http-proxies.md
burgerdev: I added a section for why I don't mean to solve this here.

rfc/014-http-proxies.md
burgerdev: You were right, my testing setup was bad. I ended up rewriting the proposal to require HTTPS.

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,158 @@
...
+
+This proposal relies on the HTTP `CONNECT` verb and the associated `HTTP_PROXY` settings, which are [widely supported across browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/CONNECT#browser_compatibility) and libraries.
+The overall idea is to deploy a local proxy that terminates TLS traffic using Contrast certificates and re-encrypts them with a web PKI certificate.
+
+### Setup example
+
+In order to make the proposal more concrete, we're going to explain it alongside an example setup.
+The Kubernetes cluster has a confidential `web service`, for example the Emojivoto one.
+It's exposed on a public IP using the Gateway API, possibly together with other services.
+Service mesh ingress is configured to not require a client certificate.
```
sespiros, line 34: One more setup prerequisite for this example: `web.cc.cloud` must be added to the web service's manifest SANs, so its mesh certificate is valid for that name. That's the [manual manifest-edit flow the deployment guide documents for load balancer IPs](https://docs.edgeless.systems/contrast/getting-started/deployment#adding-an-ip-san-to-the-manifest). `generate` only emits the resource name. Note the default * SAN doesn't cover this despite what the manifest docs suggest: a bare wildcard fails RFC 6125-style matching in Go, OpenSSL, and curl (tracked separately in CON-222).

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,158 @@
...
+- The certificate for the domain names used by backend services (`web.cc.cloud` for the example).
+
+The life of a proxied connection then looks like this:
+
+0. The client browses to `https://web.cc.cloud`.
+1. The browser issues a `CONNECT web.cc.cloud:443` request to the proxy.
+2. The proxy establishes a TCP connection to the target service.
+   If that succeeds, the proxy returns `200 OK` to the client to indicate tunnel establishment, otherwise a `503 Service Unavailable`.
+3. The browser initiates a TLS connection over the tunnel, expecting a web PKI certificate for `web.cc.cloud`.
+4. The proxy receives the `ClientHello` and initiates a similar TLS connection (ALPN) to the backend, verified with the Contrast CA certificate.
```
sespiros, line 84:
```suggestion
4. The proxy receives the `ClientHello` and initiates a similar TLS connection (ALPN) to the backend, verifying the backend's certificate chain against the configured Contrast CA certificate and its SANs against the `CONNECT` target (`web.cc.cloud`), which the backend's mesh certificate carries per the setup above.
```

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,158 @@
...
+This proposal relies on the HTTP `CONNECT` verb and the associated `HTTP_PROXY` settings, which are [widely supported across browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/CONNECT#browser_compatibility) and libraries.
+The overall idea is to deploy a local proxy that terminates TLS traffic using Contrast certificates and re-encrypts them with a web PKI certificate.
+
+### Setup example
+
+In order to make the proposal more concrete, we're going to explain it alongside an example setup.
+The Kubernetes cluster has a confidential `web service`, for example the Emojivoto one.
+It's exposed on a public IP using the Gateway API, possibly together with other services.
+Service mesh ingress is configured to not require a client certificate.
+DNS is configured to resolve the domain `web.cc.cloud` to the gateway IP.
```
sespiros, line 35: An attack scenario here: the cluster owner, who is untrusted, controls the infrastructure behind the domain's IP, so they can pass HTTP-01 or TLS-ALPN-01 challenge and obtain a browser-valid certificate for `web.cc.cloud` from a public CA. Clients that bypass the proxy (i.e off-network devices) and hit the backend directly would then see a green padlock on an impersonation of the confidential service, instead of the hard cert error they'd get today.

Remediation: the org that owns `web.cc.cloud` controls its DNS (outside the cluster owner's reach and publishes:
```
web.cc.cloud.  CAA  0 issue "letsencrypt.org; accounturi=https://acme-v02.api.letsencrypt.org/acme/acct/12345678; validationmethods=dns-01"
```
Every CA except Let's Encrypt must then refuse the name. CAs are required to honor CAA by the CA/Browser Forum Baseline Requirements to stay in browser root programs. Let's Encrypt in turn rejects HTTP-01/TLS-ALPN-01 and accepts only DNS-01, whose challenge record lives in the org's DNS zone, which the cluster owner can't write to. The accounturi parameter additionally restricts issuance to the org's own ACME account.

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,158 @@
...
+            LB -->|web.cc.cloud| App1
+            LB -->|voting.cc.cloud| App2
+    end
+
+    Proxy -->|ClientHello<br>SNI: web.cc.cloud<br>CAs: mesh-ca.pem| LB
+```
+
+### Client configuration
+
+The client browser is configured to use `proxy.internal` as its `HTTP_PROXY`.
```
sespiros, line 64: nit: browsers don't read the `HTTP_PROXY` environment variable. They take proxy settings from the OS, a PAC file, or an extension like FoxyProxy. `HTTP_PROXY`/`HTTPS_PROXY` is the CLI/library convention (curl, Go, Python), and there it's `HTTPS_PROXY` that governs https:// URLs.
```suggestion
The client browser is configured to use `proxy.internal` as its HTTPS proxy`.
```

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,158 @@
...
+Instead of a simple reverse proxy, we could come up with a full-fledged application gateway.
+Such a gateway could route requests to backends and take care of user authentication/authorization.
+
+- There's no standard, and thus it's very likely that our gateway would be unusable for most customers.
+- If the gateway terminates TLS traffic, it would need to run inside a TEE itself, making it again more complicated to expose.
+
+### Proxy at the transport layer
+
+- No standard for such proxy (SOCKS5, but not widely usable).
+- Thus, would need a listener for each remote service (like service mesh outbound)
```
sespiros, line 131: nit: both halves of this are stale now. SOCKS5 is broadly usable (Firefox and Chrome natively, curl via socks5h://, ALL_PROXY for most tools), and it wouldn't need a listener per remote service since the SOCKS request carries the target host:port in-band. The actual reason it loses to the reworked design is different: a SOCKS tunnel is transparent, so the browser would run TLS itself and still hit the mesh cert, whereas the re-encrypting CONNECT proxy can satisfy the browser with a web PKI cert. Suggest reframing the bullet around that.

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Requirements
+
+- Allow standard HTTP clients to connect to servers running behind a service mesh ingress or using the mesh certificate directly.
+- While keeping the traffic confidential up to the client's environment (local machine or internal network).
+- Without increasing risk for applications in the cluster.
+
+## Proposal
+
+This proposal relies on the HTTP `CONNECT` verb and the associated `HTTP_PROXY` settings, which are [widely supported across browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/CONNECT#browser_compatibility) and libraries.
```
sespiros, line 26:
```suggestion
This proposal relies on the HTTP `CONNECT` tunnel that browsers open for `https://` URLs, which is [widely supported across browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/CONNECT#browser_compatibility) and libraries.
```

sespiros (commented):
Thanks for the changes.

rfc/014-http-proxies.md
burgerdev: Yes, that's indeed missing.

rfc/014-http-proxies.md
burgerdev: The scenario applies when the client does not set the proxy, somehow misconfigures it, or the browser ignores it, right? That's bad, and I don't think the workaround is sufficient, at least not without additional protections on the DNS level (to make sure the CAA record is actually delivered to letsencrypt).

rfc/014-http-proxies.md
burgerdev: Wow, this keeps being full of surprises. As far as I can tell, FoxyProxy interprets the `HTTPS` vs `HTTP` prefix as the protocol to initiate with the proxy, not the backend - contrary to the interpretation of curl and go.

rfc/014-http-proxies.md
burgerdev: Maybe SOCKS is the way to go, then? The idea of this proposal, after the first rework, is that the browser keeps doing TLS, but the proxy MITMs it. We can do the same in SOCKS, no?

sespiros: ah a SOCKS proxy that terminates at the proxy, right, I keep thinking about relaying proxies. So yes this is equally viable but I guess we don't need any non-HTTP TCP? My understanding is that the HTTP CONNECT was picked for wider client-config support.

rfc/014-http-proxies.md
sespiros: hmm right "HTTPS proxy" is the wrong term again, it should be "configured to send traffic through `proxy.internal` (a standard HTTP CONNECT proxy)"

rfc/014-http-proxies.md
sespiros: Yes, and it really depends on the exact setup but I thought to mention it in any case. i.e I was thinking a scenario during which the proxy would not be local but accessible via VPN and a device would access the service while the VPN is off. We should at least cover it in documentation. 

Agreed on the second part too although in this model we are already putting trust on some corp-managed network so as a hardening step maybe requiring DNSSEC on top is sufficient. For defending against BGP hijacking there is also [MPIC](https://letsencrypt.org/2020/02/19/multi-perspective-validation). I mean even then nothing stops a CA which simply violates the requirements. That could be also covered in documentation, that backends shouldn't be directly reachable by clients in the first place.

rfc/014-http-proxies.md
```diff
@@ -0,0 +1,158 @@
...
+
+    Client->>FwdProxy: HTTP request<br>(over TLS)
+    FwdProxy->>Backend: HTTP request<br>(reencrypted)
+    Backend->>FwdProxy: HTTP response
+    FwdProxy->>Client: HTTP response<br>(reencrypted)
+
+# https://mermaid.ai/d/7acceba4-9e72-4cbd-ab06-54f125d19546
+```
+
+## Alternatives considered
```
pskopnik, line 118: I had expected "reverse proxy" in this list (there is a "reverse" mention below, but it seems to refer to the forwarding proxy). Did you disqualify it?

Its architecture/topology would be close to the forwarding proxy proposal and it may be a more familiar deployment pattern for operators. I see some advantages in the changes:

 1. Domain name and destination ownership is the same: The web PKI cert & key for `app.comp.intra` is placed only at the reverse proxy. No need for involvement of the company's interception CA, copying cert & key to two places or double issuing.
 2. Decouples the user-facing endpoint completely from the remote cluster endpoint. Responsibilities between both operators are clearly split. Users would not accidentally access the aTLS service directly.
 3. The in-cluster gateway only needs to route on L4, not L7.

But the biggest advantage I see is in implementation. There are plenty of mature reverse proxy implementation which one could plug a `contrast-atls` backend into. Caddy could be interesting because it is Go and has an extension point for upstream transports. General features provided by Caddy would remain available to operators: automatic Let's Encrypt, routing to several backends based on path or domain, auth in front, setting additional headers such as CORS.

burgerdev: Sorry for the confusion: the "forward proxy" section is a leftover from the previous design (where it didn't fit, either). This RFC just proposes a regular HTTP proxy that's visible to the clients, while forward and reverse proxies wouldn't be. I'm sticking to the nomenclature from https://www.cloudflare.com/learning/cdn/glossary/reverse-proxy/ here. According to that, a reverse proxy would be running in the untrusted datacenter, so it would need to run in a CVM and be attested (defeating the point).

But I believe what you are suggesting is an on-prem forward proxy? That would neatly solve the problem from https://github.com/edgelesssys/contrast/pull/2509/changes#r3598849781. The downside I see is that such a proxy would require configuration for mapping downstream to upstream URLs. As a local admin, if I need to do that configuration anyway, I could also just add the certificate to Caddy and not require any Edgeless software. Which is a valid alternative, I'd say.

pskopnik: Existing terminology does not map cleanly to confidential computing paradigms. (I'd stand by my interpretation of reverse proxy as the authoritatively-determined destination for traffic to a _specific_ backend).

Maybe "on-prem gateway" is better. It would be a contrast-general version of the Privatemode proxy. Topology-wise, that is identical to your set up diagram.

> As a local admin, if I need to do that configuration anyway, I could also just add the certificate to Caddy and not require any Edgeless software.

I still lack understanding of the nuances of contrast.
The "add the certificate" variant you describe would be: An admin performs remote attestation against the deployment and receives the mesh CA cert. By adding this to Caddy (or nginx, ...) with the built-in TLS upstream transport, any successful upstream connection implies the upstream endpoint is part of the attested manifest.
The `contrast-atls` upstream transport variant would be: The admin configures the gateway with a manifest history and the gateway performs the remote attestation inside the gateway eagerly on start / lazily on connection.

Indeed it seems there is no advantage to the `contrast-atls` transport, if no rotation of the mesh CA is expected. A fresh mesh CA under the same manifest is unlikely for an HA deployment, and a new manifest should not be trusted automatically. So the real advantage would come if there were a protocol to receive or verify a new manifest automatically.

charludo (commented):
IMO, this version of the proposal makes a lot of sense and I'd be OK with going with it. The one thing I really don't like is the forward proxy a) having access to the plaintext client requests and server responses, and as a consequence b) the proxy being able to unilaterally stop forwarding to the Contrast deployment, and instead forwarding to an attacker-controlled server, without the client having any chance of noticing. The reason for both is that the forward proxy is holding the web PKI secret.

I had an idea for a slightly different architecture; I hope the following makes sense, and does not derail the conversation too much ^^

- let the forward proxy ONLY handle TLS with the Contrast CA certificate
- instead of terminating the web PKI TLS for `web.cc.cloud`, it receives the so-encrypted request, then wraps it in aTLS with the Contrast CA cert.
- Inside the cluster, an off-the-shelve reverse proxy (nginx, caddy,...) handles ACME for `*.cc.cloud` (as well the the usual reverse proxy duties)

```mermaid
graph TB
    subgraph OnPrem["On-Premises Environment"]
        Client[Client]
        Bot[Bot]
        Proxy[Proxy]
        Client -->|CONNECT web.cc.cloud:443| Proxy
        Proxy -->|Additionally encrypt| Proxy
        Bot --> Proxy
    end

    subgraph Cloud["Cloud Environment"]
        LB[Gateway]
        subgraph K8s["Kubernetes Cluster"]
            RP[Reverse Proxy]
            App1[Web service]
            App2[Voting service]
        end
            LB --> RP
            RP -->|ACME for *.cc.cloud| RP
            RP --->|web.cc.cloud| App1
            RP --->|voting.cc.cloud| App2
    end

    Proxy ----->|"atls ( pki_tls ( request ) )"| LB
```

The major advantage here is that the forward proxy can not fake forwarding the traffic to the CC cluster, because it does not possess the necessary keys to terminate the web PKI TLS traffic. 

Additionally, I *think* we should then easily be able to move the forward proxy from "in-network" to "on-client".

Regardless, the proxied connection then looks like this:

0. The client browses to `https://web.cc.cloud`.
1. The browser issues a `CONNECT web.cc.cloud:443` request to the proxy.
2. The proxy establishes and verifies the aTLS session.
   If that succeeds, the proxy returns `200 OK` to the client to indicate tunnel establishment, otherwise a `503 Service Unavailable`.
3. The browser initiates a TLS connection over the tunnel, expecting a web PKI certificate for `web.cc.cloud`.
4. The forward proxy receives the encrypted message from the client. It wraps this message in the aTLS session. It forwards the now doubly encrypted message.
5. The mesh ingress terminates the outer aTLS.
6. The in-cluster reverse proxy terminates the web PKI TLS connection and forwards the request to the service.
7. The forward proxy decryptes the outer aTLS layer and forwards teh still-web-PKI-encrypted  response to the client, unable to decrypt the message contents.
8. The forward proxy continues to route traffic back and forth until the sessions close.

In short, the forward proxy becomes "key-less" (and sees no plaintext), this enables it to move to the client, and that in turn strengthens the e2e security story.

In short, the forward proxy becomes key-less, does not handle web-PKI TLS termination and thus sees no plaintext. This allows it to move to the client, and that in turn strengthens the e2e security story. The web-PKI key material lives only inside the CC cluster.

PR #2510, opened 2026-07-06
kuberesource: remove pod-role annotation from runtime

burgerdev:
The pod role is used in the manifest to apply special considerations to some pods (notably, Coordinators). Since the runtime is never a confidential pod, it does not make sense to annotate it with a pod role. https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ are a more natural fit for this.

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -120,7 +120,6 @@ func NodeInstaller(namespace string, platform platforms.Platform) (*applyappsv1.
 					PodTemplateSpec().
 						WithLabels(map[string]string{"app.kubernetes.io/name": name}).
 						WithAnnotations(map[string]string{
-							ContrastRoleLabelKey:                 string(manifest.RoleNodeInstaller),
 							"contrast.edgeless.systems/platform": platform.String(),
 						}).
 						WithSpec(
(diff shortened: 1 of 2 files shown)


sespiros (changes requested):
Thanks.

This will break https://github.com/edgelesssys/contrast/blob/de312912326847e47012c37805a4f65a01c60c04/packages/cleanup-bare-metal.sh#L48

We would either need to select node installers by matching on `*-nodeinstaller` pod-name pattern or give the installer a stable label.

internal/kuberesource/parts.go
```diff
@@ -120,7 +120,6 @@ func NodeInstaller(namespace string, platform platforms.Platform) (*applyappsv1.
 					PodTemplateSpec().
 						WithLabels(map[string]string{"app.kubernetes.io/name": name}).
 						WithAnnotations(map[string]string{
-							ContrastRoleLabelKey:                 string(manifest.RoleNodeInstaller),
 							"contrast.edgeless.systems/platform": platform.String(),
```
davidweisse, line 123: Why do we have this? As far as I can see we don't use this anywhere as well.

burgerdev: I don't know either.

burgerdev:
> Thanks.
> 
> This will break
> 
> https://github.com/edgelesssys/contrast/blob/de312912326847e47012c37805a4f65a01c60c04/packages/cleanup-bare-metal.sh#L48
> 
> We would either need to select node installers by matching on `*-nodeinstaller` pod-name pattern or give the installer a stable label.

Then ... I guess it's used after all and we can keep it.

PR #2512, opened 2026-07-07
e2e-release: optionally consume artifacts from s3

charludo:
CON-97

### DIFF
--- a/.github/actions/release_e2e_test/action.yml
+++ b/.github/actions/release_e2e_test/action.yml
@@ -5,6 +5,10 @@ inputs:
   version:
     description: Release tag to test against, e.g. v1.2.3
     required: true
+  s3_base_url:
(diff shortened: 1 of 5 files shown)


.github/actions/release_e2e_test/action.yml
```diff
@@ -5,6 +5,14 @@ inputs:
...
     description: Release tag to test against, e.g. v1.2.3
     required: true
+  artifact_source:
+    description: Where to fetch release artifacts from ('github' or 's3').
+    required: false
+    default: s3
+  artifacts_url:
+    description: Base HTTPS URL of the artifacts in S3. Required when artifact_source is 's3'.
+    required: false
+    default: ''
```
burgerdev: How about adding only `s3_url_base` as an input and using s3 when it's not empty?

charludo: fair. done.

burgerdev (approved):
Thanks!

PR #2514, opened 2026-07-08
kuberesource: use common labeling strategy

burgerdev:
This PR centralizes the definition of labels and label selectors, which used to be ad-hoc. As a result, labels for individual API resources may differ after this change, but the labels should at least be consistent.

### DIFF
--- a/internal/history/configmapstore/configmapstore.go
+++ b/internal/history/configmapstore/configmapstore.go
@@ -22,7 +22,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
@@ -30,6 +29,11 @@ import (
 
 // timeout is the timeout for Kubernetes API calls.
 const timeout = 10 * time.Second
+
+const (
+	appName      = "kvstore"
+	appComponent = "coordinator"
+)
 
 // ConfigMapStore is a Store implementation backed by Kubernetes Config Maps.
 type ConfigMapStore struct {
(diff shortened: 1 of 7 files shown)


internal/kuberesource/parts.go
```diff
@@ -688,3 +691,25 @@ func CollateralProxy(namespace, storageClassName string) []any {
 
 	return []any{statefulSet, ServiceForStatefulSet(statefulSet)}
 }
+
+// ContrastLabels should be applied to top-level resources.
+// In addition to name and component, it also labels part-of and version with compile-time constants.
+func ContrastLabels(name, component string) map[string]string {
+	// The version should be a plain semver without prefix.
```
sespiros: nit:

```suggestion
	// The version must be a label-safe semver: strip the leading "v", and assume
    // no "+build" metadata, since '+' is invalid in a label value.
```

sespiros (approved):
Thanks! Should we mark this as a breaking change, or at least add a note? Since an in place upgrade (albeit not a documented procedure) would not be possible since we are changing the label selectors on Contrast-managed resources?

internal/kuberesource/parts.go
```diff
@@ -109,18 +111,19 @@ func NodeInstaller(namespace string, platform platforms.Platform) (*applyappsv1.
 	}
 
 	d := DaemonSet(name, namespace).
-		WithLabels(map[string]string{"app.kubernetes.io/name": name}).
+		WithLabels(ContrastLabels(name, "runtime")).
```
davidweisse:
```suggestion
		WithLabels(ContrastLabels(name, component)).
```

internal/history/configmapstore/configmapstore.go
davidweisse: Should this also use `kuberesource.SelectorLabels("kvstore", "coordinator")`?

burgerdev:
> Thanks! Should we mark this as a breaking change, or at least add a note? Since an in place upgrade (albeit not a documented procedure) would not be possible since we are changing the label selectors on Contrast-managed resources?

It's marked as `changelog`. The only thing where this matters would be the ConfigMap watcher, which was using the same label selector until I just changed it due to @davidweisse's remark. Let me revert that change for now, maybe I'll come up with something in #2515.

internal/history/configmapstore/configmapstore.go
burgerdev: Ok, it turns out that this selector is worth than useless, since we're already filtering by name and we don't set the label in `RecoverConfigMaps`. Removing it here, fixing `RecoverConfigMaps` in #2515.

PR #2515, opened 2026-07-08
coordinator: remove ownerreference from configmaps

burgerdev:
The `ConfigMap`s  used to store Coordiantor history are not owned by the Coordinator `StatefulSet` anymore. Cleaning them up now requires manual deletion, but the ConfigMaps can now be used by more than one Coordinator set and restored `ConfigMap`s are consistent with the ones created by the Coordinator. 

Fixes CON-208.
Fixes CON-220.
Fixes CON-205.

### DIFF
--- a/internal/history/configmapstore/configmapstore_test.go
+++ b/internal/history/configmapstore/configmapstore_test.go
@@ -10,38 +10,19 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
 )
-
-func coordinatorPod() *corev1.Pod {
-	return &corev1.Pod{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "coordinator-0",
-			Namespace: "test",
-			OwnerReferences: []metav1.OwnerReference{
-				{
-					Kind: "StatefulSet",
-					Name: "coordinator",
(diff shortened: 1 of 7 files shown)


internal/history/configmapstore/configmapstore_test.go
```diff
@@ -67,8 +68,7 @@ func TestHas(t *testing.T) {
 	require := require.New(t)
 
 	t.Setenv("HOSTNAME", "coordinator-0")
```
charludo: These can all be removed now, right? Since `New()` doesn't read the var anymore.

internal/history/configmapstore/configmapstore_test.go
```diff
@@ -67,8 +68,7 @@ func TestHas(t *testing.T) {
...
-	s, err := New(fake.NewClientset(coordinatorPod()), "test", slog.Default())
```
charludo, line 70: same for `coordinatorPod()` I think

internal/history/configmapstore/configmapstore_test.go
burgerdev: Done

internal/history/configmapstore/configmapstore_test.go
burgerdev: Done, good catch!

burgerdev:
I discovered a regression while working on the multi-coordinator e2e test: the configmaps returned to the client were missing the `TypeMeta`. Fixed that and added a regression test.

- [x] [coordinator test](https://github.com/edgelesssys/contrast/actions/runs/28970328763)

charludo (approved):
LGTM, thanks! I'll open a PR for govulncheck...

PR #2516, opened 2026-07-08
cli/generate: normalize image ref before accessing layers cache

davidweisse:
Genpolicy stores the normalized image reference in the layers cache, converting shortnames like "busybox" into "docker.io/library/busybox". Normalizing these on our side as well avoids mismatches when accessing the layers cache when calculating pod memory.

The `--calculate-pod-memory` flag should therefore now also work for short names.

Fixes CON-216


burgerdev (approved):
Thanks!

PR #2519, opened 2026-07-09
coordinator: allow multiple coordinator variants

burgerdev:
We removed all hard-coded dependencies on the particular structure of a Coordinator, so now it's possible to deploy a second `StatefulSet` with a Coordinator role into the same namespace. As long as the Coordinator pod is labeled with the right role and its policy hash is included in the manifest, it will be considered for peer recovery and will be allowed to recover from different Coordinators.

---

Most of the PRs for this already landed, this is just adding the e2e test.

- [x] [coordinator test](https://github.com/edgelesssys/contrast/actions/runs/29006398776)
- [x] [coordinator test](https://github.com/edgelesssys/contrast/actions/runs/29104687046)

Fixes CON-207.

### DIFF
--- a/e2e/coordinator/coordinator_test.go
+++ b/e2e/coordinator/coordinator_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/edgelesssys/contrast/internal/platforms"
 	"github.com/edgelesssys/contrast/sdk"
 	"github.com/stretchr/testify/require"
+	applyappsv1 "k8s.io/client-go/applyconfigurations/apps/v1"
 )
 
 func TestCoordinator(t *testing.T) {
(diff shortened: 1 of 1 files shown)


e2e/coordinator/coordinator_test.go
```diff
@@ -237,6 +238,71 @@ func TestCoordinator(t *testing.T) {
...
+		meta, ok := r.(meta)
+		require.Truef(ok, "unexpected resource type: %T", r)
+		kind := meta.GetKind()
+		require.NotNil(kind)
+		if *kind != "StatefulSet" {
+			continue
+		}
+		s, ok := r.(*applyappsv1.StatefulSetApplyConfiguration)
+		require.Truef(ok, "unexpected resource type for StatefulSet: %T", r)
+		s.Name = toPtr(alternativeCoordinatorName)
```
sespiros, line 279: The `meta` interface only exists to read `Kind` before re-asserting to the concrete type. Since `CoordinatorBundle()` has exactly one StatefulSet, you can assert to it directly:

```suggestion
	for _, r := range rs {
		s, ok := r.(*applyappsv1.StatefulSetApplyConfiguration)
		if !ok {
			continue
		}
		s.Name = toPtr(alternativeCoordinatorName)
```

e2e/coordinator/coordinator_test.go
```diff
@@ -237,6 +238,71 @@ func TestCoordinator(t *testing.T) {
...
+		s.Spec.Template.Spec.Containers[0].WithEnv(kuberesource.EnvVar().WithName("ONLY_FOR_POLICY").WithValue("foo"))
+		selector := kuberesource.SelectorLabels(alternativeCoordinatorName, alternativeCoordinatorName)
+		s.Spec.WithSelector(kuberesource.LabelSelector().WithMatchLabels(selector))
+		s.Spec.Template.WithLabels(selector)
```
sespiros, line 284: From Claude: 
This works only because apply-config `WithLabels` *merges*, so the `pod-role=coordinator` label that `Coordinator()` puts on the template survives — and that label is what makes this pod eligible for peer recovery, i.e. the whole point of the test. That's invisible and load-bearing: the selector on line 300 (which uses `WithSelector`, a *replace*) drops the role, so a reader can't tell the pod keeps it. Set it explicitly, mirroring how `Coordinator()` itself constructs the selector:

```suggestion
		selector := kuberesource.SelectorLabels(alternativeCoordinatorName, alternativeCoordinatorName)
		selector[kuberesource.ContrastRoleLabelKey] = string(manifest.RoleCoordinator)
		s.Spec.WithSelector(kuberesource.LabelSelector().WithMatchLabels(selector))
		s.Spec.Template.WithLabels(selector)
```
This makes the role explicit on both the selector and the template, and matches the pattern in `Coordinator()` (`labelSelector[ContrastRoleLabelKey] = string(manifest.RoleCoordinator)`), so the alt coordinator is labeled the same way as the primary instead of relying on merge semantics.

sespiros (approved):
Thanks

e2e/coordinator/coordinator_test.go
burgerdev: Done

e2e/coordinator/coordinator_test.go
burgerdev: Done

PR #2520, opened 2026-07-09
collateral-proxy: move to edgelesssys/collateral-proxy

charludo:
Pending edgelesssys/collateral-proxy#1.


charludo:
@burgerdev Do you think we can merge this? :D

justfile
burgerdev: Don't we want to apply the published resources from the collateral-proxy release?

charludo: Oh right, sorry, completely forgot about that. Fixed.

burgerdev (approved):

PR #2521, opened 2026-07-09
attestation: improve error messages on failed validation

burgerdev:
A refactoring in #2494 caused the aTLS validator to only have generic names that don't reflect what they are validating. This PR adds names to validators and combinators such that the nested validation logic is clearer.

An error message from this PR looks like:

```
validator Any(snp-0-GENOA) with policy hashes [967a0eee181382a7cf8a1faf9e3903b60660f6d1c863a0766633f6205dad6950] failed: wrong policy hash for Coordinator: got 977a0eee181382a7cf8a1faf9e3903b60660f6d1c863a0766633f6205dad6950, want [967a0eee181382a7cf8a1faf9e3903b60660f6d1c863a0766633f6205dad6950]
```

Fixes CON-217.

### DIFF
--- a/internal/atls/validators/validators.go
+++ b/internal/atls/validators/validators.go
@@ -9,6 +9,7 @@ import (
 	"encoding/asn1"
 	"errors"
 	"fmt"
+	"strings"
 )
 
 // ErrOIDNotSupported is returned by a validator when it doesn't understand the OID provided as input.
@@ -16,7 +17,7 @@ var ErrOIDNotSupported = errors.New("OID not supported")
 
 // Validator is able to validate an attestation document.
 //
-// Validators are encouraged to implement the fmt.Stringer interface to improve logging.
+// Validators are required to implement the fmt.Stringer interface to improve logging.
 type Validator interface {
 	// Validate validates an attestation doc and returns an error if validation failed.
 	//
@@ -26,6 +27,8 @@ type Validator interface {
 	// If validation passes, the validator guarantees that the given reportData was present in the
 	// attestation document.
 	Validate(ctx context.Context, oid asn1.ObjectIdentifier, attDoc []byte, reportData []byte) error
+
+	fmt.Stringer
 }
 
 // ValidatorFunc creates a validator from a func.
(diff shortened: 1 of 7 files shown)


internal/manifest/validators.go
```diff
@@ -95,7 +95,16 @@ func coordinatorValidator(validatorFactory validatorFactory, coordPolicyHashes [
...
 
-	return validators.ValidatorFunc(func(ctx context.Context, oid asn1.ObjectIdentifier, attDoc []byte, reportData []byte) error {
+	v, err := validatorFactory(log, kdsGetter, attestation.ReportSetterFunc(func(attestation.Report) {}))
+	if err != nil {
+		return nil, fmt.Errorf("constructing manifest validator: %w", err)
+	}
+	innerName := fmt.Sprintf("%T", v)
+	if s, ok := v.(fmt.Stringer); ok {
+		innerName = s.String()
+	}
```
sespiros: This could be exported as `validators.Name(v)` and reused in `internal/atls/validators/validators.go`

internal/atls/validators/validators.go
```diff
@@ -61,15 +67,46 @@ func Any(vs ...Validator) Validator {
...
+	return &named{v: v, name: name}
+}
+
+func validatorName(v Validator) string {
+	name := fmt.Sprintf("%T", v)
+	switch s := v.(type) {
+	case fmt.Stringer:
+		name = s.String()
+	}
+	return name
```
sespiros:
```suggestion
    if s, ok := v.(fmt.Stringer); ok {
        return s.String()
    }
    return fmt.Sprintf("%T", v)
```

sespiros (approved):
Thanks.

internal/manifest/validators.go
burgerdev: This is gone now.

internal/atls/validators/validators.go
burgerdev: Too

internal/atls/validators/validators_test.go
sespiros, line 23:
```suggestion
		"vacuous truth": {err: ErrOIDNotSupported},
		"single validator passes through": {
                validators: []*stubValidator{{err: assert.AnError}},
                numTried:   1,
                err:        assert.AnError,
        },
```

internal/atls/validators/validators_test.go
```diff
@@ -120,3 +164,14 @@ func (s *stubValidator) Validate(_ context.Context, oid asn1.ObjectIdentifier, d
 	s.reportData = rd
 	return s.err
 }
+
+func (s *stubValidator) String() string {
+	return s.name
+}
+
+var (
+	_ Validator = ValidatorFunc(nil)
```
sespiros, line 178:
```suggestion
	_ Validator = ValidatorFunc(nil)
    _ Validator = (*anyOf)(nil)
```

sespiros (approved):
Thanks.

PR #2522, opened 2026-07-10
e2e: delete history after deleting Coordinator

burgerdev:
The history used to be owned by the Coordinator and was deleted alongside it. We recently removed ownership from the ConfigMaps, but it turned out that some tests relied implicitly on that behaviour. Update those tests to explicitly clean up the coordinator history.

This caused e2e failures, e.g. https://github.com/edgelesssys/contrast/actions/runs/29049439091.

- [ ] [attestation SNP](https://github.com/edgelesssys/contrast/actions/runs/29077900872)

### DIFF
--- a/e2e/coordinator/coordinator_test.go
+++ b/e2e/coordinator/coordinator_test.go
@@ -29,8 +29,6 @@ import (
 	"github.com/edgelesssys/contrast/internal/platforms"
 	"github.com/edgelesssys/contrast/sdk"
 	"github.com/stretchr/testify/require"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 )
 
 func TestCoordinator(t *testing.T) {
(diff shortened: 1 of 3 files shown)


e2e/coordinator/coordinator_test.go
```diff
@@ -59,13 +57,7 @@ func TestCoordinator(t *testing.T) {
...
 		// This is the documented procedure, if changes are needed here make sure to reflect that
 		// in the docs.
-		deleteOpts := metav1.DeleteOptions{
-			PropagationPolicy: toPtr(metav1.DeletePropagationForeground),
-		}
-		listOpts := metav1.ListOptions{
-			LabelSelector: labels.Set(map[string]string{kuberesource.KubernetesAppManagedByLabel: "contrast.edgeless.systems"}).AsSelector().String(),
-		}
-		require.NoError(ct.Kubeclient.Client.CoreV1().ConfigMaps(ct.Namespace).DeleteCollection(ctx, deleteOpts, listOpts))
+		require.NoError(ct.Kubeclient.DeleteHistory(t.Context(), ct.Namespace))
```
sespiros: why not use `ctx` here?

sespiros (approved):
Thanks. There is a static failure but I think it will resolve itself once https://github.com/edgelesssys/contrast/pull/2519 is merged.

e2e/coordinator/coordinator_test.go
burgerdev: Copy paste issue.

PR #2529, opened 2026-07-15
attestation/certcache: log errors from RetryHTTPSGetter

davidweisse:
We should log the errors from the individual requests of the `RetryHTTPSGetter`, otherwise we only get a `context deadline exceeded` when setting the manifest fails, as in https://github.com/edgelesssys/contrast/actions/runs/28056557011/job/83060984081. Logging the error now creates more useful logs.

- [x] [kds-pcs-downtime test](https://github.com/edgelesssys/contrast/actions/runs/29420465511)

Fixes CON-191


burgerdev (approved):
Thanks, lgtm!

PR #2530, opened 2026-07-17
.github, e2e: automatically run relevant tests on PRs

charludo:
Requires us to maintain a list of relevant paths and nix packages right in the e2e test files.


e2e/attestation/attestation_test.go
```diff
@@ -1,6 +1,9 @@
 // Copyright 2025 Edgeless Systems GmbH
 // SPDX-License-Identifier: BUSL-1.1
 
+// test-if: path:internal/attestation
+// test-if: path:internal/atls
```
burgerdev: Note to self: We might want to merge `atls` and `attestation`.

packages/by-name/scripts/select-e2e-tests/select-e2e-tests.sh
```diff
@@ -0,0 +1,152 @@
...
+    if [[ $name == openssl || $name == gpu ]] || path_matches "e2e/$name"; then
+      selected[$name]=1
+      continue
+    fi
+    while IFS= read -r d; do
+      case $d in
+      path:*) path_matches "${d#path:}" && {
+        selected[$name]=1
+        break
+      } ;;
```
burgerdev, line 171: I wonder whether we could call the go closure script here to exapnd the paths. Say we set `test-if:coordinator` but have a change in `internal/atls`. Should we run the test, or would we just end up running all tests on most PRs?

burgerdev (approved):
The shell script is a bit intimidating, but I fear the right tool for the job.

packages/by-name/scripts/select-e2e-tests/select-e2e-tests.sh
charludo: Apart from the e2e go file itself, the go closure for the e2es are virtually identical. For this to make any sense, we'd have to filter out the common denominators, and by that time what distinct files remain are now already present in the `// test-if:` directives.

burgerdev: I meant the closure of the `test-if` directives. For example, you declare `test-if: internal/atls` and that then triggers if `internal/attestation` changes, because it's in the closure of atls. But again, we might end up in a situation where small changes to a deeply nested package end up triggering all tests.

charludo: Ohhhh, OK. That might actually work? I'll try it out.

charludo: Mkay, we can't do this indiscriminately. Tests like e.g. the `coordinator` e2e would basically always run. I've added a `closure:` directive we can selectively use.

But yeah, good idea!

burgerdev (approved):
Thanks!

PR #2531, opened 2026-07-17
scripts/containers: announce upload size before crane push

sespiros:
crane push prints a log line only when a blob finishes uploading, so a multi-hundred-MB layer on a slow uplink looks like a hang for many minutes. Print the image size upfront so the silence is expected.


charludo (approved):
Yeah, nice! IMO the informational `, new layers upload silently and can take a while` isn't really necessary, but w/e. Nice to not have to wonder for sure :D

charludo (approved):

PR #2535, opened 2026-07-20
.github: fix codeql-action pin flagged by zizmor

sespiros:
The pinned commit is not a released v3 version (likely a deleted tag), so zizmor's ref-version-mismatch audit fails just fmt. Pin the current v3 release with an exact version comment like the other action pins.


charludo (approved):
Thanks for noticing. Worrying that this hasn't been noticed so far in CI OR locally.

PR #2536, opened 2026-07-20
[6/8] cli: support insecure manifests behind opt-in

msanft:
Re-roll of https://github.com/edgelesssys/contrast/pull/2357, split from #2337 as part of a stacked review series.
Closes CON-245

This PR adds CLI support for insecure manifests behind explicit opt-in:
- generate support
- verify support
- runtime class handling for insecure runtimes
- verifier and mutator updates
- tests

### DIFF
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -13,6 +13,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -42,6 +43,13 @@ const (
 	// Coordinator's env var of the same name.
 	allowInsecureEnvVar = "CONTRAST_ALLOW_INSECURE"
 )
+
+// insecureRuntimesAllowed reports whether allowInsecureEnvVar is set to a value that parses as
+// true. Commands only declare the --INSECURE flag when this returns true.
+func insecureRuntimesAllowed() bool {
+	allowed, err := strconv.ParseBool(os.Getenv(allowInsecureEnvVar))
+	return err == nil && allowed
+}
 
 // ReleaseImageReplacements contains the image replacements used by contrast.
 //
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -38,6 +38,9 @@ const (
 	latestTransitionHashFilename = "latest-transition"
 	historyFilename              = "history.yml"
 	verifyDir                    = "verify"
+	// allowInsecureEnvVar gates insecure (non-CC) runtime support and matches the
+	// Coordinator's env var of the same name.
+	allowInsecureEnvVar = "CONTRAST_ALLOW_INSECURE"
 )
 
 // ReleaseImageReplacements contains the image replacements used by contrast.
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -13,6 +13,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -20,6 +21,7 @@ import (
 	"github.com/edgelesssys/contrast/internal/attestation/certcache"
 	"github.com/edgelesssys/contrast/internal/constants"
 	"github.com/edgelesssys/contrast/internal/fsstore"
+	"github.com/edgelesssys/contrast/internal/manifest"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
@@ -38,7 +40,24 @@ const (
 	latestTransitionHashFilename = "latest-transition"
 	historyFilename              = "history.yml"
 	verifyDir                    = "verify"
+	// allowInsecureEnvVar gates insecure (non-CC) runtime support and matches the
+	// Coordinator's env var of the same name.
+	allowInsecureEnvVar = "CONTRAST_ALLOW_INSECURE"
 )
+
+// insecureRuntimesAllowed reports whether allowInsecureEnvVar is set to a value that parses as
+// true. Commands only declare the --INSECURE flag when this returns true.
+func insecureRuntimesAllowed() bool {
+	allowed, err := strconv.ParseBool(os.Getenv(allowInsecureEnvVar))
+	return err == nil && allowed
+}
+
+func validateInsecureManifest(mnf *manifest.Manifest, allowInsecure bool) error {
+	if mnf.HasInsecurePlatforms() && !allowInsecure {
+		return fmt.Errorf("manifest contains insecure platforms but --INSECURE flag not set (the flag is only available with the %s environment variable set to true)", allowInsecureEnvVar)
+	}
+	return nil
+}
 
 // ReleaseImageReplacements contains the image replacements used by contrast.
 //
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -330,9 +330,7 @@ func mapContrastWorkloads(fileMap map[string][]*unstructured.Unstructured, f fun
 
 func isContrastWorkload(resource any) (ret bool) {
 	kuberesource.MapPodSpec(resource, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
-		if kuberesource.IsContrastPod(spec) {
-			ret = true
-		}
+		ret = kuberesource.IsContrastPod(spec)
 		return spec
 	})
 	return ret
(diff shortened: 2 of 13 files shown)


cli/cmd/generate.go
```diff
@@ -325,7 +330,9 @@ func mapContrastWorkloads(fileMap map[string][]*unstructured.Unstructured, f fun
 
 func isContrastWorkload(resource any) (ret bool) {
 	kuberesource.MapPodSpec(resource, func(spec *applycorev1.PodSpecApplyConfiguration) *applycorev1.PodSpecApplyConfiguration {
-		ret = kuberesource.IsContrastPod(spec)
+		if kuberesource.IsContrastPod(spec) {
+			ret = true
+		}
```
burgerdev: That's not really necessary since we're mapping over only one resource.

cli/cmd/generate.go
```diff
@@ -93,6 +93,7 @@ subcommands.`,
 	cmd.Flags().Bool("insecure-enable-debug-shell-access", false, "enable the debug shell service in the pod CVM to get access from container to guest VM")
 	cmd.Flags().Bool("calculate-pod-memory", false, "calculate pod memory based on image layer sizes and container resource limits")
 	cmd.Flags().StringP("output", "o", "", "output file for generated YAML")
+	cmd.Flags().Bool("INSECURE", false, "allow generation for insecure (non-CC) runtimes (also requires the CONTRAST_ALLOW_INSECURE_RUNTIMES environment variable to be set)")
```
burgerdev: What's the idea behind both the flag and the env var?

msanft: Have an additional safeguard for this purpose-defeating functionality of Contrast. I think this came out of a past review but I haven't been able to find it in the linked commits. I think it'd be fine to drop. WDYT?

burgerdev: I think what would make a bit more sense would be to condition the flag declaration on the env var, such that users don't see it with `-h` and they can also not accidentally pass it.

cli/cmd/generate.go
```diff
@@ -631,6 +651,19 @@ func injectServiceMesh(resource any, memoryProfile kuberesource.MemoryProfile) e
...
 }
 
+func validateInsecurePlatforms(usedPlatforms kuberesource.PlatformCollection, allowInsecure bool) error {
+	if !slices.ContainsFunc(usedPlatforms.Platforms(), platforms.IsInsecure) {
+		return nil
+	}
+	if !allowInsecure {
+		return fmt.Errorf("insecure runtime platforms detected but --INSECURE flag not set")
+	}
+	if os.Getenv("CONTRAST_ALLOW_INSECURE_RUNTIMES") == "" {
```
sespiros: Why don't we use `CONTRAST_ALLOW_INSECURE` as a name here? Same as the coordinator side env variable?

cli/cmd/verify.go
```diff
@@ -70,6 +71,19 @@ func runVerify(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("failed to read manifest file: %w", err)
 	}
 
+	var mnfst manifest.Manifest
+	if err := json.Unmarshal(manifestBytes, &mnfst); err != nil {
+		return fmt.Errorf("unmarshalling manifest: %w", err)
```
sespiros, line 78: nit: We can now pass the already parsed manifest directly to `getCoordinatorState` so it doesn't have to unmarshal it again.

cli/cmd/generate.go
sespiros: Separate to that, sorry if this is conflicting with what was already discussed in the previous PR but I think it makes more sense to treat the env variable as a boolean on existence/non-existence so `ALLOW_INSECURE=false` is irrelevant semantically and simply its existence just allows insecure manifests. That also means that the check here should be doing `os.LookupEnv`.

burgerdev: Yep, @msanft, sorry for the back and forth: we had an internal discussion about this and decided that it would be best if we either
- expect a fixed value for the env var and reject anything else
- treat the value as a boolean and parse it

The reason is to avoid confusing users who see `CONTRAST_ALLOW_INSECURE=true` and then assume `CONTRAST_ALLOW_INSECURE=false` is the opposite.

burgerdev (approved):
lgtm, thanks!

cli/cmd/generate.go
```diff
@@ -343,6 +350,16 @@ func isCoordinator(resource any) bool {
 	return false
 }
 
+func patchCoordinatorAllowInsecure(resource any) {
+	r, ok := resource.(*applyappsv1.StatefulSetApplyConfiguration)
+	if !ok || !isCoordinator(resource) {
+		return
+	}
+	if len(r.Spec.Template.Spec.Containers) > 0 {
+		r.Spec.Template.Spec.Containers[0].WithEnv(kuberesource.NewEnvVar(allowInsecureEnvVar, "1"))
```
charludo: This isn't idempotent (i.e. running insecure generate twice appends this twice). Since all our other mutators are this should be too.

cli/cmd/generate.go
```diff
@@ -343,6 +350,16 @@ func isCoordinator(resource any) bool {
 	return false
 }
 
+func patchCoordinatorAllowInsecure(resource any) {
+	r, ok := resource.(*applyappsv1.StatefulSetApplyConfiguration)
+	if !ok || !isCoordinator(resource) {
```
charludo: `isCoordinator` needs to additionally check `r.Spec.Template.Spec != nil`

cli/cmd/generate.go
charludo: Also, can we really assume the coordinator is always at index `0`? We should probably find the coordinator(s) at any index.

cli/cmd/generate.go
```diff
@@ -586,6 +603,9 @@ func patchTargets(fileMap map[string][]*unstructured.Unstructured, imageReplacem
 		if flags.injectImageStore {
 			kuberesource.AddImageStore([]any{res})
 		}
+		if flags.allowInsecureRuntimes {
+			patchCoordinatorAllowInsecure(res)
+		}
```
charludo: This may be overly cautious, but the flag alone should not be enough in an all-secure deployment. Maybe the if should additionally check that the usedPlatforms are all insecure.

cli/cmd/common.go
```diff
@@ -38,8 +39,18 @@ const (
...
 	verifyDir                    = "verify"
+	// allowInsecureEnvVar gates insecure (non-CC) runtime support and matches the
+	// Coordinator's env var of the same name.
+	allowInsecureEnvVar = "CONTRAST_ALLOW_INSECURE"
 )
 
+// insecureRuntimesAllowed reports whether allowInsecureEnvVar is set to a value that parses as
+// true. Commands only declare the --INSECURE flag when this returns true.
+func insecureRuntimesAllowed() bool {
+	allowed, err := strconv.ParseBool(os.Getenv(allowInsecureEnvVar))
```
charludo, line 51: I don't know where the discussion around envvar presence vs value ultimately landed. However, given that in the coordinator we now have:
```go
if os.Getenv(allowInsecureEnvVar) != "" {
```
one of these sites needs to be adjusted so that the semantics are the same.

cli/cmd/verify.go
```diff
@@ -70,13 +73,21 @@ func runVerify(cmd *cobra.Command, _ []string) error {
...
+	}
+	if mnfst.HasInsecurePlatforms() && !flags.allowInsecureRuntimes {
```
charludo: I think we need the same gate in `generate`, too. IIUC, currently an *existing* manifest with insecure platforms being re-generated works even if `--INSECURE` is *not* passed.

cli/cmd/generate_test.go
```diff
@@ -160,13 +233,129 @@ func TestPatchRuntimeClassName(t *testing.T) {
...
+		"false":   {set: true, value: "false", want: false},
+		"garbage": {set: true, value: "yes please", want: false},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			t.Setenv("CONTRAST_ALLOW_INSECURE", tc.value)
+			if !tc.set {
+				os.Unsetenv("CONTRAST_ALLOW_INSECURE")
+			}
```
charludo: Use `allowInsecureEnvVar`

charludo (commented):
Thanks! Sorry for only getting to this now. I have a couple of comments + nits below. If any of these were already discussed, feel free to just ignore them, I do not mean to rip closed discussions open again.

- (maybe not in this PR) we should probably add a `kuberesource.IsBareContrastRuntimeClass(name)` function, since the checks for `contrast-cc`/`contrast-insecure` now exist in 3 places,
- a test for `patchCoordinatorAllowInsecure` would maybe be nice, I think it would have caught some of the comments

charludo (approved):
Thanks, changes LGTM

PR #2538, opened 2026-07-21
kata.runtime: 3.32.0 -> 4.0.0

charludo:
Starting with Kata 4.0.0, users will have to explicitly set `medium: Memory` on `emptyDir` volumes. `contrast generate` will fail otherwise, and the volumes will not become available.

A current definition like:
```yaml
# v1.PodSpec
spec:
  volumes:
    - emptyDir: {}
```

needs to be manually migrated to:
```yaml
# v1.PodSpec
spec:
  volumes:
    - emptyDir:
        medium: Memory
```


internal/kuberesource/parts.go
```diff
@@ -373,7 +373,7 @@ func Coordinator(namespace string) *CoordinatorConfig {
 										WithResources(
 											ResourceRequirements().
 												// As of v1.23.0, the Coordinator is roughly 270MiB unpacked.
-												WithMemoryLimitAndRequest(300),
+												WithMemoryLimitAndRequest(400),
```
burgerdev: Interesting, why only the Coordinator? I'd have expected a Kata update to increase memory requirements overall, was that limit here just too tight?

nodeinstaller/internal/kataconfig/testdata/runtime-rs/expected-configuration-qemu-insecure-gpu.toml
```diff
@@ -48,6 +48,8 @@ machine_type = 'q35'
 memory_offset = 0
 memory_slots = 10
 network_queues = 1
+overhead_memory = 128
+overhead_vcpus = 0.4
```
burgerdev: We may need to adjust the memory overhead to  https://github.com/edgelesssys/contrast/blob/07ba8ce8bba864e27d73e2830c71ac66a8e51585/internal/kuberesource/parts.go#L61-L71

packages/by-name/kata/agent/package.nix
```diff
@@ -33,6 +33,7 @@ in
     "--skip=netlink::tests::list_routes"
     "--skip=config::tests::test_from_cmdline_with_args_overwrites"
     "--skip=rpc::tests::test_get_oom_event_no_deadlock"
+    "--skip=rpc::tests::test_do_copy_file"
```
burgerdev, line 36: Did that test fail in nix? I thought I wrote it such that it's unprivileged and hermetic 😅

packages/by-name/kata/genpolicy/genpolicy_settings_dev.patch
```diff
@@ -36,7 +36,13 @@ index ea73fc37eb638cf66853e36d80ea011410e3dada..9b336eb71bff0011c5c39169043166b4
...
          },
          "UpdateInterfaceRequest": {
-@@ -363,7 +366,9 @@
+@@ -375,11 +378,14 @@
+             ]
+         },
+         "CopyFileRequest": [
+-            "$(sfprefix)"
++            "$(sfprefix)",
++            "^$(cpath)/sandbox-[a-z0-9]{16}-"
```
burgerdev: Why is this needed?

packages/by-name/kata/source/0014-agent-remove-initdata-processing.patch
```diff
@@ -1,47 +0,0 @@
...
--
--    let gc_procs = config.guest_components_procs;
--    if !attestation_binaries_available(logger, &gc_procs) {
--        warn!(
--            logger,
--            "attestation binaries requested for launch not available"
--        );
--    } else {
--        init_attestation_components(logger, config, &initdata_return_value).await?;
--    }
```
burgerdev, line 29: Don't we still need to remove all this?

internal/kuberesource/parts.go
charludo: I hit it multiple times, but not always. It was recently reduced, so might just have been slightly too small.

nodeinstaller/internal/kataconfig/testdata/runtime-rs/expected-configuration-qemu-insecure-gpu.toml
charludo: Oh shoot, good catch!

packages/by-name/kata/agent/package.nix
charludo: Yeah failed, sorry :( I haven't really looked into the exact reason. Will do.

packages/by-name/kata/genpolicy/genpolicy_settings_dev.patch
charludo: Yeah that probably needs a comment. Cliff notes: bind-mount in emptyDir_typ="shared-fs" was being blocked.

packages/by-name/kata/agent/package.nix
charludo: `do_copy_file` calls `fchown()`, which fails in the nix sandbox :/ Not sure if we can do anything about that.

nodeinstaller/internal/kataconfig/testdata/runtime-rs/expected-configuration-qemu-insecure-gpu.toml
charludo: Mh, setting `overhead_memory` to `platforms.DefaultMemoryInMebiBytes(platform)` fixes the OOM issue above / coordinator memory no longer needs a bump. let's go with that.

packages/by-name/kata/genpolicy/genpolicy_settings_dev.patch
charludo: JSON and comments... -.- Added an explanation section for all the various changes to the nix file importing the patches.

packages/by-name/kata/source/0014-agent-remove-initdata-processing.patch
charludo: Indeed. Sorry about this, and thanks for catching it... Seems I made an error in conflict resolution in git am.

packages/by-name/kata/genpolicy/package.nix
```diff
@@ -53,12 +53,24 @@
       #   # Now, edit /tmp/b/genpolicy-settings.json according to your needs.
       #   cd ..
       #   git diff --no-ext-diff --full-index --no-prefix a/genpolicy-settings.json b/genpolicy-settings.json >$CONTRAST/packages/by-name/kata/genpolicy/genpolicy_settings_dev.patch
+      #
+      # genpolicy_settings_prod.patch changes explanations:
+      #   - pgpu_resource_keys: allow the GB100/B200 and GH100/H100 PCIe GPU resource names.
+      #   - pause_container_image: use our mirrored pause image instead of the MCR one.
+      #   - ReadStreamRequest = true: allow reading container stdout/stderr (e.g. for `kubectl logs`).
+      #   - emptydir_type = "shared-fs": emptyDir volumes are host-shared via virtio-fs, we setup LUKS ourselves.
+      #   - CopyFileRequest: also allow copies into the per-sandbox shared dir, which shared-fs emptyDir bind-mounts need.
```
burgerdev: Uh oh, that's not good. I was under the impression that the combination of 

```toml
emptydir_type = "shared-fs"
shared_fs = "none"
```

means that emptyDirs are not shared, but created and mounted in the guest. We definitely don't want to allow the host populating the emptyDir.

packages/by-name/kata/source/0013-agent-remove-initdata-processing.patch
```diff
@@ -5,27 +5,28 @@ Subject: [PATCH] agent: remove initdata processing
...
--    if !attestation_binaries_available(logger, &gc_procs) {
+-    let launch_plan = build_coco_launch_plan(config, &initdata_return_value, gc_procs)?;
+-    if !attestation_components_available(logger, &launch_plan) {
 -        warn!(
 -            logger,
 -            "attestation binaries requested for launch not available"
 -        );
 -    } else {
--        init_attestation_components(logger, config, &initdata_return_value).await?;
+-        init_attestation_components(logger, &launch_plan).await?;
```
burgerdev, line 29: I wonder if that launch plan was designed for disabling the guest components, and whether we can use that instead of deleting these lines. But let's not figure this out here.

tools/tdx-measure/rtmr/rtmr.go
```diff
@@ -168,16 +168,16 @@ func CalcRtmr0(firmware []byte, gpu GPUModel, legacySerial bool) ([48]byte, erro
 	// into the matching hash set below.
 	acpiHashes := []string{
 		// Default (virtio-serial-pci + virtconsole) topology.
-		"fe2b18cedec78d5c5badd19ccce1bea1c92b5fd6d0a6c54d57e7963d14e888a5e8f614eb1e947ff55468c2ba4290bce8",
-		"13532425fd25e97766f29d820a2b7a2cbe9286ce547d53e949853d90c41be8dceeaaccac3b6f6f1eff4c0e2ad6e26842",
-		"11223fd2a2149fa13b0b56c370c8dd959110c187aa43538d696334c9873481481d7d494c93d0c0152eae6c582a84b642",
+		"a98ebc08f45482c21c329b93a51b926a299c4c366d4e79d6ba586e4a8c92a66e9d593b176ffdf32476f0dd81bb68f95b",
+		"6cd5eb8fa5c659b3ac4172a1c678aa1bba3118e107dcc5cdea0deec83b50e152703822396ef3422fd020bbf0ffdc3491",
+		"764a603f33825a43eff061cf6154516d8304eb420a08c06b7e3d7a956932b82efa591701b1854a1fb75a8deb54b2aefc",
```
burgerdev, line 173: Do you know why these would change although we did not update qemu or OVMF?

charludo: I think it might be this: https://github.com/kata-containers/kata-containers/pull/13139/changes#diff-582767aaa1f02d4214b3a2492e73435345a8b14ea58d6fd560d3e6a426ec363bR203

Rewrites the the cold-plugged default bridge even when NUMA is off.

packages/by-name/kata/genpolicy/package.nix
burgerdev: Resolved by not allowing emptyDirs anymore.

burgerdev (approved):
Great!

PR #2542, opened 2026-07-22
nodeinstaller: support containerd config v4

burgerdev:
Config version 4 has been released with containerd v2.3.0. The changes are mostly deprecations, of which we are not affected. The only config item that we're using is the `io.containerd.cri.v1` plugin. This plugin is still supported, did not move, and all the fields we're setting are not deprecated as of containerd 2.3.

New containerd config versions will now fail hard and require a new Contrast release. We're picking the conservative path here to avoid unforeseen consequences of setting wrong paths in a new config layout.

Fixes CON-81.


charludo (approved):
Thank you!

PR #2543, opened 2026-07-23
.github: cache vulnix db between runs

charludo:
Should fix failures like these: https://github.com/edgelesssys/contrast/actions/runs/30006437971/job/89204974529

### DIFF
--- a/packages/by-name/scripts/vuln-scan/vuln-scan.sh
+++ b/packages/by-name/scripts/vuln-scan/vuln-scan.sh
@@ -35,14 +35,33 @@ osv-scanner scan source \
   --format=table \
   -L "$relabeled" || exitcode=$?
 
+# vulnix downloads the NVD feeds from nvd.nist.gov on every run. Reuse a persistent cache dir when one is provided
+# Additionally, retry transient failures. Exit code 2 means "vulnerabilities found".
 vulnix_json="$workdir/vulnix.json"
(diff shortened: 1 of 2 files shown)


packages/by-name/scripts/vuln-scan/vuln-scan.sh
```diff
@@ -35,14 +35,32 @@ osv-scanner scan source \
...
+  jq -r 'if length == 0 then "vulnix: no advisories"
+         else (.[] | "vulnix: \(.name) affected by \(.affected_by | join(", "))") end' \
+    "$vulnix_json"
+  if [[ -n $sarifdir ]]; then
+    jq --arg sbom "file://$relabeled" -f "$VULNIX_SARIF_JQ" "$vulnix_json" >"$workdir/vulnix.sarif" &&
+      mv "$workdir/vulnix.sarif" "$sarifdir/vulnix.sarif"
+  fi
+  [[ $vx_rc -eq 0 ]] || exitcode=$vx_rc
+else
+  echo "::warning::vulnix skipped: could not reach NVD after 3 attempts (exit $vx_rc), OS-closure CVEs not checked this run" >&2
```
burgerdev: What if this starts failing for some other persistent reason? I'm afraid that we might just not notice.

charludo: Hm, good point. I'll make it fail.

charludo: done

burgerdev (approved):

PR #2545, opened 2026-07-23
treewide: enable NUMA support

charludo:
~~This does somehing similar to what we originally had with SNP multi-CPU, where we now add a manifest entry per NUMA node count. There's not really an alternative to this until we calculate RTMRs on-the-fly.~~

When you want to try this out, make sure that `|vCPUs| >= min(|GPUs|, |NUMA cores|)`, otherwise the VM *may* fail to boot, depending on how the it gets scheduled.

PR also includes a patch that fixes the issue that made the VM sometimes fail to launch, which was the reason I previously said you'd have to pin the topology. That's not required anymore! 🎉


packages/by-name/contrast/reference-values/package.nix
```diff
@@ -62,29 +62,33 @@ let
...
+        rtmrs = [
+          rtmr0
+          rtmr1
+          rtmr2
+          rtmr3
+        ];
+        # CET (XFAM bits 11/12) is ignored during validation because it depends on host CPU support. see validateXfamIgnoringCET.
+        xfam = "e702060000000000";
+        memoryIntegrity = false;
+      }) rtmr0List;
```
burgerdev: I'm a bit hesitant to re-introduce manifest blow up here. Should we wait with this until we calculate RTMR0 on the fly? Should we just remove the measurement from OVMF? Should we check RTMR0 after validation, similar to HOSTDATA or HWID?

burgerdev (commented):
Nice, good job getting this working! I'm wondering about the tradeoff between manifest conciseness and the efficiency gains here. Maybe @thomasten can check whether it's worth it?

packages/by-name/contrast/reference-values/package.nix
charludo: Yeah, at the very least we should not include this in the manifest by default for now. Just left it in for PM testing ease of use. I'd say let's implement/merge the feature itself (if correct NUMA-ref value are present, NUMA should work), but not make it user-visible until after TDX-Multi-CPU when we hopefully have a way to calculate the values on the fly like for SNP.

However, this: 

> Should we just remove the measurement from OVMF? 

sounds like a better alternative solution until then. I'll ping you when it's implemented!

charludo: Alright, e2es passed, should be good to review :) Added as a separate commit for now, should probably partly be a fixup of the first.

charludo:
@thomasten This should be good to test again. kata 4.0.0 did indeed bring necessary bugfixes. We still needed some (additional, new) patches for the failures you observed when testing. I was able to test with 6 GPUs successfully in the current state.

charludo:
Closed due to no performance benefits.

PR #2546, opened 2026-07-24
coordinator: keep serving if the k8s API is down

burgerdev:
Adding a local cache to the history allows to serve `contrast verify` requests without external dependencies, so the Coordinator is not affected by transient unavailability of the k8s API.

### DIFF
--- a/internal/history/history.go
+++ b/internal/history/history.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"hash"
 	"log/slog"
+	"sync"
 )
 
 const (
@@ -24,9 +25,16 @@ const (
 var hashFun func() hash.Hash = sha256.New
 
 // History is the history of the Coordinator.
+//
+// It uses a backing KV store to persist policies, manifests and transitions. All but the latest
+// transition are stored in a content-addressed scheme, and the History keeps a local cache to
+// speed up content retrieval.
 type History struct {
 	store Store
 	log   *slog.Logger
+
+	cache   map[[HashSize]byte][]byte
+	cacheMu sync.RWMutex
 }
 
 // NewWithStore creates a new History with the given storage backend.
@@ -34,6 +42,7 @@ func NewWithStore(log *slog.Logger, store Store) *History {
 	h := &History{
 		store: store,
 		log:   log,
+		cache: make(map[[HashSize]byte][]byte),
 	}
 	return h
 }
@@ -176,6 +185,10 @@ func (h *History) WalkTransitions(transitionHash [HashSize]byte, consume func([H
 }
 
 func (h *History) getContentaddressed(pathFmt string, hash [HashSize]byte) ([]byte, error) {
+	if data, ok := h.getFromCache(hash); ok {
+		return data, nil
+	}
+
 	hashStr := hex.EncodeToString(hash[:])
 	data, err := h.store.Get(fmt.Sprintf(pathFmt, hashStr))
 	if err != nil {
@@ -185,6 +198,8 @@ func (h *History) getContentaddressed(pathFmt string, hash [HashSize]byte) ([]by
 	if !bytes.Equal(hash[:], dataHash[:]) {
 		return nil, HashMismatchError{Expected: hash[:], Actual: dataHash[:]}
 	}
+
+	h.addToCache(hash, data)
 	return data, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/history/history.go
```diff
@@ -185,6 +201,10 @@ func (h *History) getContentaddressed(pathFmt string, hash [HashSize]byte) ([]by
 	if !bytes.Equal(hash[:], dataHash[:]) {
 		return nil, HashMismatchError{Expected: hash[:], Actual: dataHash[:]}
 	}
+
+	h.cacheMu.Lock()
+	defer h.cacheMu.Unlock()
+	h.cache[hash] = data
```
charludo: Maybe clone the data here/on return? Current callers behave correctly, but if some caller would modify `data` (for whatever reason), this would poison the cache. Same for the return above.

internal/history/history_test.go
```diff
@@ -712,6 +725,80 @@ func TestHistory_WalkTransitions(t *testing.T) {
...
+	require.NoError(err)
+
+	manifest := []byte(`{"some": "json"}`)
+	manifestHash, err := h.SetManifest(manifest)
+	require.NoError(err)
+
+	transition1 := &Transition{
+		ManifestHash: manifestHash,
+	}
+	transition1Hash, err := h.SetTransition(transition1)
```
charludo, line 749: unchecked err

internal/history/history_test.go
```diff
@@ -712,6 +725,80 @@ func TestHistory_WalkTransitions(t *testing.T) {
...
+
+	transition2 := &Transition{
+		ManifestHash:           manifestHash,
+		PreviousTransitionHash: transition1Hash,
+	}
+	transition2Hash, err := h.SetTransition(transition2)
```
charludo, line 756: unchecked err

internal/history/history_test.go
```diff
@@ -712,6 +725,80 @@ func TestHistory_WalkTransitions(t *testing.T) {
...
+
+	_, err = h.GetPolicy(otherPolicyHash)
+	require.NoError(err)
+	require.Equal(1, store.getCount, "GetPolicy did not request backend on cache miss")
+	_, err = h.GetPolicy(otherPolicyHash)
+	require.Equal(1, store.getCount, "GetPolicy did not cache after cache miss")
+
+	// Invalid content should not reach the cache.
+	unrelatedHash := Digest([]byte("{}"))
+	backingStore.Set(fmt.Sprintf("manifests/%x", unrelatedHash), manifest)
```
charludo: unchecked return/err

internal/history/history_test.go
```diff
@@ -712,6 +725,80 @@ func TestHistory_WalkTransitions(t *testing.T) {
...
+	require.Equal(0, store.getCount, "GetManifest did not serve cached results")
+
+	_, err = h.GetPolicy(policyHash)
+	require.NoError(err)
+	require.Equal(0, store.getCount, "GetPolicy did not serve cached results")
+
+	// Bypass the history and set content directly.
+	otherPolicy := []byte("AllowRequestsFailingPolicy := true")
+	otherPolicyHash := Digest(otherPolicy)
+	backingStore.Set(fmt.Sprintf("policies/%x", otherPolicyHash), otherPolicy)
```
charludo: unchecked return/err

internal/history/history_test.go
```diff
@@ -712,6 +725,80 @@ func TestHistory_WalkTransitions(t *testing.T) {
...
+
+	// Invalid content should not reach the cache.
+	unrelatedHash := Digest([]byte("{}"))
+	backingStore.Set(fmt.Sprintf("manifests/%x", unrelatedHash), manifest)
+	_, err = h.GetManifest(unrelatedHash)
+	require.ErrorAs(err, &HashMismatchError{})
+	require.Equal(2, store.getCount, "GetPolicy did not cache after cache miss")
+	_, err = h.GetManifest(unrelatedHash)
+	require.ErrorAs(err, &HashMismatchError{})
+	require.Equal(3, store.getCount, "GetPolicy did not cache after cache miss")
```
charludo: All these identical error messages could maybe benefit from also including the count, just for disambiguation.

burgerdev:
Thanks for the review, all comments should be addressed. Data is now cloned in both the inbound and outbound path.

charludo (approved):
LGTM, thank you!

PR #2550, opened 2026-07-27
cli: add `GPUCountValid` pre-generate verifier

charludo:
@burgerdev not sure if there's a better way. I'd like to add this verifier, because pods randomly not coming up with cryptic errors is just bad UX. The problem is, we actually need only `|vCPUs| >= min(|GPUs|, |host NUMMA nodes|)`, where the NUMA nodes are potentially fewer than the available GPUs, but we can't really get the NUMA node count here.


internal/kuberesource/parts.go
```diff
@@ -612,6 +612,43 @@ func GetCPUCount(resources *applycorev1.ResourceRequirementsApplyConfiguration)
...
+		gpuCount := GetGPUCount(container.Resources)
+		// Sidecar containers remain running alongside the actual application, keeping their GPUs attached.
+		if container.RestartPolicy != nil && *container.RestartPolicy == corev1.ContainerRestartPolicyAlways {
+			regularContainersGPU += gpuCount
+		} else {
+			if gpuCount > maxInitContainerGPU {
+				maxInitContainerGPU = gpuCount
+			}
+		}
+	}
```
burgerdev, line 645: Is that the logic used by k8s? I'd have expected all requested devices to be allocated to the VM by the device plugin.

burgerdev (approved):
LGTM, but could you please double check that initContainer and main container can share a GPU, as the currently implemented formula suggests?

charludo:
Closed due to the target PR not being merged.

PR #2554, opened 2026-07-29
treewide: use memory-backed emptydir

burgerdev:
This needs to be labeled breaking in the Kata 4 PR.


charludo (approved):
Thank you!! 

The client deployment in sets.go (line 1036+1037) are missing the medium, that's the only one I found though.

PR #2555, opened 2026-07-30
runtime: increase overhead for GPU workloads

davidweisse:
The runtime overhead on the host is not enough for running large GPU workloads. Allocating a lot of GPU memory only leaves very little memory in the host cgroup, sometimes leading to an OOM. Increasing the runtime overhead fixes this.

This was discovered because normally the cgroup memory max is not actually set. Only now when using pod memory, the memory max is set, which can then lead to an OOM.

Fixes CON-218

The issue includes a reproducer. Without the additional runtime overhead, the model load will fail halfway.

### DIFF
--- a/internal/kuberesource/parts.go
+++ b/internal/kuberesource/parts.go
@@ -60,13 +60,9 @@ func ContrastRuntimeClass(platform platforms.Platform) (*RuntimeClassConfig, err
 
 	// Consists of the default VM memory, 70MiB for the Kata shim and 100MiB for qemu overhead.
 	memoryOverhead := platforms.DefaultMemoryInMebiBytes(platform) + 170
-	if platforms.IsInsecure(platform) && platforms.IsGPU(platform) {
-		// On insecure (non-CC) GPU platforms, iommufd VFIO passthrough pins guest
-		// memory and allocates IOMMU page tables that are charged to the pod's
-		// cgroup. On CC platforms, TDX manages this memory outside the cgroup.
-		// (in the kernel / TDX module memory)
-		// Add extra headroom to avoid OOM kills. 512MB should suffice for VMs up
-		// to ~256GB of memory, which is our current limit on TDX as well.
+	if platforms.IsGPU(platform) {
+		// GPU workloads consume more memory on the host with the cause yet to be found.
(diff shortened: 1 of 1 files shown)


internal/kuberesource/parts.go
```diff
@@ -60,13 +60,16 @@ func ContrastRuntimeClass(platform platforms.Platform) (*RuntimeClassConfig, err
...
+	if platforms.IsGPU(platform) {
 		// On insecure (non-CC) GPU platforms, iommufd VFIO passthrough pins guest
 		// memory and allocates IOMMU page tables that are charged to the pod's
 		// cgroup. On CC platforms, TDX manages this memory outside the cgroup.
 		// (in the kernel / TDX module memory)
 		// Add extra headroom to avoid OOM kills. 512MB should suffice for VMs up
 		// to ~256GB of memory, which is our current limit on TDX as well.
+		//
+		// Running a CC workload with GPU will however also consume more memory on the host,
+		// on SNP as well as TDX, so we increase the overhead here as well.
```
burgerdev: But will an insecure GPU pod now require even more memory? cc @msanft

msanft: I don't get the "even more" - can you elaborate a bit here?

msanft: I'm not a particular fan of having a comment with an explanation for the insecure runtimes but not for the CC runtimes. Perhaps we should just say "cause to be found" for the latter, but make that explicit. Or just say that we assume it to be the same cause as for the insecure runtimes.

msanft (approved):
LGTM besides the comment nit

internal/kuberesource/parts.go
thomasten: Should not require more because IIUC the Pod Memory just enables the cgroup memory limit. I don't know why it isn't enabled without it. For the insecure runtime, it seems to be enabled.

This may also be the actual reason for the observed behavior and the existing comment may be wrong/unrelated.

thomasten: As there are doubts about the validity of the existing comment, I'd remove it and just state that GPU workloads consume more memory (+ "cause to be found" if you like)

msanft (approved):

thomasten (approved):

PR #2559, opened 2026-07-31
apitypes: init

charludo:
Just moving things around, no behavioral changes.

### DIFF
--- a/apitypes/apitypes.go
+++ b/apitypes/apitypes.go
@@ -1,0 +1,10 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+// Package apitypes contains the wire-format types of the Contrast HTTP API.
+//
+// The types here define the request/response bodies exchanged with the Coordinator.
+package apitypes
+
+// Port is the listening port of the HTTP API server.
+const Port = "1314"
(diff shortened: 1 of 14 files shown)


apitypes/apitypes.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2025 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+// Package apitypes contains the wire-format types of the Contrast HTTP API.
+//
+// The types here define the request/response bodies exchanged with the Coordinator.
+package apitypes
+
+// Port is the listening port of the HTTP API server.
+const Port = "1314"
```
burgerdev, line 10: Does the port need to be exported? I was thinking that we might want to co-host the http api on 1313 (dispatch on presence of aTLS ALPN entry), so we should at least make it easy to change.

burgerdev (approved):
lgtm

apitypes/apitypes.go
charludo: Currently, is must be, yes, since we already have three consumers, and this just moves the const into a different package.

The co-hosting idea makes sense, but I'd rather do that separately. This is really just a refactor, and at least until we publish the SDK as stable, IMO we can still change this at any point.

PR #2561, opened 2026-08-03
ci: exclude upgrade-gpu-operator on Metal-QEMU-TDX-GPU

thomasten:
gpu-operator on this cluster is managed manually. It runs with another config than what upgrade-gpu-operator.sh applies. Until now, the gpu-operator version was the same, so the script exited early and did not do anything. Now, a later version is installed in the cluster.


charludo (approved):
Thanks! Also noticed #2563 due to our nightlies.

thomasten:
I expected the gpu-operator update would be transparent to Contrast, sorry :D
As soon as we can confirm that the new version is stable and also works with B300, we should probably upgrade it on the other cluster as well and update the docs to recommend this version.

PR #2564, opened 2026-08-04
 coordinator: add `POST /v1/manifest` handler 

charludo:


apitypes/manifest.go
```diff
@@ -0,0 +1,49 @@
...
+
+// SetManifestRequest is the request body of POST /v1/manifest.
+type SetManifestRequest struct {
+	// Manifest is the JSON-encoded manifest to set.
+	Manifest []byte `json:"manifest"`
+	// Policies are the policies referenced by the manifest.
+	Policies [][]byte `json:"policies"`
+	// PreviousTransitionHash is the expected hash of the latest transition, used for
+	// compare-and-swap. If unset, the update is not conditional.
+	PreviousTransitionHash []byte `json:"previous_transition_hash,omitempty"`
```
burgerdev: The mandatory signature is over the candidate transition, which includes the previous transition hash. Thus, the signature is sufficient to guarantee atomicity, and we don't need this argument.

apitypes/manifest.go
```diff
@@ -0,0 +1,49 @@
...
+	// Signature is the workload owner's signature over the manifest.
+	//
+	// Over HTTP this is the only supported way to authorize an update to an existing
+	// manifest, because the Coordinator can't authenticate the caller by its client
+	// certificate as it does for aTLS-based gRPC calls.
+	Signature []byte `json:"signature,omitempty"`
```
burgerdev: We did not care too much about 0/1/N signatures in proto because that's all the same encoding: https://protobuf.dev/programming-guides/proto3/#conditionally-safe-changes. That does not hold in JSON, so we should go for a slice of signatures to allow for future multi-party manifest updates. The code today should assert that it's either 0 or 1, though.

apitypes/apiv1/manifest.go
```diff
@@ -0,0 +1,49 @@
...
+// SetManifestResponse is the response body of POST /v1/manifest.
+type SetManifestResponse struct {
+	// Version is the Coordinator version.
+	Version string `json:"version"`
+	// RootCA is the PEM-encoded certificate of the deployment's root CA.
+	RootCA []byte `json:"root_ca"`
+	// MeshCA is the PEM-encoded certificate of the deployment's mesh CA.
+	MeshCA []byte `json:"mesh_ca"`
+	// SeedSharesDoc is only set when the initial manifest was set.
+	SeedSharesDoc *SeedShareDocument `json:"seed_shares_doc,omitempty"`
```
burgerdev, line 31: These are all not trustworthy until covered by attestation - we could create reportdata content similar to `/attest`, though.

coordinator/internal/httpapi/manifest.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/edgelesssys/contrast/apitypes"
+	"github.com/edgelesssys/contrast/internal/constants"
+	"github.com/edgelesssys/contrast/internal/userapi"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// maxSetManifestBodySize limits the accepted request body size. A manifest and its policies
+// are much smaller than this, but they're the largest input the Coordinator accepts.
+const maxSetManifestBodySize = 16 << 20 // 16 MiB
```
burgerdev, line 23: Assuming a typical policy entry of 128B, this allows for > 100k entries. I had trouble estimating that without doing the calculation, maybe leave a hint.

Slightly unrealistic upper bound: [150k pods](https://kubernetes.io/docs/setup/best-practices/cluster-large/), and each pod having a unique policy.

Totally unrealistic upper bound: [130k nodes](https://cloud.google.com/blog/products/containers-kubernetes/how-we-built-a-130000-node-gke-cluster/) * 110 pods/node

coordinator/internal/httpapi/manifest.go
```diff
@@ -0,0 +1,133 @@
...
+	contentType := r.Header.Get("Content-Type")
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		writeJSONError(w, http.StatusBadRequest, err)
+		return
+	}
+	if mediaType != "application/json" {
+		writeJSONError(w, http.StatusUnsupportedMediaType, errContentType)
+		return
+	}
```
burgerdev: This is probably common to all endpoints. A pattern I liked was the [mux middleware](https://pkg.go.dev/github.com/gorilla/mux#MiddlewareFunc), but we can also just have a common `checkMediaType` function.

apitypes/manifest.go
charludo: done

apitypes/apiv1/manifest.go
charludo: So, client sends nonce in request, response gets a new field with `sha256(nonce || transitionDigest || sha256(rootCA) || sha256(meshCA) || sha256(seedShareDoc))`?

Should I already implement this, or is it pending a design decision re: Teams thread?

coordinator/internal/httpapi/manifest.go
charludo: done, went with a common function, seemed simpler for the current needs

apitypes/apiv1/manifest.go
burgerdev: Yes, I fear this endpoint needs to be discussed first (together with the other RPC methods that rely on  attestation).

charludo: Alright. Then let's leave it unapproved for now, until everything else is done it's not going anywhere anyways. The PRs layered on top don't really depend on this decision luckily.

charludo:
Closed for now, waiting on #2638 to be finalized and can be reworked afterwards.

PR #2566, opened 2026-08-04
treewide: b300 support

msanft:
In anticipation of Privatemode's B300 use, we can already prepare some (but not all) places in Contrast for B300 GPU support. For now, this PR uses the `GB110_B300_SXM6_AC` device ID from [this table](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/README.md#compatible-gpus) and the [name normalization logic](https://github.com/NVIDIA/sandbox-device-plugin/blob/8e76fe813d18a6bd86b429c0095379dddf2e7833/pkg/device_plugin/device_plugin.go#L223) in the sandbox-device-plugin.
Closes CON-290


e2e/gpu/gpu_test.go
```diff
@@ -34,6 +34,14 @@ const (
 	nvidiaLibPath     = "/usr/local/nvidia/lib64"
 )
 
+var supportedGPUClasses = []corev1.ResourceName{
+	"nvidia.com/GB100_B200",
+	"nvidia.com/GB110_B300_SXM6_AC",
+	"nvidia.com/GH100_H100_PCIE",
+	// TODO(msanft): remove when the GPU operator is updated in all clusters.
+	"nvidia.com/pgpu",
```
thomasten: I don't remember what the reason was in Contrast to use pgpu instead of specific names. The user docs also recommend this. Was that related to the gpu-operator and will this change with the new version?

msanft: > Was that related to the gpu-operator and will this change with the new version?

Exactly. Hence the comment

thomasten: I vaguely remember some hacky patch inside Contrast to make specific names work esp. for Privatemode, which code path we didn't want to be taken for the general audience. But I might remember that wrong.

justfile
```diff
@@ -269,8 +269,24 @@ populate target=default_deploy_target platform=default_platform set=default_set:
         dmesgFlag="--add-dmesg"
     fi
     gpuFlags=()
-    if [[ {{ platform }} == "Metal-QEMU-TDX-GPU" ]] ; then
-        gpuFlags=("--gpu-class" "nvidia.com/GB100_B200")
+    if [[ "$target" == "gpu" ]] ; then
+        gpuClass=$(kubectl get nodes -o json | jq -er '
+            [
```
charludo, line 274: Why switch to the check of target here? IMO we still want to be able to get a GPU deployment for e.g. the `custom` target above. Could we not instead check for the platform being one of the GPU platforms?

e2e/gpu/gpu_test.go
```diff
@@ -220,7 +213,7 @@ func TestGPU(t *testing.T) {
 
 			require.Len(gpuLines, int(expectedGPUAmount), "expected %d GPUs, got %d:\n%s", expectedGPUAmount, len(gpuLines), stdout)
 			for i, line := range gpuLines {
-				require.Contains(line, gpuName, "GPU %d should be %s", i, gpuName)
+				require.Regexp(`^GPU [0-9]+: NVIDIA .+ \(UUID: GPU-[^)]+\)$`, line, "GPU %d has unexpected nvidia-smi output", i)
```
charludo, line 244: I'd prefer if we mapped explicitly what we expect here, something along the lines of
```go
  var expectedModel = map[corev1.ResourceName]string{
      "nvidia.com/GB100_B200":         "B200",
      "nvidia.com/GB110_B300_SXM6_AC": "B300",
      "nvidia.com/GH100_H100_PCIE":    "H100",
  }
```
since otherwise, this checks dropped from "this is an NVIDIA B200" (or whatever) to just "this is some NVIDIA card", so the test now succeeds with the wrong/unexpected GPU.

e2e/gpu/gpu_test.go
```diff
@@ -34,6 +34,14 @@ const (
 	nvidiaLibPath     = "/usr/local/nvidia/lib64"
 )
 
+var supportedGPUClasses = []corev1.ResourceName{
+	"nvidia.com/GB100_B200",
+	"nvidia.com/GB110_B300_SXM6_AC",
+	"nvidia.com/GH100_H100_PCIE",
+	// TODO(msanft): remove when the GPU operator is updated in all clusters.
+	"nvidia.com/pgpu",
+}
```
charludo: There's now four places where we maintain this list (here, justfile, both genpoicy patches). Would you mind adding a 5-line devdoc just stating "if you want to add a new GPU, here's all the places you need to consider?"

e2e/gpu/gpu_test.go
```diff
@@ -245,6 +238,37 @@ func TestMain(m *testing.M) {
...
+			if ok && quantity.Value() > maxQuantity {
+				maxQuantity = quantity.Value()
+			}
+		}
+		if maxQuantity > 0 {
+			return string(gpuClass), min(maxQuantity, int64(2)), nil
+		}
+	}
+
+	// error case
```
charludo:
```suggestion
	// No supported class found, so build useful diagnostics.
```

charludo (commented):
Thanks for your efforts! This generally reads good to me, I have a couple nits in the comments + the below:

- there's a leftover comment in genpolicy/package.nix:58 that should probably also go in the "fix outdated comments" commit.

justfile
msanft: Good point, I'll just check if it's a `*-GPU` platform instead.

e2e/gpu/gpu_test.go
msanft: I think we (i.e. @charludo, iirc) removed that in Kata at some point when we didn't need it anymore?

charludo: I'm not sure exactly what patch you mean, do you remember when it was dropped?

In any case, iiuc, specific names should work: https://github.com/edgelesssys/contrast/blob/main/packages/by-name/kata/source/0018-shim-guess-CDI-devices-without-direct-match.patch

charludo (approved):
Thanks for the changes! LGTM, also since nothing existing seems to break.

e2e/gpu/gpu_test.go
thomasten: #2180 says that that patch "is brittle and should not be relied upon". I don't know why this is or whether this has changed.

docs/docs/howto/cluster-setup/bare-metal.md
```diff
@@ -247,7 +247,7 @@ helm install --wait --generate-name \
   --set sandboxDevicePlugin.image=nvidia-sandbox-device-plugin \
   --set sandboxDevicePlugin.version=8e76fe81 \
   --set 'sandboxDevicePlugin.env[0].name=P_GPU_ALIAS' \
-  --set 'sandboxDevicePlugin.env[0].value=pgpu' \
+  --set 'sandboxDevicePlugin.env[0].value=' \
```
thomasten: I think setting this to empty is wrong for this version of the gpu-operator

packages/by-name/scripts/upgrade-gpu-operator/upgrade-gpu-operator.sh
thomasten, line 1: This script now mixes stuff for the old gpu-operator and the new gpu-operator. The result will probably not work for either.

packages/by-name/scripts/upgrade-gpu-operator/upgrade-gpu-operator.sh
```diff
@@ -6,6 +6,98 @@ set -euo pipefail
...
 BLACKWELL=false
 GPU_OPERATOR_VERSION=""
+GPU_OPERATOR_NAMESPACE="gpu-operator"
+B300_GPU_RESOURCE="nvidia.com/GB110_B300_SXM6_AC"
+# This image supplies BusyBox for the init container; the deployed 8e76fe81 image is distroless.
+B300_PCI_IDS_IMAGE="nvcr.io/nvidia/cloud-native/nvidia-sandbox-device-plugin:v0.0.4@sha256:abb7cc6d692f9e9c0e79b8d7eac39dc2c4ea7b7c4b1ad7094109b5bc846f8080"
+
+# TODO: Remove this workaround once the sandbox device plugin includes PCI ID 0x3182.
+# https://github.com/NVIDIA/gpu-operator/pull/2625
+patch_b300_pci_ids() {
```
thomasten: The patch seems overly complicated. Can't we just put an up-to-date pci.ids in a ConfigMap and mount that into the main container to override the file?

msanft: Yes. I haven't got to review the patch yet, but I've verified it works for the new version of the operator.

dev-docs/nvidia/cdi.md
```diff
@@ -154,6 +154,11 @@ The Kata agent receives `CreateContainer` requests and inspects their annotation
 Each CDI annotation is passed to the CDI library, which extends the OCI runtime config according to the CDI spec generated by the NVIDIA CTK.
 Here, the journey ends with containers equipped with their expected devices.
 
+## Adding support for a GPU model
+
+When adding a model-specific GPU resource name, update the mapping in the [GPU end-to-end test](../../e2e/gpu/gpu_test.go), the discovery list in the [root recipe file](../../justfile), and both the [development](../../packages/by-name/kata/genpolicy/genpolicy_settings_dev.patch) and [production](../../packages/by-name/kata/genpolicy/genpolicy_settings_prod.patch) genpolicy settings patches.
```
msanft:
```suggestion
When adding a model-specific GPU resource name, update the mapping in the [GPU E2E test](../../e2e/gpu/gpu_test.go), the discovery list in the [justfile](../../justfile), and both the [development](../../packages/by-name/kata/genpolicy/genpolicy_settings_dev.patch) and [production](../../packages/by-name/kata/genpolicy/genpolicy_settings_prod.patch) genpolicy settings patches.
```

docs/docs/howto/cluster-setup/bare-metal.md
```diff
@@ -227,105 +227,27 @@ sed -n '/^# Install the GPU Operator$/,/^$/p' ../../../../packages/by-name/scrip
...
-    "path": "/spec/rules/11/matchAny/1/matchFeatures/0/matchExpressions/nvidia.com~1gpu.family/value/-",
-    "value": "blackwell"
-  }
-]'
-```
-
-</details>
+GPU Operator 26.3 supports Blackwell GPU detection and confidential-computing configuration without additional node feature or CC manager patches.
+The sandbox device plugin bundled with this release still needs a newer PCI database to identify B300 devices (`10de:3182`).
+Contrast's `upgrade-gpu-operator` maintenance script installs that database when invoked with `--blackwell`.
```
msanft: Don't reference the internal script in public documentation. Instead, we should mirror the past state and explain the procedure to patch manually briefly here.

e2e/gpu/gpu_test.go
```diff
@@ -43,35 +66,21 @@ func TestGPU(t *testing.T) {
...
+	gpuConfigs, err := gpuConfigsFromNodes(nodes.Items)
+	require.NoError(t, err)
 
-	resources := kuberesource.GPU(deploymentName, gpuClass, expectedGPUAmount)
+	var resources []any
+	for i, config := range gpuConfigs {
+		t.Logf("Using GPU resource %s (%s) with quantity %d", config.resource, config.model, config.quantity)
+		// The hostpath CSI volume is node-local. Exercise the block-device
+		// regression once without preventing other GPU models from running on
+		// their respective nodes.
```
msanft:
```suggestion
```
unnecessary comment

packages/by-name/scripts/upgrade-gpu-operator/upgrade-gpu-operator_test.sh
msanft, line 1: Drop this

internal/kuberesource/parts.go
```diff
@@ -87,6 +87,20 @@ func NodeInstallers(namespace string, platforms []platforms.Platform) ([]*applya
 	return out, nil
 }
 
+func nvidiaGPUToleration() *applycorev1.TolerationApplyConfiguration {
```
msanft: Add a comment to remove this when the node is untainted.

docs/docs/howto/cluster-setup/bare-metal.md
```diff
@@ -227,106 +227,69 @@ sed -n '/^# Install the GPU Operator$/,/^$/p' ../../../../packages/by-name/scrip
...
-0000:18:00.0 3D controller [0302]: NVIDIA Corporation GB100 [B200] [10de:2901] (rev a1)
-        Subsystem: NVIDIA Corporation Device [10de:1999]
-        Kernel driver in use: vfio-pci
-        Kernel modules: nvidiafb, nouveau
-```
-
-In this case, the device ID is `2901`. The device ID then needs to be added to the list of supported GPUs:
+GPU Operator 26.3 supports Blackwell GPU detection and confidential-computing configuration without additional node feature or CC manager patches.
+The sandbox device plugin bundled with this release still needs a newer PCI database to identify B300 devices (`10de:3182`).
+A complete current `pci.ids` exceeds the ConfigMap size limit, so retain only the NVIDIA vendor records and PCI class definitions before mounting the database into the sandbox device plugin:
```
msanft:
```suggestion
Version 26.3 of the GPU operator supports Blackwell GPU detection and confidential-computing configuration without additional node feature or CC manager patches. For B300 GPUs, however, a small patch is necessary as the sandbox device plugin bundled with this release carries an outdated PCI database lacking the B300 device ID (`10de:3182`).

<details>
<summary>Enabling Support for B300 GPUs</summary>
...
```

internal/kuberesource/parts_test.go
```diff
@@ -4,11 +4,37 @@
...
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	applyappsv1 "k8s.io/client-go/applyconfigurations/apps/v1"
 )
 
+func TestGPUBlockDevice(t *testing.T) {
```
msanft: Drop this test

.github/workflows/bm_maintenance_platform.yml
```diff
@@ -66,9 +66,10 @@ jobs:
         run: |
           GPU_OPERATOR_VERSION=$(jq -r '."gpu-operator".currentValue' ./tools/bm-maintenance/versions.json)
           OPTS=(--version "$GPU_OPERATOR_VERSION")
-          if [[ "${PLATFORM_NAME}" != "Metal-QEMU-TDX-GPU" ]]; then
-            nix run ".#${SET}.scripts.upgrade-gpu-operator" -- "${OPTS[@]}"
+          if [[ "${PLATFORM_NAME}" == "Metal-QEMU-TDX-GPU" ]]; then
+            OPTS+=(--blackwell)
           fi
+          nix run ".#${SET}.scripts.upgrade-gpu-operator" -- "${OPTS[@]}"
```
thomasten: gpu-operator in this cluster should not be managed by Contrast CI because it should be managed the same way in pm-testing as in pm-prod.

msanft: Hmm, this would be a breaking change compared to what we had before. I can do that though. Then we can drop Blackwell support in this script altogether?

thomasten: I don't understand what is the breaking change here, but if any, I did it in #2561, no? :D

msanft: Oh, i didn't know about that PR. I thought we run the upgrade routinely in the PM test cluster. I'll then just drop Blackwell support here if we don't have any means to test it anyways? Perhaps document this in PM instead

thomasten: sounds good :)

msanft:
IMO, this is ready to be merged now. @charludo can we coordinate this with the release somehow?

dev-docs/e2e/dgx-007/manifest.json
```diff
@@ -27,14 +28,32 @@
...
+    },
+    {
+      "op": "copy",
+      "from": "/tdx/0",
+      "path": "/tdx/-"
+    },
+    {
+      "op": "replace",
+      "path": "/tdx/1/MrSeam",
+      "value": "ab62561a173acbd18ee50ff37750db44184c6cf5e886df74247cc575e163b04c34b9e18374757c235affa614d4127f6b"
```
sespiros, line 73: In the case of multi-runtime-class, won't we have 2 entries under `/tdx` already? `copy` will copy the value to index 2 but then this line will overwrite a valid `mrseam` I think.

charludo (approved):
Sorry it took so long. Only thing I noticed is docs/docs/howto/workload-deployment/GPU-configuration.md:45 still instructs to use `nvidia.com/pgpu`. But that is OK in a follow up, from my side, just merge 🎉

PR #2568, opened 2026-08-04
sdk: negotiate API version

charludo:


charludo:
@sespiros only requesting review on this one for now, since all other API PRs require it.

sdk/internal/transport/transport.go
```diff
@@ -0,0 +1,114 @@
+// Copyright 2026 Edgeless Systems GmbH
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build contrast_unstable_api
+
+// Package transport implements the HTTP plumbing shared by the SDK's API versions.
+//
+// It's internal so that the versioned API packages can share one connection setup
+// without exposing it as part of the SDK's public surface.
+package transport
```
sespiros: Why the generic name? Is this going to have something other than http? I find the name a bit confusing and there are 3 clients now; before we had the sdk client which used an `http.Client`, now the sdk client uses a "transport" which actually is another `Client` struct itself wrapping an `http.Client`. 

Some suggestions:
- rename the package to `sdk/internal/httpapi`
- extend the package doc to say why the layer exists (shared HTTP request plumbing for the versioned API packages because those won't be able to miport the sdk back)

sdk/internal/transport/transport.go
```diff
@@ -0,0 +1,114 @@
...
+	// HTTPClient is used to contact the Coordinator.
+	HTTPClient *http.Client
+	// BaseURL is the Coordinator's HTTP API root, e.g. "http://coordinator:1314".
+	BaseURL string
+	// Log receives diagnostics that aren't part of a returned error.
+	Log *slog.Logger
+}
+
+// URL resolves an API path like "/v1/manifest" against the client's base URL.
+func (c *Client) URL(path string) (string, error) {
```
sespiros: Does this need to be exported?

apitypes/apitypes.go
```diff
@@ -8,3 +8,11 @@ package apitypes
 
 // Port is the listening port of the HTTP API server.
 const Port = "1314"
+
+// APIError is the body returned by the Contrast HTTP API if a request was not successful.
+type APIError struct {
+	// Version is the Coordinator version.
+	Version    string `json:"version"`
+	StatusCode int    `json:"status_code"`
```
sespiros: StatusCode is added here but nothing populates it yet (`writeJSONError` still sets only Version and Err), so as of this PR every `/attest` error body carries `"status_code":0`. Pull `writeJSONError` in this PR, or add the field together with the code that sets it.

Separately: `DoJSON` parses the `APIError` and then flattens it into a `fmt.Errorf` string, so callers can only string-match. The SDK is supposed to be the library others build on, so make this a public typed error that works with `errors.As` (e.g. apitypes.APIError implementing error), carrying the HTTP status and the parsed wire fields. It can't live under sdk/internal or callers can't name it. See my other comment.

sdk/internal/httpapi/httpapi.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		details := httpResp.Status
+		var apiErr apitypes.APIError
+		if err := json.Unmarshal(errBody, &apiErr); err == nil {
+			details = apiErr.Err
+		} else {
+			c.Log.Error("parsing error response", "err", err, "response", string(errBody))
+		}
+		return nil, fmt.Errorf("HTTP API call failed with %d (%s): %s", httpResp.StatusCode, http.StatusText(httpResp.StatusCode), details)
+	}
```
sespiros, line 110: mentioned in the error definition as well. i.e

```suggestion
    if httpResp.StatusCode != http.StatusOK {
          errBody, err := io.ReadAll(httpResp.Body)
          if err != nil {
                  return nil, fmt.Errorf("reading response (status code %d): %w", httpResp.StatusCode, err)
          }
          apiErr := &apitypes.APIError{Err: httpResp.Status}
          if err := json.Unmarshal(errBody, apiErr); err != nil {
                  c.Log.Error("parsing error response", "err", err, "response", string(errBody))
          }
          apiErr.StatusCode = httpResp.StatusCode // response status is authoritative
          return nil, apiErr
    }
```

then clients can do:
```
_, err := client.GetAttestation(ctx, nonce)
var apiErr *apitypes.APIError
if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusPreconditionFailed {
      // coordinator has no manifest yet / needs recovery
}
```

sdk/internal/httpapi/httpapi.go
```diff
@@ -0,0 +1,114 @@
...
+
+// URL resolves an API path like "/v1/manifest" against the client's base URL.
+func (c *Client) URL(path string) (string, error) {
+	if c.BaseURL == "" {
+		return "", ErrBaseURLUnset
+	}
+	base, err := url.Parse(c.BaseURL)
+	if err != nil {
+		return "", fmt.Errorf("parsing base URL %q: %w", c.BaseURL, err)
+	}
```
sespiros, line 48: `WithBaseURL("coordinator:1314")` (scheme forgotten) parses as an opaque URL: the path splice below is a no-op and `URL("/attest")` returns `"coordinator:1314"` with the API path silently dropped. Since the docs advertise `http://coordinator:1314`, better to reject this early:

```suggestion
      base, err := url.Parse(c.BaseURL)
      if err != nil {
              return "", fmt.Errorf("parsing base URL %q: %w", c.BaseURL, err)
      }
      if (base.Scheme != "http" && base.Scheme != "https") || base.Host == "" {
              return "", fmt.Errorf("base URL %q: expected http(s)://host[:port][/prefix], set it with WithBaseURL", c.BaseURL)
      }
```

sdk/internal/transport/transport.go
```diff
@@ -0,0 +1,114 @@
...
+	ref, err := url.Parse(path)
+	if err != nil {
+		return "", fmt.Errorf("parsing path %q: %w", path, err)
+	}
+	// Ensure a base URL with a path prefix, as used by reverse proxies, keeps that
+	// prefix: "https://proxy/contrast" + "/v1/manifest" resolves to
+	// "https://proxy/contrast/v1/manifest".
+	base.Path = strings.TrimSuffix(base.Path, "/") + ref.EscapedPath()
+	base.RawQuery = ref.RawQuery
+	return base.String(), nil
```
sespiros: Assigning `ref.EscapedPath()` (encoded) into `base.Path` (decoded) corrupts escaped paths: `URL("/v1/foo%2Fbar")` returns `/v1/foo%252Fbar`. Latent while all paths are literals, but joining in escaped space via `JoinPath` fixes it and keeps the proxy-prefix behavior (verified for prefix, trailing-slash, %2F, and query cases):

```suggestion
    // Ensure a base URL with a path prefix, as used by reverse proxies, keeps that
    // prefix: "https://proxy/contrast" + "/v1/manifest" resolves to
    // "https://proxy/contrast/v1/manifest". Joining the escaped forms keeps
    // percent-encoded segments intact.
    joined := base.JoinPath(ref.EscapedPath())
    joined.RawQuery = ref.RawQuery
    return joined.String(), nil
```

sdk/internal/transport/transport.go
sespiros: Also the edge cases raised below could be caught by a simple matrix test in this package.

sdk/capabilities.go
```diff
@@ -0,0 +1,69 @@
...
+//
+// It's held behind a pointer so that copying a [Client] doesn't copy the mutex.
+type negotiation struct {
+	mu      sync.Mutex
+	version string
+}
+
+// NegotiateAPIVersion returns the newest API version supported by both this SDK and the Coordinator.
+//
+// The result is cached, so this costs at most one request per [Client].
```
sespiros: nit: worth clarifying that the first successful result is cached?

sdk/internal/transport/transport.go
```diff
@@ -0,0 +1,114 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("reading response (status code %d): %w", httpResp.StatusCode, err)
+		}
+		details := httpResp.Status
+		var apiErr apitypes.APIError
+		if err := json.Unmarshal(errBody, &apiErr); err == nil {
+			details = apiErr.Err
+		} else {
+			c.Log.Error("parsing error response", "err", err, "response", string(errBody))
+		}
```
sespiros: nit: A body that is valid JSON but not an APIError (e.g. {} from a proxy) sets details to the empty string and the returned error ends in a dangling ": ". Keep the status fallback unless there's an actual message:

(note the else if condition)
```suggestion
      details := httpResp.Status
      var apiErr apitypes.APIError
      if err := json.Unmarshal(errBody, &apiErr); err != nil {
              c.Log.Error("parsing error response", "err", err, "response", string(errBody))
      } else if apiErr.Err != "" {
              details = apiErr.Err
      }
```

sdk/verify.go
```diff
@@ -21,14 +19,17 @@ import (
...
 	"github.com/edgelesssys/contrast/internal/manifest"
+	"github.com/edgelesssys/contrast/sdk/internal/transport"
 	"github.com/spf13/afero"
 )
 
+// attestPath is the path of the Coordinator's attestation endpoint, which is unversioned.
+const attestPath = "/attest"
+
 // Client is used to interact with a Contrast deployment.
 type Client struct {
```
sespiros: `Client`, `New` and all the `With*` builders now live in `verify.go` but with the client growing negotiation and versioned APIs, a `client.go` holidng the type constructor and builders would make the package easier to read I think.

sespiros (changes requested):
Thanks, did a pass and left a few comments.

sdk/internal/transport/transport.go
charludo: Fair point, done

sdk/internal/transport/transport.go
charludo: No, indeed not. Done.

apitypes/apitypes.go
charludo: I originally wanted to keep this PR completely behavior-change free. That's out the window now anyways, so... Done :)

sdk/internal/httpapi/httpapi.go
charludo: Oh yeah, good idea. Done.

sdk/internal/httpapi/httpapi.go
charludo: done

sdk/internal/transport/transport.go
charludo: done

sdk/internal/transport/transport.go
charludo: Fair, added

sdk/capabilities.go
charludo: done

sdk/internal/transport/transport.go
charludo: indeed, done

sdk/verify.go
charludo: Yeah fait. extracted

sespiros (approved):
Thanks, all addressed. One leftover: since writeJSONError now sets status_code here, the handler test could assert the body field in this PR too (right now that assertion only comes with the next PR in the stack).

sdk/capabilities.go
```diff
@@ -0,0 +1,69 @@
...
+// supportedAPIVersions are the API versions this SDK can speak, newest first.
+var supportedAPIVersions = []string{apiv1.Version}
+
+// negotiation caches the result of API version negotiation.
+//
+// It's held behind a pointer so that copying a [Client] doesn't copy the mutex.
+type negotiation struct {
+	mu      sync.Mutex
+	version string
+}
```
burgerdev: Is there a reason why these are not direct client fields?

sdk/capabilities.go
```diff
@@ -0,0 +1,69 @@
...
+// This endpoint is deliberately unversioned. It's how clients discover which versions
+// exist, so it must be reachable without knowing a version first.
+const capabilitiesPath = "/capabilities"
+
+// supportedAPIVersions are the API versions this SDK can speak, newest first.
+var supportedAPIVersions = []string{apiv1.Version}
+
+// negotiation caches the result of API version negotiation.
+//
+// It's held behind a pointer so that copying a [Client] doesn't copy the mutex.
```
burgerdev: Where does the requirement for copying the client come from? Can we drop it? Linters usually warn when copying a mutex.

sdk/client.go
```diff
@@ -0,0 +1,110 @@
...
+	// validatorsFromManifestOverride is used by tests to replace the validators.
+	validatorsFromManifestOverride func(*certcache.CachedHTTPSGetter, *manifest.Manifest, *slog.Logger) (validators.Validator, error)
+}
+
+// New returns a new [Client].
+//
+// Logging is disabled by default, and a memory-backed cache is used.
+// For HTTP interactions, [http.DefaultClient] is used by default.
+//
+// The Coordinator's address must be set via [Client.WithBaseURL] before any API call.
```
burgerdev: Do we have a use case for not setting the base URL? Otherwise, I'd pass it as a mandatory parameter to `New`.

sdk/capabilities.go
charludo: The existing `GetAttestation`, `ValidateAttestation` and (for consistency... well, not that well motivated...) the new `NegotiateAPIVersion` were all value receivers. Switched them to pointer receivers.

sdk/capabilities.go
charludo: Made them so, now that copyability is gone.

sdk/client.go
charludo: Hm... No, not really. Made them mandatory.

charludo: (other than not having to touch the existing tests 😄 )

sespiros (approved):

PR #2576, opened 2026-08-07
scripts: fix cleanup-bare-metal RuntimeClass extraction

thomasten:
The first result from the second command was appended without whitespace and thus not recognized as used RuntimeClass.


charludo (approved):
😵‍💫

PR #2579, opened 2026-08-10
.github: move vuln-scan out of release nightly

charludo:


.github/workflows/vuln_scan.yml
```diff
@@ -61,3 +65,23 @@ jobs:
...
+    if: ${{ failure() && github.event_name == 'schedule' }}
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - name: Notify teams channel
+        if: ${{ github.run_attempt == 1 }}
```
sespiros: This could be folded into the job level if to save ourselves of a checkout that will be skipped anyway. The rest of our notifications are done this way also.

.github/workflows/vuln_scan.yml
```diff
@@ -61,3 +65,23 @@ jobs:
...
+        uses: ./.github/actions/post_to_teams
+        with:
+          webhook: ${{ secrets.TEAMS_CI_WEBHOOK }}
+          title: "Nightly vulnerability scan failed"
+          message: "Check the run for findings that no ignore file covers."
```
sespiros: This message assumes a specific failure when the job could fail on SBOM build, cachix or nix errors.

.github/workflows/vuln_scan.yml
```diff
@@ -1,12 +1,16 @@
...
+  schedule:
+    - cron: "45 2 * * *" # 02:45 UTC, every day
   workflow_call:
     secrets:
       GITHUB_TOKEN_IN:
         required: false
       CACHIX_AUTH_TOKEN:
         required: true
+      TEAMS_CI_WEBHOOK:
+        required: false
```
sespiros: These 2 lines are not needed and can be removed. The notification only works on the scheduled path anyway which works.

sespiros (changes requested):
Thanks, some small inline comments.

sespiros:
One more thing that I realized is that if we skip scanning on release branches, we are missing potential advisories which get published for one of our releases (well our latest release) after a release is cut, which is what we want.

charludo:
Thanks for the review, implemented the changes.

> One more thing that I realized is that if we skip scanning on release branches, we are missing potential advisories which get published for one of our releases (well our latest release) after a release is cut, which is what we want.

Hm, I think adding a comment about checking (this) nightly run before cutting a release should be enough, right?

sespiros:
> Hm, I think adding a comment about checking (this) nightly run before cutting a release should be enough, right?

Depends on what we want to do here. Say a release is out for a week already and a vulnerability gets published for one of its dependencies. Even if that dependency gets bumped in main, the released version keeps being vulnerable. Intuitively I would expect 2 notifications one saying that a vulnerability affects main and another one saying that our active release is affected.

charludo:
> > Hm, I think adding a comment about checking (this) nightly run before cutting a release should be enough, right?
> 
> Depends on what we want to do here. Say a release is out for a week already and a vulnerability gets published for one of its dependencies. Even if that dependency gets bumped in main, the released version keeps being vulnerable. Intuitively I would expect 2 notifications one saying that a vulnerability affects main and another one saying that our active release is affected.

But that's unrelated to this PR, right? We're not currently doing that. Up until this point, the vuln-scan 1) after a PR is merged and 2) nightly in the release workflow, and only on the tip of main. With this PR, 1) stays the same, and 2) moves the workflow step out into its own workflow (+ adds notification).

Interesting idea to keep scanning on release branches. I guess we could do that.... But should be seldomly enough to not flood us, and probably only for a supported subset of most recent versions that still get patch updates. That needs a discussion about which versions we DO support though, I don't think we can decide that here

sespiros:
> But that's unrelated to this PR, right? We're not currently doing that. Up until this point, the vuln-scan 1) after a PR is merged and 2) nightly in the release workflow, and only on the tip of main. With this PR, 1) stays the same, and 2) moves the workflow step out into its own workflow (+ adds notification).
> 
> Interesting idea to keep scanning on release branches. I guess we could do that.... But should be seldomly enough to not flood us, and probably only for a supported subset of most recent versions that still get patch updates. That needs a discussion about which versions we DO support though, I don't think we can decide that here

Yeap maybe open an issue on linear about it. Also there seems to be a failing flake check.

sespiros (approved):

PR #2580, opened 2026-08-10
genpolicy: support insecure registries

davidweisse:
This allows specifying insecure registries for `contrast generate`, e.g., running a local registry and then calling:
```
contrast generate --insecure-registry localhost:5000 ...
```
We need this for the policy test suite to speed up policy generation with a local registry.

Fixes CON-163

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -93,6 +93,7 @@ subcommands.`,
 	cmd.Flags().Bool("insecure-enable-debug-shell-access", false, "enable the debug shell service in the pod CVM to get access from container to guest VM")
 	cmd.Flags().Bool("calculate-pod-memory", false, "calculate pod memory based on image layer sizes and container resource limits")
 	cmd.Flags().StringP("output", "o", "", "output file for generated YAML")
+	cmd.Flags().StringArray("insecure-registry", []string{}, "registries to access via plain HTTP instead of HTTPS (can be passed more than once)")
 	must(cmd.MarkFlagFilename("policy", "rego"))
 	must(cmd.MarkFlagFilename("settings", "json"))
 	must(cmd.MarkFlagFilename("manifest", "json"))
@@ -205,6 +206,10 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("patch targets: %w", err)
 	}
 	fmt.Fprintln(cmd.OutOrStdout(), "✔️ Patched targets")
+
+	if len(flags.insecureRegistries) > 0 {
+		fmt.Fprintln(cmd.OutOrStdout(), "⚠️ Using insecure registries for policy generation!")
+	}
 
 	if err := generatePolicies(cmd.Context(), flags, fileMap, extraFile.Name(), log); err != nil {
 		return fmt.Errorf("generate policies: %w", err)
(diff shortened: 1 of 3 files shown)


cli/cmd/generate.go
```diff
@@ -93,6 +93,7 @@ subcommands.`,
 	cmd.Flags().Bool("insecure-enable-debug-shell-access", false, "enable the debug shell service in the pod CVM to get access from container to guest VM")
 	cmd.Flags().Bool("calculate-pod-memory", false, "calculate pod memory based on image layer sizes and container resource limits")
 	cmd.Flags().StringP("output", "o", "", "output file for generated YAML")
+	cmd.Flags().StringArray("insecure-registry", []string{}, "insecure registries to allow for generating policies (can be passed more than once)")
```
charludo:
```suggestion
	cmd.Flags().StringArray("insecure-registry", []string{}, "insecure registries to allow for generating policies (can be passed more than once)")
	must(cmd.Flags().MarkHidden("insecure-registry"))
```

Unless there's a usecase for external users for this flag?

cli/genpolicy/genpolicy.go
```diff
@@ -24,13 +24,14 @@ import (
...
-	cachePath    string
+	rulesPath          string
+	settingsPath       string
+	cachePath          string
+	insecureRegistries []string
 }
 
 // New creates a new Runner for the given configuration.
-func New(rulesPath, settingsPath, cachePath string, bin []byte) (*Runner, error) {
+func New(rulesPath, settingsPath, cachePath string, insecureRegistries []string, bin []byte) (*Runner, error) {
```
charludo, line 34: This is getting to the point where we might want to introduce [functional options](https://go-patterns.dev/creational/functional-options).... Though to be fair, this is FAR from the only location with many-arged functions, so definitely not in this PR.

You know what, ignore this, this is just a reminder to myself to create a ticket.

cli/cmd/generate.go
charludo: Edit: ...and there very well might be. IDK, I'm not in the loop here. Was that something agreed upon previously?

charludo: This sounds somewhat like we're bypassing TLS verification, when in reality we only try HTTP and HTTPs-only registries will fail. Maybe copy upstream's wording? 

```suggestion
	cmd.Flags().StringArray("insecure-registry", []string{}, "registries to access via plain HTTP instead of HTTPS (can be passed more than once)")
```

cli/genpolicy/genpolicy_test.go
```diff
@@ -39,6 +40,9 @@ while [ $# -gt 0 ]; do
 	--layers-cache-file-path=*)
 	  printf "%%s" "[]" >${1#--layers-cache-file-path=}
 	;;
+	--insecure-registry=*)
+	  printf "%%s" "${1#--insecure-registry=}" >>insecure_registries
```
charludo, line 44:
```suggestion
	  printf "%%s" "${1#--insecure-registry=}" >insecure_registries
```
Or rather, is this intentional? Asking because the other options all append

cli/cmd/generate.go
charludo: Also, we probably should print a serious warning ala insecure debug shell when the flag is used

charludo (commented):
Nice, thank you! I almost exclusively have nits, nothing substantial.

cli/cmd/generate.go
davidweisse: Nothing discussed previously, but I can at least imagine a similar use case as ours, i.e., using a registry mirror to speed up the pull time.

cli/genpolicy/genpolicy_test.go
davidweisse: This is intentional to test multiple insecure registries on the command line.

cli/cmd/generate.go
charludo: Yeah... I guess this isn't security-compromising, the images are pinned idea way. If you want to keep it as-is maybe ping Markus here for after his vacation? Just to make sure.

cli/genpolicy/genpolicy_test.go
charludo: Ah, duh.

charludo (approved):
Thanks for the fixups!

PR #2581, opened 2026-08-10
packages/by-name/nvidia-driver: init

charludo:
Got annoyed with not having a script to update the hashes for the GPU drivers.

### DIFF
--- a/packages/by-name/nvidia-driver/package.nix
+++ b/packages/by-name/nvidia-driver/package.nix
@@ -1,0 +1,16 @@
+# Copyright 2026 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+{
+  callPackage,
+  kata,
+  linuxPackagesFor,
+}:
+
+# The kernel modules have to be built against the very kernel the PodVM boots, so this
+# must stay in sync with the `boot.kernelPackages` set in packages/nixos/kata.nix.
+callPackage ./driver.nix {
+  inherit (linuxPackagesFor (kata.kernel-uvm.override { withGPU = true; }))
+    nvidiaPackages
+    ;
+}
--- a/packages/by-name/nvidia-driver/package.nix
+++ b/packages/by-name/nvidia-driver/package.nix
@@ -7,10 +7,8 @@
   linuxPackagesFor,
 }:
 
-# The kernel modules have to be built against the very kernel the PodVM boots, so this
-# must stay in sync with the `boot.kernelPackages` set in packages/nixos/kata.nix.
 callPackage ./driver.nix {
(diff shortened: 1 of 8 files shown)


packages/scripts.nix
```diff
@@ -569,6 +569,39 @@
...
+        driverFile="packages/by-name/nvidia-driver/driver.nix"
+        cd "$(git rev-parse --show-toplevel)"
+
+        if [[ $# -gt 0 ]]; then
+          echo "Pinning NVIDIA driver to $1" >&2
+          sed -i -E "s|^([[:space:]]*)version = \"[0-9.]+\";|\1version = \"$1\";|" "$driverFile"
+        fi
+
+        echo "Realising the kernel config the driver is built against" >&2
+        nix eval --raw .#base.nvidia-driver.src.outputHash >/dev/null
```
sespiros: Why is this needed? and also you probably want:

```suggestion
        nix eval --raw .#legacyPackages.x86_64-linux.base.nvidia-driver.src.outputHash >/dev/null
```

packages/by-name/nvidia-driver/package.nix
```diff
@@ -0,0 +1,16 @@
...
+# SPDX-License-Identifier: BUSL-1.1
+
+{
+  callPackage,
+  kata,
+  linuxPackagesFor,
+}:
+
+# The kernel modules have to be built against the very kernel the PodVM boots, so this
+# must stay in sync with the `boot.kernelPackages` set in packages/nixos/kata.nix.
```
sespiros: Could we enforce this by construction somehow?

sespiros (commented):
Thanks, this looks helpful, my only worry is that "keep the kernel derivation in sync"

packages/scripts.nix
charludo: Oops yes, thanks, fixed.

It's not strictly required to do manually. If we remove the line, `nix-update` below realizes the package anyways. The step can just take a while, and `nix-update` seems stuck then.

packages/by-name/nvidia-driver/package.nix
charludo: yeeeeeeesmmaybe... But it would be kinda ugly... We could instead import the driver.nix file directly in nixos/gpu.nix, and then here in the package use our mkNixosConfig and consume just the nvidia package from it.

The current version just gives the package the kernel to build against.

Maybe we could also extract the gpu kernel into a standalone output and consume it both from the package and from the nixos config, as a single source of truth?

charludo: Pushed an alternative. I think I prefer it. LMK and I'll squash.

sespiros (approved):
This seems to work, thanks! 

P.S. I did stare at this kernel-uvm-gpu package for a while to figure out why `@args` is needed but it seems to work :D

PR #2582, opened 2026-08-10
policy-test: init test runner

davidweisse:
This creates the new subpackage `policy-test` which contains the genpolicy test suite. Currently, all it does is:
1. Read the `pod.yml` and run `contrast generate` on it.
2. Extract the policy annotation and convert it to a `Policy` interface (currently rego, i.e., `*OPAPolicy`)
3. Read the agent RPC requests from `tesdata/` and compare them against the policy (`CreateSandboxRequest`, `CreateContainerRequest`, ...)

The current test case is taken from the `/tmp/policy.jsonl` from inside a running container, similar to the one in the `pod.yml`. For now these are just a few simple cases to test, before we can automate getting these and properly normalize them (e.g., replacing image references, patching namespaces, etc.).

The rego policy checker is based on the upstream test runner at https://github.com/kata-containers/kata-containers/tree/main/src/tools/genpolicy/tests/policy. Note that the rego rules don't always return true/false, but can return a metadata response, for example:

```rego
CreateContainerRequest := {"ops": ops, "allowed": true} if { ... }
```

In this case, if the request is allowed, we have to apply the json patches contained in `ops` to the saved state of the policy engine. This is used to store information like sandbox names or ids over multiple requests.

To run the policy test suite, run the `just policy` target. This currently does not use a local registry, so generating the policy one time at the beginning will take a few seconds. The next step is to integrate a local registry into this process, to speed up the policy generation.

Fixes CON-250


policy-test/main.go
```diff
@@ -0,0 +1,162 @@
...
+	var id *initdata.Initdata
+	manipulator := func(i *initdata.Initdata) error {
+		id = i
+		return nil
+	}
+	if err := cmd.ManipulateInitdata(fileMap, manipulator); err != nil {
+		return "", err
+	}
+
+	policy, ok := id.Data["policy.rego"]
```
charludo: Since this keeps the last match, are we certain that the assets will remain single-workload?

policy-test/rego.go
```diff
@@ -0,0 +1,159 @@
...
+		if v {
+			log.Printf("%s: allowed", tc.Kind)
+			return nil
+		}
+		log.Println("Rego prints:")
+		for _, msg := range printHook.messages {
+			log.Println(msg)
+		}
+		return fmt.Errorf("%s: blocked by policy", tc.Kind)
+	case map[string]any:
```
charludo: Was the printing jut for your debugging? If not, we should probably also have it in the `map[string]any` branch.

policy-test/main.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := os.WriteFile(filepath.Join(workDir, "pod.yml"), podYaml, 0o644); err != nil {
+		return err
+	}
+
+	policy, err := generatePolicy(c.Context(), workDir, flags)
+	if err != nil {
+		return fmt.Errorf("generate policy: %w", err)
+	}
+
+	p := NewOPAPolicy(policy)
```
charludo: Should this not go into the loop? Otherwise state  accumulates between testcases

policy-test/rego.go
```diff
@@ -0,0 +1,159 @@
...
+		rego.EnablePrintStatements(true),
+		rego.PrintHook(printHook),
+	)
+
+	res, err := r.Eval(ctx)
+	if err != nil {
+		return err
+	}
+
+	if len(res) != 1 {
```
charludo: Might make sense to add a special case for the `0` case, something along the lines of `<kind>: no rule <query> in policy`. Just not to lose the extra info in that case.

policy-test/main.go
```diff
@@ -0,0 +1,162 @@
...
+		if err != nil {
+			return fmt.Errorf("read test case file %s: %w", file.Name(), err)
+		}
+		var tc []TestCase
+		if err := json.Unmarshal(fileData, &tc); err != nil {
+			return fmt.Errorf("unmarshal test case %s: %w", file.Name(), err)
+		}
+		for _, testCase := range tc {
+			if err := p.AllowRequest(c.Context(), testCase); err != nil {
+				return fmt.Errorf("test case %s failed: %w", file.Name(), err)
```
charludo: We shouldn't return on the first error. Abort the one file, and collect all the errors in `errors.Join()` to return after the loop.

policy-test/policy.go
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: BUSL-1.1
+
+package main
+
+import "context"
+
+type TestCase struct {
+	Kind    string `json:"kind"`
+	Request any    `json:"request"`
+}
```
charludo, line 12: What do you think about already adding an `Allowed bool` field here? Seems currently, we only test that a policy is allowed, but that means e.g. that a policy allowing everything also passes the test. I assume we'll definitely want negative test cases.

That would need post-processing on the json though I guess :/ Or did you have an alternative in mind?

policy-test/rego.go
```diff
@@ -0,0 +1,159 @@
...
+		// The agent logs these responses during policy evaluation, but they are not the actual request. We can ignore them here.
+		// TODO(daidweisse): Should we verify these against the returned ops from the policy evaluation?
+		return nil
+	}
+
+	query := fmt.Sprintf("data.agent_policy.%s", tc.Kind)
+
+	printHook := &collectingPrintHook{}
+
+	r := rego.New(
```
charludo, line 71: nit: recompiles rego per request, maybe can be moved out?

policy-test/rego.go
```diff
@@ -0,0 +1,159 @@
...
+
+// AllowRequest evaluates the given test case against the policy.
+func (p *OPAPolicy) AllowRequest(ctx context.Context, tc TestCase) error {
+	req, ok := tc.Request.(map[string]any)
+	if !ok {
+		return fmt.Errorf("request is not a map[string]any, got: %T", tc.Request)
+	}
+	if _, isMetadata := req["ops"]; isMetadata {
+		// The agent logs these responses during policy evaluation, but they are not the actual request. We can ignore them here.
+		// TODO(daidweisse): Should we verify these against the returned ops from the policy evaluation?
```
charludo:
```suggestion
		// TODO(davidweisse): Should we verify these against the returned ops from the policy evaluation?
```
😁

policy-test/main.go
```diff
@@ -0,0 +1,162 @@
...
+
+func execute(c *cobra.Command, _ []string) error {
+	flags, err := parseFlags(c)
+	if err != nil {
+		return fmt.Errorf("parse flags: %w", err)
+	}
+
+	workDir, err := os.MkdirTemp("", "contrast-policy-test-*")
+	if err != nil {
+		return err
```
charludo: nit: wrap in `fmt.Errorf`, same for the other errors like this on lines 55, 68

charludo (commented):
Thanks, this is great! 🎉 

Apart from my comments below, a couple general thoughts:
- not sure if we want to export `ManipulateInitdata` publicly... Maybe it belongs in an internal package instead?
- I don't think there's a ticket for running this in CI yet, right? Which we can probably do after the local registry PR 
- why use `log`? I think most everywhere uses `slog`? (But on the other hand... w/e)
- please add some devdocs info about how to obtain the testcases. (Can also be a separate PR, esp. if that's not settled yet.)
- My biggest questionmark is currently how negative testcases will be obtained and implemented here

packages/by-name/policy-test/package.nix
```diff
@@ -0,0 +1,56 @@
...
+}:
+
+buildGoModule (_finalAttrs: {
+  pname = "policy-test";
+  version = builtins.readFile ../../../version.txt;
+
+  inherit (contrast.cli)
+    preConfigure
+    postConfigure
+    ;
```
charludo: Ah, regarding the `cd` issue: The following should work:

```suggestion
 preConfigure = ''
    (
      modRoot=. # The cli's preConfigure installs relative to $modRoot.
      ${contrast.cli.preConfigure}
      ${contrast.cli.postConfigure}
    )
  '';
```
by just changing the `modRoot` in that one phase. Though it'd probably be nicer to also split `postConfigure` off and repeat the `modRoot` there.

Then you can remove the `cd` and set `modRoot = "policy-test";` on the package itself.

policy-test/main.go
davidweisse: I think this only makes sense, i.e., one deployment file = one policy. And then have one asset for every resource we want to test.

policy-test/policy.go
davidweisse: Haven't thought too much about that either, but we do have an issue for that. Adding an `Allowed` field is definitely necessary though.

davidweisse: Considering negative test cases, I realized the current logging logic does not really make sense (i.e., logging in case of policy failure). Instead, I rewrote the interface to be closer to upstream, returning a bool and the corresponding Rego prints (`(allowed bool, prints string, err error)`). Replacing Rego with something else should still be no problem: The bool indicates policy evaluation, the `prints` string any policy evaluation logging/error, and the error indicating any internal processing error.

policy-test/rego.go
davidweisse: We shouldn't log this at all here, see other comment. Instead, log this at the caller, where we can actually decide if logging makes sense.

policy-test/rego.go
davidweisse: Yes, this significantly speeds up the rego evaluation. I think I was just too lazy to look up how to do this the first time.

packages/by-name/policy-test/package.nix
davidweisse: The configure phase itself does `cd $modRoot`, so moving both into the `preConfigure` phase as in the suggestion works :+1: (without splitting it up). Although this is still somewhat hacky...

davidweisse:
> * not sure if we want to export `ManipulateInitdata` publicly... Maybe it belongs in an internal package instead?

Hm, I don't like the idea of having this in an internal package either (like `internal/initdata`, if that's what you mean), because the logic is closely tied to the CLI with, e.g., the `fileMap` logic. Since we only need to extract the policy, I just hardcoded a `MapPodSpec` for now. We don't necessarily need all the features of `manipulateInitdata` here.

> * please add some devdocs info about how to obtain the testcases. (Can also be a separate PR, esp. if that's not settled yet.)

There is a ticket for automating this process, i.e., periodically generating testcases, so I don't think it's necessary to document at this point.

policy-test/main.go
```diff
@@ -0,0 +1,186 @@
...
+
+	var errs []error
+	for _, file := range dirs {
+		if file.IsDir() {
+			continue
+		}
+		log.Printf("===== Running test case: %s", file.Name())
+		fileData, err := os.ReadFile(filepath.Join(dataDir, file.Name()))
+		if err != nil {
+			return fmt.Errorf("read test case file %s: %w", file.Name(), err)
```
charludo: same thing with the early return on errors here. Should be collected and skipped.

policy-test/main.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		var tc []TestCase
+		if err := json.Unmarshal(fileData, &tc); err != nil {
+			return fmt.Errorf("unmarshal test case %s: %w", file.Name(), err)
```
charludo: same here

policy-test/main.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		p, err := NewOPAPolicy(policy)
+		if err != nil {
+			return fmt.Errorf("create OPA policy: %w", err)
```
charludo: here too

policy-test/main.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	dataDir := "./policy-test/testdata/"
+	dirs, err := os.ReadDir(dataDir)
+	if err != nil {
+		return fmt.Errorf("read test data dir: %w", err)
+	}
+
+	var errs []error
+	for _, file := range dirs {
```
charludo, line 73: This may be paranoid, but: wdyt about counting testcases and erroring also when no those are 0? right now, if someone e.g. moves the files to a subdir, the test succeeds, right?

policy-test/rego.go
```diff
@@ -0,0 +1,162 @@
...
+		rego.PrintHook(printHook),
+	)
+
+	res, err := r.Eval(ctx)
+	if err != nil {
+		return false, "", err
+	}
+
+	if len(res) == 0 {
+		return false, "", fmt.Errorf("no rule %s found in policy", query)
```
charludo: maybe we should also return prints from this function? SO in the error case, we can debug the error directly?

charludo (approved):
Thanks, great work! Just some leftover re: errors in parsing files etc cancelling all test cases.

policy-test/policy.go
charludo: Good call, nice

policy-test/rego.go
davidweisse: In this particular case (i.e., len(res) = 0), there are no prints afaik. But in general I agree to add the prints to this block when checking the result length, and only if it is to catch regressions.

charludo (approved):
Thanks for the fixups!

PR #2583, opened 2026-08-10
policy-test: use local registry

davidweisse:
This uses the `--insecure-registry` flag from #2580 to configure a local registry via `crane registry serve` during the policy test suite evaluation. To use this, first run `just policy-registry` and wait until it says `Registry is ready.`. Then run `just policy`. The first target will start a local registry via `crane registry serve` and push the needed images into it while also writing to the image replacements file.

One problem is: genpolicy also pulls the configured pause image (ghcr.io/edgelesssys/kuberenetes/pause:3.6 in our case), which cannot be replaced by the image replacements file, since it isn't present in any of the deployment YAMLs. We configure `ghcr.io` as an insecure registry and set up an HTTP proxy that reroutes requests to `ghcr.io` to the local registry. For this to work, we of course push the pause image to the local registry beforehand.

Doing some testing, I get the following results:
- The first image pull will always take longer, as it has to pull the layers. About 2-3s to pull busybox from a local registry.
- Subsequent pulls only need the manifest/config, this takes ~120ms.
- Policy evaluation takes ~30-50ms for each RPC request.

Fixes CON-251

This PR is currently rebased on both #2580 and #2582, so only the most recent commit is relevant.


davidweisse:
Update: Instead of the weird proxy setup, we can dynamically overwrite the pause image in the genpolicy settings file before running `contrast generate`. We can then explicitly pass that file to generate via a command line flag.

Also, with the rebase, the rego evaluation timing goes down to <10ms (~10-20ms max for `CreateContainerRequest` and <1ms for for others)

policy-test/main.go
```diff
@@ -57,6 +61,14 @@ func execute(c *cobra.Command, _ []string) error {
 		return fmt.Errorf("write pod.yml: %w", err)
 	}
 
+	// Patch the pause image in genpolicy-settings.json to use the insecure registry if specified.
+	if flags.insecureRegistry != "" {
+		genpolicySettings = bytes.ReplaceAll(genpolicySettings, []byte("ghcr.io/edgelesssys/kubernetes/pause"), []byte(flags.insecureRegistry+"/kubernetes/pause"))
```
charludo: This fails silently if no match is found. Maybe add a check to ensure the new pause image is written to the settings.

justfile
```diff
@@ -2,19 +2,20 @@ default_cli := "contrast.cli"
 default_deploy_target := "openssl"
 default_platform := "${default_platform}"
 default_set := "${set}"
+default_registry := "${container_registry}"
```
charludo: Is there a reason to add the new `default_registry` here, or could we just do
```just
push target set=default_set registry=container_registry:
```
below?

policy-test/main.go
```diff
@@ -121,8 +133,10 @@ func generatePolicy(ctx context.Context, workDir string, flags *flags) (string,
 	args := []string{
 		"--workspace-dir=" + workDir,
 		"--reference-values=Metal-QEMU-SNP",
+		"--settings=" + filepath.Join(workDir, "genpolicy-settings.json"),
 		"--genpolicy-cache-path=" + filepath.Join(workDir, "layers-cache.json"),
 		"--image-replacements=" + flags.imageReplacementsFile,
+		"--insecure-registry=" + flags.insecureRegistry,
```
charludo: I think cobra's `StringArray` turns the empty default into `[""]`, so something is always passed here. Maybe 
```go
if flags.insecureRegistry != "" {
    args = append(args, "--insecure-registry="+flags.insecureRegistry)
}
```
or similar?

justfile
```diff
@@ -554,12 +555,38 @@ lint:
...
+    just push "initializer" "{{ default_set }}" "localhost:5000"
+    busybox="busybox@sha256:dc2d74b28e4cf8984fa52af1f39bc7c3d9c73760b41a74d629f5d11b1ab28616"
+    crane copy $busybox localhost:5000/$busybox
+    printf "$busybox=localhost:5000/$busybox\n" >> ./{{ workspace_dir }}/just.containerlookup
+    crane copy ghcr.io/edgelesssys/kubernetes/pause:3.6 localhost:5000/kubernetes/pause:3.6
+    echo "Registry is ready."
+    tail -n 0 --pid $PID -f ./{{ workspace_dir }}/registry.log
+
 # Run the policy test suite.
-policy: initializer
```
charludo, line 574: Shouldn't we still push the initializer here? If it didn't change, it's close to a no-op, but pushing it only in `policy-registry` means needing to restart the registry when we made a change and want to re-test.

justfile
```diff
@@ -554,12 +555,38 @@ lint:
...
+policy-registry:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if curl -sf localhost:5000/v2/ > /dev/null; then
+        echo "Registry already running on port 5000."
+        exit 1
+    fi
+    mkdir -p ./{{ workspace_dir }}
+    echo "Setting up registry..."
+    crane registry serve --address localhost:5000 > ./{{ workspace_dir }}/registry.log 2>&1 &
```
charludo: Probably need to wait (shortly) for readiness

justfile
```diff
@@ -554,12 +555,38 @@ lint:
 unit:
     CGO_ENABLED=1 go test -tags=contrast_unstable_api -v -race ./...
 
+# Run a local registry needed for the policy test suite.
+policy-registry:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if curl -sf localhost:5000/v2/ > /dev/null; then
+        echo "Registry already running on port 5000."
+        exit 1
```
charludo: This should be `exit 0`, we're basically saying "nothing to do, already up!".
Alternatively, exit 1 and "port 5000 in use, can't start".

justfile
charludo: Lots of places with the `5000` now, how about `policy_registry := "localhost:5000"` at the top of the file?

justfile
```diff
@@ -554,12 +555,38 @@ lint:
...
+    crane registry serve --address localhost:5000 > ./{{ workspace_dir }}/registry.log 2>&1 &
+    PID=$!
+    trap 'kill $PID 2>/dev/null; wait $PID 2>/dev/null || true' EXIT
+    just push "initializer" "{{ default_set }}" "localhost:5000"
+    busybox="busybox@sha256:dc2d74b28e4cf8984fa52af1f39bc7c3d9c73760b41a74d629f5d11b1ab28616"
+    crane copy $busybox localhost:5000/$busybox
+    printf "$busybox=localhost:5000/$busybox\n" >> ./{{ workspace_dir }}/just.containerlookup
+    crane copy ghcr.io/edgelesssys/kubernetes/pause:3.6 localhost:5000/kubernetes/pause:3.6
+    echo "Registry is ready."
+    tail -n 0 --pid $PID -f ./{{ workspace_dir }}/registry.log
```
charludo: I *think* this won't work on darwin...? I have no way to check though.

packages/by-name/policy-test/package.nix
```diff
@@ -35,13 +36,9 @@ buildGoModule (_finalAttrs: {
   modRoot = "policy-test";
 
   preConfigure = ''
-    # The preConfigure hook works on modRoot, but we want to run it in the repository root.
-    actualModRoot=$modRoot
-    modRoot=.
-    ${contrast.cli.preConfigure}
+    install -D ${kata.genpolicy.settings-dev}/genpolicy-settings.json policy-test/assets/genpolicy-settings-kata.json
```
charludo, line 100: Interestingly, this now leaves the CLI's asset at the `THIS FILE IS REPLACED...` state. It currently does not matter, because you always pass `--settings=` in `policy-test/main.go`, but maybe worth a comment here.

charludo (commented):
Thanks! I have a bunch of small comments, but generally, this LGTM. Tested it out locally and seems to work.

- One thing I didn't catch on the last PR - maybe the `policy-test` should be moved from top-level into `tools/`...? Tools also doesn't feel quite right though. IDK. 
- I still don't really like having to run two terminals/just commands. `crane` uses https://github.com/google/go-containerregistry, so why not do this ourselves? Super rough sketch:
    ```go
    import (
      "net"
      "net/http"
      "github.com/google/go-containerregistry/pkg/registry"
      "github.com/google/go-containerregistry/pkg/crane"
      "github.com/google/go-containerregistry/pkg/v1/layout"
  )

  ln, err := net.Listen("tcp", "127.0.0.1:0")
  if err != nil { return fmt.Errorf("listen: %w", err) }
  srv := &http.Server{Handler: registry.New(registry.Logger(log.New(io.Discard, "", 0)))}
  go srv.Serve(ln)
  defer srv.Close()
  addr := ln.Addr().String()

  for _, img := range []string{busyboxRef, pauseRef} {
      if err := crane.Copy(img, addr+"/"+strings.SplitN(img, "/", 2)[1], crane.Insecure); err != nil {...}
  }

  idx, err := layout.ImageIndexFromPath(initializerOCIDir)
  crane.PushIndex(idx, addr+"/contrast/initializer:latest", crane.Insecure)
     ```
    where `initializerOCIDir` comes from nix, e.g. through another `install -D` in `assets/`. We could even pull busybox and pause images in the nix build and persist them, so no re-download between runs. The only not-so-nice thing about this would be to have to either write the image-replacements from go, or (probably better), allow `generatePolicies` to optionally take a map of image replacements directly.

The second point should probably be decided on before implementing the review comments (@burgerdev), since going the `go-containerregistry` route would make a lot of them irrelevant.

justfile
burgerdev: The short delay until crane starts serving has been a recurring pain point in Constellation's Bazel builds. If possible, I'd really prefer to run the registry from Go code, as @charludo suggested.

justfile
burgerdev: Correct, this would need to use `gtail` via Nix.

davidweisse:
Okay, next try: I decided to embed the images directly into the nix build. Though I'm not sure how to prevent manually writing the image replacements file. We can then push the images to a local registry using `go-containerregistry` from the go code.

Unfortunately, extracting the images using go and pushing them again apparently changes the image digest, so I had to adjust the test case. Hopefully we can avoid issues there when we have a way to fetch and normalize the test cases, and then dynamically replace the needed images. That way we don't have to rely on having the same hashes in the hardcoded test cases and during the test suite execution.

packages/by-name/policy-test/package.nix
```diff
@@ -5,8 +5,47 @@
...
+  writeText,
 }:
 
+let
+  busybox = dockerTools.pullImage {
+    imageName = "busybox";
+    imageDigest = "sha256:dc2d74b28e4cf8984fa52af1f39bc7c3d9c73760b41a74d629f5d11b1ab28616";
+    hash = "sha256-kouAB8WzSQpp60F/C83gXkWFKs/jl+Iq09YX8EiJezk=";
+  };
+  pause = dockerTools.pullImage {
```
burgerdev: If you `skopeo copy --all` the remote images to an OCI layout in a fixed-output derivation you might be able to keep the digests intact.

davidweisse: Yep, that works. Also noticed an error in the digest calculation of OCI images. The digest in the image URI should be the digest of the manifest, not the digest of the index. Not sure why this worked for the initializer before, but I got an error when trying to pull the busybox image with `busybox@sha256:<index-digest>`.

packages/by-name/policy-test/package.nix
```diff
@@ -5,8 +5,67 @@
...
+    }:
+    runCommand "pull-oci-image-${imageName}"
+      {
+        outputHash = hash;
+        outputHashAlgo = "sha256";
+        outputHashMode = "recursive";
+        nativeBuildInputs = [ skopeo ];
+      }
+      ''
+        skopeo copy \
```
charludo, line 31: `--preserve-digests` seems like something we might want here as well. Then skopeo should fail at build time if it can't, for whatever reason, do that.

packages/by-name/policy-test/package.nix
```diff
@@ -30,25 +89,27 @@ buildGoModule (_finalAttrs: {
...
   '';
 
   env.CGO_ENABLED = 0;
   dontFixup = true;
 
   ldflags = [
     "-s"
+    "-X main.BusyboxImage=${busybox}"
+    "-X main.PauseImage=${pause}"
+    "-X main.InitializerImage=${initializer}"
```
charludo: This are unused now

charludo (approved):
Thank you, this is great! I have two small comments, everything else LGTM 🎉

packages/by-name/policy-test/package.nix
davidweisse: Hmm, does not work. I think `--preserve-digests` only works when the source and destination formats are the same. Turns out, busybox is saved as an OCI image in the docker registry (which is why the digest is preserved when we copy it as an OCI image) but the pause image is a docker image (so the digest cannot be preserved if we copy it to an OCI image).

Not sure what to do about that. Either we leave it how it is, because the pause image is not actually there in any RPC requests (I think?), or we have to differentiate based on the image (and then copy to `docker-archive` instead).

charludo: Ah, too bad. Yeah then let's just leave it for now.

PR #2584, opened 2026-08-12
ci: update bm-tcb-specs for dgx-007

thomasten:
I deployed the latest module from https://github.com/intel/confidential-computing.tdx.tdx-module on this machine.


charludo (approved):
Thanks!

PR #2585, opened 2026-08-13
certcache: on transient proxy error, fall back to upstream

charludo:
The collateral proxy caches whatever the vendor returned, including errors, for up to an hour.
A single 429 or 5XX from AMD KDS / Intel PCS therefore keeps being replayed to every client long after the vendor recovered.

Not the fix for the current failure, but a safety net for the next time the cache is poisoned this way.

### DIFF
--- a/internal/attestation/certcache/cached_client.go
+++ b/internal/attestation/certcache/cached_client.go
@@ -91,7 +91,7 @@ func (c *CachedHTTPSGetter) fetch(ctx context.Context, url string) (map[string][
 		return c.ContextHTTPSGetter.GetContext(ctx, url)
 	}
 
-	proxyCtx, cancel := context.WithTimeout(ctx, retryAttemptsProxy*retryInterval)
+	proxyCtx, cancel := context.WithTimeout(ctx, c.proxyBudget(ctx))
 	header, body, err := c.ContextHTTPSGetter.GetContext(proxyCtx, c.redirectToProxy(url))
 	cancel()
 	if err == nil {
@@ -100,12 +100,30 @@ func (c *CachedHTTPSGetter) fetch(ctx context.Context, url string) (map[string][
 		return header, body, nil
 	}
 	var httpErr *httpError
-	if errors.As(err, &httpErr) {
+	if errors.As(err, &httpErr) && !transientStatus(httpErr.code) {
 		return nil, nil, err
 	}
 	c.proxyRetryAfter.Store(c.clock.Now().Add(proxyRetryCooldown).UnixNano())
-	c.logger.Warn("collateral proxy not reachable, falling back to direct upstream fetching", "url", url, "error", err, "cooldown", proxyRetryCooldown)
+	c.logger.Warn("collateral proxy unhealthy, falling back to direct upstream fetching", "url", url, "error", err, "cooldown", proxyRetryCooldown)
 	return c.ContextHTTPSGetter.GetContext(ctx, url)
+}
+
+// proxyBudget is the time to spend on the proxy attempt, capped to half of what the caller granted us.
+func (c *CachedHTTPSGetter) proxyBudget(ctx context.Context) time.Duration {
+	budget := time.Duration(retryAttemptsProxy) * retryInterval
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		return budget
+	}
+	return min(budget, deadline.Sub(c.clock.Now())/2)
+}
+
+// transientStatus reports whether an HTTP status from the proxy indicates a temporary
+// condition on the proxy path, rather than an answer the vendor endpoint would repeat.
+//
+// Deliberately wider then [shouldRetry], which does not retry 429, but still needs to be served from cache.
+func transientStatus(code int) bool {
+	return code >= 500 || code == http.StatusTooManyRequests
 }
 
 func (c *CachedHTTPSGetter) proxyInCooldown() bool {
--- a/internal/attestation/certcache/cached_client_test.go
+++ b/internal/attestation/certcache/cached_client_test.go
@@ -7,9 +7,13 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"log/slog"
+	"net/http"
+	"net/http/httptest"
 	neturl "net/url"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -28,7 +32,6 @@ func TestMain(m *testing.M) {
 }
 
 func TestAlwaysRevalidate(t *testing.T) {
-	const proxy = "http://collateral-proxy.default.svc"
 	for _, tc := range []struct {
 		name string
 		url  string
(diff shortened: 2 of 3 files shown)


internal/attestation/certcache/cached_client.go
```diff
@@ -100,14 +100,20 @@ func (c *CachedHTTPSGetter) fetch(ctx context.Context, url string) (map[string][
 		return header, body, nil
 	}
 	var httpErr *httpError
-	if errors.As(err, &httpErr) {
+	if errors.As(err, &httpErr) && !transientStatus(httpErr.code) {
```
sespiros, line 103: When the proxy answers 5XX, the retries eat the caller's whole time budget, so line 108 is reached with nothing left and the vendor is never contacted.

In the coordinator's issue path:

```
t=0s   collateral fetching gets 20s total
       (ATLSIssuerOptionalEndorsementFetchTimeout = 30 - 5 - 5)
t=0s   proxy answers 500; shouldRetry says 5XX is retriable, so we retry.
       The proxy attempt is capped at 25s (retryAttemptsProxy * retryInterval).
t=20s  the caller's 20s runs out first -> retrying stops, error returns.
t=20s  line 108 falls back to the vendor with an already-expired context
       -> returns instantly, AMD is never contacted.
```

The proxy may spend 25s, the caller only has 20s, so the retries win and the fallback loses.
Confirmed with the real `RetryHTTPSGetter` wiring (`upstreamHits=0` means the vendor was never asked):

```
proxy 500: proxyHits=5 upstreamHits=0  err=retries failed before context deadline exceeded
proxy 429: proxyHits=1 upstreamHits=1  err=<nil>
```

429 works because `shouldRetry` won't retry it, so it returns immediately with ~20s still on the clock.
5XX also works for the CLI and validator, which call `Get()` with no deadline at all.

Not a disaster: the cooldown is still set, so the next fetch within 2 minutes skips the proxy and goes direct.
But one attestation attempt per cooldown window burns its full 20s and fails, and on that path this line does nothing.

Fix is to stop letting the proxy attempt consume the whole budget. On line 94:

```go
proxyCtx, cancel := context.WithTimeout(ctx, c.proxyBudget(ctx))
```

```go
// proxyBudget is the time to spend on the proxy attempt: the full retry budget, but never
// more than half of what the caller granted us, so the fallback below is still affordable.
func (c *CachedHTTPSGetter) proxyBudget(ctx context.Context) time.Duration {
	budget := time.Duration(retryAttemptsProxy) * retryInterval
	deadline, ok := ctx.Deadline()
	if !ok {
		return budget
	}
	return min(budget, deadline.Sub(c.clock.Now())/2)
}
```

Simpler alternative: give the proxy a single attempt (`retryInterval`), since the proxy hop is meant to be a fast cache read.

internal/attestation/certcache/cached_client.go
```diff
@@ -100,14 +100,20 @@ func (c *CachedHTTPSGetter) fetch(ctx context.Context, url string) (map[string][
...
 	c.proxyRetryAfter.Store(c.clock.Now().Add(proxyRetryCooldown).UnixNano())
-	c.logger.Warn("collateral proxy not reachable, falling back to direct upstream fetching", "url", url, "error", err, "cooldown", proxyRetryCooldown)
+	c.logger.Warn("collateral proxy unhealthy, falling back to direct upstream fetching", "url", url, "error", err, "cooldown", proxyRetryCooldown)
 	return c.ContextHTTPSGetter.GetContext(ctx, url)
 }
 
+// transientStatus reports whether an HTTP status from the proxy indicates a temporary
+// condition on the proxy path, rather than an answer the vendor endpoint would repeat.
+func transientStatus(code int) bool {
+	return code >= 500 || code == http.StatusTooManyRequests
```
sespiros, line 126: This says 429 is transient; `shouldRetry` (`retry_client.go:109`) says it isn't worth retrying.
That divergence is correct and should stay, so say so here rather than leaving the next reader to unify them.
Unifying would make 429 eat the caller's budget too, breaking the one case that currently works.

```go
// Deliberately wider than shouldRetry in retry_client.go, which doesn't retry 429: a
// rate-limited proxy should be bypassed, not hammered.
func transientStatus(code int) bool {
```

internal/attestation/certcache/cached_client.go
```diff
@@ -100,14 +100,20 @@ func (c *CachedHTTPSGetter) fetch(ctx context.Context, url string) (map[string][
...
 		return nil, nil, err
 	}
 	c.proxyRetryAfter.Store(c.clock.Now().Add(proxyRetryCooldown).UnixNano())
-	c.logger.Warn("collateral proxy not reachable, falling back to direct upstream fetching", "url", url, "error", err, "cooldown", proxyRetryCooldown)
+	c.logger.Warn("collateral proxy unhealthy, falling back to direct upstream fetching", "url", url, "error", err, "cooldown", proxyRetryCooldown)
 	return c.ContextHTTPSGetter.GetContext(ctx, url)
 }
 
+// transientStatus reports whether an HTTP status from the proxy indicates a temporary
+// condition on the proxy path, rather than an answer the vendor endpoint would repeat.
```
sespiros, line 122: The current wording ("an answer the vendor endpoint would repeat") doesn't hold: the proxy relays vendor statuses verbatim and caches any status up to its 1h default TTL.
That stale-error case is the strongest argument for this change, worth naming.

````suggestion
// transientStatus reports whether an HTTP status from the proxy warrants bypassing it.
// It's either the proxy's own error (502 when it can't reach the vendor and has nothing
// cached), or a vendor error the proxy is replaying from its cache for up to an hour.
````

File: internal/attestation/certcache/cached_client_test.go [hunk 4]
```diff
@@ -312,25 +313,46 @@ func TestProxyFallback(t *testing.T) {
...
+		client, _ := newHostGetterClient(getter)
 
 		_, _, err := client.Get(directCRL)
 		assert.Error(err)
 		assert.Equal(1, getter.hits[proxyHost])
 		assert.Equal(0, getter.hits[kdsHost]) // upstream not contacted
 		assert.False(client.proxyInCooldown())
 	})
+
+	for _, code := range []int{http.StatusInternalServerError, http.StatusBadGateway, http.StatusTooManyRequests} {
```
sespiros, line 325: These wire `fakeHostGetter` directly, so the `*httpError` arrives unwrapped.
In production it comes through `IntervalRetrier.Do` as `retries failed before %w: last err: %w`.
That still unwraps (I confirmed `errors.As` finds 500/502/429 through the double-`%w`), but nothing pins it: drop the `%w` on `lastErr` and the 5XX fallback dies silently with every test green.

Add one case through the retrier:

```go
t.Run("5XX survives the retrier's error wrapping", func(t *testing.T) {
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		w.WriteHeader(http.StatusInternalServerError)
	}))
	defer srv.Close()

	g := NewRetryHTTPSGetter(srv.Client(), 50*time.Millisecond, slog.New(slog.DiscardHandler))
	ctx, cancel := context.WithTimeout(t.Context(), 250*time.Millisecond)
	defer cancel()
	_, _, err := g.GetContext(ctx, srv.URL)

	var httpErr *httpError
	require.ErrorAs(t, err, &httpErr)
	assert.True(t, transientStatus(httpErr.code))
})
```

Also worth a comment on line 338: `hits[proxyHost] == 1` only holds because the retrier is bypassed, the real path hits the proxy 5 times.

internal/attestation/certcache/cached_client_test.go
```diff
@@ -312,25 +313,46 @@ func TestProxyFallback(t *testing.T) {
...
+			assert.Equal([]byte("crl-bytes"), body)
+			assert.Equal(1, getter.hits[proxyHost])
+			assert.Equal(1, getter.hits[kdsHost])
+			assert.True(client.proxyInCooldown())
+		})
+	}
 }
 
-func newHostGetterClient(getter *fakeHostGetter, collateralProxyBase string) (*CachedHTTPSGetter, *testingclock.FakeClock) {
+const proxyBase = "http://collateral-proxy.default.svc"
```
sespiros, line 392: Now that this is package-level, delete `const proxy = "http://collateral-proxy.default.svc"` from `TestAlwaysRevalidate` (line 33) and `TestRedirectToProxy` (line 57) and use `proxyBase` in both.
Same string in three places otherwise.

sespiros (approved):
Thanks, approving but leaving all the comments from AI review as suggested improvements.

`docs/docs/howto/collateral-proxy.md:55` is not in the diff and needs a separate edit:

> The proxy is a soft dependency: if a component can't reach it, **or it answers with a transient error (5XX, 429),** the component logs a warning and falls back to fetching directly from the vendor endpoint, then retries the proxy after a short cooldown.

The PR description is a half-finished sentence.
`no changelog` keeps it out of release notes, but it's the record of why this exists. Suggested replacement:

> The collateral proxy caches whatever the vendor returned, including errors, for up to an hour.
> A single 429 or 5XX from AMD KDS / Intel PCS therefore keeps being replayed to every client long after the vendor recovered.
> Treat those statuses like an unreachable proxy: fall back to the vendor directly and put the proxy in cooldown.
> Not the fix for `<link to the failing run>`, but a safety net for the next time the cache is poisoned this way.

internal/attestation/certcache/cached_client_test.go
charludo: thanks, added

internal/attestation/certcache/cached_client.go
charludo: Thanks, this is an important catch. Implemented the suggested fix to let the proxy only have half the total budget.

charludo:
Thanks for the review @sespiros! Should be good to go again.

sespiros (approved):
Thanks.

PR #2589, opened 2026-08-14
ci: update olimar PIID after firmware update

sespiros:
Updating olimar's BIOS regenerated the platform's SGX provisioning keys, which changed its PPID, QE ID and Platform Instance ID. The platform had to be registered with Intel again before it could produce quotes at all, and the new PIID no longer matches what we pin.

Extracted from the post-update pckid_retrieval.csv as described in dev-docs/e2e/bare-metal-runner.md.


charludo (approved):
Oh wow. Thanks for tackling this!!

PR #2592, opened 2026-08-18
apitypes: depend on the standard library only

charludo:


sespiros (approved):
Thanks. Note #2568 will need a rebase and separately I am thinking maybe it's actually better if we merged individual PRs in a stack instead of waiting for the whole stack to be approved because they might get stale.

One more thing to think about is that the encoding that produces `transitionDigest` is still an internal thing, so an internal refactor there would silently break deployed clients that build on `apitypes`. This is a general comment on the API design. So it's not only about the imports but also about the "contract" `apitypes` is supposed to capture.

charludo:
> I am thinking maybe it's actually better if we merged individual PRs in a stack instead of waiting for the whole stack to be approved because they might get stale.

Hm, maybe. TBF the stacks a re surprisingly easy to rebase. But if you'd prefer we get done with approved PRs ASAP (which I also get as an argument), we can also do that.

> One more thing to think about is that the encoding that produces `transitionDigest` is still an internal thing, so an internal refactor there would silently break deployed clients that build on `apitypes`. This is a general comment on the API design. So it's not only about the imports but also about the "contract" `apitypes` is supposed to capture.

Yeah. Probably should have a test in the `apitypes` package with a comment and warning that these tests must not be altered within a version of the API. So we get failing tests if we change something that's incompatible.

PR #2593, opened 2026-08-20
ci: update bm-tcb-specs for dgx-007

msanft:
Extracted the CI-critical fix from #2566 as discussed, also addressing @sespiros comment.


charludo (approved):
❤️ 

Awesome, thank you so much. Let's wait for the tests and then 🚀

PR #2594, opened 2026-08-20
.github: hot-fixes for patch release

charludo:
First commit: would have failed the prepare post release PR step. Not blocking the release itself, but the post-PR.
Second commit: temporarily disable TDX-GPU tests

PR #2596, opened 2026-08-20
.github: make the post-release action kind-aware

sespiros:
Follow-up to #2594.

That PR added a third boolean to `prepare_post_release_pr`, alongside `create_docs_release` and `bump_version`, all three wired to the same `inputs.kind == 'minor'` at the call site.
The action was asking its caller three questions that have only ever had one answer between them, so it takes `kind` once and decides internally.
Across roughly 30 post-release PRs since 1.5.2 those booleans have only ever been all-on or all-off (the 4-commit and 2-commit PRs), never mixed.
No behaviour change for minors, and `release_promote.yml` still relies on the default.

The second commit restores the guard the hot-fix had to drop.
Skipping the version check on a patch left `checkout_main_ref` unvalidated on that path, which is the mistake the check exists to catch.
Main can never carry a patch version but is always ahead of one, so a patch asserts that ordering instead of an exact match.
By construction this is weaker than the minor check: it catches a ref that is behind the release, not one that is ahead and still wrong.

Should land after #2594.
Main-only, no backport needed.

### DIFF
--- a/.github/actions/prepare_post_release_pr/action.yml
+++ b/.github/actions/prepare_post_release_pr/action.yml
@@ -26,7 +26,7 @@ inputs:
     required: true
   kind:
     description: >
-      Release kind, minor or patch. A patch skips the versioned docs snapshot, the post-release version bump, and the check that checkout_main_ref carries the released version.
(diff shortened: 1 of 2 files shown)


charludo (approved):
Yes!! Thank you.

PR #2597, opened 2026-08-20
ci: run bare metal maintenance on the right nodes

sespiros:
Two nodes have carried `ci.contrast.edgeless.systems/main-runner=true` since a second bare-metal machine joined the TDX-GPU cluster on 2026-08-14, so the maintenance jobs pick one at random.

`nix-gc` collects the runner's nix store and dies on the other machine (timed out 08-18 and twice on 08-20, blocking v1.23.1, hence b0753f7d07).
It now resolves the runner's own node by matching `/etc/machine-id` against the node's `machineID`, so no labeling is needed.
`cleanup` and `cleanup-containerd` pass but only clean the node they land on, so they now run once per `main-runner` node, and a failure on one no longer skips the rest.

Revert b0753f7d07 once the `bare metal maintenance` run this merge triggers is green on TDX-GPU.
Why nix-gc dies on that node is still unknown, but the last commit dumps job and pod state on failure, so the next occurrence will say.

### DIFF
--- a/.github/workflows/bm_maintenance_platform.yml
+++ b/.github/workflows/bm_maintenance_platform.yml
@@ -103,7 +103,13 @@ jobs:
             kubectl delete job -n maintenance-cleanup cleanup-maintenance --ignore-not-found
             sed -e "s#@@REPLACE_IMAGE@@#${IMAGE}#g" -e "s#@@REPLACE_K3S_VERSION@@#${K3S_VERSION}#g" -e "s#@@REPLACE_NODE@@#${node}#g" \
               ./tools/bm-maintenance/cleanup.yml | kubectl apply -f -
-            kubectl wait -n maintenance-cleanup --for=condition=complete --timeout=600s job/cleanup-maintenance || failed=1
+            if ! kubectl wait -n maintenance-cleanup --for=condition=complete --timeout=600s job/cleanup-maintenance; then
+              failed=1
+              echo "::error::cleanup failed on ${node}"
+              kubectl get pods -n maintenance-cleanup -o wide || true
+              kubectl describe job -n maintenance-cleanup cleanup-maintenance || true
+              kubectl events -n maintenance-cleanup || true
+            fi
             kubectl logs -n maintenance-cleanup job/cleanup-maintenance || true
           done
           [[ ${failed} -eq 0 ]]
@@ -142,7 +148,13 @@ jobs:
             kubectl delete job -n maintenance-containerd-cleanup containerd-cleanup-maintenance --ignore-not-found
             sed -e "s#@@REPLACE_IMAGE@@#${IMAGE}#g" -e "s#@@REPLACE_NODE@@#${node}#g" \
               ./tools/bm-maintenance/cleanup-containerd.yml | kubectl apply -f -
-            kubectl wait -n maintenance-containerd-cleanup --for=condition=complete --timeout=600s job/containerd-cleanup-maintenance || failed=1
+            if ! kubectl wait -n maintenance-containerd-cleanup --for=condition=complete --timeout=600s job/containerd-cleanup-maintenance; then
+              failed=1
+              echo "::error::containerd cleanup failed on ${node}"
+              kubectl get pods -n maintenance-containerd-cleanup -o wide || true
+              kubectl describe job -n maintenance-containerd-cleanup containerd-cleanup-maintenance || true
+              kubectl events -n maintenance-containerd-cleanup || true
+            fi
             kubectl logs -n maintenance-containerd-cleanup job/containerd-cleanup-maintenance || true
           done
           [[ ${failed} -eq 0 ]]
@@ -184,6 +196,12 @@ jobs:
         run: |
           kubectl wait -n maintenance-nix-gc --for=condition=complete --timeout=600s job/nix-garbage-collection
           echo "result=success" >> "$GITHUB_OUTPUT"
+      - name: Collect diagnostics
+        if: failure()
+        run: |
+          kubectl get pods -n maintenance-nix-gc -o wide || true
+          kubectl describe jobs -n maintenance-nix-gc || true
+          kubectl events -n maintenance-nix-gc || true
       - name: Collect logs and cleanup
         if: always()
         run: |
--- a/.github/workflows/bm_maintenance_platform.yml
+++ b/.github/workflows/bm_maintenance_platform.yml
@@ -77,7 +77,7 @@ jobs:
   cleanup:
     name: "Cleanup ${{ inputs.platform-name }}"
     runs-on: ${{ inputs.runner }}
-    timeout-minutes: 15
+    timeout-minutes: 25
     outputs:
       result: ${{ steps.update.outputs.result }}
     permissions:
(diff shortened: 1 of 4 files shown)


.github/workflows/bm_maintenance_platform.yml
```diff
@@ -118,16 +124,23 @@ jobs:
...
-      - name: Wait for cleanup job
-        id: update
-        run: |
-          kubectl wait -n maintenance-containerd-cleanup --for=condition=complete --timeout=600s job/containerd-cleanup-maintenance
+          nodes=$(kubectl get nodes -l ci.contrast.edgeless.systems/main-runner=true --field-selector spec.unschedulable=false -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')
+          if [[ -z "${nodes}" ]]; then
+            echo "::error::no node labeled ci.contrast.edgeless.systems/main-runner=true"
+            exit 1
+          fi
+          while IFS= read -r node; do
```
charludo: (sorry, sanity check: we expect `nodes` to always have `len==1`, right? otherwise I'd comment that a failure on one node means we never try the other nodes)

.github/workflows/bm_maintenance_platform.yml
```diff
@@ -86,18 +86,24 @@ jobs:
...
+            kubectl delete job -n maintenance-cleanup cleanup-maintenance --ignore-not-found
+            sed -e "s#@@REPLACE_IMAGE@@#${IMAGE}#g" -e "s#@@REPLACE_K3S_VERSION@@#${K3S_VERSION}#g" -e "s#@@REPLACE_NODE@@#${node}#g" \
+              ./tools/bm-maintenance/cleanup.yml | kubectl apply -f -
+            kubectl wait -n maintenance-cleanup --for=condition=complete --timeout=600s job/cleanup-maintenance
```
charludo: I wonder if we can race a second kubectl wait with `--for-condition=failed` on all of these and have them cancel the attempt long before the 600s timeout in case of an ctual error

dev-docs/e2e/bare-metal-runner.md
```diff
@@ -78,6 +78,14 @@ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin
 
 These steps depend on the Kubernetes distribution used for this runner.
 
+Besides `ci.contrast.edgeless.systems/main-runner=true`, which several bare-metal nodes of a cluster may carry, the runner's own node needs a label with the runner name registered with GitHub (`dgx-007`, `olimar`, ...).
+The nix garbage collection job selects on it, because only the machine hosting the runner has the nix store it collects.
+Once the node has joined the cluster, apply the label with:
+
+```bash
+kubectl label node <node> ci.contrast.edgeless.systems/runner-host=<runner name>
```
charludo: Can we just use `kubernetes.io/hostname=<runner name>` and not have to worry about labeling the hosts? Or is there a downside

charludo (approved):
Thanks! In general, LGTM. I'm not overly familiar with the available labeling strategies (both technically and best-practice) though, maybe Markus should also take a look at this?

.github/workflows/bm_maintenance_platform.yml
sespiros: The cleanup-containerd job still selects nodes based on the main-runner label being true which currently is true for 2 nodes on TDX-GPU. And yes a failure one node means we never try the other nodes so I pushed a change.

.github/workflows/bm_maintenance_platform.yml
sespiros: Good idea, but it wouldn't fire as-is: these Jobs are restartPolicy: OnFailure with the default backoffLimit: 6, so Failed only appears once container restarts reach 6, which with kubelet's 10/20/40/80/160/300s backoff is ~610s, just past the 600s timeout. It'd need backoffLimit: 2 as well. Maybe a follow-up? The most annoying thing with this actually was not that but rather that the logs were cryptic, just a "timed out waiting on condition". Pushed another commit.

dev-docs/e2e/bare-metal-runner.md
sespiros: Good call, ended up doing it. So the Job now selects on kubernetes.io/hostname. The catch is the value, since the GitHub runner name doesn't exist anywhere in the k8s API, dgx-007's node is scw-…, palutena's is scw-…, so `hostname=$RUNNER_NAME` would match on the three k3s boxes and silently match nothing on exactly the two clusters that have the problem.

So the workflow first resolves this machine to its node by matching `/etc/machine-id` against the node's `machineID`, then substitutes that name. No labeling at all any more, so the runner-host label and the manual kubectl label step are gone from the PR. It requires exactly one match, because two nodes of a cloned Scaleway pool image do share a machine ID.

.github/workflows/bm_maintenance_platform.yml
charludo: Ah, gotcha.

.github/workflows/bm_maintenance_platform.yml
charludo: Ah, too bad. Thanks for digging that up, and the follow up LGTM.

.github/workflows/bm_maintenance_platform.yml
```diff
@@ -86,19 +86,33 @@ jobs:
...
-          sed -i "s#@@REPLACE_IMAGE@@#${IMAGE}#g" ./tools/bm-maintenance/cleanup.yml
-          sed -i "s#@@REPLACE_K3S_VERSION@@#${K3S_VERSION}#g" ./tools/bm-maintenance/cleanup.yml
-          kubectl apply -f ./tools/bm-maintenance/cleanup.yml
-      - name: Wait for cleanup job
-        id: update
-        run: |
-          kubectl wait -n maintenance-cleanup --for=condition=complete --timeout=600s job/cleanup-maintenance
+          nodes=$(kubectl get nodes -l ci.contrast.edgeless.systems/main-runner=true --field-selector spec.unschedulable=false -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')
+          if [[ -z "${nodes}" ]]; then
+            echo "::error::no node labeled ci.contrast.edgeless.systems/main-runner=true"
```
charludo:
```suggestion
            echo "::error::no node labeled ci.contrast.edgeless.systems/main-runner=true or all runners are cordoned"
```

.github/workflows/bm_maintenance_platform.yml
```diff
@@ -86,19 +86,33 @@ jobs:
...
+            exit 1
+          fi
+          failed=0
+          while IFS= read -r node; do
+            kubectl delete job -n maintenance-cleanup cleanup-maintenance --ignore-not-found
+            sed -e "s#@@REPLACE_IMAGE@@#${IMAGE}#g" -e "s#@@REPLACE_K3S_VERSION@@#${K3S_VERSION}#g" -e "s#@@REPLACE_NODE@@#${node}#g" \
+              ./tools/bm-maintenance/cleanup.yml | kubectl apply -f -
+            kubectl wait -n maintenance-cleanup --for=condition=complete --timeout=600s job/cleanup-maintenance || failed=1
```
charludo:
```suggestion
            if ! kubectl wait -n maintenance-cleanup --for=condition=complete --timeout=600s job/cleanup-maintenance; then
              failed=1
              echo "::error::cleanup failed on ${node}"
              kubectl get pods -n maintenance-cleanup -o wide || true
              kubectl describe job -n maintenance-cleanup cleanup-maintenance || true
              kubectl events -n maintenance-cleanup || true
            fi
```
+ drop the `if: failure()` step below.

(same for the other while loop below).

Reason: If node A fails and node B succeeds, previously we'd only sees B's job.

charludo (commented):
Thanks! LGTM, to more small things re: errors.

.github/workflows/bm_maintenance_platform.yml
```diff
@@ -87,18 +86,32 @@ jobs:
...
-        id: update
-        run: |
-          kubectl wait -n maintenance-cleanup --for=condition=complete --timeout=600s job/cleanup-maintenance
+          nodes=$(kubectl get nodes -l ci.contrast.edgeless.systems/main-runner=true --field-selector spec.unschedulable=false -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')
+          if [[ -z "${nodes}" ]]; then
+            echo "::error::no node labeled ci.contrast.edgeless.systems/main-runner=true, or all of them are cordoned"
+            exit 1
+          fi
+          failed=0
+          while IFS= read -r node; do
```
burgerdev: I think this could be just `for node in $nodes`.

.github/workflows/bm_maintenance_platform.yml
```diff
@@ -119,16 +132,31 @@ jobs:
...
-        id: update
-        run: |
-          kubectl wait -n maintenance-containerd-cleanup --for=condition=complete --timeout=600s job/containerd-cleanup-maintenance
+          nodes=$(kubectl get nodes -l ci.contrast.edgeless.systems/main-runner=true --field-selector spec.unschedulable=false -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')
+          if [[ -z "${nodes}" ]]; then
+            echo "::error::no node labeled ci.contrast.edgeless.systems/main-runner=true, or all of them are cordoned"
+            exit 1
+          fi
+          failed=0
+          while IFS= read -r node; do
```
burgerdev: dito

burgerdev (approved):
Changing the jobs to target specific nodes makes perfect sense - there's just not a good way to run jobs but schedule like a daemonset that I know of.

However, I want to stress that having more than one `main-runner` was not how this was designed - the name itself is a giveaway. 
- There should not be two action runners in one cluster, because it violates assumptions about concurrent execution. (containerd restarts will mess with concurrent tests)
- There should not be two nodes labeled `main-runner`, because our CSI driver does not support that.
- There's also the implicit assumption that the Github action runs on the node that's labeled `main-runner`.

sespiros:
@burgerdev thanks, I implemented the fixes.

>     * There should not be two action runners in one cluster, because it violates assumptions about concurrent execution. (containerd restarts will mess with concurrent tests)
> 
>     * There should not be two nodes labeled `main-runner`, because our CSI driver does not support that.
> 
>     * There's also the implicit assumption that the Github action runs on the node that's labeled `main-runner`.

The second node didn't get the label by accident, iiuc this adds a `main-runner=true` nodeSelector to all Contrast pods: https://github.com/edgelesssys/contrast/blob/904f6e55a0cfc4b1033fb9e88a0409fef9fd4e6e/internal/kuberesource/mutators.go#L783 and then also https://github.com/edgelesssys/contrast/commit/7891c4af85f1866a04ddc58395ee20b2aaf0eeea makes `TestGPU` deploy one workload per GPU model found in the cluster, so the B300 test can only be scheduled if that node carries the label.

PR #2599, opened 2026-08-21
upgrade-gpu-operator: upgrade in place instead of reinstalling

sespiros:
Fixes CON-311.

helm uninstall + helm install never refreshes a CRD, so discovery validated v26.3.3's ClusterPolicy against a schema from 2025-07-21. The API server silently pruned the two new fields, the operator deployed no device plugin and still reported itself ready, and every GPU test since failed to schedule.

Use helm upgrade --install so the chart's pre-upgrade CRD hook runs. Drop the version-equality skip that made the broken state permanent, bound the wait loop that held a runner for 6h, verify the settings actually landed, and build the package natively instead of pinning it to x86_64-linux.

Do not merge before the v1.23.1 patch release is cut. Merging fires bare metal maintenance, which with the skip gone resets discovery to P_GPU_ALIAS="". It's currently pinned to the pgpu alias, cluster-side only, because tmp/v1.23.1 hardcodes that name. main accepts either.

### DIFF
--- a/packages/by-name/scripts/upgrade-gpu-operator/upgrade-gpu-operator.sh
+++ b/packages/by-name/scripts/upgrade-gpu-operator/upgrade-gpu-operator.sh
@@ -71,18 +71,38 @@ helm upgrade --install "$RELEASE" --wait \
 
 # The API server drops any ClusterPolicy field the installed CRD doesn't declare,
 # without erroring, and the operator then reports itself ready having deployed no
-# device plugin. Assert the settings above actually survived, so a stale CRD fails
-# here instead of silently producing a GPU-less node.
-if ! kubectl get clusterpolicy -o json |
-  jq -e '.items[0].spec
-         | .sandboxWorkloads.mode == "kata"
-           and ([.kataSandboxDevicePlugin.env[]? | select(.name == "P_GPU_ALIAS")] | length == 1)' >/dev/null; then
-  echo "ERROR: the ClusterPolicy did not retain the kata sandbox device plugin settings." >&2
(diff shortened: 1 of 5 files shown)


packages/by-name/scripts/upgrade-gpu-operator/upgrade-gpu-operator.sh
```diff
@@ -29,26 +29,23 @@ if [[ -z $GPU_OPERATOR_VERSION ]]; then
...
-    echo "GPU Operator version $GPU_OPERATOR_VERSION is already installed. Skipping installation."
-    exit 0
-  fi
-  helm delete -n gpu-operator "$helm_release"
+# Reuse the existing release name if the operator is already installed. Clusters
+# set up before this script used --generate-name, so their release is called
+# something like gpu-operator-1787258651. Upgrading that release in place keeps
+# them on one release instead of stranding the old one next to a new one.
+RELEASE=gpu-operator
+existing=$(helm list -n gpu-operator -o json 2>/dev/null | jq -r '.[].name' | grep '^gpu-operator' | head -1 || true)
```
charludo: How about:
```suggestion
existing=$(helm list -n gpu-operator -o json | jq -r 'first(.[].name | select(startswith("gpu-operator"))) // empty')
```
this avoids (through `|| true` and `2>/dev/null`) that an actual error leads to creating a duplicate release

charludo (approved):
Thanks. Generally LGTM. Though I guess we'll need to try it and see if something breaks.

charludo (approved):

PR #2602, opened 2026-08-24
apitypes: add stable error codes

charludo:
@sespiros sorry it took so long, this is another PR that came out of your comments on the loop doc.

### DIFF
--- a/apitypes/apitypes.go
+++ b/apitypes/apitypes.go
@@ -6,33 +6,5 @@
 // The types here define the request/response bodies exchanged with the Coordinator.
 package apitypes
 
-import (
-	"fmt"
-	"net/http"
-)
-
 // Port is the listening port of the HTTP API server.
 const Port = "1314"
-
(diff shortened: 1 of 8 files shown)


sespiros (approved):
LGTM, 2 comments highlighted in AI review which I think are nice additions:

1. Pin the wire format: a marshal fixture asserting {"code":"no_manifest",...} (à la TestCapabilitiesResponseFieldName), and assert Code once in the handler test. The strings are contract now; a constant rename would pass CI today.
2. ErrNoStateToRecover isn't mapped in errorCode(), so it becomes invalid_argument (412 → 4xx fallback). Latent until Recover goes HTTP, but map it or leave a TODO.

charludo:
@sespiros thanks for the review, implemented the two suggestions. Feel free to take another look. Otherwise, I'm thinking of integrating this into the stack.... right now it's a "branch" on the stack stem ^^

sespiros:
@charludo nice, lgtm!

PR #2603, opened 2026-08-24
apitypes: move the v1 wire types into a subpackage

charludo:
...and this is the last one of the loop doc PRs.

### DIFF
--- a/apitypes/apitypes.go
+++ b/apitypes/apitypes.go
@@ -3,7 +3,9 @@
 
 // Package apitypes contains the wire-format types of the Contrast HTTP API.
 //
-// The types here define the request/response bodies exchanged with the Coordinator.
+// This package holds only what is independent of the API version.
+// Everything that belongs to a specific version lives in a subpackage named after it, e.g.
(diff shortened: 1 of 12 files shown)


sespiros (approved):
LGTM, an AI comment:

`TestNoExternalImports` only scans the root directory, so `apitypes/apiv1` just escaped the stdlib-only guard, and apiv1 legitimately imports root apitypes for the alias, which the current rule would reject. Suggest: walk all apitypes/... dirs, allow stdlib + the apitypes/ prefix, reject the rest.

charludo:
> LGTM, an AI comment:
> 
> `TestNoExternalImports` only scans the root directory, so `apitypes/apiv1` just escaped the stdlib-only guard, and apiv1 legitimately imports root apitypes for the alias, which the current rule would reject. Suggest: walk all apitypes/... dirs, allow stdlib + the apitypes/ prefix, reject the rest.

Good catch, done

sespiros:
@charludo lgtm, thanks

PR #2604, opened 2026-08-24
dev-shells: add helm

sespiros:
The repo ships upgrade-gpu-operator, whose whole job is driving helm, but helm isn't on PATH in the dev shell, so debugging it means reaching outside with nix shell. kubectl and yq-go are already here for the same kind of cluster poking.

This pulls helm from the flake's nixpkgs pin, so it's the same 3.20.2 the script's own package uses rather than whatever nixpkgs-unstable has.

### DIFF
--- a/dev-shells/by-name/development.nix
+++ b/dev-shells/by-name/development.nix
@@ -12,6 +12,7 @@
   gotools,
   just,
   kubectl,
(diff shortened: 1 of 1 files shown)


charludo (approved):
<3

PR #2605, opened 2026-08-24
dev-docs/e2e: admit release.yml to the WIF attribute condition

sespiros:
The condition pinned assertion.ref to refs/heads/main, written when release_nightly.yml was the only consumer. release.yml became a second consumer the next day (4f6126ecb4) and can never satisfy it: it refuses to run outside a tmp/vX.Y.Z branch, so its OIDC ref claim is never refs/heads/main.

That has been true since May, but nothing exercised it. The darwin release e2e entries ran for the first time during v1.23.1 (run 32400490123), where the auth step was rejected with unauthorized_client.

Pin the new clause to workflow_dispatch and to release.yml's workflow_ref so a release branch can't be used through any other trigger or workflow.

The live provider has been updated to match.


charludo (approved):
Thanks, esp. also for already updating the providers

PR #2606, opened 2026-08-24
.github: publish the release before deleting the working branch

sespiros:
The working branch is the draft release's target_commitish, so deleting it first makes every later edit to the release fail with "HTTP 422: Release.target_commitish is invalid". That includes the --draft=false publish itself, so v1.23.1 was left as a draft after the tag and the release branch had already been pushed.

Latent since dfcdfa4070, which folded this cleanup in from release_publish.yml. There it ran on the release:published trigger, so it could not run before the release left draft state.


charludo (approved):
Thanks!

PR #2607, opened 2026-08-24
.github: generate release notes once, before they are reviewed

sespiros:
Three fixes found while investigating why v1.23.1 could not publish.

Notes for a patch diffed against the previous minor, replaying the whole preceding minor release. The diff base now follows the release kind.

The publish job regenerated the notes after the approval gate, discarding the operator's review. It recomputed text the draft already had, and could not reproduce anything a human added, which security releases always need: the fix comes from a private fork with no PR in this repository, so v1.19.1 and v1.23.1 name the advisory by hand.

The post-release PR to main now merges after publishing, since it advertises the release in contrast-releases.json and the docs.

Depends on #2606.


dev-docs/release.md
```diff
@@ -20,13 +49,13 @@ This draft release can be promoted to actual release.
...
+6. Review the release notes. If label/title/description changes are necessary, change them on the original PR itself, then regenerate the notes on the draft (see [Editing the release notes](#editing-the-release-notes)).
 
-7. Review the release notes. If label/title/description changes are necessary, change them on the original PR itself.
+7. Approve the `Publish release` job in the GitHub Actions workflow run.
 
-8. Approve the `Publish release` job in the GitHub Actions workflow run.
+8. Check that the publish job succeeds.
 
-9. Check that the publish job succeeds and confirm the release notes on GitHub were regenerated against the now-existing tag.
+9. Review and merge the auto generated update PR for `main`. It advertises the new version in `contrast-releases.json` and the docs, so it must only land once the release exists.
```
charludo: Oh interesting reason. The reason i had in mind was actually something much more basic: when the release step fails, but the PR is already merged, it's an unnecessary PITA to revert 😄 

But your reason might be more convincing ^^

dev-docs/release.md
```diff
@@ -1,5 +1,34 @@
 # How to release
 
+## Editing the release notes
+
+The notes are written once, when the draft release is created, and the publish
+job never touches them again. Everything you change on the draft before
+approving `Publish release` is what ships.
+
```
charludo:
```suggestion
The notes are generated once, when the draft release is created.
```

(the rest of the paragraph is just the LLM regurgitating its task prompt.)

dev-docs/release.md
```diff
@@ -1,5 +1,34 @@
...
+Prefer fixing the source: set the right labels and titles on the PRs, then hit
+_Generate release notes_ on the draft. Check the _Previous tag_ the UI picked
+before you do. It guesses, and for a patch it tends to guess the previous
+minor, which replays the whole preceding minor release. It should be the
+previous patch in the same series, so `v1.23.0` for `v1.23.1`.
+
```
charludo: That's not a prefer, that's a hard rule that we also mention in the instructions below.

dev-docs/release.md
```diff
@@ -1,5 +1,34 @@
...
+Edit the body by hand when the notes can't come from PRs. A security fix is
+developed on a temporary private fork, so it reaches the release branch with no
+pull request in this repository and nothing generates a line for it. Add it
+yourself, as `v1.19.1` and `v1.23.1` do:
+
```
charludo:
```suggestion
Edit the body by hand when the notes can't be auto-generated, e.g. for PRs from a GHSA temporary private fork.

```

dev-docs/release.md
```diff
@@ -1,5 +1,34 @@
...
+### :warning: Security fixes
+
+* Fixes [GHSA-xxxx-xxxx-xxxx](https://github.com/edgelesssys/contrast/security/advisories/GHSA-xxxx-xxxx-xxxx)
+<!--    Please read the advisory to check if your existing Contrast deployment is affected. If so, upgrade to vX.Y.Z or apply the workarounds described in the advisory. -->
+```
+
+Keep the guidance line inside the comment markers while the advisory is still a
+draft, and remove the markers when you publish the advisory ([GHSA](ghsa.md),
+_Publish_ step 4). The advisory link is dead until then, which is expected: the
+patch release goes out before the advisory does.
```
charludo:
```suggestion
Uncomment the guidance line once the GHSA is published. The link is dead until then, which is OK.
```

(though, we haven't actually done that now; maybe we even don't want to? v1.19.1 also didn't.)

charludo (commented):
Thanks! Just had the same thing happen again with 1.24.0, so this is very welcome indeed 😄 

Only part of this I'm not happy with is the `Editing the release notes` docs section. It's pretty clearly LLM generated and an order of magnitude more verbose than necessary. IMO the terseness of our dev docs is a strength. Esp. in the `release.md`, we just have a bunch of very clear instructions. It would be nice if the new section also followed that.

(I also think that 99% of the information value would already be convered by: "ensure correct PR labls and title, then regenerate notes against correct tag. add manual additions, e.g. GHSA private fork PRs which aren't auto-discovered. Here's a template for that. Remove comment after publish".)

dev-docs/release.md
sespiros: Made it leaner, thanks!

dev-docs/release.md
sespiros: I decided to drop that comment completely actually.

charludo (approved):
Thanks!

Last thing: I'd prefer if the "Editing the release notes" was at the end / below the actual release instructions, but that is purely personal preference, and this is approved either way.

PR #2609, opened 2026-08-24
ci/release: test the version that was built, not the branch's

sespiros:
The build job bumps version.txt to the release version before building and caching, but the e2e jobs evaluate whatever the branch says. On a release that's a pre-version nothing ever built, so nothing is cached: Linux runners quietly rebuild and pass, Mac runners can't build Linux artifacts and fail.

Apply the same bump before evaluating.

That's the last fix that I had to workaround with the manual uploading to cachix of the assets needed by the darwin runners.

### DIFF
--- a/.github/actions/release_e2e_test/action.yml
+++ b/.github/actions/release_e2e_test/action.yml
@@ -5,6 +5,10 @@ inputs:
   version:
     description: Release tag to test against, e.g. v1.2.3
     required: true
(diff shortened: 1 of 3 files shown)


charludo (approved):
Thanks! So we probably also unnecessarily rebuild things after the release due to missing/wrong cache entry? Nice that that's fixed as well :D

PR #2610, opened 2026-08-25
.github: push nightly images and relabel instead of pushing to release tag

charludo:
Instead of nightly pushing to the ghcr release tag, nightly now pushes to a dated tag, then promote relabels to the release tag.

The second commit is unrelated and fixes a regex that left a handful of versioned docs entries mangled.


.github/workflows/release_promote.yml
```diff
@@ -179,6 +179,51 @@ jobs:
...
+          set -euo pipefail
+          crane=$(nix build --no-link --print-out-paths ".#${SET}.nixpkgs.crane" | head -n1)/bin/crane
+
+          while IFS= read -r replacement; do
+            [[ -n "${replacement}" ]] || continue
+            # Right-hand side is `<registry>/<repo>:<tag>@sha256:<digest>`.
+            pinned="${replacement#*=}"
+            ref="${pinned%%:*}@${pinned#*@}"
+            echo "Tagging ${ref} as ${VERSION}"
+            "${crane}" tag "${ref}" "${VERSION}"
```
sespiros: crane already accepts `<repo>:<tag>@sha256:<digest>` and resolves by digest, so the surgery buys nothing. Verified with a deliberately wrong tag:
```
$ crane digest alpine:3.20
sha256:d9e853e87e55526f6b2917df91a2115c36dd7c696a35be12163d44e6e2a4b6bc
$ crane digest "index.docker.io/library/alpine:3.19@sha256:d9e853e8…b6bc"
sha256:d9e853e87e55526f6b2917df91a2115c36dd7c696a35be12163d44e6e2a4b6bc
```
It also removes a latent trap: `${pinned%%:*}` strips from the first colon, so it silently mangles any registry carrying a port (localhost:5000/...). Unreachable from this workflow, but this is the kind of loop that gets copied.


```suggestion
            # Right-hand side is `<registry>/<repo>:<tag>@sha256:<digest>`; crane
            # resolves it by digest, so the tag the nightly baked in is ignored.
            pinned="${replacement#*=}"
            echo "Tagging ${pinned} as ${VERSION}"
            "${crane}" tag "${pinned}" "${VERSION}"
```

.github/workflows/release_promote.yml
```diff
@@ -179,6 +179,51 @@ jobs:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Tag promoted digests with the release version
+        env:
+          VERSION: ${{ needs.prepare.outputs.EFFECTIVE_VERSION }}
+          SET: base
+        run: |
+          set -euo pipefail
+          crane=$(nix build --no-link --print-out-paths ".#${SET}.nixpkgs.crane" | head -n1)/bin/crane
+
+          while IFS= read -r replacement; do
```
sespiros, line 225: If the artifact is ever empty or truncated, the while read loop is a no-op, retag-images goes green, and publish ships a release whose images carry no release tag, discovered only by users. Checking before the loop fails fast instead of half-tagging. The hardcoded 6 mirrors the expected array in the Verify nightly tests passed step above: deliberate breakage when the image list changes, rather than silent under-tagging. Verified: 0 lines → exit 1, comment-only file → exit 1, 3 of 6 → exit 1, full 6 → exit 0. (|| true is required, grep -c exits 1 on zero matches and set -e would kill the step before the error message printed.)


```suggestion

          # Keep in sync with the image list in .github/actions/release_artifacts.
          count=$(grep -cE '^[^#].*=' image-replacements.txt || true)
          if (( count != 6 )); then
            echo "::error::Expected 6 image replacements, found ${count}."
            exit 1
          fi

          while IFS= read -r replacement; do
```

packages/scripts.nix
```diff
@@ -270,7 +270,7 @@
     ];
     text = ''
       tag='[a-zA-Z0-9_.-]+'
-      sha='@sha256:[a-fA-F0-9]{64}'
+      sha='@sha256:([a-fA-F0-9]{64}|\.\.\.)'
```
sespiros: It suggests doing * so that if we do a patch release on an already mangled older version, it fixes it.

```suggestion
      sha='(@sha256:([a-fA-F0-9]{64}|\.\.\.))*'
```

You can also fix past versioned docs:
```
sed -i -E 's/(@sha256:[a-f0-9]{64})@sha256:\.\.\./\1/' \
  docs/versioned_docs/version-*/howto/troubleshooting.md
```

.github/workflows/release_promote.yml
```diff
@@ -179,6 +179,51 @@ jobs:
...
+            [[ -n "${replacement}" ]] || continue
```
sespiros: Skip potential comments.

```suggestion
            case "${replacement}" in ''|'#'*) continue ;; esac
```

sespiros (approved):
LGTM, just AI reviewer suggestions.

packages/scripts.nix
charludo: I wasn't sure if we want to touch the past docs... But I guess there's no real reason not to.

.github/workflows/release_promote.yml
charludo: fair! done

charludo:
@sespiros implemented the feedback, if you want to take another look. Otherwise LMK or just hit merge 😄

sespiros:
@charludo thanks, you might want to split the commits in a cleaner way, the last one got the regex fix but otherwise lgtm/feel free to merge.

charludo:
> @charludo thanks, you might want to split the commits in a cleaner way, the last one got the regex fix but otherwise lgtm/feel free to merge.

Oops, indeed, I forgot there's two commits here  🤦🏼‍♀️  Fixed, merging now.

PR #2611, opened 2026-08-26
apitypes: turn into proper go module

charludo:


go.mod
```diff
@@ -111,3 +112,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.4.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
+
+replace github.com/edgelesssys/contrast/apitypes => ./apitypes
```
sespiros, line 116: The replace => ./apitypes only works inside this repo, so anyone go getting contrast from outside will try to download apitypes v0.0.0 from GitHub and fail, no such tag exists. After merging we need to push a tag apitypes/v0.1.0 and point the root go.mod at it, before the next release. Should probably become a release-checklist item.

sespiros (approved):
LGTM, just AI comments again, 1 inline and:

Downloading a module only gives you that folder, so the published apitypes has no LICENSE file in it (and pkg.go.dev won't render unlicensed modules). Needs a LICENSE copy inside apitypes/.

go.mod
burgerdev: Actually, the root module should not be imported by other projects!

burgerdev:
> LGTM, just AI comments again, 1 inline and:
> 
> Downloading a module only gives you that folder, so the published apitypes has no LICENSE file in it (and pkg.go.dev won't render unlicensed modules). Needs a LICENSE copy inside apitypes/.

pkg.go.dev won't render BUSL-licensed code, either. But we can license SDK types as MIT. What I don't understand yet is how the final go module and dependency structure would look like.

go.mod
sespiros: Ok then we need to make the sdk its own module too so a consumer can simply import sdk and apitypes?

charludo:
> > LGTM, just AI comments again, 1 inline and:
> > Downloading a module only gives you that folder, so the published apitypes has no LICENSE file in it (and pkg.go.dev won't render unlicensed modules). Needs a LICENSE copy inside apitypes/.
> 
> pkg.go.dev won't render BUSL-licensed code, either. But we can license SDK types as MIT. What I don't understand yet is how the final go module and dependency structure would look like.

@sespiros is right and the SDK should move to standalone module as well, though in my mind that's downstream from untangling the SDK from the BUSL-licensed internal Contrast packages.

IMO that's not for this PR though.

go.mod
burgerdev: Oops, yes.

packages/by-name/contrast/contrast/package.nix
```diff
@@ -81,6 +83,8 @@ buildGoModule (finalAttrs: {
   checkPhase = ''
     runHook preCheck
     go test -tags=${lib.concatStringsSep "," finalAttrs.tags} -race ./...
+    # apitypes is a separate module, so it is not covered by the pattern above.
+    (cd apitypes && go test -tags=${lib.concatStringsSep "," finalAttrs.tags} -race ./...)
```
burgerdev:
```suggestion
    go test -tags=${lib.concatStringsSep "," finalAttrs.tags} -race ./apitypes/...
```

go.mod
burgerdev: We could avoid the tag with a pseudo-version, https://go.dev/doc/modules/version-numbers#in-development, for now.

charludo: done

packages/by-name/contrast/contrast/package.nix
charludo: done

PR #2614, opened 2026-08-27
  tdx-measure: derive RTMR[0] ACPI digests from QEMU blobs

sespiros:
## Summary

  Replace the hard-coded RTMR[0] ACPI digests with digests derived from QEMU's measured `fw_cfg` blobs.

  - add qtest and `fw_cfg` access to `tdx-measure`
  - dump QEMU's measured ACPI blobs during the Nix build
  - mirror Kata's normal and legacy-serial device topologies
  - calculate the ACPI extensions from the dumped blobs
  - verify both topologies against the existing reference digests

  The vCPU count remains fixed at one. Parameterization will be handled separately.

  ## Testing

  - `just lint`
  - `go test ./tools/tdx-measure/...`
  - `nix build -L --system x86_64-linux .#base.kata.qemu-acpi-blobs.tests.regression --no-link`
  - two-vCPU TDX `TestMultipleCPUs` E2E on the follow-up spike
  
  CON-131

### DIFF
--- a/packages/by-name/kata/qemuACPIBlobs/package.nix
+++ b/packages/by-name/kata/qemuACPIBlobs/package.nix
@@ -6,6 +6,7 @@
   stdenvNoCC,
   qemu-cc,
   source,
+  tdx-measure,
 }:
 
 {
--- a/packages/by-name/kata/qemuACPIBlobs/package.nix
+++ b/packages/by-name/kata/qemuACPIBlobs/package.nix
@@ -1,0 +1,152 @@
+# Copyright 2026 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+{
+  lib,
+  stdenvNoCC,
+  qemu-cc,
+  source,
+}:
+
+{
+  # MADT depends on the vCPU count.
+  vcpus,
+  legacySerial ? false,
+  # qemu-cc omits memory-dependent ACPI data; 130 MiB is the DMA minimum.
+  memoryMiB ? 1024,
+}:
+
+# Generates OVMF-measured ACPI blobs for Kata's q35 topology with qtest.
+let
+  # Force review of the mirrored topology on Kata upgrades.
+  mirroredKataVersion = "4.1.0";
+
+  qemuAcpiDump = tdx-measure.overrideAttrs (oldAttrs: {
+    pname = "qemu-acpi-dump";
+    subPackages = [ "internal/cmd/qemu-acpi-dump" ];
+    ldflags = [ "-s" ];
+    doCheck = false;
+    meta = oldAttrs.meta // {
+      mainProgram = "qemu-acpi-dump";
+    };
+  });
+
+  mkBlob =
+    {
+      vcpus,
+      legacySerial,
+      memoryMiB,
+    }:
+    let
+      mem = "${toString memoryMiB}M";
+
+      # Legacy serial replaces virtio-serial-pci and shifts later PCI slots.
+      serialDevices =
+        if legacySerial then
+          [
+            "-chardev"
+            "null,id=charconsole0"
+            "-serial"
+            "chardev:charconsole0"
+          ]
+        else
+          [
+            "-device"
+            "virtio-serial-pci,disable-modern=false,id=serial0"
+          ];
+
+      # Device order mirrors Kata and determines ACPI PCI slots.
+      qemuArgs = [
+        # TDX disables SMM and PIC; both affect ACPI.
+        "-machine"
+        "q35,accel=qtest,smm=off,pic=off"
+        # qtest cannot use -cpu host; ACPI uses the -smp count.
+        "-cpu"
+        "qemu64"
+        "-m"
+        mem
+        # Match Kata's OVMF hot-plug bridge.
+        "-device"
+        "pcie-root-port,id=rp-pci-bridge-0,bus=pcie.0,chassis=16,slot=0,addr=2,multifunction=off,bus-reserve=0x1,pref64-reserve=1m,mem-reserve=1m,io-reserve=4k"
+        "-device"
+        "pcie-pci-bridge,bus=rp-pci-bridge-0,id=pci-bridge-0"
+      ]
+      ++ serialDevices
+      ++ [
+        # Storage devices preserve Kata's PCI topology; null-co provides their backends.
+        "-blockdev"
+        "driver=null-co,node-name=drv-image,size=1073741824"
+        "-device"
+        "virtio-blk-pci,disable-modern=false,drive=drv-image"
+        "-device"
+        "virtio-scsi-pci,id=scsi0,disable-modern=false"
+        "-blockdev"
+        "driver=null-co,node-name=drv-initdata,size=1073741824"
+        "-device"
+        "virtio-blk-pci,disable-modern=false,drive=drv-initdata"
+        "-blockdev"
+        "driver=null-co,node-name=drv-imagepuller,size=1073741824"
+        "-device"
+        "virtio-blk-pci,disable-modern=false,drive=drv-imagepuller"
+        # Balloon preserves vhost-vsock's PCI slot without a host backend.
+        "-device"
+        "virtio-balloon-pci"
+        # hubport replaces tap; virtio-net-pci remains ACPI-visible.
+        "-netdev"
+        "hubport,id=network-0,hubid=0"
+        "-device"
+        "virtio-net-pci,netdev=network-0,disable-modern=false"
+        "-rtc"
+        "base=utc,driftfix=slew,clock=host"
+        "-vga"
+        "none"
+        "-no-user-config"
+        "-nodefaults"
+        "-nographic"
+        "--no-reboot"
+        "-object"
+        "memory-backend-ram,id=dimm1,size=${mem}"
+        "-numa"
+        "node,memdev=dimm1"
+        "-smp"
+        "${toString vcpus},cores=${toString vcpus},threads=1,sockets=1,maxcpus=${toString vcpus}"
+      ];
+
+      manifest = builtins.toJSON {
+        inherit vcpus legacySerial memoryMiB;
+        kataVersion = source.version;
+      };
+    in
+    assert lib.assertMsg (source.version == mirroredKataVersion)
+      "qemu-acpi-blobs mirrors Kata ${mirroredKataVersion}; review qemuArgs before updating to ${source.version}";
+    assert lib.assertMsg (vcpus > 0) "qemu-acpi-blobs requires vcpus > 0";
+    assert lib.assertMsg (
+      memoryMiB >= 130
+    ) "qemu-acpi-blobs requires memoryMiB >= 130 for fw_cfg DMA scratch space";
+    stdenvNoCC.mkDerivation {
+      name = "qemu-acpi-blobs${lib.optionalString legacySerial "-legacy-serial"}-${toString vcpus}vcpu";
+
+      dontUnpack = true;
+
+      nativeBuildInputs = [
+        qemu-cc
+        qemuAcpiDump
+      ];
+
+      buildPhase = ''
+        runHook preBuild
+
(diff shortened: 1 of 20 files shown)


tools/tdx-measure/internal/qtest/client.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"io"
+	"net"
+	"strings"
+	"time"
+)
+
+// Client is a qtest protocol client.
+type Client struct {
+	context context.Context
```
burgerdev: > Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. -- https://pkg.go.dev/context#pkg-overview

See also https://go.dev/blog/context-and-structs.

tools/tdx-measure/internal/qtest/client.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if _, err := io.WriteString(c.conn, command+"\n"); err != nil {
+		return "", fmt.Errorf("sending qtest command %q: %w", command, err)
+	}
+
+	response, err := c.reader.ReadString('\n')
+	if err != nil {
+		return "", fmt.Errorf("reading response to qtest command %q: %w", command, err)
+	}
+	response = strings.TrimSuffix(strings.TrimSuffix(response, "\n"), "\r")
```
burgerdev: Where would the `\r` come from? It looks like the responses are terminated by a sole `\n`. https://gitlab.com/qemu-project/qemu/-/blob/master/system/qtest.c#L404

tools/tdx-measure/internal/qtest/fwcfg.go
```diff
@@ -0,0 +1,154 @@
...
+	directory map[string]fwConfigFile
+}
+
+type fwConfigFile struct {
+	selector uint16
+	size     uint32
+}
+
+// OpenFWConfig reads the fw_cfg file directory. requiredFile identifies the
+// byte order used by QEMU's qtest DMA address ports.
```
burgerdev: Is this an emulation byte-order thing? We know that both guest and host are Intel - does that help?

tools/tdx-measure/internal/qtest/session.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	processDone := make(chan error, 1)
+	go func() {
+		processDone <- command.Wait()
+		close(processDone)
+	}()
+
+	connection, err := acceptConnection(ctx, listener, processDone)
+	if err != nil {
```
burgerdev, line 70: Is that actually correct? From what I understood, qemu listens on that socket and the test driver is supposed to connect? 

https://gitlab.com/qemu-project/qemu/-/blob/master/system/qtest.c#L877

tools/tdx-measure/internal/acpi/files.go
```diff
@@ -0,0 +1,17 @@
...
+
+// TableLoader is the BIOS linker/loader fw_cfg path.
+const TableLoader = "etc/table-loader"
+
+// OVMFMeasuredFiles returns OVMF's fixed measurement order.
+func OVMFMeasuredFiles() []string {
+	return []string{
+		TableLoader,
+		"etc/acpi/rsdp",
+		"etc/acpi/tables",
```
burgerdev: Why is one a constant, but the others are not?

packages/by-name/kata/qemu-acpi-blobs/package.nix
```diff
@@ -0,0 +1,156 @@
...
+        mem
+        # Match Kata's OVMF hot-plug bridge.
+        "-device"
+        "pcie-root-port,id=rp-pci-bridge-0,bus=pcie.0,chassis=16,slot=0,addr=2,multifunction=off,bus-reserve=0x1,pref64-reserve=1m,mem-reserve=1m,io-reserve=4k"
+        "-device"
+        "pcie-pci-bridge,bus=rp-pci-bridge-0,id=pci-bridge-0"
+      ]
+      ++ serialDevices
+      ++ [
+        # Block backends are not ACPI-visible; null-co avoids host disks.
```
burgerdev: Why do we need these at all, then?

packages/by-name/kata/qemu-acpi-blobs/package.nix
```diff
@@ -0,0 +1,156 @@
+# Copyright 2026 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+{
+  buildGoModule,
+  # MADT depends on the vCPU count.
+  vcpus ? 1,
```
burgerdev: Shouldn't this be a function parametrized with `vcpus`? Like

https://github.com/edgelesssys/contrast/blob/f5acb3c21c7500828a1b80e7beaf30304f4c6ce7/packages/by-name/kata/calculateSnpLaunchDigest/package.nix#L4-L16

tools/tdx-measure/internal/qtest/client.go
sespiros: Thanks. Fixed.

tools/tdx-measure/internal/qtest/client.go
sespiros: Fixed.

tools/tdx-measure/internal/qtest/fwcfg.go
sespiros: [It is big endian](https://gitlab.com/qemu-project/qemu/-/blob/v11.0.0/docs/specs/fw_cfg.rst#L186). Thanks for spotting it.

tools/tdx-measure/internal/qtest/session.go
sespiros: Yes it is correct, and there are 2 modes actually. When you don't specify `server=on`, the "client" is the one which opens the socket and then QEMU connects to it. QEMU’s own libqtest follows the same sequence, it creates the listening socket before spawning QEMU (https://gitlab.com/qemu-project/qemu/-/blob/v11.0.0/tests/qtest/libqtest.c#L487-512), then accepts QEMU’s connection (https://gitlab.com/qemu-project/qemu/-/blob/v11.0.0/tests/qtest/libqtest.c#L547-555). I added a comment clarifying this distinction.

tools/tdx-measure/internal/acpi/files.go
sespiros: Thanks, fixed, that was a refactor artifact.

packages/by-name/kata/qemu-acpi-blobs/package.nix
sespiros: Good point, fixed!

packages/by-name/kata/qemu-acpi-blobs/package.nix
sespiros: The comment is misleading. Technically the block backends are not ACPI-visible yes but the point here is that the attached PCI devices do modify the tables so we need to reproduce exactly Kata's PCI topology. Actually removing these modify all rdsp, tables, table-loader blobs. Clarirified this in the comment.

Separately, QEMU requires each virtio-blk-pci device to have a backend hence the null-co to satisfy the requirement without actual storage.

packages/by-name/kata/qemu-acpi-blobs/package.nix
```diff
@@ -3,18 +3,21 @@
...
   tdx-measure,
 }:
 
+{
+  # MADT depends on the vCPU count.
+  vcpus,
+  legacySerial ? false,
+  # qemu-cc omits memory-dependent ACPI data; 130 MiB is the DMA minimum.
+  memoryMiB ? 1024,
+}:
```
burgerdev, line 19: And now that this is a function, it needs to be camel case (`qemuACPIBlobs`?) cc @charludo

burgerdev (approved):
Thanks!

packages/by-name/kata/qemuACPIBlobs/package.nix
```diff
@@ -3,18 +3,21 @@
...
   tdx-measure,
 }:
 
+{
+  # MADT depends on the vCPU count.
+  vcpus,
+  legacySerial ? false,
+  # qemu-cc omits memory-dependent ACPI data; 130 MiB is the DMA minimum.
+  memoryMiB ? 1024,
+}:
```
sespiros: Good point, I will push the change.

packages/by-name/kata/qemuACPIBlobs/package.nix
```diff
@@ -0,0 +1,176 @@
...
+  mirroredKataVersion = "4.1.0";
+
+  qemuAcpiDump = buildGoModule {
+    pname = "qemu-acpi-dump";
+    version = "0.1.0";
+
+    src = ../../../../tools/tdx-measure;
+
+    proxyVendor = true;
+    vendorHash = "sha256-FNOA2wkqCk65e2qV+mezlmv2NZ96/K/fV8ocpVDwpAs=";
```
charludo: `just codegen` doesn't update this. It's also identical to `tdx-measure/package.nix`. Can't you just use 
```nix
tdx-measure.overrideAttrs (_: { pname = ...; subPackages = [ "internal/cmd/qemu-acpi-dump" ]; doCheck = false; })
```
or similar here?

charludo (commented):
Sorry for being late to the party.

Does the regression test actually run anywhere?

packages/by-name/kata/qemuACPIBlobsTest/package.nix
```diff
@@ -0,0 +1,6 @@
+# Copyright 2026 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+{ qemuACPIBlobs }:
+
+(qemuACPIBlobs { vcpus = 1; }).tests.regression
```
charludo: passing `vcpus = 1` here, but...

packages/by-name/kata/qemuACPIBlobs/package.nix
```diff
@@ -0,0 +1,169 @@
...
+  passthru.tests.regression = tdx-measure.overrideAttrs (_oldTdxMeasure: {
+    pname = "qemu-acpi-blobs-test";
+    # Reference digests use one vCPU; memory does not affect qemu-cc ACPI.
+    ACPI_BLOBS_DEFAULT_DIR = mkBlob {
+      vcpus = 1;
+      legacySerial = false;
+      inherit memoryMiB;
+    };
+    ACPI_BLOBS_LEGACY_SERIAL_DIR = mkBlob {
+      vcpus = 1;
```
charludo: ...but that does not take effect without:
```suggestion
      inherit vcpus;
```
(also the same just above this!)

charludo (approved):
Changes LGTM, apart from the comment(s) below. Also, I assume the test will be extended to some vcpu counts >1 once the other PR(s) land?

sespiros:
> Changes LGTM, apart from the comment(s) below. Also, I assume the test will be extended to some vcpu counts >1 once the other PR(s) land?

No, the tests just test that for 1 vcpu the ones that get generated match the hardcoded we had before. If these change we could do the same procedure we did before to calculate the new normal and adjust the tests. So I simplified it further, removed the passthru relationship completely to make it clearer that this is not parameterized actually like `qemuACPIBlobs`.

charludo (approved):

charludo:
> > Changes LGTM, apart from the comment(s) below. Also, I assume the test will be extended to some vcpu counts >1 once the other PR(s) land?
> 
> No, the tests just test that for 1 vcpu the ones that get generated match the hardcoded we had before. If these change we could do the same procedure we did before to calculate the new normal and adjust the tests. So I simplified it further, removed the passthru relationship completely to make it clearer that this is not parameterized actually like `qemuACPIBlobs`.

Ahhh sorry I was comparing this in my mind against the `snp-measure` tool/test but forgot that there, we can easily get the values to compare against from elsewhere, and that's not the case here. Makes sense, 1 it is :D

PR #2615, opened 2026-08-27
ci: replace dgx-007 with b13-gs01 as the TDX-GPU runner

sespiros:
b13-gs01 takes over the TDX-GPU cluster.

The runner swap is already done, so `update bm tcb specs` is failing on main until this merges (next scheduled attempt 01:00 UTC). `b13-gs01-kubeconfig` exists in GCP and the CI service account (needed for the darwin jobs) is bound to it.

Validated with four consecutive green `Metal-QEMU-TDX-GPU / gpu` runs on the new runner.

The dev-docs commit is separate, and covers what the bare-metal runner guide gets wrong once the node isn't k3s. Found while bringing b13-gs01 up.

CON-327


dev-docs/e2e/runner-decommision.md
```diff
@@ -32,4 +32,27 @@ just get-ghcr-read-token
...
+3. `gcloud secrets versions add "${RUNNER_NAME}-kubeconfig" --project=796962942582 --data-file=./"${RUNNER_NAME}-kubeconfig"`.
+4. Replace `/home/github/.kube/config` on the runner.
+5. Tell devs in the Teams channel to re-run `just get-credentials <platform>`.
+
+## Re-provision collateral-proxy
+
+*only applicable if the runner was part of a multi-node cluster*
+
+Its volume is on `csi-hostpath-sc`, a directory on the node that provisioned it, so when that node
+leaves the pod never schedules again. Its state is a cache, so just rebuild it:
```
charludo: Why only applicable then? New node new proxy, right?

IMO all three lines can go

sespiros: done, thanks, force-pushed.

dev-docs/e2e/bare-metal-runner.md
```diff
@@ -1,5 +1,15 @@
...
+> [!IMPORTANT]
+> One cluster, one runner, one node labelled `ci.contrast.edgeless.systems/main-runner=true`,
+> all the same machine. Two runners break the assumption that e2e tests never run concurrently,
+> since the node-installer restarts containerd. Two `main-runner` nodes break the `hostpath` CSI
+> driver, whose volumes are node-local. And `update_bm_tcb_specs.yml` resolves
+> `dev-docs/e2e/$(hostname)/manifest.json` on the runner.
+>
+> Adding a machine to a cluster that already has a runner means replacing the old one: remove its
+> registration and its `main-runner` label before the new runner goes online.
+
```
burgerdev: Inside a callout I'd prefer to focus on what to do, not why. Maybe something like this:


```suggestion
## Remove old runner (if applicable)

> [!IMPORTANT]
> Each testing cluster must have exactly one node labelled `ci.contrast.edgeless.systems/main-runner=true`. 

If there's already a runner in the cluster, it needs to be removed before the new runner can be added.
Two runners break the assumption that e2e tests never run concurrently,
since the node-installer restarts containerd. Two `main-runner` nodes break the `hostpath` CSI
driver, whose volumes are node-local. 

Remove the label `ci.contrast.edgeless.systems/main-runner=true` from all nodes in the cluster before continuing.
```

But that should go into the k8s setup section, because the doc's default assumption is a one-node cluster with k3s.

dev-docs/e2e/bare-metal-runner.md
```diff
@@ -254,17 +287,24 @@ If it's anything other than a btrfs, setup a btrfs builder volume.
...
 echo "Setting up btrfs nix builder volume..."
 sudo mkdir -p /mnt/nixbld
-sudo fallocate -l 20G /mnt/btrfs.img
+sudo fallocate -l 50G /mnt/btrfs.img
 sudo mkfs.btrfs -f /mnt/btrfs.img
 
 # Create fstab entry to mount the file as btrfs
-sudo echo -e "# btrfs for nix builder \n/mnt/btrfs.img /mnt/nixbld btrfs loop,defaults 0 0" >> /etc/fstab
+echo "# btrfs for nix builder" | sudo tee -a /etc/fstab
+echo "/mnt/btrfs.img /mnt/nixbld btrfs loop,defaults 0 0" | sudo tee -a /etc/fstab
```
burgerdev: A heredoc could be clearer.

dev-docs/e2e/bare-metal-runner.md
```diff
@@ -254,17 +287,24 @@ If it's anything other than a btrfs, setup a btrfs builder volume.
...
 sudo systemctl daemon-reload
 sudo mount -a
 
+# `mount -a` skips a malformed entry silently
+findmnt /mnt/nixbld
```
burgerdev: Maybe hint that the output should show a mount at that location, not at an ancestor?

dev-docs/e2e/bare-metal-runner.md
```diff
@@ -313,6 +353,19 @@ CONFIG="${CONFIG//127.0.0.1/$(hostname)}"
 echo "${CONFIG}" > $(hostname)-kubeconfig
 ```
 
+On a Scaleway Kosmos cluster there's nothing on the machine to rewrite, the API server is the
```
burgerdev: Please split `## Developer access` into `### k3s` and `### Scaleway Kosmos`. That way, you won't have to explain the difference, just describe the process.

dev-docs/e2e/bare-metal-runner.md
```diff
@@ -313,6 +353,19 @@ CONFIG="${CONFIG//127.0.0.1/$(hostname)}"
...
+managed one. Download the cluster kubeconfig instead and save it as `${RUNNER_NAME}-kubeconfig`.
```
burgerdev: This important requirement got lost (emphasis mine):

> Generate a Kubernetes configuration in the Scaleway portal, _using a dedicated API key for this machine_.

dev-docs/e2e/runner-decommision.md
```diff
@@ -32,4 +32,22 @@ just get-ghcr-read-token
 
 *only applicable if the runner was part of a multi-node cluster*
 
-1. Create a new kubeconfig.
+On a managed control plane the kubeconfig carries a cluster-wide admin token, not a per-machine
+credential, so removing the node revokes nothing. Deleting `/home/github/.kube/config` before the
+host is reinstalled suffices.
+
+Rotate only if the token is compromised. It can't be scoped, so every copy breaks at step 1,
+merged `~/.kube/config` files on developer machines included:
```
burgerdev: No, we don't want our machine suppliers to retain access to the cluster (token is somewhat compromised from the point it was added to the machine). Please rotate unconditionally.

dev-docs/e2e/runner-decommision.md
```diff
@@ -32,4 +32,22 @@ just get-ghcr-read-token
...
+
+1. `scw k8s cluster reset-admin-token "${CLUSTER_ID}"`.
+2. Download a fresh kubeconfig, checking it has a static token and not an `exec` credential
+   (see [bare-metal runner setup](./bare-metal-runner.md#developer-access)).
+3. `gcloud secrets versions add "${RUNNER_NAME}-kubeconfig" --project=796962942582 --data-file=./"${RUNNER_NAME}-kubeconfig"`.
+4. Replace `/home/github/.kube/config` on the runner.
+5. Tell devs in the Teams channel to re-run `just get-credentials <platform>`.
```
burgerdev: This should not be part of decommissioning, because the cluster should not be usable for CI anymore after the main runner was removed. The instructions for adding a runner cover the case of a replaced runner.

dev-docs/e2e/runner-decommision.md
```diff
@@ -32,4 +32,22 @@ just get-ghcr-read-token
...
+
+## Re-provision collateral-proxy
+
+```bash
+just collateral-proxy-redeploy
+```
```
burgerdev: Why is this needed?

sespiros: There were tests failures because collateral-proxy wasn't ready. The root cause was that collateral-proxy's storage was pinned to the old host so it needed a redeploy to the new node.

Moved to the setup guide next to the CSI driver install.

burgerdev (approved):
Thx!

PR #2618, opened 2026-08-31
osv-scanner, vulnix: ignore new findings

charludo:


osv-scanner.toml
```diff
@@ -57,3 +57,7 @@ reason = "wait for kata 4.1.0" # TODO(charludo): remove after kata 4.1.0 update
 [[IgnoredVulns]]
 id = "RUSTSEC-2026-0235" # rkyv
 reason = "wait for kata 4.1.0" # TODO(charludo): remove after kata 4.1.0 update
+
+[[IgnoredVulns]]
+id = "RUSTSEC-2026-0258" # h2
+reason = "only relevant to genpolicy, waiting for kata update" # TODO(charludo): remove after kata update
```
burgerdev: The fact that ignoring this is ok is due to the vuln being a DoS vector, I'd put that into the reason field.

vulnix-whitelist.toml
```diff
@@ -64,3 +64,6 @@ cve = [ "CVE-2026-40467", "CVE-2026-40468", "CVE-2026-40469", "CVE-2026-40553" ]
 
 ["patch"]
 cve = [ "CVE-2026-56288", "CVE-2026-56289" ]
+
+["coreutils"]
+cve = [ "CVE-2026-56391", "CVE-2026-56392" ]
```
burgerdev, line 69: Can we add justfications for these, too?

sespiros (approved):

PR #2619, opened 2026-09-01
rfc: backwards compatibility

charludo:
[rendered version](https://github.com/edgelesssys/contrast/blob/rfc-014/rfc/014-backwards-compatibility.md)

### DIFF
--- a/rfc/014-backwards-compatibility.md
+++ b/rfc/014-backwards-compatibility.md
@@ -1,0 +1,162 @@
+# RFC 014: Backwards Compatibility
+
+## Background
+
+Contrast's components currently communicate via gRPC APIs using aTLS, where the attestation evidence is embedded inside the TLS handshake.
+This makes it practically infeasible for third parties to reimplement components to fit their needs, for example a custom CLI replacement or an alternative initializer.
+Even for us, the current setup complicates additional, alternative components like the planned CVM support.
+The gRPC/aTLS approach additionally makes it difficult to front Contrast with a reverse proxy, which customers have expressed a desire for in the past.
+
+For these reasons, we're currently planning to introduce, and migrate to, an HTTP API.
+As a necessary precondition, this RFC defines the versioning scheme to be used by said HTTP API.
+Currently, the gRPC/aTLS couplings are versioned solely through Contrast release versions, and even then backwards compatibility isn't clearly defined and can break, as the description in [this issue](https://linear.app/edgelesssys/issue/CON-113/spike-get-an-overview-of-code-that-might-break-backwards-compatibility) shows.
+
+The goal of this RFC is to prevent a similar situation for the planned HTTP API.
+It defines how endpoints should be versioned, and how components should negotiate a shared version.
+As a result, Coordinators and clients of different versions will be able to communicate,
+and we will be able to introduce additional functionality easily and in a properly versioned manner, while guaranteeing that old functionality isn't broken unexpectedly.
+
+## Proposal
+
+Ownership of the client logic used to communicate with the Coordinator moves to the SDK for the HTTP API.
+This differs from the current layout, where each component owns their own, bespoke implementation of gRPC API client logic.
+The SDK becomes the sole owner of the client logic, and all Contrast components currently communicating via the gRPC User API (especially the CLI and initializer) will consume these SDK functions.
+Thus we eventually (once gRPC is fully gone) will end up with exactly one implementation of the protocol layer to version and test.
+
+In the medium-term, the goals are:
+- Every component still speaks gRPC/aTLS exactly as today.
+- Every component can additionally speak HTTP through the SDK.
+- Components communicate by the newest mutually supported HTTP API version, or fall back to gRPC, automatically.
+
+For the split gRPC/aTLS + HTTP approach we already have a precedent in the existing codebase with the Coordinator's `POST /attest` endpoint
+and the corresponding `GetAttestation` / `ValidateAttestation` functions in the SDK doing application-level attestation over plain HTTP.
+We will extend these patterns to the UserAPI.
+
+A table of currently planned HTTP API endpoints is shown below.
+
+| Operation | gRPC (aTLS) | HTTP | Status |
+| --- | --- | --- | --- |
+| Capabilities | n/a | `GET /capabilities` | released |
+| GetManifests | `userapi.GetManifests` | `POST /attest` / `POST /v1/attest` | released / planned |
+| SetManifest | `userapi.SetManifest` | `POST /v1/manifest` | planned |
+| Recover | `userapi.Recover` | `POST /v1/recover` | planned |
+| NewMeshCert | `meshapi.NewMeshCert` | `POST /v1/mesh/cert` | planned |
+
+While the design of the API and its security mechanisms will be the topic of a separate RFC, the endpoints and handling will share a common shape:
+- Each handler should reuse the internal logic from the existing gRPC methods.
+- For each handler, a corresponding SDK client method is exposed publicly.
+- Contrast components should use the new methods if possible, or fallback to gRPC.
+
+### Versioning scheme
+
+The old API contract is implicitly defined by aTLS internals, the proto definitions, and the gRPC library version.
+The new API contract is constituted solely by the HTTP verbs and paths, the JSON message shapes, and the error types the API might return.
+This contract lives fully in a new `apitypes` package.
+
+Additive changes are possible within an API version: new, optional fields may be added.
+Non-additive, breaking changes are prohibited within minor versions, and include:
+- The addition of new, mandatory fields.
+- The removal of fields.
+- The repurposing of fields.
+- The changing of API paths.
+
+In order to make a change like this, a new major version of the API is required.
+Rather than changing such implementation details in the `apitypes` package directly and releasing the changes as a new major version, a subpackage for each API version is added, namely `apitypes/apiv1` and so forth.
+The purpose of this is to allow the SDK to easily import and support past API versions.
+For clearer public-facing versioning, the SDK `semver` tracks the API version.
+New API versions require new major SDK versions, additive intra-version changes only require minor version bumps.
+
+The SDK exposes each API version via the client library, for example:
+```go
+func (c *Client) V1() *apiv1.API {
+	return apiv1.New(c.httpapi)
+}
+```
+
+In this way, an arbitrary number of API versions can be exposed simultaneously, and versions can be deprecated as needed.
+Consumers can thus choose to interact with a specific version of the API, for example:
+```go
+c := client.New(...)
+c.V1().SetManifest(ctx, req)
+```
+However, the SDK additionally exposes "unversioned" variants of the endpoints:
+```go
+c := client.New(...)
+c.SetManifest(ctx, req)
+```
+These variants autonegotiate the newest shared version between client and Coordinator before calling the corresponding versioned functions.
+Generally, this mechanism is to be preferred.
+
+### Automatic version negotiation
+
+From the above table, note that `/capabilities` is the only newly introduced endpoint that's unversioned.
+This is intentional: the endpoint serves as a way for clients to discover what versions the Coordinator speaks.
+
+Under `GET /capabilities`, the Coordinator handles capability inquiries by returning `sdk.supportedAPIVersions`.
+This is simply a list of all API versions the SDK knows about and supports, currently:
+```go
+var supportedAPIVersions = []string{apiv1.Version}
+```
+The client then simply selects the newest version both it and the Coordinator support, and makes requests via that version's functions.
+The capabilities response may be cached for reuse.
+
+In the case of the example from above, the "unversioned" `SetManifest` version looks something like this:
+```go
+func (c *Client) SetManifest(...) {
+    version, _ := c.NegotiateAPIVersion(...)
+    switch version {
+    case apiv1.Version:
+	    return c.V1().SetManifest(ctx, req)
+    default:
+	    return nil, fmt.Errorf("SetManifest is not implemented for API version %q", version)
+    }
+}
+```
+
+To mitigate downgrade attacks on this negotiation, two mechanisms are supported.
+
+First, users may pin the minimum required API version in the manifest via a new `MinimumAPIVersion` field.
+If the newest API version supported by both client and Coordinator is less than this field, communication via HTTP API isn't possible, and the client should throw an appropriate error.
+
+Even though we could, at least for the time being, fall back to the gRPC API in this case, this would ignore that a pinned minimum version clearly indicates the user's desire to use the HTTP API,
+thus likely meaning the user also expects the Coordinator to speak this version; moreover, the error-on-minimum-unmet behavior would need to be implemented anyways once gRPC is deprecated.
+Falling back to gRPC *in the case of a pinned minimum version* would therefore be a potentially unexpected, temporary behavior and shouldn't be introduced.
+
+Second, the hash of the Coordinator `/capabilities` response is included in `ConstructReportData`.
+This allows clients to verify that the capabilities response they received during version negotiation was genuine.
+
+### gRPC deprecation path
+
+Once the HTTP API is fully implemented and tested, the gRPC API can be deprecated in the medium-term future, after a sufficient transition period.
+The planned timeline for this looks as follows.
+
+0. Fully implement the HTTP API.
+1. Use the HTTP API by default in all our e2e tests.
+2. After 2 weeks of no (API-related) e2e failures, remove the `contrast_unstable_api` tags and default to the HTTP API with the next minor release.
+   Publicly announce the change in the changelog, mark as breaking change, and include a notice informing readers of the deprecation and planned removal for the second-next minor release.
+   Also print an appropriate warning in the CLI whenever gRPC fallback is being used.
+3. With the specified minor release, remove all gRPC related code, again marking the PR as breaking change.
+
+### Testing strategy
+
+A new e2e test is introduced with the purpose of testing that communication between clients and Coordinators with different supported API versions succeeds.
+For each supported API version in the SDK, plus a gRPC-only case, a corresponding test case starts a Coordinator supporting all versions *up to* the specified version.
+All user API functionalities (`GetManifests`, `SetManifest`, `Recover`) are tested for each supported SDK version (via the explicitly versioned SDK methods),
+as well as using the automatic version negotiation paths for overlap between versions, and disjunct supported versions.
+
+While the gRPC API is still supported, automatic negotiation failure should fall back to gRPC.
+Afterward, this simply becomes an expected error case, and the gRPC-only Coordinator test case can be removed.
+
+Additionally, the release e2e test should gain an new step in which the to-be-released deployment files are tested via an older, released version of the Contrast CLI, and vice versa.
+This is intended to prevent accidental breaking changes between the *same* versions of the API in *different* Contrast releases.
+
+## Out of scope
+
+Coordinator-to-Coordinator traffic (for example peer recovery) can keep using aTLS for now, since both ends are always ours.
+
+## Alternatives considered
+
+In regards to the gRPC deprecation path, we could also consider taking this as reason to bump to Contrast `v2.0.0`.
+We could do so essentially immediately after step fully implementing and testing the HTTP API.
+While we would have to support `v1` with security updates for at least some time,
+this would eliminate the need for the temporary gRPC and HTTP double support within the same version.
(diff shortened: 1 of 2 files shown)


rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+
+| Operation | gRPC (aTLS) | HTTP | Status |
+| --- | --- | --- | --- |
+| SetManifest | `userapi.SetManifest` | `POST /v1/manifest` | add handler: #2564, optionally use: #2575 |
+| GetManifests | `userapi.GetManifests` | `POST /attest` (released) + `POST /v1/attest` | has long-been implemented, moved into new version scheme in: #2574 |
+| Recover (user) | `userapi.Recover` | planned | gRPC-only today |
+| NewMeshCert | `meshapi.NewMeshCert` | planned `POST /v1/mesh/cert` | gRPC-only today, initializer-side fallback |
+| Recover (peer) | `meshapi.Recover` | eventually, so different-version Coordinators can recover each other | single-path |
+| Capabilities | n/a | `GET /capabilities` (released 1.23) | released, needs downgrade protection |
+| Transit engine | n/a | `/v1/transit/{encrypt,decrypt}/{name}` | already HTTP-only, needs integration into new versioning scheme |
```
burgerdev:
```suggestion
```

The transit API is not part of our HTTP API, but implements the Vault API and follows Vault's versioning scheme.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+The full table of planned and implemented endpoints is shown below (courtesy of @sespiros).
+
+| Operation | gRPC (aTLS) | HTTP | Status |
+| --- | --- | --- | --- |
+| SetManifest | `userapi.SetManifest` | `POST /v1/manifest` | add handler: #2564, optionally use: #2575 |
+| GetManifests | `userapi.GetManifests` | `POST /attest` (released) + `POST /v1/attest` | has long-been implemented, moved into new version scheme in: #2574 |
+| Recover (user) | `userapi.Recover` | planned | gRPC-only today |
+| NewMeshCert | `meshapi.NewMeshCert` | planned `POST /v1/mesh/cert` | gRPC-only today, initializer-side fallback |
+| Recover (peer) | `meshapi.Recover` | eventually, so different-version Coordinators can recover each other | single-path |
+| Capabilities | n/a | `GET /capabilities` (released 1.23) | released, needs downgrade protection |
```
burgerdev: I'd remove the PR references for now and speak of all the endpoints as planned, since it does not look like we can start with the implementation based on this RFC.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Compatibility surface
+
+The old contract was (is) implicitly defined by aTLS internals, the proto definitions, the gRPC library version.
+The new contract is constituted solely by the HTTP verbs and paths, the JSON message shapes, and the error types the API might return.
+
+### Future API evolution
+
+- within an `/vX/API` version, only additive changes are allowed (that is, new, optional fields). Existing fields may never be removed or repurposed. Paths can never change.
+- for breaking changes, a new API version path is added. In this process, a new `apitypes` version (for example `apitypes/v2/`) path is added. Only when adding this new version may fields be removed, repurposed, or otherwise be made incompatible. Components must continue to serve all past API versions, using the appropriate-version API types package.
```
burgerdev: I don't think "serving all past API versions" is a sensible requirement - don't we want to phase them out, eventually?

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+
+## gRPC deprecation path
+
+Once the HTTP API is fully implemented and tested, we should move towards deprecating the gRPC API in the medium-term future. Proposed plan:
+
+1. Use the HTTP API by default in all our e2e tests.
+2. After 2 weeks of no (API-related) e2e failures, remove the `contrast_unstable_api` tags and default to the HTTP API with the next minor release. Publicly announce the change in the changelog, mark as breaking change, include a notice informing readers of the deprecation and planned removal for the second-next minor release. Also print an appropriate warning in the CLI whenever gRPC is being used.
+3. With the specified minor release, remove all gRPC related code, again marking the PR as breaking change.
+Alternatively, we could - depending on what other (breaking) features land in the meantime - also consider taking this as reason to bump to Contrast v2.0.0. This we could do essentially immediately after step 1 of the 3 steps given above (as long as we continue to backport patch releases to v1.x.0 for a while).
```
burgerdev: I'd remove that here. If you feel this needs mentioning, put it in an "alternatives considered" section. That way, it's clear that this is _not_ the primary plan.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Testing strategy
+
+A new e2e test is introduced (#2577) with the purpose of testing the HTTP API, both via the autonegotiated, unversioned entrypoint and the explicitly versioned `/v1/` path.
```
burgerdev: This assumes a particular SDK layout which we don't have yet?

burgerdev: Do I understand correctly that the e2e test is run for all API versions with the current cli and coordinator code? How does that interact with downgrade attack prevention?

If this is using the current code anyway, a unit test for older protocols might be sufficient (together with the test using released artifacts mentioned below).

tools/vale/styles/config/vocabularies/edgeless/accept.txt
```diff
@@ -175,6 +180,7 @@ systemd
 Tailscale
 tardev
 tarfs
+tbd
```
burgerdev: That's almost always correct to flag.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+
+The full table of planned and implemented endpoints is shown below (courtesy of @sespiros).
+
+| Operation | gRPC (aTLS) | HTTP | Status |
+| --- | --- | --- | --- |
+| SetManifest | `userapi.SetManifest` | `POST /v1/manifest` | add handler: #2564, optionally use: #2575 |
+| GetManifests | `userapi.GetManifests` | `POST /attest` (released) + `POST /v1/attest` | has long-been implemented, moved into new version scheme in: #2574 |
+| Recover (user) | `userapi.Recover` | planned | gRPC-only today |
+| NewMeshCert | `meshapi.NewMeshCert` | planned `POST /v1/mesh/cert` | gRPC-only today, initializer-side fallback |
+| Recover (peer) | `meshapi.Recover` | eventually, so different-version Coordinators can recover each other | single-path |
```
burgerdev: This is "out of scope" at the end of the doc.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+
+We should also take care that the new HTTP contracts live only in the SDK.
+CLI and initializer become consumers of the SDK instead of implementing their own HTTP client logic.
+Thus we eventually (once gRPC is fully gone) will end up with exactly one implementation of the protocol layer to version and test.
+
+## Backwards compatibility
+
+### During the migration period
+
+- gRPC/aTLS servers keep running. We could also play it safe and just keep them the default in all components, and feature gate HTTP.
```
burgerdev: How about falling back to gRPC on 5xx HTTP errors?

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+- (when all HTTP endpoints are implemented and the unstable feature flag is removed) clients prefer their newest supported HTTP API version. They ask the server about its supported versions via the `/capabilities` endpoint and choose the newest common one (TODO: downgrade attacks relevant?), falling back to gRPC if the server doesn't speak HTTP (yet, or support not enabled). We use an explicit capabilities/version endpoint so a transient error isn't misunderstandable as "not supported."
+- AFAICT, there's no real benefit to pinning the API version for example in the manifest. The client should always drive selection.
```
burgerdev: Is this still true?

burgerdev (commented):
Thanks! The overall style could be a little less "researchy"/ambiguous and more prescriptive.

rfc/014-backwards-compatibility.md
charludo: This is what #2568 implemented. The SDK exposes both the "general" `client.SetManifest(...)` (e.g.), as well as for each version a `client.V1().SetManifest(...)`. We can just call these explicit functions here.

charludo: Clarified this in the text. 

> If this is using the current code anyway, a unit test for older protocols might be sufficient

Hm, interesting. Would that really be a unit test though? Seems more like a "headless e2e", testing the integration between multiple components?

If so, not sure if this really has a lot of benefit over actually running a full e2e.

rfc/014-backwards-compatibility.md
charludo: I'm not so sure. I'd like to prevent the possibility of an attacker forcing "down" to gRPC. If `5xx` would mean fall back to gRPC, that would be trivial.

I've mentioned something similar in the text. To be honest, this might all just be academic... We're not likely to introduce a `v2` of the API while gRPC is still supported.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+| --- | --- | --- | --- |
+| Capabilities | n/a | `GET /capabilities` | released |
+| GetManifests | `userapi.GetManifests` | `POST /attest` / `POST /v1/attest` | released / planned |
+| SetManifest | `userapi.SetManifest` | `POST /v1/manifest` | planned |
+| Recover | `userapi.Recover` | `POST /v1/recover` | planned |
+| NewMeshCert | `meshapi.NewMeshCert` | `POST /v1/mesh/cert` | planned |
+
+While the design of the API and its security mechanisms will be the topic of a separate RFC, the endpoints and handling will share a common shape:
+- Each handler should reuse the internal logic from the existing gRPC methods.
+- For each handler, a corresponding SDK client method is exposed publicly.
```
burgerdev, line 47: An interesting detail we might want to discuss: the `/attest` endpoint is split into two sdk functions, and for a good reason - one for fetching the data and one for validating it. That allows fetching the response by an untrusted component and verifying the response somewhere safe. 

Afaict, this is the only API that needs to support such access patterns (or even could support them without a big hassle). The other APIs need two round trips to first establish a key and then use it, which would be inconvenient to relay. This is also the only API useful for data owners, and we might be able to assume workload owners have a more or less direct connection to the Coordinator. 

However, if we version the APIs and allow access without the SDK, people could build something if they need it.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+The old API contract is implicitly defined by aTLS internals, the proto definitions, and the gRPC library version.
+The new API contract is constituted solely by the HTTP verbs and paths, the JSON message shapes, and the error types the API might return.
+This contract lives fully in a new `apitypes` package.
+
+Additive changes are possible within an API version: new, optional fields may be added.
+Non-additive, breaking changes are prohibited within minor versions, and include:
+- The addition of new, mandatory fields.
+- The removal of fields.
+- The repurposing of fields.
+- The changing of API paths.
```
burgerdev, line 61: This is all true, but we also need to keep the processing logic and any guarantees an endpoint gives compatible. May not need a deep exploration here, but it is a necessary requirement.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+
+In order to make a change like this, a new major version of the API is required.
+Rather than changing such implementation details in the `apitypes` package directly and releasing the changes as a new major version, a subpackage for each API version is added, namely `apitypes/apiv1` and so forth.
```
burgerdev, line 64: We could structure this a little bit more and add packages for use cases, similar to `userapi` vs. `meshapi`. Could be an `adminv1`, a `dataownerv1`, a `workloadv1` and `peerv1`. But maybe this is overkill for now and we can organize that later.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+c := client.New(...)
+c.V1().SetManifest(ctx, req)
+```
+However, the SDK additionally exposes "unversioned" variants of the endpoints:
+```go
+c := client.New(...)
+c.SetManifest(ctx, req)
+```
+These variants autonegotiate the newest shared version between client and Coordinator before calling the corresponding versioned functions.
+Generally, this mechanism is to be preferred.
```
burgerdev, line 88: I'm not sure whether this works in practice. If we introduce a `c.V2()`, it is by definition incompatible with the V1 API, so how would a `c.SetManifest` be able to call both? (If there was a way to massage one version's type into the other, we would have done so instead of releasing a breaking change)

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+    default:
+	    return nil, fmt.Errorf("SetManifest is not implemented for API version %q", version)
+    }
+}
+```
+
+To mitigate downgrade attacks on this negotiation, two mechanisms are supported.
+
+First, users may pin the minimum required API version in the manifest via a new `MinimumAPIVersion` field.
+If the newest API version supported by both client and Coordinator is less than this field, communication via HTTP API isn't possible, and the client should throw an appropriate error.
```
burgerdev, line 119: Not all callers have access to a manifest, or might have different manifests to choose from.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+
+Even though we could, at least for the time being, fall back to the gRPC API in this case, this would ignore that a pinned minimum version clearly indicates the user's desire to use the HTTP API,
+thus likely meaning the user also expects the Coordinator to speak this version; moreover, the error-on-minimum-unmet behavior would need to be implemented anyways once gRPC is deprecated.
+Falling back to gRPC *in the case of a pinned minimum version* would therefore be a potentially unexpected, temporary behavior and shouldn't be introduced.
+
+Second, the hash of the Coordinator `/capabilities` response is included in `ConstructReportData`.
+This allows clients to verify that the capabilities response they received during version negotiation was genuine.
```
burgerdev, line 126: This could actually become part of the HTTP key attestation protocol (to be designed).

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+
+### gRPC deprecation path
+
+Once the HTTP API is fully implemented and tested, the gRPC API can be deprecated in the medium-term future, after a sufficient transition period.
+The planned timeline for this looks as follows.
+
+0. Fully implement the HTTP API.
+1. Use the HTTP API by default in all our e2e tests.
+2. After 2 weeks of no (API-related) e2e failures, remove the `contrast_unstable_api` tags and default to the HTTP API with the next minor release.
+   Publicly announce the change in the changelog, mark as breaking change, and include a notice informing readers of the deprecation and planned removal for the second-next minor release.
```
burgerdev, line 136: This needs a much longer timeline to accommodate PM clients - at least on the Coordinator side.

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+   Also print an appropriate warning in the CLI whenever gRPC fallback is being used.
+3. With the specified minor release, remove all gRPC related code, again marking the PR as breaking change.
+
+### Testing strategy
+
+A new e2e test is introduced with the purpose of testing that communication between clients and Coordinators with different supported API versions succeeds.
+For each supported API version in the SDK, plus a gRPC-only case, a corresponding test case starts a Coordinator supporting all versions *up to* the specified version.
```
burgerdev, line 143:
```suggestion
For each supported API version in the SDK, plus a gRPC-only case, a corresponding test case starts the latest released Coordinator supporting all versions *up to* the specified version.
```

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+All user API functionalities (`GetManifests`, `SetManifest`, `Recover`) are tested for each supported SDK version (via the explicitly versioned SDK methods),
+as well as using the automatic version negotiation paths for overlap between versions, and disjunct supported versions.
+
+While the gRPC API is still supported, automatic negotiation failure should fall back to gRPC.
```
burgerdev, line 147: Doesn't that contradict line 120ff?

rfc/014-backwards-compatibility.md
```diff
@@ -0,0 +1,162 @@
...
+Afterward, this simply becomes an expected error case, and the gRPC-only Coordinator test case can be removed.
+
+Additionally, the release e2e test should gain an new step in which the to-be-released deployment files are tested via an older, released version of the Contrast CLI, and vice versa.
```
burgerdev, line 150: Imho we can skip the "new CLI, older Coordinator" tests for simplicity (and also declare that configuration unsupported).

rfc/014-backwards-compatibility.md
charludo: Not sure I quite understand the conclusion - we should try to split functions into validation vs networking, with both halves exposed, so consumers can pick & choose?

rfc/014-backwards-compatibility.md
charludo: True. Added a sentence, but it is the vaguest of the bunch. 

Though at least the e2e and release tests *should* catch errors in that class.

rfc/014-backwards-compatibility.md
charludo: Hm, I'm a bit torn. A bit more organization does sound like a generally good thing. Though not sure how worth it is, given that the number of actual functions we plan to expose is low. In any case, #2626 would probably be the best place, if we want to do this...

rfc/014-backwards-compatibility.md
charludo: Hm. We could make it a flag that defaults to the manifest entry if present. But, then it's up to the caller again to make sure to add the flag. ~~(Or add a config file 🫣)~~

Though this does not affect the detection, only the prevention.

rfc/014-backwards-compatibility.md
charludo: Will clarify in the text, but no. Basically, we view gRPC as `apiv0`, and assume all components speak it. We negotiate the highest sharedly spoken version between components, which *might* be `apiv0`. But if a user has pinned `apivX` with `x>0`, then we don't negotiate down to / fall back to gRPC.

Basically, 120ff was exclusively about the pinned to `>0` case.

rfc/014-backwards-compatibility.md
charludo: Is there a reason for this? It costs us close to nothing to keep shipping older versions of the API in the CLI.

rfc/014-backwards-compatibility.md
charludo: Eh. Right. Fix is relatively simple though, just don't unmarshal inside the API functions. The types are public anyways, so the caller can also do it.

Added two clarifying lines in the text.

rfc/014-backwards-compatibility.md
charludo: Lengthened to 6 months - is that OK? Though this should only affect step `3.`, right? We can already default to HTTP earlier, juts not remove gRPC.

rfc/014-backwards-compatibility.md
burgerdev: We can save one half of the test matrix, which may also be the more tedious one to test.

rfc/014-backwards-compatibility.md
burgerdev: That was the idea, yes, and I think my suggestion in the last sentence was wrong: validation according to manifest needs to be done by the SDK and can't simply be reimplemented elsewhere. But I don't think we need to support this for APIs that are workload(-owner)-facing.

rfc/014-backwards-compatibility.md
burgerdev: Did you push these lines? At least I don't find them.

rfc/014-backwards-compatibility.md
burgerdev: I think manifest, flag and reportdata content are sufficiently many options.

rfc/014-backwards-compatibility.md
burgerdev: Ok, but this means that downgrade protection is only available once client and server removed gRPC, correct?

PR #2620, opened 2026-09-01
debugshell: expose to host

burgerdev:
This PR adds new functionality to the `debugshell` used to log into the VM image rootfs. If the workload is configured to allow debug access (`contrast generate --insecure-enable-debug-shell-access`), the SSH server started inside the VM is now listening on a VSOCK in addition to a local TCP port. This allows connecting to the VM without adding a sidecar container or requiring containers to start at all. 

---

### Developer notes

Most code changes in this PR are for adding a script to the debugshell container that can be used to access a pod from the host, with something like the following pod:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-ssh
spec:
  containers:
    - name: main
      image: burgerdev.azurecr.io/burgerdev/contrast/debugshell@sha256:c218531b154b3ff3011da6e16bfaa0304e92fe55ff2414f3d8af1a7f2e48c6ab
      imagePullPolicy: Always
      command: ["debugshell-host", "custom-mr", "coordinator-0", "cat", "/etc/hosts"]
      env:
      - name: TMPDIR
        value: /tmp
      volumeMounts:
        - name: ctr
          mountPath: /run/containerd/containerd.sock
        - name: sbs
          mountPath: /run/vc/sbs
      securityContext:
        privileged: true # for /dev/vsock
  volumes:
    - name: ctr
      hostPath:
        path: /var/run/k3s/containerd/containerd.sock
        type: Socket
    - name: sbs
      hostPath:
        path: /run/vc/sbs
        type: Directory
``` 

### Discussion

The documented workflow for using the debug shell already exposes the shell to the operator:

https://github.com/edgelesssys/contrast/blob/5ba870c47304b6ad4ad2849c18c62f9e17a3162a/docs/docs/howto/debugshell.md?plain=1#L18-L19

Exposing the SSH server over VSOCK would not change that materially.

~However, using the debugshell feature without the policy settings change is possible. In that case, the user can observe the host journal but not exec into the VM. This subset of functionality is broken by this PR.~

As @sespiros observed, the above is not true due to the debug console being exposed. Thus, no new exposure is added by this change.

Fixes CON-271


tools/debugshell/main.go
```diff
@@ -45,18 +46,37 @@ func main() {
...
+			log.Fatalf("Error listening on VSOCK: %v", err)
+		}
+		if err := s.Serve(l); err != nil {
+			log.Fatalf("Error: %v", err)
+		}
+	})
+
+	wg.Go(func() {
 		<-ctx.Done()
 		if err := s.Close(); err != nil {
```
charludo, line 81: Doesn't `s.Close()` make both `ListenAndServe` and `Serve` above return `ssh.ErrServerClosed`? -> we return error here -> the systemd service restarts `on-failure`. 

(Sorry, I know this is tangential and not really the meat of the PR, I was just wondering about the `fix waitgroup handling` commit changes)

tools/debugshell/main.go
```diff
@@ -45,18 +46,37 @@ func main() {
...
+		// k3s crictl inspectp | flatten
+		// $.[3]."status"."id" = "e2e110885c4cae0cfd15c16e9be4c1072dd023200be18c9ad5c2b61543483422"
+		// $.[3]."status"."metadata"."name" = "coordinator-0"
+		// $.[3]."status"."metadata"."namespace" = "custom-mr"
+
+		log.Printf("Starting debug shell server on VSOCK")
+		l, err := vsock.Listen(22, nil)
+		if err != nil {
+			log.Fatalf("Error listening on VSOCK: %v", err)
+		}
```
charludo, line 73: Is the vsock mandatory now? From the PR description, it read more as "in addition, optionally"

packages/by-name/scripts/debugshell-host/debugshell-host.sh
```diff
@@ -0,0 +1,30 @@
...
+  jq -rs '.[0].status.id')
+
+# URL format: vsock://$CID:1024
+cid=$(cat "/run/vc/sbs/$sandbox/persist.json" | jq -r .AgentState.URL | sed -e 's|vsock://||' -e 's|:.*$||')
+
+key=$(mktemp)
+trap 'rm -f "$key"' EXIT
+rm "$key"
+ssh-keygen -t ed25519 -f "$key" -C "" -N ""
+ssh -o "ProxyCommand=socat - VSOCK-CONNECT:$cid:22" -o StrictHostKeyChecking=false -i "$key" root@localhost "$@"
```
charludo:
```suggestion
key=$(mktemp -d)
trap 'rm -rf "$key"' EXIT
ssh-keygen -t ed25519 -f "$key/id_ed25519" -C "" -N ""
ssh -o "ProxyCommand=socat - VSOCK-CONNECT:$cid:22" -o StrictHostKeyChecking=false -i "$key/id_ed25519" root@localhost "$@"
```
IMO cleaner than the create -> immediate delete temp, and also cleans the pubkey

packages/by-name/scripts/debugshell-host/debugshell-host.sh
```diff
@@ -0,0 +1,30 @@
...
+namespace=$1
+podname=$2
+shift 2
+
+sandbox=$(crictl --config /dev/null --runtime-endpoint /run/containerd/containerd.sock inspectp |
+  jq '.[]' |
+  jq "select(.status.metadata.name == \"$podname\")" |
+  jq "select(.status.metadata.namespace == \"$namespace\")" |
+  jq 'select(.status.state == "SANDBOX_READY")' |
+  jq -rs '.[0].status.id')
```
charludo, line 21: Nit: If no sandbox matches, this prints `null` but does not error. We then fail at `cat /run/vc/sbs/null/persist.json`. The error is probably still understandable if one looks at the code, but still, could be nicer to fail here already :D

packages/by-name/scripts/debugshell-host/debugshell-host.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# URL format: vsock://$CID:1024
+cid=$(cat "/run/vc/sbs/$sandbox/persist.json" | jq -r .AgentState.URL | sed -e 's|vsock://||' -e 's|:.*$||')
```
charludo, line 24: same here, maybe check cid is numeric

packages/by-name/scripts/debugshell-host/debugshell-host.sh
```diff
@@ -0,0 +1,30 @@
...
+podname=$2
+shift 2
+
+sandbox=$(crictl --config /dev/null --runtime-endpoint /run/containerd/containerd.sock inspectp |
+  jq '.[]' |
+  jq "select(.status.metadata.name == \"$podname\")" |
+  jq "select(.status.metadata.namespace == \"$namespace\")" |
+  jq 'select(.status.state == "SANDBOX_READY")' |
+  jq -rs '.[0].status.id')
+
```
charludo: Also not sure if `jq` shouldn't rather be invoked just once here rahter than piped repeatedly (also a nit, of course)

packages/by-name/scripts/debugshell-host/debugshell-host.sh
charludo: Nit: the existing wrapper in the debugshell's package.nix uses `StrictHostKeyChecking=no` and `UserKnownHostsFile=/dev/null`, would prefer if those matched.

packages/by-name/scripts/debugshell-host/package.nix
charludo, line 1: This package is a bit strange, because it's not really a reusable package. due to the `/etc/passwd` and `/etc/group` files being property of the container, not the script.

How about we make this a bog-standard `writeShellApplication`, and in the containers.nix, the debugshell's `copytoRoot`gains a 
```nix
(
  pkgs.runCommand "debugshell-nss" { } ''
    mkdir -p $out/etc $out/tmp
    echo "root:x:0:0::/tmp:/bin/sh" >$out/etc/passwd
    echo "root:x:0:root" >$out/etc/group
  ''
)
```
?

charludo (commented):
Thank you! 🎉  I have a couple of comments, plus:

[Our current docs](https://github.com/edgelesssys/contrast/blob/main/docs/docs/howto/debugshell.md?plain=1#L16-L19) still describe a 2-step process for enabling the debug shell, but after this PR, step 2 alone seems to be sufficient, so that probably also should be updated here.

packages/containers.nix
```diff
@@ -146,12 +146,16 @@
...
       contrastPkgs.debugshell
+      contrastPkgs.scripts.debugshell-host
       openssh
       contrastPkgs.tdx-tools
     ];
     config = {
       Entrypoint = [ "/bin/debugshell" ];
       Cmd = [ "journalctl --no-tail --no-pager -f" ];
+      Env = [
+        "TMPDIR=/tmp"
```
sespiros, line 157: I think there are a bunch of failing CI because this breaks the policy for the sidecar. Why is this needed? `mktemp` defaults to `/tmp`, we can either drop or add PATH like the rest of the container images.

tools/debugshell/main.go
```diff
@@ -45,18 +46,37 @@ func main() {
...
-		defer wg.Done()
+		// Reachable at
+		// cat /run/vc/sbs/$SANDBOXID/persist.json  | jq -r .AgentState.URL
+		// vsock://4157183119:1024
+		// Sandboxid at
+		// k3s crictl inspectp | flatten
+		// $.[3]."status"."id" = "e2e110885c4cae0cfd15c16e9be4c1072dd023200be18c9ad5c2b61543483422"
+		// $.[3]."status"."metadata"."name" = "coordinator-0"
+		// $.[3]."status"."metadata"."namespace" = "custom-mr"
+
```
sespiros, line 68:
```suggestion
              // Reachable from the host: resolve the pod's sandbox ID, read the VM's CID
              // from the Kata shim state, then dial this port.
              //
              //   sandbox=$(crictl pods --namespace "$NS" --name "$POD" --state ready -q)
              //   cid=$(jq -r .AgentState.URL "/run/vc/sbs/$sandbox/persist.json" | sed -e 's|vsock://||' -e 's|:.*||')
              //   ssh -o ProxyCommand="socat - VSOCK-CONNECT:$cid:22" root@localhost
              //
              // This is what packages/by-name/scripts/debugshell-host automates.
```

tools/debugshell/main.go
```diff
@@ -45,18 +46,37 @@ func main() {
...
+		log.Printf("Starting debug shell server on VSOCK")
```
sespiros:
```suggestion
		log.Printf("Starting debug shell server on VSOCK port 22")
```

packages/by-name/scripts/debugshell-host/debugshell-host.sh
```diff
@@ -0,0 +1,30 @@
...
+if [[ $# -lt 2 ]]; then
+  echo "Usage: $0 <namespace> <pod> [args...]" >&2
+  exit 1
+fi
+
+namespace=$1
+podname=$2
+shift 2
+
+sandbox=$(crictl --config /dev/null --runtime-endpoint /run/containerd/containerd.sock inspectp |
```
sespiros:
```suggestion
sandbox=$(crictl --config /dev/null --runtime-endpoint "${CONTAINER_RUNTIME_ENDPOINT:-/run/containerd/containerd.sock}" inspectp |
```

`crictl` already reads `CONTAINER_RUNTIME_ENDPOINT`, but passing `--runtime-endpoint` explicitly overrides it, which is why the example pod has to remap the k3s socket with a hostPath. Defaulting to the env var keeps today's behaviour and lets k3s users just set the variable.

sespiros (commented):
I don't have any implementation comments but couldn't we build what we need on top of kata-agent's debug console? 

This is already exposed via vsock in port 1026 and configurable without modifying the runtime since https://github.com/edgelesssys/contrast/pull/2374.

```bash
debug=true just default openssl
ssh dev-machine
bash kata-debug-shell.sh openssl-sse/coordinator-0
```

The ssh one probably will be easier to work with but I am curious if you were aware of this. In that light, this is not a breaking change since you can already get a root shell without a policy change through the kata-agent debug console.

And follow-up on this, if idd ssh over vsock is the way to go, we could drop `debug_console` and `debug_console_vport` from the kernel cmdline completely?

burgerdev:
@sespiros 

> I don't have any implementation comments but couldn't we build what we need on top of kata-agent's debug console?
> 
> This is already exposed via vsock in port 1026 and configurable without modifying the runtime since #2374.
> 
> ```shell
> debug=true just default openssl
> ssh dev-machine
> bash kata-debug-shell.sh openssl-sse/coordinator-0
> ```
> 
> The ssh one probably will be easier to work with but I am curious if you were aware of this. In that light, this is not a breaking change since you can already get a root shell without a policy change through the kata-agent debug console.

Thanks for the reminder, since the console is already exposed in debug mode we are not opening an additional hole here - updated the PR body.

The mechanism here is intended for headless use (in scripts), which is very tedious to achieve with the interactive console.

> And follow-up on this, if idd ssh over vsock is the way to go, we could drop `debug_console` and `debug_console_vport` from the kernel cmdline completely?

We could, but if we keep it we can still get in if only one of the SSH daemon and the agent fails. Since it's not costly and debug-only anyway, should we just keep it?

packages/by-name/scripts/debugshell-host/debugshell-host.sh
burgerdev: You'd still need to mount the socket, though. I figured it would be easier to just change the mount source, but this does not hurt.

packages/by-name/scripts/debugshell-host/debugshell-host.sh
burgerdev: I did that for readability, but maybe I actually want a jq script here.

packages/by-name/scripts/debugshell-host/debugshell-host.sh
burgerdev: Nice, thanks!

packages/containers.nix
burgerdev: That's weird, it should be possible to declare that here. Although I agree that we don't need it.

tools/debugshell/main.go
burgerdev: This is a leftover, I meant to delete it. Would you prefer to leave a comment as you suggested?

tools/debugshell/main.go
burgerdev: Yes, mandatory. This would fail if VSOCK was not supported (in which case the agent would not work either), or if port 22 was already listened on (which we control). Is that acceptable?

tools/debugshell/main.go
burgerdev: The context expires on SIGINT and SIGTERM, so I'd say exiting non-zero is expected?

The waitgroup handling fix is about wrong usage of `wg.Go` (does not need `wg.Done` - https://pkg.go.dev/sync#example-WaitGroup).

davidweisse:
> This allows connecting to the VM without adding a sidecar container or requiring containers to start at all.

How is this supposed to work? In a normal workflow, adding `--insecure-debug-shell-access` automatically injects the debugshell container. Then again, for the policy test suite I may need to call genpolicy directly, in which case I can just add the `"contrast.insecure-debug" = "true"` to the initdata manually, which works.

tools/debugshell/main.go
charludo: Yep, sgtm

packages/by-name/scripts/debugshell-host/debugshell-host.sh
```diff
@@ -0,0 +1,29 @@
...
+  jq "select(.status.metadata.namespace == \"$namespace\")" |
+  jq 'select(.status.state == "SANDBOX_READY")' |
+  jq -rs '.[0].status.id')
+
+# URL format: vsock://$CID:1024
+cid=$(cat "/run/vc/sbs/$sandbox/persist.json" | jq -r .AgentState.URL | sed -e 's|vsock://||' -e 's|:.*$||')
+
+key=$(mktemp -d)
+trap 'rm -rf "$key"' EXIT
+ssh-keygen -t ed25519 -f "$key/id_ed25519" -C "" -N ""
```
davidweisse, line 28: When i directly execute a command like
```
debugshell-host custom-daw test cat /tmp/policy.jsonl
```
this pollutes the output with `Generating public/private ed25519 key pair...`.

tools/debugshell/main.go
sespiros: I think it's helpful to leave the comment.

sespiros:
> We could, but if we keep it we can still get in if only one of the SSH daemon and the agent fails. Since it's not costly and debug-only anyway, should we just keep it?

I am fine with keeping it, it's only on debug runtimes anyway and we wouldn't stray from upstream.

burgerdev:
Addressed all comments, manually checked that the debugshell-host works from the container like this:

```sh
# my dev machine
env debug=true just default custom
# dev server
docker run -it --rm --privileged --entrypoint debugshell-host -v /run/k3s/containerd:/run/containerd -v /run/vc:/run/vc:ro burgerdev.azurecr.io/burgerdev/contrast/debugshell@sha256:78ce6314ef1e460a72f53009d86c1464a4765af1e2fdc113ec1d0200017f6b84 custom-mr coordinator-0
```

packages/by-name/scripts/debugshell-host/find_sandbox_id.jq
```diff
@@ -0,0 +1,9 @@
+first(
+  .[]
+  | select(
+      .status.metadata.name == $podname
+      and .status.metadata.namespace == $namespace
+      and .status.state == "SANDBOX_READY"
+    )
+) // error("no SANDBOX_READY pod found for \($podname)/\($namespace)")
```
davidweisse, line 8:
```suggestion
) // error("no SANDBOX_READY pod found for \($namespace)/\($podname)")
```
...right?

packages/by-name/scripts/debugshell-host/package.nix
```diff
@@ -0,0 +1,31 @@
...
+# server running in a local Kata VM.
+# It needs access to the containerd socket at the default location.
+#
+# Usage: debugshell-host $NAMESPACE $POD [command...]
+writeShellApplication {
+  name = "debugshell-host";
+  text = ''
+    export JQ_FIND_SANDBOX_ID=${./find_sandbox_id.jq}
+  ''
+  + builtins.readFile ./debugshell-host.sh;
```
davidweisse, line 23:
```suggestion
  runtimeEnv = {
    JQ_FIND_SANDBOX_ID = "${./find_sandbox_id.jq}";
  };
  text = builtins.readFile ./debugshell-host.sh;
```
nit

packages/containers.nix
```diff
@@ -146,6 +146,15 @@
...
       coreutils
       contrastPkgs.debugshell
+      contrastPkgs.scripts.debugshell-host
+      (runCommand "debugshell-rootfs" { } ''
+        mkdir -p \
+          $out/etc \
+          $out/tmp
+
+          echo "root:x:0:0::/tmp:/bin/sh" > $out/etc/passwd
+          echo "root:x:0:root" > $out/etc/group
```
davidweisse, line 156:
```suggestion
        echo "root:x:0:0::/tmp:/bin/sh" > $out/etc/passwd
        echo "root:x:0:root" > $out/etc/group
```

davidweisse (approved):

packages/by-name/scripts/debugshell-host/debugshell-host.sh
```diff
@@ -0,0 +1,36 @@
...
+sandbox=$(crictl --config /dev/null --runtime-endpoint "${CONTAINER_RUNTIME_ENDPOINT:-/run/containerd/containerd.sock}" inspectp |
+  jq -r -f "$JQ_FIND_SANDBOX_ID" --arg namespace "$namespace" --arg podname "$podname")
+
+# URL format: vsock://$CID:1024
+cid=$(cat "/run/vc/sbs/$sandbox/persist.json" | jq -r .AgentState.URL | sed -e 's|vsock://||' -e 's|:.*$||')
+
+if ! echo -n "$cid" | grep -qE "^[0-9]+$"; then
+  # The above pipeline succeeded, so running only parts of it should succeed, too.
+  echo "Could not parse numeric CID from agent URL $(jq -r .AgentState.URL <"/run/vc/sbs/$sandbox/persist.json")" >&2
+  exit 2
```
sespiros, line 25:
```suggestion
# URL format: vsock://$CID:1024
url=$(jq -r .AgentState.URL "/run/vc/sbs/$sandbox/persist.json")
cid=${url#vsock://}
cid=${cid%%:*}

if [[ ! $cid =~ ^[0-9]+$ ]]; then
  echo "Could not parse numeric CID from agent URL $url" >&2
  exit 2
fi
```

sespiros (approved):
Thanks. The label still says `breaking change`, probably it should be feature so that the release notes get generated correctly. Also the text still says "connecting to the VM without adding a sidecar container" which is not true through `generate` iiuc/users always get the sidecar because they only know about `--insecure-enable-debug-shell-access`.

There are a few docs changes needed either in this PR or a follow-up. If it's a follow-up let's at least track them in a ticket so that the docs don't go silently very stale:

**`docs/docs/howto/debugshell.md`**

- Line 5 says the service is "reachable from within the pod sandbox". It's reachable from the node now: vsock 22 after this PR, and vsock 1026 since #2374.
- Lines 16-19 present the policy change and the flag as two required steps. The policy change is only needed for `kubectl exec` into the sidecar. The flag on its own already gives the VM journal via `kubectl logs` (the "Collect pod VM logs" section says as much) and, since #2374, a root shell from the node.
- The architecture diagram shows the sidecar-only path.
- The danger callout could be sharper: with the flag set, anything on the node that can open an `AF_VSOCK` socket gets root in the VM, with no policy involved.

**`dev-docs/serial-console.md`**

- The premise is that host shell access needs `set=debug` or `withDebug`. That stopped being true for `kata-debug-shell.sh` with #2374: it works on a release runtime with just the initdata flag. Only the actual serial console (`console=ttyS0` plus the autologin getty from `debug.nix`) still needs the debug runtime.
- Line 17 points at `packages/kata-debug-shell/kata-debug-shell.sh`, which doesn't exist. It's `packages/by-name/kata-debug-shell/`.

PR #2621, opened 2026-09-01
cli: don't add wildcard SAN

burgerdev:
The plain wildcard SAN `*` can't be used with the majority of TLS implementations, which makes it ineffective for use outside the cluster where standard implementations (in particular, browsers) are usually used. Inside the cluster, where more permissive TLS libraries may be in use, its presence does more harm than good, considering that workloads may impersonate each other. Thus, don't add a plain `*` as SAN anymore.

Fixes CON-222.

### DIFF
--- a/docs/docs/architecture/components/manifest.md
+++ b/docs/docs/architecture/components/manifest.md
@@ -56,9 +56,7 @@ Allowed values are:
 - URIs
 - DNS names
 
-By default, the Contrast CLI will add two SANs for each workload on generate: The pod name as DNS name under which the pod can be reached inside the cluster,
-and a wildcard DNS name `*` to allow the certificate to be used with any other hostname (for example an external load balancer).
-As DNS is untrusted in the context of Contrast, issuing a wildcard certificate won't weaken the security of your workload.
+The Contrast CLI adds the workload resource name (for example, the `Deployment` name) to the manifest as a DNS SAN.
 In addition, the Coordinator will add the pod IP address to the certificate.
 
 If the pod is exposed via a different IP than the pod IP, for example a load balancer, and you want to include that IP in the certificate for verification,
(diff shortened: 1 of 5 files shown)


docs/docs/architecture/components/manifest.md
```diff
@@ -56,9 +56,7 @@ Allowed values are:
 - URIs
 - DNS names
 
-By default, the Contrast CLI will add two SANs for each workload on generate: The pod name as DNS name under which the pod can be reached inside the cluster,
-and a wildcard DNS name `*` to allow the certificate to be used with any other hostname (for example an external load balancer).
-As DNS is untrusted in the context of Contrast, issuing a wildcard certificate won't weaken the security of your workload.
+The Contrast CLI adds the pod controller name (for example, the `Deployment` name) to the manifest.
```
sespiros:
```suggestion
The Contrast CLI adds the workload resource name (for example, the `Deployment` name) to the manifest as a DNS SAN.
```

sespiros (approved):
Thanks, I had _this commit_ as part of some other forgotten branch with "improvements" somewhere.

A couple of more stale docs/tests:
- https://github.com/edgelesssys/contrast/blob/5ba870c47304b6ad4ad2849c18c62f9e17a3162a/docs/docs/getting-started/deployment.md?plain=1#L563
- https://github.com/edgelesssys/contrast/blob/5ba870c47304b6ad4ad2849c18c62f9e17a3162a/docs/docs/howto/workload-deployment/workload-communication.md?plain=1#L37
- https://github.com/edgelesssys/contrast/blob/5ba870c47304b6ad4ad2849c18c62f9e17a3162a/internal/manifest/manifest_test.go#L24

burgerdev:
Addressed, thanks!

PR #2623, opened 2026-09-01
.github: allow both minor and patch releases in nightly pipeline

charludo:
This removes the "manual" release pipeline. The nightly one was manually triggerable already anyways, so the only real differentiator was that the "old" pipeline could handle patch releases.

Now the nightly pipeline:
- does the minor draft release flow every night, unchanged to how it currently is;
- but if (up to 2 minor versions old) version branches received updates, a draft patch release is also created (i.e. the entire pipeline is excercised from the release branch rather than main)
- the next morning, we can promote any of the draft releases

So *in theory*, in the future, if we want to release a patch release (and a minor release), we:
- cherry-pick the wanted changes onto the release branch(es) and push
- wait for the night
- promote all of them

I'm sure this is going to go super smooth, first try!! 👀


.github/workflows/release_nightly.yml
```diff
@@ -193,3 +249,119 @@ jobs:
...
+          QUEUE: ${{ steps.decide.outputs.queue || steps.input-queue.outputs.queue }}
+        run: |
+          next=$(jq -r '.[0] // empty' <<<"${QUEUE}")
+          if [[ -z "${next}" ]]; then
+            echo "No further patch nightlies to dispatch."
+            exit 0
+          fi
+          rest=$(jq -c '.[1:]' <<<"${QUEUE}")
+          echo "Dispatching patch nightly on ${next} (queue after it: ${rest})."
+          gh workflow run release_nightly.yml --repo "${GH_REPO}" --ref "${next}" \
```
sespiros, line 371: I haven't run this but if next nightly dispatches another nightly for an older release, these would have the old version of `release_nightly.yml` so they would fail. I think this won't be a problem anymore once we do the first release with this included but until then if we need to do a patch release for 1.24, we would have a broken workflow.

.github/workflows/release_nightly.yml
```diff
@@ -193,3 +249,119 @@ jobs:
...
+        with:
+          persist-credentials: false
+          fetch-tags: true
+          fetch-depth: 0
+      - name: Decide patch nightlies
+        if: github.ref_name == 'main'
+        id: decide
+        env:
+          # How many release branches below the current minor are considered for patch nightlies.
+          DEPTH: ${{ vars.PATCH_NIGHTLY_MINOR_DEPTH || 2 }}
```
sespiros: why 2? SECURITY.md supports only latest minor.

.github/workflows/release_nightly.yml
```diff
@@ -123,6 +177,8 @@ jobs:
   nightly:
     name: "release-requirement: e2e nightly"
     uses: ./.github/workflows/e2e_nightly.yml
```
sespiros, line 180: So now if this runs on a release branch, that means that the `bm_maintenance_platform.yml` workflow will run which might bring the clusters to older versions i.e the gpu-operator for example. We can have a `skip-maintenance` boolean input on e2e_nightly.yml and have it set for KIND==patch

.github/workflows/release_promote.yml
```diff
@@ -275,16 +309,38 @@ jobs:
...
+    env:
+      GH_TOKEN: ${{ github.token }}
+      GH_REPO: ${{ github.repository }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           persist-credentials: true
           # edgelessci's token is required to push to the protected release/* branch.
           token: ${{ secrets.NUNKI_CI_COMMIT_PUSH_PR }}
+      - name: Verify nightly ref is on release branch
```
sespiros: This runs after the approval gate and after promote, retag-images and prepare-post-release-pr already created the draft, the public ghcr tags and the PR, so a failure leaves all three to clean up by hand. Move it to `prepare` next to the other checks. 

Also it can only fail if `release/*` was rewound, which branch protection blocks, so dropping it works too.

.github/workflows/release_nightly.yml
```diff
@@ -3,15 +3,31 @@ name: release nightly
...
+      queue:
+        description: JSON array of release branches to dispatch patch nightlies for after this run. Normally only set by chained dispatches.
+        type: string
+        required: false
+        default: "[]"
+      dispatch_next:
+        description: Dispatch follow-up patch nightlies after this run (decided on main, dequeued from `queue` on release branches).
+        type: boolean
+        required: false
+        default: true
```
sespiros: Default this to false: scheduled runs are covered by `event_name == 'schedule'` and the chain can pass `-f dispatch_next=true` in the dispatch step. Otherwise every manual re-run from main also kicks off the patch nightlies.

.github/workflows/release_nightly.yml
```diff
@@ -193,3 +249,119 @@ jobs:
...
+            without_v="${latest#v}"
+            next_patch="v${without_v%.*}.$((${without_v##*.} + 1))"
+            valid_bases+=("${next_patch}")
+            if [[ "${head_sha}" == "$(git rev-parse "${latest}^{commit}")" ]]; then
+              echo "${branch}: no unreleased changes since ${latest}."
+              continue
+            fi
+            # Skip if the last successful nightly on this branch already built this exact commit and its draft is still around.
+            last_good=$(gh run list --repo "${GH_REPO}" --workflow=release_nightly.yml \
+              --branch "${branch}" --status success --limit 1 --json headSha --jq '.[0].headSha // empty')
```
sespiros: Decide skips a branch when the latest successful run built the current sha, but promote takes the latest *completed* run. A chained run cancelled while pending (the next schedule entering the concurrency group does that) sits between the two: promote refuses it and Decide won't re-dispatch, so the branch can't be promoted until someone runs a nightly on it by hand. Make Decide look at the latest completed run and skip only if it succeeded, so the next night self-heals.

.github/workflows/release_nightly.yml
```diff
@@ -3,15 +3,31 @@ name: release nightly
...
+        type: string
+        required: false
+        default: "[]"
+      dispatch_next:
+        description: Dispatch follow-up patch nightlies after this run (decided on main, dequeued from `queue` on release branches).
+        type: boolean
+        required: false
+        default: true
+      notify:
+        description: Notify the Teams channel about e2e failures. Set by chained dispatches, where nobody watches the run.
```
sespiros, line 19: The build, release-e2e and dispatch-next jobs here don't notify, only e2e_nightly's do. For chained runs "where nobody watches" a failure in them only surfaces when promote refuses the run, and a broken dispatch surfaces as a missing draft. A notify-teams job here with needs on those jobs would close that.

.github/workflows/release_promote.yml
sespiros: This is not touched by this pr but for a patch, image-replacements comes from the release branch's release_artifacts, but retag-images' `count != 6` is kept in sync with main's list. Fine today (both branches build six), breaks the first time the list changes on main only. `> 0` or a comment on the check would do.

sespiros (changes requested):
Nice cleanup overall. I'm a bit anxious about changing the release pipeline right before your OOO given the corner cases below, so could we either merge it when you're back, or do the backport to release/v1.24 and one manually triggered chained run before you leave?

Blocking for me: the maintenance skip and dev-docs/release.md, which still describes release.yml. The rest are suggestions.

Also could you split the release.yml removal (plus the parse_version/release_build_linux cleanup) from the nightly/promote rework? We rebase-merge, so the rework would be a readable commit on its own.

charludo:
Thanks for the review! I know this sucks to review 😓 

> dev-docs/release.md

Ah, crap! Yes, definitely needed

> Also could you split the release.yml removal (plus the parse_version/release_build_linux cleanup) from the nightly/promote rework?

Yep sounds good, will do.

I'll push fixups for your comments today, but yes, we don't need to merge this "now". I guess the best way to truly test this would be to merge this before the next proper minor release that's not time-critical, then immediately push some small bug fix and do a patch release...

.github/workflows/release_nightly.yml
charludo: > SECURITY.md supports only latest minor

Ah, neat, did not know that. Yeah makes sense, will limit to 1. 2 was just chosen arbitrarily.

.github/workflows/release_nightly.yml
charludo: Ohhhhhh good catch, thank you. 

Actually, this might be a problem... Though I suppose it always was when needing to do a patch release.

.github/workflows/release_nightly.yml
charludo: Yeah good point. IDK about backporting this... Esp since the TDX runner is still unavailable. Merge in 3 weeks sounds better to me 😅

.github/workflows/release_nightly.yml
charludo: done

.github/workflows/release_nightly.yml
charludo: done, I guess it might occasionally break, but not worse than before.

.github/workflows/release_promote.yml
charludo: done

.github/workflows/release_nightly.yml
charludo: done, good idea

.github/workflows/release_nightly.yml
charludo: Oh yeah right. Thanks. Done.

.github/workflows/release_nightly.yml
charludo: done

.github/workflows/release_promote.yml
charludo: Yeah seems like a good tradeoff. Done.

charludo:
@sespiros comments should be addressed. Sorry for force-pushing, that's due to the splitting in two commits.

.github/workflows/release_nightly.yml
```diff
@@ -193,3 +249,152 @@ jobs:
...
+          fi
+          rest=$(jq -c '.[1:]' <<<"${QUEUE}")
+          echo "Dispatching patch nightly on ${next} (queue after it: ${rest})."
+          gh workflow run release_nightly.yml --repo "${GH_REPO}" --ref "${next}" \
+            -f queue="${rest}" -f notify=true -f dispatch_next=true
+
+  notify-teams:
+    name: Notify teams channel of failure
+    runs-on: ubuntu-24.04
+    needs: [process-inputs, cleanup-nightly-drafts, release-x86_64-linux, release-aarch64-darwin, nightly, test, dispatch-next]
```
sespiros: e2e_nightly.yml already sends a card when e2e fails, so this job sends a second one for the same failure. Drop nightly from here.

Also, e2e_nightly.yml turns its notify off completely for patch runs (the !inputs.skip-maintenance). It should only stop reporting the maintenance jobs, not the e2e ones.

dev-docs/release.md
```diff
@@ -76,47 +39,47 @@ If you need to include new changes merged into main since the last successful ni
...
 
-    ```sh
-    export REL_VER=v0.1.1
-    export CUR_VER="$(echo $REL_VER | awk -F. -v OFS=. '{$NF -= 1 ; print}')"
-    echo "Releasing $CUR_VER -> $REL_VER"
-    ```
+1. Ensure all needed PRs were backported to the current release branch, and all backport PRs were merged.
 
-3. Checkout the current release branch:
+2. Wait for the next night, or trigger the patch nightly yourself as described above.
```
sespiros: s/above/below :P

.github/workflows/release_nightly.yml
sespiros: Worth a comment here: the dispatched run uses the release branch's copy of this workflow, e2e_nightly.yml and the actions, so changes to those only reach patch nightlies once backported.

.github/workflows/release_nightly.yml
```diff
@@ -159,8 +217,6 @@ jobs:
       fail-fast: false
     name: "release-requirement: e2e release on ${{ matrix.platform.name }} (${{ matrix.cli-arch }})"
     runs-on: ${{ matrix.cli-arch == 'aarch64-darwin' && 'macos-latest' || matrix.platform.linux-runner }}
-    # Serialize matrix entries that share a cluster (linux + darwin) to
-    # avoid clashing on coordinator state and resource cleanup.
     concurrency:
       group: e2e-release-${{ matrix.platform.name }}-${{ github.run_id }}
```
sespiros: I think it's worth keeping the comment here.

sespiros, line 223: I think it's worth keeping the comment here.

sespiros (approved):
Thanks, just a few more minor things. Let's mark this as do not merge and mention in the PR description how we are going to merge this.

.github/workflows/release_nightly.yml
charludo: Yup my bad

dev-docs/release.md
charludo: Indeed :D

.github/workflows/release_nightly.yml
charludo: Right. Done.

charludo:
@sespiros done. LMK if you want to re-re-review, or if it's OK to merge :D

PR #2630, opened 2026-09-03
packages/cleanup-namespaces: conservatively skip namespace on kubectl error

thomasten:
The kubectl command to get the `keep` label may fail. In this case, be conservative and skip the deletion of the ns.


burgerdev (approved):
Thanks, lgtm!

PR #2633, opened 2026-09-04
ci: auto-fix govulncheck findings

charludo:
When `govulncheck` fails due to found errors, a new CI step and script runs which attempts to autofix the findings, *on main*. I.e. after pushing to a feature branch and failing govulncheck there, we check if it also fails on main, and if so, if no existing autofix PR is open.

Only then do we open a bot PR which, depending on what's required,
- bumps imported go module versions,
- updates the go patch version,
- runs our `codegen` script

This will *usually* work just fine. However, in some cases (e.g. bump of the required minimum go version inside a module), it might fail because additional, not automate-able changes are required. The PR will still be opened, but needs human fixups.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -71,6 +71,90 @@ jobs:
         run: |
           nix run ".#${SET}.scripts.govulncheck"
 
+  govulncheck-autofix:
+    name: "Fix vulnerabilities and open a PR"
+    needs: govulncheck
+    if: ${{ failure() && !github.event.pull_request.head.repo.fork }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 60
+    permissions:
+      contents: read
+      pull-requests: read
+    concurrency:
+      group: govulncheck-autofix
+      cancel-in-progress: false
+    env:
+      SET: base
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: main
+          token: ${{ secrets.NUNKI_CI_COMMIT_PUSH_PR }}
+          persist-credentials: true
+      - name: Check for an existing autofix PR
+        id: existing
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          number=$(gh pr list --head automated/govulncheck --state open --json number --jq '.[0].number // empty')
+          if [[ -n "${number}" ]]; then
+            echo "PR #${number} is already open, leaving it alone." | tee -a "$GITHUB_STEP_SUMMARY"
+          fi
+          echo "number=${number}" >> "$GITHUB_OUTPUT"
+      - uses: ./.github/actions/setup_nix
+        if: ${{ steps.existing.outputs.number == '' }}
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          cachixToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - name: Apply the mechanical fixes
+        id: fix
+        if: ${{ steps.existing.outputs.number == '' }}
+        run: |
+          set +e
+          nix run ".#${SET}.scripts.govulncheck-fix" > "${RUNNER_TEMP}/report.md"
+          echo "status=$?" >> "$GITHUB_OUTPUT"
+      - name: Build the PR body
+        if: ${{ steps.existing.outputs.number == '' }}
+        run: |
+          {
+            echo "Automated update created by the [static checks workflow](https://github.com/edgelesssys/contrast/blob/main/.github/workflows/static.yml)."
+            echo
+            cat "${RUNNER_TEMP}/report.md" 2>/dev/null || echo "No report produced."
+            echo
+            echo "\`generate\` ran after the fixes, so this PR also contains the follow-up changes it makes: \`go mod tidy\`, vendor hashes, the \`go\` directive across all modules and \`modernize\` rewrites. A dependency that requires a newer Go moves the \`go\` directive, which can touch far more files than the table above lists."
+          } > "${RUNNER_TEMP}/body.md"
+          cat "${RUNNER_TEMP}/body.md" >> "$GITHUB_STEP_SUMMARY"
+      - name: Create PR
+        if: ${{ steps.existing.outputs.number == '' && contains(fromJSON('["0", "4"]'), steps.fix.outputs.status) }}
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8.1.1
+        with:
+          title: "deps: fix vulnerabilities reported by govulncheck"
+          body-path: ${{ runner.temp }}/body.md
+          commit-message: "deps: fix vulnerabilities reported by govulncheck"
+          base: main
+          draft: false
+          labels: "dependencies"
+          branch: automated/govulncheck
+          delete-branch: true
+          committer: edgelessci <edgelessci@users.noreply.github.com>
+          author: edgelessci <edgelessci@users.noreply.github.com>
+          token: ${{ secrets.NUNKI_CI_COMMIT_PUSH_PR }}
+      - name: Notify teams channel about findings that need a human
+        if: ${{ steps.fix.outputs.status == '4' }}
+        uses: ./.github/actions/post_to_teams
+        with:
+          webhook: ${{ secrets.TEAMS_CI_WEBHOOK }}
+          title: "govulncheck found vulnerabilities that cannot be fixed automatically"
+          message: "Check the run summary for the list."
+      - name: Notify teams channel about the autofix failing
+        if: ${{ steps.fix.outputs.status != '' && steps.fix.outputs.status != '0' && steps.fix.outputs.status != '4' }}
+        uses: ./.github/actions/post_to_teams
+        with:
+          webhook: ${{ secrets.TEAMS_CI_WEBHOOK }}
+          title: "The govulncheck autofix failed"
+          message: "No PR was opened, the vulnerabilities are still unfixed. Check the run for the error."
+
   golangci-lint:
     runs-on: ubuntu-24.04
     timeout-minutes: 15
--- a/packages/govulncheck-fix.sh
+++ b/packages/govulncheck-fix.sh
@@ -1,0 +1,223 @@
+#!/usr/bin/env bash
+# Copyright 2026 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+set -euo pipefail
+
+exec 3>&1 1>&2
+
+readonly overlay="overlays/nixpkgs.nix"
+readonly goAttr="legacyPackages.x86_64-linux.base.nixpkgs.go"
+
+readonly sep=$'\x1f'
+
+readonly exitNeedsHuman=4
+
+dryRun=false
+if [[ ${1:-} == "--dry-run" ]]; then
+  dryRun=true
+elif [[ $# -gt 0 ]]; then
+  echo "usage: govulncheck-fix [--dry-run]"
+  exit 2
+fi
+
+workdir=$(mktemp -d)
+trap 'rm -rf "${workdir}"' EXIT
+
+tags="${GOVULNCHECK_TAGS}"
+
+# Skips are (osvs, module, current, reason) and actions are (kind, dir, module, current, fixed, osvs).
+skip() { printf '%s\n' "$1${sep}$2${sep}$3${sep}$4" >>"${workdir}/skips"; }
+
+relative() {
+  if [[ $1 == "${PWD}" ]]; then
+    printf '.'
+  else
+    printf '%s' "${1#"${PWD}"/}"
+  fi
+}
+
+: >"${workdir}/findings.jsonl"
+while IFS= read -r dir; do
+  echo "  govulncheck ${dir}"
+  CGO_ENABLED=0 govulncheck -C "${dir}" -tags "${tags}" -format json ./... |
+    jq -c --arg dir "${dir}" 'select(.finding) | .finding + {dir: $dir}' >>"${workdir}/findings.jsonl"
+done < <(go list -f '{{.Dir}}' -m)
+
+if [[ ! -s ${workdir}/findings.jsonl ]]; then
+  echo "No vulnerabilities found."
+  exit 0
+fi
+
+# A single OSV yields module-level, package-level and symbol-level.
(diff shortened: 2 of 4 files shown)


packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+        { print }
+        /^  # }\);$/ && !done {
+          print ""
+          print "  # Remove this once nixpkgs has caught up."
+          print "  " attr " = prev." attr ".overrideAttrs ("
+          print "    finalAttrs: _prevAttrs: {"
+          print "      version = \"0\";"
+          print "      src = final.fetchurl {"
+          print "        url = \"https://go.dev/dl/go${finalAttrs.version}.src.tar.gz\";"
+          print "        hash = \"\";"
```
sespiros: `fetchurl` maps `hash = ""` to `lib.fakeHash`, so nix-update looks for `sha256-AAAA…=` in the file and never finds it: the version is replaced, the hash stays `""`, and the resulting go override cannot verify. Print the fake hash literal in the template:

```awk
print "        hash = \"sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=\";"
```

Repro: delete the existing `go_1_26` override from `overlays/nixpkgs.nix`, insert the template, then

```bash
nix-update --flake --version=1.26.6 --override-filename=overlays/nixpkgs.nix legacyPackages.x86_64-linux.base.nixpkgs.go
git diff overlays/nixpkgs.nix
```

With `""` only the version line changes, with the fake hash literal the version and hash lines match the `go_1_26` override on main (checked here with the aarch64-darwin attribute, same overlay). The test runs on this branch (#2634, #2635) only went through `go get`, so this is the cheapest check of the toolchain path.

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+# A single OSV yields module-level, package-level and symbol-level.
+# Only the symbol-level one (trace[0].function set) means we call the code path.
+jq -s -r '
+  [ .[] | select(.trace[0].function != null) ]
+  | group_by([ .dir, .trace[0].module ])
+  | map([
+      .[0].dir,
+      .[0].trace[0].module,
+      .[0].trace[0].version,
+      ([ .[] | .fixed_version // empty ] | sort | last // ""),
```
sespiros: `sort | last` compares strings, so two advisories in one module fixed in v1.9.x and v1.10.x bump to the lower one and the report lists both as fixed:

```console
$ jq -n '["v1.9.0","v1.10.0"] | sort | last'
"v1.9.0"
$ jq -n '["go1.26.9","go1.26.10"] | sort | last'
"go1.26.9"
```

Emit all fixed versions from jq and pick with `sort -V | tail -n1`, as the stdlib branch already does.

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+
+tags="${GOVULNCHECK_TAGS:-contrast_unstable_api}"
+
+# Skips are (osvs, module, current, reason) and actions are (kind, dir, module, current, fixed, osvs).
+skip() { printf '%s\n' "$1${sep}$2${sep}$3${sep}$4" >>"${workdir}/skips"; }
+
+: >"${workdir}/findings.jsonl"
+while IFS= read -r dir; do
+  echo "  govulncheck ${dir}" >&2
+  { CGO_ENABLED=0 govulncheck -C "${dir}" -tags "${tags}" -format json ./... || true; } |
```
sespiros: In `-format json` govulncheck exits 0 with findings (exit 3 only comes from the text handler), so `|| true` only hides build and vulndb errors, which then read as "No vulnerabilities found." and exit 0. Drop the `|| true` and let pipefail surface them.

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+  if [[ -z ${fixed} ]]; then
+    continue
+  fi
+
+  if [[ ${module} == "stdlib" || ${module} == "toolchain" ]]; then
+    printf '%s\n' "${osvs}${sep}${current}${sep}${fixed}" >>"${workdir}/stdlib"
+    continue
+  fi
+
+  if [[ ${current%%.*} != "${fixed%%.*}" ]]; then
```
sespiros: v0 and v1 share the import path, so `go get mod@v1.x` from v0.x works and only v2+ needs the `/vN` path:

```console
$ go get github.com/spf13/cobra@v0.0.5
$ go get github.com/spf13/cobra@v1.0.0
go: upgraded github.com/spf13/cobra v0.0.5 => v1.0.0
```

Skip only when `${fixed%%.*}` is neither v0 nor v1 and differs from `${current%%.*}`.

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+      skip "${osvs}" "${module}" "${current}" \
+        "\`go get ${module}@${fixed}\` failed, may need a toolchain bump first"
+      continue
+    fi
+  fi
+  printf '%s\n' "${kind}${sep}${dir}${sep}${module}${sep}${current}${sep}${fixed}${sep}${osvs}" \
+    >>"${workdir}/applied"
+done <"${workdir}/ordered"
+
+writeReport
```
sespiros: `writeReport` runs before `generate`, so when generate fails the go.mod bumps are already in the tree, the report says "Fixed automatically", create-pull-request opens the PR, and Teams gets "cannot be fixed automatically". Run `generate` before writing the report and exit with a distinct code for failures so the workflow can skip the PR and word the notification correctly.

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+  dryRun=true
+elif [[ $# -gt 0 ]]; then
+  echo "usage: govulncheck-fix [--dry-run]" >&2
+  exit 2
+fi
+
+workdir=$(mktemp -d)
+trap 'rm -rf "${workdir}"' EXIT
+
+tags="${GOVULNCHECK_TAGS:-contrast_unstable_api}"
```
sespiros: The check derives tags from `contrastPkgs.contrast.contrast.tags`, the fixer hardcodes `contrast_unstable_api`, so they can drift. Pass them the same way:

```nix
runtimeEnv.GOVULNCHECK_TAGS = lib.concatStringsSep "," contrastPkgs.contrast.contrast.tags;
```

and drop the default in the script.

.github/workflows/static.yml
```diff
@@ -71,6 +71,82 @@ jobs:
...
+      - name: Apply the mechanical fixes
+        id: fix
+        if: ${{ steps.existing.outputs.number == '' }}
+        run: |
+          set +e
+          nix run ".#${SET}.scripts.govulncheck-fix" > "${RUNNER_TEMP}/report.md"
+          echo "status=$?" >> "$GITHUB_OUTPUT"
+      - name: Build the PR body
+        if: ${{ steps.existing.outputs.number == '' }}
+        run: |
```
sespiros, line 119: #2634 touched 20 files (go directive 1.25.6 to 1.26.0 everywhere, `toPtr` to `new(...)` in five Go files) while the table names one module. That is `generate` doing its job, but the bot PR should say so: add a line to the body that `generate` ran (tidy, vendorHash, go directive sync, modernize) and can move the go directive when a dependency requires a newer Go.

.github/workflows/static.yml
```diff
@@ -71,6 +71,82 @@ jobs:
         run: |
           nix run ".#${SET}.scripts.govulncheck"
 
+  govulncheck-autofix:
```
sespiros, line 74: static.yml only runs on push to main and PRs, so a new advisory waits for traffic. Intended, or do we want a schedule on the govulncheck job?

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+  awk -F"${sep}" '$1 != "stdlib"' "${workdir}/actions"
+} >"${workdir}/ordered"
+
+while IFS="${sep}" read -r kind dir module current fixed osvs; do
+  if [[ ${kind} == stdlib ]]; then
+    echo "Pinning go to ${fixed} in ${overlay}" >&2
+    if ! grep -q "${goPin} = prev.${goPin}.overrideAttrs" "${overlay}"; then
+      awk -v attr="${goPin}" '
+        { print }
+        /^  # }\);$/ && !done {
```
sespiros, line 178: The insert is a silent no-op if the `  # });` marker line changes, and nix-update then runs against a file without the pin. Assert it after the awk:

```bash
grep -q "${goPin} = prev.${goPin}.overrideAttrs" "${overlay}"
```

(spyros: this seems very nitty but posting regardless)

.github/workflows/static.yml
```diff
@@ -71,6 +71,82 @@ jobs:
...
+          author: edgelessci <edgelessci@users.noreply.github.com>
+          token: ${{ secrets.NUNKI_CI_COMMIT_PUSH_PR }}
+      - name: Notify teams channel about findings that need a human
+        if: ${{ steps.fix.outputs.status != '0' && steps.fix.outputs.status != '' }}
+        uses: ./.github/actions/post_to_teams
+        with:
+          webhook: ${{ secrets.TEAMS_CI_WEBHOOK }}
+          title: "govulncheck found vulnerabilities that cannot be fixed automatically"
+          message: "Check the run summary for the list."
+          additionalActions: '[{"type": "Action.OpenUrl", "title": "Run", "url": "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"}]'
```
sespiros: `post_to_teams` already adds a "GitHub workflow run" button with this URL, drop `additionalActions`.

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+  if [[ -s ${workdir}/applied ]]; then
+    printf '### Fixed automatically\n\n'
+    printf '| Advisory | Module | From | To | Where |\n|---|---|---|---|---|\n'
+    while IFS="${sep}" read -r kind dir module current fixed osvs; do
+      if [[ ${kind} == stdlib ]]; then
+        printf '| %s | go (toolchain) | %s | %s | %s |\n' \
+          "${osvs}" "${current}" "${fixed}" "${overlay}"
+      else
+        printf '| %s | %s | %s | %s | %s |\n' \
+          "${osvs}" "${module}" "${current}" "${fixed}" "${dir}"
```
sespiros: The "Where" column prints `/home/runner/work/contrast/contrast/tools/debugshell`. Use `${dir#"$PWD/"}`.

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env bash
+# Copyright 2026 Edgeless Systems GmbH
+# SPDX-License-Identifier: BUSL-1.1
+
+set -euo pipefail
+
+exec 3>&1 1>&2
```
sespiros, line 7: After `exec 3>&1 1>&2` every `>&2` in the script is a no-op, drop them.

packages/govulncheck-fix.sh
```diff
@@ -0,0 +1,203 @@
...
+  writeReport
+  echo "Nothing to fix automatically." >&2
+  if [[ -s ${workdir}/skips ]]; then
+    exit 1
+  fi
+  exit 0
+fi
+
+if [[ ${dryRun} == true ]]; then
+  cp "${workdir}/actions" "${workdir}/applied"
```
sespiros, line 162: Dry run prints "### Fixed automatically" for changes it did not make. Use "Would fix automatically" when `dryRun` is set.

(spyros: nittiest of nits, feel free to ignore if it doesn't bother you :D)

.github/workflows/static.yml
```diff
@@ -71,6 +71,82 @@ jobs:
         run: |
           nix run ".#${SET}.scripts.govulncheck"
 
+  govulncheck-autofix:
+    name: "Fix vulnerabilities and open a PR"
+    needs: govulncheck
+    if: ${{ failure() && github.event.pull_request.head.repo.fork != true }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 90
```
sespiros: The generate job does the same work with 60 minutes and the recorded run took 6.5, align to 60.

(spyros: also feels too nitty)

.github/workflows/static.yml
```diff
@@ -71,6 +71,82 @@ jobs:
         run: |
           nix run ".#${SET}.scripts.govulncheck"
 
+  govulncheck-autofix:
+    name: "Fix vulnerabilities and open a PR"
+    needs: govulncheck
+    if: ${{ failure() && github.event.pull_request.head.repo.fork != true }}
```
sespiros: The rest of the file uses `!github.event.pull_request.head.repo.fork`, use the same form here.

sespiros (changes requested):
AI review came back with too many comments, I did a pass over them and posting (mostly) as they are (here and inline):

- Drop `wip! test on PR branch` before merge, it points the checkout at `ch/auto-govulnfix`. The do-not-merge check does not match `wip!`, so nothing blocks it.
- Add a commit body and PR description with the design choices: always fixes `main` regardless of the triggering ref, never touches an existing autofix PR, exit 1 means "needs a human". `ci:` fits the diff better than `.github:` since most of it is `packages/` and `justfile`.

.github/workflows/static.yml
charludo: Intended, since we're also not running `govulncheck` on a scedule afaik. At least in my experience these findings also happen in the midst of the working day, when we are pushing anyways, so not much gained by nightly runs.

packages/govulncheck-fix.sh
charludo: done

packages/govulncheck-fix.sh
charludo: done, thanks for noticing

packages/govulncheck-fix.sh
charludo: oh, nice. moar things working automatically :D

packages/govulncheck-fix.sh
charludo: good idea, thx

.github/workflows/static.yml
charludo: added a line.

packages/govulncheck-fix.sh
charludo: well now that you've mentioned it, I can't unsee it... :D

packages/govulncheck-fix.sh
charludo: Oh shoot,  indeed. Fixed. Thanks for noticing.

charludo:
Thanks for the review @sespiros! SHould be ready to go again. Since right now no new findings exist, I've taken the liberty to already remove the `wip!` commit.

sespiros (approved):
Thanks!

PR #2638, opened 2026-09-10
rfc: HTTP API Security

charludo:
[rendered version](https://github.com/edgelesssys/contrast/blob/rfc-015/rfc/015-http-api-security.md)


rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+A client that validates the report therefore holds an authenticated public key it can encrypt to.
+This is the replacement mechanism for aTLS binding the TLS key into the report, and is what allows us to fulfill requirement 2.iii for requests sending secrets to the Coordinator.
+
+The key is generated ephemerally per exchange.
+A fresh report is minted on every call regardless, since the client's nonce is in it, so per-exchange generation introduces little extra attestation cost.
+The Coordinator holds the private key in a two minutes TTL-bounded, size-bounded pending-key cache until the follow-up request arrives.
+Per requirement 4, losing this cache is benign.
+The follow-up request fails with a distinguishable error and the client repeats the exchange.
+
+For the algorithmic primitives, [HPKE](https://www.rfc-editor.org/rfc/rfc9180.html) base mode with `DHKEM(X25519, HKDF-SHA256)`, `HKDF-SHA256`, and `AES-256-GCM` is chosen.
```
burgerdev: `x25519` is imho a nice ECDH choice, but it's often debated in context of FIPS and other requirement docs that don't consider it acceptable. Other code touching asymmetric crypto already uses `P-256`.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+An implementation exists in [`crypto/hpke`](https://pkg.go.dev/crypto/hpke).
```
burgerdev, line 55: PQ is upcoming, https://datatracker.ietf.org/doc/draft-ietf-hpke-pq/, then we can go for `MLKEM768-X25519`. I saw that they are already implemented in Go, but we should maybe wait for the published standard.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+
+This is largely solved already by [RFC 012](012-application-level-verification.md)'s `POST /attest`.
+The client generates a 32-byte random nonce and sends it.
+The Coordinator returns a TEE report whose `REPORTDATA` covers that nonce along with digests of the state being reported.
+The Coordinator continues to authenticate itself solely through the TEE signature over that report.
+
+This direction is amended by the inclusion of an ephemeral HPKE public key generated by the Coordinator, returned in the response and covered by the report data:
+
+```
+reportData = sha256(nonce || transitionHash || sha256(rootCA) || sha256(meshCA) || hpkePubKey)
```
burgerdev: Why is this not just `nonce || pubkey`, and then a signature over state? How does this work in the recovery case?

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+   1. Freshness: a response, or an authenticated request, must be bound to *this specific* exchange.
+   2. Authenticity: the Coordinator must be identifiable as one allowed by the expected manifest.
+   3. Confidentiality: secrets must be readable only by their intended recipient, independent of the transport.
+   4. Authorization: privileged requests must be provably issued by a party the manifest permits.
+3. Server-side state must be optional for correctness. Losing it (restart, second replica, or a load balancer moving a client) may cost a retry, but must not cause a security or correctness failure.
+4. The wire contract must live in the `apitypes` package and be versioned per [RFC 014](014-backwards-compatibility.md).
+
+## Design
+
+### Case 1: Coordinator authenticates itself to a client
```
burgerdev: Hm, I was not expecting a breakdown by attested end, but rather a more generic protocol, more closely mapping to the TLS handshake, where client and server optionally attest.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+For the algorithmic primitives, [HPKE](https://www.rfc-editor.org/rfc/rfc9180.html) base mode with `DHKEM(X25519, HKDF-SHA256)`, `HKDF-SHA256`, and `AES-256-GCM` is chosen.
+An implementation exists in [`crypto/hpke`](https://pkg.go.dev/crypto/hpke).
+
+Note that the key isn't bound to Coordinator state.
+Therefore a Coordinator awaiting recovery can still issue one, which is what [`POST /v1/recover`](#post-v1recover-recover) depends on.
+
+#### Proof of possession
+
+Proof of possession of the HPKE private key isn't required for the guarantees this API needs, and having the Coordinator sign something with the HPKE key isn't readily implementable,
+as any scheme of the form `PRIVK_SIGN(report)` requires that the attested key is a signing key.
```
burgerdev: Generally speaking, HPKE keys are asymmetric keys that can also be used for signing. It should be possible to use an `ecdsa.PrivateKey`, convert it to an `ecdh.PrivateKey` with https://pkg.go.dev/crypto/ecdsa#PrivateKey.ECDH and then plug that into `NewDHKEMPrivateKey`.

What I don't know yet:
- Is there a concern using the same key for signing _and_ encryption? Probably not, especially given that the key is used a handful of times.
- How to prove possession of a hybrid key? Is both halves individually sufficient? `mldsa` is not part of the stdlib.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+
+Furthermore, it's unclear what proof of possession adds against the threat model:
+
+- An attacker who substitutes a key of their own needs a TEE report over that key that validates against the client's reference values, which only a genuine Coordinator can produce.
+- An attacker who relays honestly passes through the genuine Coordinator's key, so anything sealed to it stays unreadable to them.
+- An attacker who replays an older report fails the nonce check.
+- A key bound into the report but not held by the Coordinator could only be placed there by a malicious Coordinator, which measurements are what exclude.
+
+If an explicit confirmation is wanted anyway, adding a separate ephemeral signing key alongside the KEM key, both covered by the same report data, would provide one.
+This isn't proposed for the initial implementation.
```
burgerdev: Yeah, this is probably correct. Let's move the PoP discussion into the "Alternatives" section?

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+If the signature can be verified against the key list in the manifest, this serves as authorization.
+This replaces the `validatePeer` mTLS client certificate check.
+
+Manifest updates reuse the existing signature scheme.
+`validateSignature` already verifies an ECDSA signature by a `WorkloadOwnerPubKeys` entry over the transition hash.
+
+Recovery has no such scheme to reuse.
+Seed share owners authenticate currently by presenting their key as an mTLS client certificate, which `validatePeer` compares against `SeedshareOwnerPubKeys`.
+Because seed share owner keys are RSA-4096 (while workload owner keys are ECDSA P-384), the manifest-update scheme can't be copied directly.
+The new signature should instead be [RSA-PSS](https://www.rfc-editor.org/rfc/rfc8017.html#section-8.1) with SHA-256.
```
burgerdev, line 73: What we should do instead is phase out RSA keys (looking at PQ readyness). iirc, the only reason we went for OAEP back then was that HPKE was not yet implemented in Go. Hybrid signing still unclear, though, as for the PoP.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+#### Setting the initial manifest
+
+Currently setting the initial manifest requires no authentication of the client to the Coordinator, following the trust-on-first-use model of [RFC 002](002-secure-set-manifest-endpoint.md).
+The Coordinator trusts the first manifest it's given, and the client then verifies, among everything else, that the manifest now in force is the one it set.
+An attacker who races the legitimate client and sets a manifest first causes that verification to fail.
+
+To further harden this and prevent the TOFU window issue, a public key can be delivered to the Coordinator out of band, in its pod specification,
+and the Coordinator can then require the initial `SetManifest` to be signed by the corresponding private key.
+Since the Coordinator's pod specification is covered by its policy, and the policy hash appears in `HOSTDATA`, a key placed there becomes part of the launch measurement and is itself attested.
+This isn't proposed for the initial implementation.
```
burgerdev: This is a separate improvement that I'd consider unrelated.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+
+Recovery has no such scheme to reuse.
+Seed share owners authenticate currently by presenting their key as an mTLS client certificate, which `validatePeer` compares against `SeedshareOwnerPubKeys`.
+Because seed share owner keys are RSA-4096 (while workload owner keys are ECDSA P-384), the manifest-update scheme can't be copied directly.
+The new signature should instead be [RSA-PSS](https://www.rfc-editor.org/rfc/rfc8017.html#section-8.1) with SHA-256.
+
+#### Setting the initial manifest
+
+Currently setting the initial manifest requires no authentication of the client to the Coordinator, following the trust-on-first-use model of [RFC 002](002-secure-set-manifest-endpoint.md).
+The Coordinator trusts the first manifest it's given, and the client then verifies, among everything else, that the manifest now in force is the one it set.
```
burgerdev, line 78: I think we should just check the signature/client cert against the incoming manifest on first set, which would make this easier to port to HTTP.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+```
+POST /v1/mesh/cert                    (no Authorization header)
+-> 401 Unauthorized
+  WWW-Authenticate: contrast-attestation-v1 nonce="<base64>"
+
+POST /v1/mesh/cert                    (Authorization header, evidence in body)
+-> 200 OK
+```
+
+This reuses the standard [`WWW-Authenticate`](https://www.rfc-editor.org/rfc/rfc9110.html#name-www-authenticate) and [`Authorization`](https://www.rfc-editor.org/rfc/rfc9110.html#name-authorization) header semantics with a Contrast-specific authentication scheme, so intermediaries treat the exchange as ordinary HTTP authentication.
```
burgerdev, line 97: We might need to think about what goes into the `Authorization` header, since HTTP header length is finite.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+
+The client's report binds two client keys to the challenge:
+
+```
+reportData = sha512(certPubKey || kemPubKey || nonce)
+```
+
+`certPubKey` is the ECDSA key the issued certificate will attest, and `kemPubKey` is an X25519 key used only to receive the sealed response.
+The aTLS path uses one key for both roles, but separating avoids using one key across two cryptographic purposes.
+Proof of possession of `certPubKey` comes from the request signature, similar to the protocol-level PoP [`draft-reddy-rats-key-binding`](https://datatracker.ietf.org/doc/html/draft-reddy-rats-key-binding) prescribes for certificate enrollment.
```
burgerdev: Consider pointing to a concrete revision of the draft, since it's under development and might change drastically. It also makes heavy use of EAT structures, which we may want to avoid until we implement a proper standard.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+The current implementation appends the observed peer address to the certificate's DNS names:
+
+```go
+if host, _, err := net.SplitHostPort(p.Addr.String()); err == nil {
+    dnsNames = append(dnsNames, host)
+}
+```
+
+Behind a terminating proxy this observes the proxy's address rather than the workload's.
+Using `X-Forwarded-For` would make certificate contents attacker-influenced, so the HTTP path should drop source-derived SANs and use only the manifest-declared ones.
```
burgerdev, line 126: The IP is already not trustworthy (the host can route it as it pleases), but anyone maliciously using that IP would still need to have the private key for the cert. Anyway, for the proxy reason, it's probably correct to not add the IP here.

rfc/015-http-api-security.md
burgerdev: On the other hand, we argued above that we don't need to prove possession. There's also nothing left to sign, is there?

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+h' = sha256(hex(h))
+sig = ECDSA-Sign(workloadOwnerKey, h')
+```
+
+The initial manifest requires no signature.
+The client confirms the result with a follow-up `/v1/attest`.
+
+The mTLS peer-key path (`validatePeer` against `WorkloadOwnerPubKeys`) is dropped, meaning the signature becomes the only authorization mechanism for updates.
+The gRPC path accepts either today, so this doesn't introduce a new mechanism.
+The Coordinator is no longer attested during the call, verification moves to the set-then-verify sequence the CLI effectively performs anyway.
```
burgerdev, line 174: Well, right now it's one call, but with this change we'll need two, and potentially lose the atomicity ("I received the mesh CA cert for the manifest I just set"). If we included a nonce in the request we could just return the same thing as `/v1/attest` if we wanted.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+Security is unchanged for updates.
+
+#### `POST /v1/recover` (Recover)
+
+This requires two requests.
+
+1. `POST /v1/attest` with a fresh nonce, to obtain the attested HPKE public key.
+2. `POST /v1/recover` carrying the seed and salt sealed to that key, plus a seed share owner signature.
+
+The client consumes only the HPKE key and ignores `CoordinatorState`, which is empty while stale.
```
burgerdev: It should be possible to distinguish stale from unset.

rfc/015-http-api-security.md
```diff
@@ -0,0 +1,240 @@
...
+
+```go
+type RecoverRequestV1 struct {
+    // Nonce from step 1, identifying which pending key to unseal with.
+    Nonce []byte `json:"nonce"`
+    // {seed, salt} sealed to the Coordinator's HPKE key, as returned by hpke.Seal.
+    Ciphertext []byte `json:"ciphertext"`
+    // RSA-PSS-SHA256 over sha256("contrast recover v1" || nonce || hpkePubKey || ciphertext).
```
burgerdev: We could sign `Force`, too.

rfc/015-http-api-security.md
burgerdev: We should also try to consolidate our hashing strategies. Some have a nonce at the beginning, some at the end. I think a good idea is what you pick up in the `Recover` API:

```
hash("contrast $what $version" || nonce || ...)
```

rfc/015-http-api-security.md
thomasten: I think we should use it now. Go's default TLS config does it. Privatemode uses it for its secret exchange.

rfc/015-http-api-security.md
charludo: swapped for `P-256`.

rfc/015-http-api-security.md
charludo: Sounds good. Swapped the primitives out in the text.

rfc/015-http-api-security.md
charludo: moved

rfc/015-http-api-security.md
charludo: Removed here, left the mention in the out of scope section.

rfc/015-http-api-security.md
charludo: Hm, fair. Changed in the text to include only the nonce in the header, the evidence in the body.

rfc/015-http-api-security.md
charludo: done

rfc/015-http-api-security.md
charludo: Yeah true. Removed the first sentence.

rfc/015-http-api-security.md
charludo: Removed here, points to the PoP reasoning at the bottom. I guess this topic is a nice eventual followup 😄

rfc/015-http-api-security.md
charludo: added a status field

rfc/015-http-api-security.md
charludo: added

rfc/015-http-api-security.md
charludo: Sounds good. Unified in the RFC.

PR #2639, opened 2026-09-11
 tdx: support reference values for multiple vCPU counts

sespiros:
TDX ACPI measurements depend on the guest CPU topology. Until now, Contrast generated TDX reference values for one vCPU only.

This PR:

- generates non-GPU TDX launch measurements for every supported vCPU count from 1 through 220
- stores the resulting RTMR0 values as alternatives in a single TDX reference-value profile
- verifies RTMR0 against those alternatives while validating all other TDX measurements normally
- keeps TDX GPU reference values at one vCPU because multi-vCPU GPU guests are unsupported
- enables the existing multiple-CPUs E2E test for `Metal-QEMU-TDX`

While aggregating the generated measurements, the build fails if two vCPU counts produce the same RTMR0 or if any reference value other than RTMR0 differs.

## Testing

- `go test ./internal/attestation/tdx ./internal/manifest`
- `just e2e multiple-cpus Metal-QEMU-TDX`


internal/attestation/tdx/validator.go
```diff
@@ -51,6 +51,37 @@ func (v *StaticValidateOptsGenerator) TDXValidateOpts(_ *tdx.QuoteV4) (*validate
...
+type RTMR0ValidateOptsGenerator struct {
+	Opts          *validate.Options
+	AllowedRtmr0s [][]byte
+}
+
+// TDXValidateOpts returns validation options for an allowed RTMR0 value.
+func (v *RTMR0ValidateOptsGenerator) TDXValidateOpts(report *tdx.QuoteV4) (*validate.Options, error) {
+	if len(v.AllowedRtmr0s) == 0 {
+		return v.Opts, nil
+	}
```
burgerdev, line 64: Without looking at the implementation, I can imagine the following behaviours of an empty list of allowed `RTMR[0]`s (from more to less expected):

- Never pass validation because `RTMR[0]` does not match.
- Fall back to `RTMR[0]` in `v.Opts`, which may be set or unset.
- Ignore `RTMR[0]` completely during validation.

I think this needs to be more clear for the caller, and especially in the manifest.

internal/attestation/tdx/validator.go
```diff
@@ -51,6 +51,37 @@ func (v *StaticValidateOptsGenerator) TDXValidateOpts(_ *tdx.QuoteV4) (*validate
 	return v.Opts, nil
 }
 
+// RTMR0ValidateOptsGenerator accepts multiple RTMR0 values.
+type RTMR0ValidateOptsGenerator struct {
```
burgerdev, line 55: Is there any chance that we'll need to do this again for another RTMR?

internal/manifest/manifest.go
```diff
@@ -385,9 +398,10 @@ type SNPValidatorOptions struct {
 // TDXValidatorOptions contains the verification and validation options to be used
 // by a TDX Validator.
 type TDXValidatorOptions struct {
-	VerifyOpts   *tdxverify.Options
-	ValidateOpts *tdxvalidate.Options
-	AllowedPIIDs [][]byte
+	VerifyOpts    *tdxverify.Options
+	ValidateOpts  *tdxvalidate.Options
+	AllowedRtmr0s [][]byte
```
burgerdev, line 403: Nit: this should be `AllowedRTMR0s`, but we already have `Rtmrs` in the manifest so we should probably wait for a new manifest version before changing the capitalization.

packages/by-name/contrast/reference-values/package.nix
```diff
@@ -63,28 +65,57 @@ let
...
-            xfam = "e702060000000000";
-            memoryIntegrity = false;
-          }
-        )
-      ];
+            }
+          ) vcpuCounts;
+          referenceValues = runCommand "tdx-reference-values.json" { nativeBuildInputs = [ jq ]; } ''
+            set -o pipefail
+            for launchDigests in ${lib.escapeShellArgs (map toString launchDigests)}; do
```
burgerdev, line 85: Does this mean we get a set of reference values per CPU count? I was expecting a list of RTMR0s per set of refvals.

packages/by-name/contrast/reference-values/package.nix
```diff
@@ -63,28 +65,57 @@ let
...
+          vcpuCounts = if withGPU then [ 1 ] else lib.range 1 220;
+          launchDigests = map (
+            vcpus:
+            kata.calculateTdxLaunchDigests {
+              inherit
+                os-image
+                ovmf
+                withGPU
+                vcpus
+                ;
```
charludo, line 79: Together with the "TDX reference values differ outside RTMR0" guard below, this makes me think maybe we should only calculate `RTMR0` in the loop, for vCPU counts `2..220`? That would make the check superfluous, and not require recomputing everything else 219 times. If possible. It's a bit of a nit tbh.

internal/attestation/tdx/validator.go
```diff
@@ -51,6 +51,37 @@ func (v *StaticValidateOptsGenerator) TDXValidateOpts(_ *tdx.QuoteV4) (*validate
...
+	Opts          *validate.Options
+	AllowedRtmr0s [][]byte
```
charludo, line 57: Shouldn't this maybe be a map? Assigning CPU count to index in the list *works*, but can break relatively easily if someone e.g. reorders them.

packages/by-name/contrast/reference-values/package.nix
```diff
@@ -63,28 +65,57 @@ let
...
+                      $shared + {
+                        rtmr0Alternatives: (map(.rtmrs[0]) | .[1:])
+                      }
+                    ]
+                  }
+              end
+            ' > "$out"
+          '';
+        in
+        (builtins.fromJSON (builtins.readFile referenceValues)).tdx;
```
charludo, line 118: This writes the values to file, only to immediately re-read them into nix. Maybe we can avoid that roundtrip?

charludo (commented):
Nice! question, have you considered only adding the required (based on vCPU count) values into the manifest, instead of the full list? We had that for a while for the SNP values at least.

Also, just a hunch, but does this break on runtime-rs?

internal/attestation/tdx/validator.go
sespiros: Currently we launch all of our guest vms with the same kernel (RTMR1), initrd & guest kernel command line (RTMR2). Only RTMR0 varies because of the vCPU-dependent ACPI tables. So unless that changes and we want to allow multiple combinations of kernels/kernel config, I don't see a need. That would also require further design because independent per-RTMR allowlists could accept arbitrary combinations, while only particular combinations may be valid.

I also think that if we were to allow further and more complex parameterization we would have to eventually switch away from having fixed accepted values in the manifest to a more dynamic attestation model where we would have a list of accepted hashes per measured component instead and we would retrieve a quote + an event log we would have to replay etc.

internal/attestation/tdx/validator.go
sespiros: Good point. Aside of my other comment wrt how we can make this more generic, I will include Rtmrs[0] in `AllowedRtmr0s` when constructing the options and make an empty allowlist an error.

packages/by-name/contrast/reference-values/package.nix
sespiros: We keep one set of reference values, with the other RTMR0s in Rtmr0Alternatives. Following Charlotte’s suggestion, we now calculate the full set once and only RTMR0 for the remaining CPU counts.

packages/by-name/contrast/reference-values/package.nix
sespiros: Added a fixup to calculate the shared measurements once and only RTMR0 for the other CPU counts. This also removes the check that the other measurements stay unchanged.

internal/attestation/tdx/validator.go
sespiros: This list is only an allowlist: validation checks whether the reported RTMR0 matches any entry. We don’t use list indices as CPU counts, so reordering it doesn’t change validation.

packages/by-name/contrast/reference-values/package.nix
sespiros: Pushed a fixup.

sespiros:
> Nice! question, have you considered only adding the required (based on vCPU count) values into the manifest, instead of the full list? We had that for a while for the SNP values at least.

I tried selecting only the RTMR0s needed by the workloads by parsing the deployment files, but that complicates updates. generate preserves reference values in existing manifests, so changing CPU limits could leave the new VM’s measurement untrusted. Updating the list automatically could overwrite a user’s manual restrictions, since we can’t distinguish those from generated values.

I’d keep all supported CPU counts allowed by default for now. Narrowing the list needs a clear update workflow first. @burgerdev any ideas on this?

> Also, just a hunch, but does this break on runtime-rs?

I haven't tested runtime-rs, should I? But TIL runtime-rs calculates amount of guest CPUs differently so yes this could break our calculations here. I also did a quick poc for this and turns out runtime-rs also introduces other mismatches so I don't think this should be part of this PR.

internal/attestation/tdx/validator.go
burgerdev: Alright, let's not worry about it now and improve support for refval choice lists in manifest format v2.

internal/attestation/tdx/validator.go
burgerdev: In the manifest, these should be a list (for encoding stability), but in the validator it should be a lookup table, no? `map[HexString]struct{}` for example. I did not measure the impact, though.

burgerdev (commented):
>I tried selecting only the RTMR0s needed by the workloads by parsing the deployment files, but that complicates updates. generate preserves reference values in existing manifests, so changing CPU limits could leave the new VM’s measurement untrusted. Updating the list automatically could overwrite a user’s manual restrictions, since we can’t distinguish those from generated values.
> 
> I’d keep all supported CPU counts allowed by default for now. Narrowing the list needs a clear update workflow first. @burgerdev any ideas on this?

I think adding 200+ hashes to the manifest is more than awkward. These are not hardware reference values to be filled in by the user, but software measurements provided by the CLI. Are you sure these don't get updated? That would also affect all other software measurements. 

I would not worry about users messing with the `RTMR[0]` - why would they, and how?

internal/attestation/tdx/validator.go
sespiros: I benchmarked this using 48-byte SHA-384 values and `map[HexString]struct{}`, including hex encoding of the reported measurement. Five runs on Apple M4 / Go 1.26.5.

Median lookup times, matching the last entry:

| Allowed hashes | `[][]byte` scan | Hex-string map |
|---:|---:|---:|
| 1 | 2.6 ns | 49 ns |
| 2 | 4.0 ns | 50 ns |
| 4 | 6.9 ns | 50 ns |
| 16 | 23 ns | 48 ns |
| 220 | 313 ns | 46 ns |

The scan allocates nothing; the hex-map lookup allocates 96 bytes for encoding. Building the map from existing hex strings costs another ~2.9 µs for 220 entries.

With the new selection logic, the allowlist contains only the distinct CPU counts needed by the deployment, so it should generally be small. I'd keep the slice for now.

These are microbenchmarks with synthetic SHA-384 digests, not end-to-end attestation timings.

burgerdev: Nice, thanks!

sespiros:
> I think adding 200+ hashes to the manifest is more than awkward. These are not hardware reference values to be filled in by the user, but software measurements provided by the CLI. Are you sure these don't get updated? That would also affect all other software measurements.
> 
> I would not worry about users messing with the `RTMR[0]` - why would they, and how?

I pushed a change that implements the suggestion, ptal.

burgerdev (approved):
Thanks, lgtm

PR #2642, opened 2026-09-14
scripts: allow fetching and normalizing agent rpcs

davidweisse:
This adds two scripts: one for fetching the agent rpcs from the `policy-test/assets/pod.yml` by running `genpolicy`, deploying it, and extracting `/tmp/policy.jsonl` via `debugshell-host`; and one for then post-processing this to replace any dynamic data like container ids, kata paths, etc. with fixed content. Getting new test cases for the policy test suite would then look something like this:

```sh
# If not already present
just runtime write-namespace (apply "runtime")

just policy-rpc
nix run .#base.scripts.normalize-agent-rpcs workspace/policy.jsonl > policy-test/testdata/testcase.json
```

The kata runtime replaces the storage source for pause containers with `"pause"`, but genpolicy doesn't. Therefore, when we run genpolicy with an image that has a digest (normally it doesn't: `ghcr.io/edgelesssys/kubernetes/pause:3.6`, but with the local registry we have: `.../pause@sha256:...`), the policy check will fail. The patch aligns the genpolicy behavior with the runtime so that the storage source also contains `"pause"` and the "digest check" will pass.

Fixes CON-252


packages/by-name/scripts/normalize-agent-rpcs/normalize-agent-rpcs.jq
```diff
@@ -0,0 +1,127 @@
...
+            "11:22:33:44:55:66"
+        )
+
+    else
+        .
+    end
+)
+
+# Add allowed field
+| map(. + {allowed: true})
```
burgerdev, line 105: Why do we have that field here instead of in the modification patches?

packages/by-name/scripts/normalize-agent-rpcs/normalize-agent-rpcs.jq
```diff
@@ -0,0 +1,127 @@
...
+# Normalize object key ordering
+| walk(
+    if type == "object" then
+        to_entries
+        | sort_by(.key)
+        | from_entries
+    else
+        .
+    end
+)
```
burgerdev: Could this be accomplished with `jq -S`?

packages/by-name/kata/source/package.nix
```diff
@@ -191,6 +191,12 @@ rec {
       # emitted.
       # Upstream PR: https://github.com/kata-containers/kata-containers/pull/13692
       ./0027-runtime-rs-omit-guest-pull-driver-metadata.patch
+
+      # Changing the pause container image reference to one with digest creates a policy mismatch
+      # because the runtime fixes the storage source to "pause" and the policy matches the digest.
+      # Align policy generation with the runtime by setting the image ref for the pause container to "pause"
+      # before populating the policy data.
+      ./0028-genpolicy-change-pause-container-image-ref-to-pause.patch
```
burgerdev: I think we should merge this into the guest pull verification patch.

packages/by-name/scripts/normalize-agent-rpcs/normalize-agent-rpcs.jq
davidweisse: If you mean the JSON patches that come in a future PR: the test consists of multiple requests, and each one is allowed by default. A modification would then change something and mark that specific request as not allowed. Otherwise, we would have to 1. allow every request on every patch file or 2. have a separate base patch that adds the allow field to every request and apply that patch before any others. Currently we can just process the test file as is, so not sure what's better/easier to read.

burgerdev (approved):
Thanks, lgtm!

PR #2644, opened 2026-09-17
sbom: describe the release CLI

sespiros:
Use cli-release as the CLI SBOM root so its component reference and hash identify the variant shipped in releases. The development CLI embeds different genpolicy settings and produces a different output hash.


burgerdev (approved):
Thanks!

PR #2649, opened 2026-09-21
policy-test: add adversarial tests

davidweisse:
This allows creating adversarial tests for the policy test suite via JSON patch files. The test data directory structure is changed in the following way:
```
testdata
└ pod
  └ resource.yml
  └ base.json
  └ specific_test.json
└ deployment
└ ...
```

The `resource.yml` is used to generate the policy. The `base.json` contains all default RPC requests and is used as the base test case. Other tests are JSON patch files that get applied to `base.json` after which the policy gets evaluated again.

To make error handling easier (especially with subtests inside subtests), the test runner now uses Go's testing library, so the main runner is now a `func TestPolicy(t *testing.T)`.

Fixes CON-254


policy-test/main_test.go
```diff
@@ -44,159 +43,68 @@ type testImage struct {
...
-	if err != nil {
-		return fmt.Errorf("listen: %w", err)
-	}
-	defer lis.Close()
+	lis, err := (&net.ListenConfig{}).Listen(t.Context(), "tcp", "127.0.0.1:0")
+	require.NoError(err)
+	registryAddr := lis.Addr().String()
+
+	t.Cleanup(func() {
+		require.NoError(srv.Close())
```
burgerdev, line 57: `require` inside `Cleanup` panics, iirc. I would not worry too much about the errors here.

policy-test/main_test.go
```diff
@@ -252,3 +160,58 @@ func generatePolicy(ctx context.Context, runner *genpolicy.Runner, yaml []byte)
...
+		req.NoError(err)
+		files, err := os.ReadDir(filepath.Join(dataDir, name))
+		req.NoError(err)
+
+		// Ensure the base test case is run first, and fail the entire subtest if this already fails.
+		if !t.Run("base.json", func(t *testing.T) {
+			runTestCase(t, policy, baseData)
+		}) {
+			t.FailNow()
+		}
```
burgerdev: You could `require.True` this `Run`.

burgerdev (approved):
lgtm

burgerdev (approved):
lgtm

policy-test/main_test.go
davidweisse: Not sure where this would panic, but don't we do the same thing here? https://github.com/edgelesssys/contrast/blob/96045b2854d6bbc576105df9b569c651bd2f6459/e2e/proxy/proxy_test.go#L80-L84

